PR #10, opened 2019-10-15
Mccalluc/bootstrap

mccalluc:
@ngehlenborg : Does this styling for the banner have your buy-in? (Differs from old demo in just using legible text for the "HuBMAP" link, and just listing links across the top, instead of burying in menu.) Also, in the body, this is how I'll be flagging the UI work that still needs to be done.

banner


mccalluc:
Going ahead with merge to unblock work, but comments still welcome!

ngehlenborg:
This is looking good. I assume the plan is to remove the styles as the content/format/etc. gets implemented?

mccalluc:
yep! but I think we want to be clear about what's been implemented, and what's just on the screen, because something needs to be on the screen.

PR #56, opened 2019-11-20
Mccalluc/dockerize

mccalluc:
This provides a Dockerfile, and script, `docker.sh`, which shows how to run it: Most importantly, mounting the `app.conf`.

The [base image](https://github.com/tiangolo/uwsgi-nginx-flask-docker) is not CentOS; instead it looks like it starts from [python:3.7-stretch](https://github.com/tiangolo/uwsgi-nginx-docker/blob/master/python3.7/Dockerfile). I did first take a look at our ingest-ui for inspiration, but `uwsgi-nginx-flask-docker` seems to be well thought-out, well regarded, well documented, and maintained.

It is opinionated, and to take advantage of defaults it expects a certain project structure: Most of the changes in this PR are file renames, from `portal/*` to `app/app/*`.

@shirey : If this can be approved, I could build and push the image to dockerhub, and then you all would pull it down and deploy? Or do you have some other hand-off in mind? Going forward, I would ping you towards the end of the sprint to let you know a new image had been deployed, and you would take it from there?


mccalluc:
- Down the road, it will be just uWSGI. (NGINX will be handled by gateway.) But for now, this is fine.
- If quickstart is run first, config is in place... but if someone runs docker.sh first, it's not obvious.
- Add option for PORT on command line.
- Consider pip freeze for extra stability.

mccalluc:
Thanks! Static resources aren't being served in docker: Let me figure that out, and then I'll merge.

PR #77, opened 2019-12-13
Deployment changes

yuanzhou:
This PR includes:

- added docker-compose files for local dev and deployment on testing server
- updated instructions


yuanzhou:
Sounds good @mccalluc. Whatever works best for you would work for me. 

I think plain docker command can do whatever a docker-compose.yml can do. In my use case, I have several docker-compose projects and each project may have multiple services running on different ports with multiple volume mounts. The docker compose file makes defining these services very easy for different deployment environments  and I can specify a shared docker network for containers to talk to each other. From starting/stopping services perspective, docker compose make the management of multiple docker containers lot easier with simple and short commands instead of writing long docker command with many options and arguments. In your case, if you are only developing one docker project, the plain docker command or a script works as good as a docker-compose.yml.

Regarding the redirect error, I think you are right. `portal-ui` is what I used in the docker-compose file for the hostname. The reason I have to use it is because the gateway has its own nginx as a reverse proxy, it forwards requests to backend python services running individual containers via uwsgi. In the case of your portal-ui, you have a container that runs nginx+uwsgi+flask. What I did was creating ssl certificates to the gateway nginx config file for domain `portal.test.hubmapconsortium.org` and nginx does a proxy to the `portal-ui` hostname on port 80, that's how nginx talks to the container by using the container's hostname. And it makes minimum changes to your code or container. I don't understand how flask could have picked up the backend hostname instead of the sub-domain and used that in the redirect. 

One solution would be making the domain name as a configurable in the `app.conf` and use that when the flask `/login` endpoint talks to Globus. Any insights? We need to address this issue especially when deploying with other docker projects on the same machine.

mccalluc:
I should see if I can get my image to expose the uwsgi... If you have a base image you'd prefer, I'd be open to changing, but it might be nice for now to keep on the nginx in the image, so it's easier to preview.

(I think I'll try to change the way it references the image... I think it can point to a tag on github...)

So, still in my court for now. Thanks.

mccalluc:
Ok, I think is the nginx config:
```
# cat /etc/nginx/conf.d/nginx.conf
server {
    listen 80;
    location / {
        try_files $uri @app;
    }
    location @app {
        include uwsgi_params;
        uwsgi_pass unix:///tmp/uwsgi.sock;
    }
    location /static {
        alias /app/app/static;
    }
}
```

Let me see if I can expose `uwsgi.sock` on a port.

PR #82, opened 2019-12-17
Docker compose fix

yuanzhou:
@mccalluc I finally got a chance to checkout the modifications you made based on PR #77 and thanks for incorporating the deployment changes. I like the idea of pulling from DockerHub image for testing deployment with docker-compose. This way, you'll have control to make all the code changes and push the updated image (with a release version/tag) to DockerHub. Then I'll just pull the updated docker-compose files for deployment on testing server.

My PR includes:

- fix to the volume mount path for `app.conf`. Since you moved the docker-compose yaml files to its own directory 'compose/', the path to `app.conf` needs to be updated as well.
- due to the error with globus settings, I couldn't use either port 5001 or 8686 for local dev compose settings, so I changed to port 5000, which is in the whitelist. Tested the local dev deployment on my machine and it works well.
- updated instructions

I also deployed again on the testing server, still having the Globus mismatch URL which uses `portal-ui/login` as the redirect URL. Can you change the flask code and make the backend URL configurable in `app.conf`? Without changes, it works fine on local dev, but testing has different infrastructure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.0.6 - in progress
+### Changed
+- Fixed Docker Compose.
+
 ## [v0.0.5](https://github.com/hubmapconsortium/portal-ui/tree/v0.0.5) - 2019-12-16
 ### Added
 - Format details page headers.
--- a/README.md
+++ b/README.md
@@ -18,12 +18,16 @@ update `app.conf` with the Globus client ID and client secret,
 run `quick-start.sh` again,
 and visit [localhost:5000](http://localhost:5000).
 
-## Tag and release
+## Tag, release, and deployment
 To tag a new version for github and [dockerhub](https://hub.docker.com/repository/docker/hubmap/portal-ui),
-increment `VERSION` and run:
+checkout a release branch, increment `VERSION` and run:
 ```sh
 ./push.sh
 ```
+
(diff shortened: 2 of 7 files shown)


yuanzhou:
Sounds great! Thanks for making the tweaks!

mccalluc:
@yuanzhou : Can you look over my changes, and see if they make sense on your side? (A different approach would be to always have some explicit SERVER_NAME, but I think that might be more confusing.)

Go ahead and merge if it looks good.

yuanzhou:
@mccalluc I tested with deploying the docker-compose-fix  branch on testing server which includes the changes we both made, and I'm getting 404 error. I tried both `SERVER_NAME = 'https://portal.test.hubmapconsortium.org'` and `SERVER_NAME = 'portal.test.hubmapconsortium.org'` in the app.conf . And before I deployed, I removed the old container as well as the old image. The new deployment was pulling from DockerHub using the image specified in compose file

I got the same 404 after I shell into the contianer, nginx is running, but returns 404

I'm also wondering if you can make it clear in the config whether to specify the scheme (http VS https) or not needed at all. Maybe something just similar to `ENTITY_API_BASE = 'https://entity-api.test.hubmapconsortium.org'`, we call it `PORTAL_BASE_URL` or something up to you. This way, we can even specify some /path after the domain/hostname in case the ddeployment uses directories.

mccalluc:
(Not 100% sure this will do it, but right now we're deploying from images on github.)

PR #83, opened 2019-12-17
Add links to known issues from json schema

mccalluc:


context/app/validation_utils.py
```diff
@@ -0,0 +1,17 @@
+import jsonschema
+
+
+def for_each_validation_error(data, schema, callable):
```
ilan-gold: Maybe add some comments as to what is going on here with respect to why we are appending `_TODO`

context/app/routes_main.py
```diff
@@ -44,11 +44,7 @@ def details(type, uuid):
 
     entity = client.get_entity(uuid)
     entity_schema = load_yaml(open(current_app.root_path + '/schemas/entity.yml'))
```
ilan-gold: maybe the `entity.yml` should be a variable set at the top of this file.  also, maybe a `with open(...)`

context/app/validation_utils.py
mccalluc: yes: done.

context/app/routes_main.py
mccalluc: Added `with open`. Tried moving it up and out, but then `current_app` isn't available. Not sure about the best approach, but just added a TODO for now.

ilan-gold: ok - i meant just the `'/schemas/entity.yml'` as a variable since the `current_app` shouldn't change.

PR #84, opened 2019-12-18
Mccalluc/v0.0.6

mccalluc:
Image pushed to Dockerhub


mccalluc:
This image did not work for Zhou either. Going to merge this, and then try to get a local setup where I can replicate the error he sees.

PR #103, opened 2020-01-05
Tighter routing of MD fixes docker

mccalluc:


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Fix redirect problem.
```
mccalluc: This broke because the CWD when running under docker is different than when running under Flask dev server. I fixed the immediate problem, and then came back later and fixed tests.

PR #110, opened 2020-01-08
fix compose

### LINKED ISSUE #89: Mimic real deployment with compose
Summary from Zhou:
> To sum it up:
>
> 1. add a new nginx as a reverse proxy locally with docker, running on localhost:5000 with port mapping host port 5000 to nginx container port 80, no need to use 443.
> 2. proxy the requests from http://localhost:5000 to http://portal-ui, portal-ui is the hostname and container name that runs another nginx with uwsgi/flask to serve the portal UI
> 3. try with or without SERVER_NAME = 'http://localhost:5000'
If everything works, you should get globus redirect url as http://localhost:5000/login instead of http://portal-ui/login
> You can also try change the container's hostname to localhost

mccalluc:
Fix #89

The important part is `proxy_set_header  Host $http_host;` in the nginx config: No changes required in the flask code.

I've also changed dev to build from source rather than relying on image. Conceivably, you could do this for test, too, but I think the docker images make things more predictable.

I'll go ahead and merge, but will ping you and Bill on Slack: It'd be really nice to have this up on your infrastructure by Friday morning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.0.8 - In progress
 ### Changed
-- Get the Docker Compose to better emulate the production environment.
+- Get Globus login redirect working with Docker Compose reverse proxy.
 
 ## [v0.0.7](https://github.com/hubmapconsortium/portal-ui/tree/v0.0.7) - 2020-01-08
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## v0.0.8 - In progress
+### Changed
+- Get the Docker Compose to better emulate the production environment.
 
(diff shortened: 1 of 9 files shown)


yuanzhou:
@mccalluc sorry that I just saw your message. I'll re-deploy and test.

yuanzhou:
On HTTPS deployment, we'll also need to add ` proxy_set_header X-Forwarded-Proto $scheme;`

So the nginx config looks like:
````
location / {
        proxy_pass http://portal-ui;
 
        # If proxy_set_header is missing,
        # the redirect url passed to Globus will use "portal-ui" instead.
        # (Plain "$host" does not include port.)
        proxy_set_header  Host $http_host;
        proxy_set_header X-Forwarded-Proto $scheme;
    }
````

PR #111, opened 2020-01-09
simplify compose demo

mccalluc:
I've stripped the demo down to the minimum, and it still works. @yuanzhou, Is the test deployment config under version control? Could you link to it, probably from compose/README.md?


yuanzhou:
For testing deployment or future production deployment, do you prefer to build from the context or pull from the tagged image? Pulling the tagged image requires the `test.yaml` or `prod.yaml` to be updated once you release a new tag.

mccalluc:
@yuanzhou : I may be confused: Do you still want to use compose configurations in this repo, or can the sample here just be for demos on localhost?

I was thinking there must be some other place where you have docker compose for everything in the gateway, and it's there that you would increment the version number when I have a new release. (You sent me your `portal-ui.conf`: Where is the docker compose that references that nginx config?)

yuanzhou:
@mccalluc I can see why this is confusing. It's good to have the demos in this repo for localhost testing. We'll probably need this down the road for production deployment testing on local.

The gateway has the `portal-ui.conf` in that repo because the gateway serves as the reverse procy for all the APIs running behind it. But the actual docker compose yaml files are in each individual repos separately. That's why initially I added the `test.yml` in your repo. 

I thought you didn't want to keep that `test.yml` and later `prod.yaml` in your repo. If the plan is to deploy the portal-ui along with other APIs, then we should probably keep those yamls in this repo. They won't fit the gateway repo for version control purposes.

yuanzhou:
For example, the entity-api repo has these docker-compose yaml files: https://github.com/hubmapconsortium/entity-api/tree/master/docker

The gateway has nginx conf files for each API service on each deployment environment: https://github.com/hubmapconsortium/gateway/tree/master/nginx

For building the starting up the services, I can either go to the gateway and use that script to handle the build/start/stop of all services or I can go to the individual repo to issue the docker compose commands to build/start/stop that individual service.

yuanzhou:
The `test.yml` I'm currently using for portal-ui container is:

````
version: "3.7"

services:

  portal-ui:
    image: hubmap/portal-ui:v0.0.7
    # Avoid accidentally creating zombie processes
    init: true
    # Specifying a restart policy to avoid downtime
    restart: always
    networks:
      - gateway_hubmap

networks:
  # This is the network created by gateway to enable communicaton between multiple docker-compose projects
  gateway_hubmap:
    external: true
````

It's based on the `base.yml`:

````
version: "3.7"

services:

  portal-ui:
    hostname: portal-ui
    container_name: portal-ui
    volumes:
      # Mount the app config in container to keep it outside the image.
````
Please fee free to modify.

mccalluc:
Brought back test.yml. Thanks for explanation.

PR #122, opened 2020-01-23
new issue templates

mccalluc:
- I feel ok with leaving out the body templates: If I'm a good person, I'll be sufficiently precise without the paperwork, and if I'm not, then it won't help.
- Does this feel like the right set of top-level types? I feel like making a distinction between an actionable "enhancement" and a "proposal" that needs buy-in is useful. Thoughts?


ilan-gold:
I like that idea a lot - that has been one of the hardest things with what I am working on right now to see what is enhancing vs. proposing.  I agree about the bodies.

PR #124, opened 2020-01-24
Mccalluc/fill in cypress

### LINKED ISSUE #115: Cypress.io tests

mccalluc:
... and this fills out the earlier test with a basic click-through. Fix #115


mccalluc:
Sorry for the failures earlier: The request for the the JSON data for vitessce is blocked by the hubmap gateway for IPs not on the whitelist. Constructing the Vitessce config inside the API client is a little weird, but for now, it's the one place that watches whether we're in mock API mode or not.

(I had started going down the route we use in Vitessce of setting up routes for the mock API responses, but this is different because this API is on a different host, which ties our hands, and because most of the API requests are sent from the server, rather than the client.)

context/app/api/client.py
```diff
@@ -113,5 +111,39 @@ def get_provenance(self, uuid):
...
+                            "target": [0, 0, 0]
+                        }
+                    },
+                    "x": 0, "y": 0, "w": 12, "h": 2
+                },
+            ]
+        }
 
-# TODO: More functions
+    # TODO: More functions
```
jkmarx: Not a fan of in code TODOs especially vague ones.

cypress/integration/portal-ui.spec.js
```diff
@@ -1,12 +1,46 @@
...
-    // so we won't end up depending on outside resources by accident.
-    cy.server({ force404: true });
+  it('has nice 404', () => {
+    // Homepage
+    cy.visit('/no-such-page', {failOnStatusCode: false});
+    cy.contains('404: Not Found');
+    cy.contains('If this page should exist');
   });
 
   it('works', () => {
```
jkmarx: for the description, 'perhaps correctly loads hubmap landing page'.

context/app/api/client.py
```diff
@@ -113,5 +111,39 @@ def get_provenance(self, uuid):
 
         return provenance
 
+    def get_vitessce_conf(self):
+        if self.is_mock:
```
jkmarx: Is this a temporary code until the APIs are up?

context/app/api/client.py
mccalluc: Will delete!

cypress/integration/portal-ui.spec.js
mccalluc: Scope of the test has expanded, but will retitle.

context/app/api/client.py
mccalluc: Mostly: 
- Some APIs are "up"... but they are only accessible from white-listed IPs, so I wanted some way to run the app when out of office.
- One place that isn't on the white list is the Travis server. If the requests were client side, we could mock them in Cypress, but the server side is a black box.

mccalluc:
Addressed Jennifer's comments in #130.

PR #133, opened 2020-02-07
Mccalluc/v0.0.9

mccalluc:
Scatterplot is mostly working: If you hover the mouse over a dot, you'll hit #454, but it renders, and the HTTP stuff works.


yuanzhou:
Merged and deployed on test server.  One UI thing I noticed is when you go to the scatterplot page, for example: https://portal.test.hubmapconsortium.org/browse/sample/86467b8dea173a57f3e0e566434cf170

there's a modal window shows up then disappears very quickly. I know it's a loading behavior,  it can be more smooth with fading in/out maybe. Minor user-experience improvement.

PR #137, opened 2020-02-27
Jkmarx/add webpack and react

jkmarx:
More work and discussion to be done. Unsure how we should handle the production vs development folder. This just gets react working locally so I can start ui component development within the portal.

**NOTE**: I'll be branching from here to incorporate the new materials UI theme.


mccalluc:
@jkmarx : Sorry to miss this -- Just added myself as reviewer, so it's on my radar now. I think the plan is for me to branch from this, make changes necessary to preserve existing functionality while react work is in progress (probably just add a new route?), and merge to master. Later you'll pick up these changes from master in your branch, and submit your own work?

If that's not right, let me know. Thanks!

mccalluc:
We chatted in person. Closing this, and a PR for the branch that was taken from this will be made in a day or two. (but re-open if this is a mistake!)

PR #141, opened 2020-03-02
Mccalluc/handle 401

mccalluc:
This commonly happens when you have been away from the site for a day or two and then come back.

- Should the Flask session be expiring on its own without this?
- Spent a little time on trying to mock the session.clear() so we could check that it is called... but decided what I have is good enough after wrestling just a little while.

### DIFF
--- a/context/app/main.py
+++ b/context/app/main.py
@@ -18,6 +18,11 @@ def access_denied(e):
     return render_template('errors/403.html', types={}), 403
 
 
+def unauthorized(e):
+    '''A 401 probably means Globus credentials have expired.'''
+    return render_template('errors/401.html', types={}), 401
+
+
 def gateway_timeout(e):
     '''A 504 means the API has timed out.'''
     return render_template('errors/504.html', types={}), 504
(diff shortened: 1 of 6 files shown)


context/app/main.py
```diff
@@ -18,6 +18,14 @@ def access_denied(e):
     return render_template('errors/403.html', types={}), 403
 
 
+def unauthorized(e):
```
ilan-gold: Why call this unauthorized if it's expired?

ilan-gold:
As for the session expiring, I don't think it's necessary as long as it's clear in the code that globus handles that.

context/app/main.py
mccalluc: "Unauthorized" is the name for the HTTP code... It seems to be caused by the globus credentials expiring, but conceivably it could have another cause that I haven't hit yet.

mccalluc:
> flask session expiring independently
@ilan-gold : I think I do want to keep that. Without it, the error message is telling the user to re-login, but it looks like they are already logged in, and that's just confusing.

PR #142, opened 2020-03-03
Handle case where credentials not granted

### LINKED ISSUE #129: Institution login works, but globus credentials missing --> Pretty error
When Jennifer tried to log in, the institution login worked, but globus credentials were not granted.

I should run the portal locally, and let you login on my machine and get the stack trace for the error you hit. This will be quicker than trying to get the log from PSC, and then getting a fix out and seeing if it changes the behavior.

mccalluc:
@shirey : I haven't found a way to distinguish between users who have not been granted access and those whose login has just expired. Is there a better way to do this?

Fix #129.


mccalluc:
Yes, there is a better way:

> Access granted would be based on membership in a group.  Our AuthHelper.getUserInfo (in the commons repo) can return a list of groups that a user (based on token) are a member of.  For most of our stuff I've been giving access via the "HuBMAP-Read" group.  I'll send you info on how to use that helper class in a separate thread.

mccalluc:
@shirey : If I get a 401, I now fire off a request to the nexus API you pointed me to.
- Should I be calling this directly? Or should I be referencing the commons code, instead of copy and paste?
- I'm currently checking the name of the group... Would it be better / more stable to check the long hex ID?
- Not sure what the other params on the API call are for... Are they needed?
- How do you test something like this? Is there a way for me to force expiration of globus creds? Could I get a second login without creds so I could confirm that the API response is what I think it should be?

(Don't feel obliged to answer in detail... if this seems mostly in the right direction, and the rest we'll learn in action, great.)

@ilan-gold : Could you give a code review?

context/app/main.py
```diff
@@ -19,11 +20,10 @@ def access_denied(e):
 
 
 def unauthorized(e):
-    '''A 401 probably means Globus credentials have expired.'''
-    # Go ahead and clear the flask session for the user.
-    # Without this, the button still says "Logout", as if they were still logged in.
-    session.clear()
```
ilan-gold: Why get rid of this?

mccalluc:
@shirey : never mind: What I should do is check groups immediately after login, and if they don't have the right groups, give an error at that point. (If you have thoughts about my use of this API, in general, that's still welcome!)

PR #159, opened 2020-03-14
Jkmarx/add footer header

jkmarx:
- Add footer.
- Include react header and footer to base template
- Add extra styling to make it cohesive


mccalluc:
@jkmarx : I can do a code review right now, but but without tests, and docker, and development working I do not think it should be merged. I'm not expecting you to do any more on this -- but if you think you do want to do more, and we need more feedback from me, let me know?

Thank you for everything: I hope the new job is great, and good luck these next couple months.

jkmarx:
@mccalluc I agree. Sorry, I assigned you before Travis kept failing. The current issue is with the missing provVis scheme.json. Wondering how you'd like to proceed. Since there isn't urgency to this, I'd like to continue working on this if we can resolve the dependency issue.

context/app/static/js/index.css
```diff
@@ -28,3 +29,18 @@ li {
...
     font-size: 2.6rem;
 }
+
+/* https://css-tricks.com/couple-takes-sticky-footer/ */
+.main-content {
+    min-height: calc(100vh - 120px);
+}
+
+.container-lg {
+    max-width: 1240px;
```
mccalluc: Did Nils ask for this? I feel like with the visualizations we want to do, if someone makes their window absurdly wide, perhaps we should respect their intentions.

context/app/templates/pages/base.html
```diff
@@ -8,16 +8,20 @@
...
     <link rel="stylesheet" href="{{ url_for('static', filename='table.css') }}">
-    <link rel="stylesheet" href="{{ url_for('static', filename='sticky-footer.css') }}">
     {% block extra_stylesheets %}{% endblock %}
 
     <title>{% block title_text %}{% endblock %} | HuBMAP</title>
+    <script>
+      let isAuthenticated = false;
+      {% if is_authenticated %}
+        isAuthenticated = true;
+      {% endif %}
```
mccalluc: perhaps
```suggestion
      const isAuthenticated = {{ 'true' if is_authenticated else 'false' }};
```

context/app/templates/pages/base.html
```diff
@@ -28,9 +32,9 @@
         </div>
       </div>
     </div>
-
-    {% include 'fragments/footer.html' %}
-
-    {% block extra_scripts %}{% endblock %}
+    <footer id="react-footer"></footer>
+    {% block extra_scripts %}
+      <script src="{{ url_for('static', filename='public/bundle.js') }}"></script>
```
mccalluc: Should this be outside the extra_scripts block? Won't we want bundle.js on every page?

context/app/static/js/index.css
jkmarx: Yes, but we didn't specifically speak about the visualizations. I'd agree with you on it and the  tool tabs can be handled separately with wise.

mccalluc:
Ok, fixed my set-up, and the rendering looks good. I'm going to look at the Travis errors and see if I have any ideas there, then look at the code. We also need to change the banner links to `search`.

PR #162, opened 2020-04-13
Fix capitalization of globus group

mccalluc:
@shirey : Coming back to this after a month. Looking back at your original note on slack:

> If you can view through the Globus app then you have the correct privileges.  Anyone in the HuBMAP-Read group (just confirmed, you are a member) has read access to everything in the landing zone except the mentioned Stanford data and a few data sets with sequence info.  Not sure of the details of how their API uses the client id- their app internally uses their own client id/secret.

I feel like lower-case "r" in "read" must have worked at some point, but it doesn't now. Please confirm that this capitalization will be correct going forward. (Or if I need to be more flexible, let me know.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Hit Elasticsearch instance and start configuring facets.
 - Add react, webpack, eslint, and styling dependencies.
 - Make the Elasticsearch endpoint part of the config.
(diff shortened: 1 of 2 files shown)


mccalluc:
@shirey : Thanks! Closing this in favor of #164, where I use only the ID.

PR #183, opened 2020-04-19
Remove active API url from default config

### LINKED ISSUE #172: Don't use default value for ENTITY_API_BASE
From Zhou:

> Don't use default value for ENTITY_API_BASE in app.conf, I'll always explicitly specify the URL based on deployment

mccalluc:
@yuanzhou : I think the change in default_config.py is what you wanted? Fix #172. Merge if it looks good to you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 # Changelog
 
 ## v0.0.12 - In progress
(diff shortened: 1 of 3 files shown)


yuanzhou:
Looks good. Thanks.

PR #186, opened 2020-04-24
Mccalluc/nexus cookie

mccalluc:
Put the nexus token in a cookie so the client side has it accessible. Demonstrate that it works with search.


mccalluc:
@john-conroy :
> The js in readCookie is on the older side, but otherwise looks good.

Thanks -- If there's a utilities library you'd like to include, file an issue? I'll fix the build, then merge.

mccalluc:
Close this PR in favor of #195.

Issue #180, opened 2020-04-17
Create a favicon


ngehlenborg:
@tsliaw: please create a few options when you have time.

ngehlenborg:
Here are some materials that could be helpful: https://drive.google.com/drive/folders/1i-JdnsSphQYzpnh7KMciazarJhT4ctkd

mccalluc:
Done!

ngehlenborg:
It looks like the generic detail pages are missing the favicon. @john-conroy: can you fix that when you create per-type detail pages?

<img width="903" alt="image" src="https://user-images.githubusercontent.com/1216518/81865153-8da17800-953b-11ea-84a9-abebb8f62f02.png">

https://portal.test.hubmapconsortium.org/browse/dataset/4f80da2501a9547fbbb6f2a1b9458633

mccalluc:
Close in favor of #293

PR #192, opened 2020-04-24
Added favicon (#180)

tsliaw:
This worked in placing the favicon in the public folder and getting the favicon to show, but I'm not sure the organization of what I have is appropriate, so please let me know of any changes I'll have to make. (#180)


context/app/templates/pages/base.html
```diff
@@ -9,6 +9,7 @@
     <!-- <link rel="stylesheet" href="{{ url_for('static', filename='todo-styles.css') }}"> -->
     <link rel="stylesheet" href="{{ url_for('static', filename='table.css') }}">
     <link rel="stylesheet" href="{{ url_for('static', filename='sticky-footer.css') }}">
+    <link rel="shortcut icon" image="image/ico" href="../../static/public/assets/favicon.ico">
```
mccalluc, line 12: - Incuding "shortcut" is [no longer a good practice](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types). (See the "icon" row.)
- I don't think the `image` is necessary... and if you do want a MIME type, the attribute should be `type`.
- You're trying to use a relative path from the source code, but what would `../../` mean if you were on `/`?

john-conroy: I believe you should use `url_for()` to access files in static/public similar to the stylesheets imported above the favicon

mccalluc:
@tsliaw : The original author of a PR should merge, so this is good whenever.

PR #200, opened 2020-04-28
Remove rel from ccf-ui a tag

john-conroy:
Safe to remove `rel="noopener noreferrer"` from ccf-ui link as suggested by @mccalluc


mccalluc:
@john-conroy : On second thought, I think the warnings from eslint are almost always well justified. Thank you for this change, but I'd suggest we stay with the status quo, and if the CCF browser ends up needing more from us, we can cross that bridge when we come to it.

PR #201, opened 2020-04-28
Mccalluc/0.0.12

mccalluc:
@yuanzhou : A new docker image is out: Can you deploy? Going forward, we're going to try making two releases a week..., and I'll be more active on the slack, after doing manual QA on the release, to let you know that it's good for production.

If there are aspects of this process which should change so there's less work for you, let me know?


mccalluc:
@yuanzhou : The docker image was made before the merge from master which broke the build, so you can redeploy like usual. -- I'll figure out what the build problem is and fix it on this branch. 

@tsliaw : Have you seen this?
```
10% building 0/0 modules 0 active(node:4876) UnhandledPromiseRejectionWarning: Error: Unsupported MIME type: image/x-icon
```

yuanzhou:
Will deploy on both DEV and TEST shortly.

PR #206, opened 2020-04-29
Mccalluc/fix mime bug

mccalluc:
In fixing merge conflicts in PR #192, tests started to fail, but it got merged before those were fixed. I'm not sure exactly what changed, but looking at favicons-webpack-plugin, what it's really for is when you want to check in a PNG with your source code, but have webpack generate the ICO for the build. Since we already have an ICO, we don't need to involve webpack, and can serve it just like other Django statics.

(When this is in master, we should merge master into the PRs that are out, and that should fix them.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.0.12 - In progress
 ### Added
-- Added favicon
+- Added favicon, just as Django asset.
 ### Changed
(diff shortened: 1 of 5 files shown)


tsliaw:
@mccalluc I don't see the favicon updated on my end when I pulled this branch when using ./quickstart.sh. I do see it when running docker, so is this good enough?

Issue #207, opened 2020-04-30
Easier deployments: docker tag "latest"


mccalluc:
Right now, we need to make PRs which update the docker compose configuration for deployment -- It will be easier if we just use `latest` tag so that configuration can be stable, and we can just hit a script on their side to kill the old container and start a new one.

Need scripts (?) on our side to tag images.

mccalluc:
redeploy.sh will push our own images now.

PR #215, opened 2020-05-03
Mccalluc/docker latest tags

mccalluc:
Towards #207 -- Will wait to close that until we can actually deploy `latest` ourselves. @yuanzhou : Confirm that the change to the docker compose config is what we'll want?


yuanzhou:
Looks good to me. I'll work on the deployment script that uses the latest tag as well.

Question: if this PR gets merged, do you need to me redeploy on DEV and TEST as a result?

PR #217, opened 2020-05-03
Mccalluc/new schema

mccalluc:
Copying and pasting, even if it's done by a script, isn't ideal, but it's also less trouble for now that pypi/npm/git-submodule/etc. If this just seems wrong to you, we can do something else.

On the plus side, the new schemas report much fewer errors. I should talk with Bill and get on the same page with him about what we should do with the errors that remain.


mccalluc:
@john-conroy : Please review!

Issue #202, opened 2020-04-28
On details pages, make react the default


mccalluc:
Looks good, but there are some gaps to be filled before it can be the default.
- [x] Breadcrumb links should be fixed. (Right now it says "Mock Donor ID 567 / Mock Sample ID 32546".)
- [ ] Visualizations should be filled in.
- [x] There needs to be a way to just pretty print the json... we're not actually interested in the structure as JSON per-se, so I think something like the nested rectangles on the jinja version would be good.

If the way forward on the third item isn't clear, feel free to make a stop gap, and file a new issue for design.

john-conroy:
@mccalluc Could you explain the intended functionality of the breadcrumb links?

mccalluc:
The idea is if we are on dataset, to be able to navigate "up" to the sample it came from, the organ the sample came from, and the donor. (And there might be navigation in the other direction to, but since that's a list, not sure not sure if the same idiom works.)

One larger purpose is that being able to navigate to the donor should take the burden off this page for displaying all the donor information, and being able to navigate to the sample takes off the burden for displaying sample information... Does that help?

mccalluc:
Done.

PR #219, opened 2020-05-04
John conroy/react details default

john-conroy:
Towards #202 . I'll wait to close the issue until the visualizations can be filled in.


context/app/static/js/App.jsx
```diff
@@ -40,3 +41,14 @@ export default function App(props) {
...
+App.propTypes = {
+  flaskData: PropTypes.exact({
+    entity: PropTypes.object,
+    flashed_messages: PropTypes.array,
+    provenance: PropTypes.object,
+    vitessce_conf: PropTypes.object,
+  }).isRequired,
+};
+
+export default App;
```
mccalluc, line 54: Is this a better practice than exporting at the definition?

context/app/static/js/components/Details.jsx
```diff
@@ -69,3 +57,13 @@ export default function Details(props) {
     </Container>
   );
 }
+
+Details.propTypes = {
+  /* eslint-disable react/forbid-prop-types */
+  assayMetaData: PropTypes.object.isRequired,
+  provData: PropTypes.object.isRequired,
+  vitData: PropTypes.object.isRequired,
+  /* eslint-disable react/forbid-prop-types */
```
mccalluc, line 66:
```suggestion
  /* eslint-enable */
```

context/app/static/js/components/RecursiveList.jsx
```diff
@@ -0,0 +1,83 @@
...
+                  {!isRoot && <PropertyName> {propertyName}: </PropertyName>}
+                  {Object.values(property).map((objProperty, index) => (
+                    <RecursiveList
+                  // eslint-disable-next-line react/no-array-index-key
+                      key={index}
+                      property={objProperty}
+                      propertyName={Object.getOwnPropertyNames(property)[index]}
+                      isRootChild={isRoot}
+                    />
+                  )) }
```
mccalluc, line 63: This almost matches the block above... Would it be possible to factor it out and define it up at the top of the function, and then reference it here, either bare, or wrapped in `<DataPanel>`?

context/app/static/js/components/RecursiveList.jsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function RecursiveList(props) {
+  const {
+    property, propertyName, isRoot, isRootChild,
+  } = props;
+
+
+  return (
+    /* eslint-disable no-nested-ternary */
```
mccalluc, line 29: It is getting a little hard to read. Consider handling the `!property` and `isLeaf(property)` cases as guard clauses: That would take off two layers of nesting.

context/app/static/js/components/RecursiveList.jsx
```diff
@@ -0,0 +1,83 @@
...
+  font-size: 16px;
+`;
+
+const PropertyName = styled.span`
+  font-weight: bolder;
+  font-size: 1rem;
+`;
+
+function isLeaf(property) {
+  return (typeof property === 'number' || typeof property === 'string' || typeof property === 'boolean' || isEmptyArrayOrObject(property));
```
mccalluc, line 19: consider using a set?
```suggestion
  return ( new Set(['number', 'string', 'boolean']).has(typeof property) || isEmptyArrayOrObject(property));
```

context/app/static/js/components/RecursiveList.jsx
```diff
@@ -0,0 +1,83 @@
...
+              isRootChild={isRootChild}
+            />
+          ) : (
+            isRootChild
+              ? (
+                <DataPanel propertyName={propertyName} isRootChild={isRootChild}>
+                  {Object.values(property).map((objProperty, index) => (
+                    <RecursiveList
+                    // eslint-disable-next-line react/no-array-index-key
+                      key={index}
```
mccalluc, line 46: Would this be any better?
```suggestion
                      key={objProperty}
```

context/app/static/js/helpers/functions.js
```diff
@@ -0,0 +1,9 @@
+export function isEmptyArrayOrObject(val) {
+  let ret = false;
+  if (val.constructor.name === 'Object') {
+    ret = Object.keys(val).length === 0;
+  } else if (val.constructor.name === 'Array') {
+    ret = val.length === 0;
+  }
+  return ret;
```
mccalluc, line 8: Avoid `let` and just return as soon as possible?
```suggestion
  if (val.constructor.name === 'Object') {
    return Object.keys(val).length === 0;
  }
  if (val.constructor.name === 'Array') {
    return val.length === 0;
  }
  return false;
```

context/app/static/js/App.jsx
john-conroy: More of a preference than a better practice. Not strongly opinionated on this one so if you believe I should revert the change where possible, I can.

context/app/static/js/components/RecursiveList.jsx
john-conroy: The key would then be one of an object, array, number or string. Could change the key to the 'propertyName' ?

context/app/static/js/App.jsx
mccalluc: No, I don't care one way or the other, just curious. Just leave it. Thanks!

john-conroy:
Resolved the issues @mccalluc mentioned. UI inside the expansion panels isn't great by any means, but I am weary of putting in much more time into it since it's a stop gap before the design is finished.

mccalluc:
@tsliaw : Can you look at this? Unless you find something abhorrent, I'd suggest that this be merged, and you take it as a starting point for iterations on the design.

john-conroy:
Thanks! I think those issues are best handled separately.

Issue #184, opened 2020-04-20
Segregate Donor, Sample, and Dataset search


mccalluc:
Eventually, there may be separate indexes, but https://github.com/hubmapconsortium/portal-search/pull/11 provides a stop-gap.

mccalluc:
@tsliaw : This is part of what we talked about this morning.

Issue #178, opened 2020-04-17
Faceted browsing UI treats all entities as 'dataset' when linking from the search results to the detail page


ngehlenborg:
Problem: when clicking any entry in the search results table, the link includes _dataset_ as the entity type, even if the entity is a _donor_ or _sample_.

PR #221, opened 2020-05-05
Redirect to the right type, if we are on the wrong one

### LINKED ISSUE #92: Redirect to appropriate document type
Right now, the view is determined by the type in the url.
- You can view any entity with the any of the templates by changing the url.
- If you just have a uuid, there's nothing that gets you to the right view.

Proposal:
- All pages will check the type of an entity, and redirect to the appropriate one.
- We can add an `entity` page which will only redirect.

mccalluc:
Fix #92. @ilan-gold, besides the code, does this seem like the right thing to do?
- Not a replacement for pointing people to the right url from search (#184 / #178), but seems like good thing in its own right.
- Just above the added lines, there's still a 404 if the entity type is unrecognized... Should I get rid of that? 
- Default is 302... but we could send a 301... But permanent redirects can be very annoying, if you ever get them wrong.

### DIFF
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -1,5 +1,6 @@
 from flask import (Blueprint, render_template, abort, current_app,
-                   session, flash, request, get_flashed_messages)
+                   session, flash, request, get_flashed_messages,
+                   redirect, url_for)
 
 from yaml import safe_load as load_yaml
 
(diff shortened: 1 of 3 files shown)


context/app/routes_main.py
```diff
@@ -39,6 +40,10 @@ def details(type, uuid):
     client = _get_client()
 
     entity = client.get_entity(uuid)
+    actual_type = entity['entity_type'].lower()
+    if type != actual_type:
```
ilan-gold, line 44: Something I just noticed, but isn't `type` a protected keywork in python?

mccalluc: Good catch.

PR #226, opened 2020-05-05
Mccalluc/redeploy

mccalluc:
@ilan-gold / @john-conroy : Script works for me. For now, I'm comfortable being the person-who-redeploys, but I don't need to be a single point of failure, either.

@yuanzhou : Thanks for the instructions! In the longer term, the script probably makes more sense as part of the configuration on those machines, but this works for now. There is one more envvar we need to add, and we'll still leave the production deployment to you, but you'll be hearing from us less frequently. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v0.0.14 - In progress
 ### Added
 - Redirect to the correct entity type, if we've landed on the wrong one.
+- Redeploy script.
 
 ## [v0.0.13](https://github.com/hubmapconsortium/portal-ui/tree/v0.0.13) - 2020/05/03
 ### Added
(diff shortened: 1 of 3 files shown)


yuanzhou:
@mccalluc per the script, can you confirm that you didn't have to delete the old image before pulling the new one? Otherwise, looks good to me. And I'm happy to deploy on PROD. Also link this PR to https://github.com/hubmapconsortium/devops/issues/2

mccalluc:
@yuanzhou : Right: I didn't have to delete the image... and I wouldn't want to, because most of the layers are going to be reused between versions... Over time, you will accumulate unused image layers, but I'm guessing that you'll just be migrating to a different machine before the old layers are using enough disk to be a problem.

mccalluc:
> Sorry - just saw this. Where is this run?

On your local machine: The script incorporates the ssh.

mccalluc:
@yuanzhou : 
> Also link this PR to hubmapconsortium/devops#2

That link doesn't work for me. It sounds like you're good with this, but I'll ping you on slack to ask if there's a follow-up for me.

PR #231, opened 2020-05-06
John conroy/fix docker react

### LINKED ISSUE #155: React in docker should work by design, not by accident
Follow-up to #152.

If you build by hand, there is a `static/public/bundle.js` and the react code works in docker... except it shouldn't.
- Add a docker ignore so files which are built during local development aren't accidentally incorporated in the image.
- Add npm steps to the docker file so that assets like this a built as part of the image.
- ... and once that's working and merged, tidy it up by using a multi-stage build: We shouldn't need node_modules at run time, etc.

(Assigning @jkmarx ... but just as another possible option for the next couple days: Testing would also be good.)

john-conroy:
Fixes #155 . I opted not to suppress any messages in the static build stage.


context/Dockerfile
```diff
@@ -1,10 +1,19 @@
+FROM node:12.16.3-alpine as builder
+WORKDIR /work
+COPY . .
+RUN npm install
+RUN npm run prepublish
```
mccalluc, line 5: I started #232 but didn't finish. Let's just add a TODO here since it should be a quick fix.
```suggestion
# TODO: Get rid of prepublish... wget the prov-vis schema shouldn't be necessary. See PR #232.
RUN npm run prepublish
```

context/.dockerignore
```diff
@@ -1,3 +1,7 @@
 # Be careful that credentials in development environment do not leak into image!
 **/app.conf
 **/instance/
+**/node_modules/
+app/static/public/
```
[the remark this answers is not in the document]
mccalluc: Never mind: Makes sense after reading the Dockerfile.

quick-start.sh
```diff
@@ -25,7 +25,6 @@ fi
 
 grep 'TODO' "$APP_CONF" && die "Replace 'TODO' in $APP_CONF."
 
-npm install
-npm run dev-build
+cd $CONTEXT && npm install && npm run dev-build && cd ../
```
mccalluc, line 28:
```suggestion
cd $CONTEXT && npm install && npm run dev-build && cd -
```

test.sh
```diff
@@ -40,7 +40,7 @@ end pytest
 start docker
 ./docker.sh 5001
 server_up 5001
-npm run cypress:run
+cd context && npm run cypress:run && cd ../
```
mccalluc, line 43:
```suggestion
cd context && npm run cypress:run && cd -
```

john-conroy:
I think this works for now. I'll monitor the situation moving forward.

PR #244, opened 2020-05-11
Allow different filters and columns depending on type

### LINKED ISSUE #178: Faceted browsing UI treats all entities as 'dataset' when linking from the search results to the detail page

### LINKED ISSUE #184: Segregate Donor, Sample, and Dataset search

mccalluc:
Fix #184. Fix #178. In time, this can be moved into the webpacked code, but not needed for this one issue.

@tsliaw : Let me know if you have questions about how you would use this to fill in the details for different types, or if it's good, approve.

### DIFF
--- a/context/app/templates/pages/search.html
+++ b/context/app/templates/pages/search.html
@@ -17,6 +17,111 @@
 <script src="https://unpkg.com/@hubmap/portal-search@v0.0.4/umd/@hubmap/portal-search.min.js"></script>
 
   <script>
+    const commonFilters = [
+      {
+        type: 'RefinementListFilter',
+        props: {
+          id: 'created_by_user_displayname',
+          title: 'Creator',
+          field: 'created_by_user_displayname.keyword',
+          operator: 'OR',
+          size: 5
+        }
+      },
+      {
+        type: 'RefinementListFilter',
+        props: {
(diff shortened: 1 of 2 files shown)


context/app/templates/pages/search.html
```diff
@@ -17,6 +17,111 @@ <h1 class="no-content">{% block title_text %}Search{% endblock %}</h1>
...
+      }
+    ];
+    const filtersByType = {
+      '': commonFilters,
+      donor: commonFilters,
+      sample: commonFilters,
+      dataset: commonFilters
+    };
+    const resultFieldsByType = {
+      '': ['description', 'status', 'entity_type'],
```
tsliaw, line 119: Is there a reason why the fields you chose to display in the table are these ones? I was asking because it seems there's a lack of description for a lot of the entries and that status seemed to only apply to datasets.

mccalluc: No reason at all. Please do find a better set of fields!

Issue #223, opened 2020-05-05
Import field documentation from ingest-validation-tools


mccalluc:
There are nice explanations of each of the fields there... We should use them.

mccalluc:
Done. In the UI.

PR #246, opened 2020-05-11
Mccalluc/expose fields

mccalluc:
Towards #223. Demonstrate how we could pull some information from submodules.

@john-conroy and @tsliaw : Code review, and give some thought to how this information would be used in UI. Maybe there's a flask endpoint that exposes all the name-description dicts as JSON? or incorporate them in the webpack build?

Exposing the submission docs is a lower priority (Viewing them on github is fine for now), and they might take a different approach.


john-conroy:
> a flask endpoint that exposes all the name-description dicts as JSON

Sounds fine. I'm approving for now, but I'll think about it more early next week.

PR #251, opened 2020-05-11
John conroy/home search to react

### LINKED ISSUE #171: unpkg is unresponsive: need to host our own packages
The request for portal-search returns "Rate exceeded."

john-conroy:
Create components to mimic the existing Home/Search templates to get us to a baseline of everything except for error pages inside our react app. Also moves dependencies from unpkg to node modules and fixes #171 and #198 .


context/app/static/js/Routes.jsx
```diff
@@ -0,0 +1,41 @@
...
+import React from 'react';
+import NoticeAlert from './components/NoticeAlert';
+import Details from './components/Details';
+import Home from './components/Home';
+import Search from './components/Search/Search';
+
+function Routes(props) {
+  const { flaskData } = props;
+  const {
+    flashed_messages, entity, provenance, vitessce_conf,
```
mccalluc, line 11: I'd say these should be renamed on assignment to be consistent with JS naming conventions? But not 100% sure it's worth the trouble, either. Thoughts?

context/app/static/js/Routes.jsx
```diff
@@ -0,0 +1,41 @@
...
+  } = flaskData;
+  const urlPath = window.location.pathname;
+
+  if (urlPath.indexOf('/browse') > -1) {
```
mccalluc, line 15:
```suggestion
  if (urlPath.startsWith('/browse')) {
```

context/app/static/js/Routes.jsx
```diff
@@ -0,0 +1,41 @@
...
+    );
+  }
+
+  if (urlPath === '/') {
+    return (
+      <Home />
+    );
+  }
+
+  if (urlPath.indexOf('/search') > -1) {
```
mccalluc, line 34:
```suggestion
  if (urlPath === '/search') {
```

context/app/static/js/index.jsx
```diff
@@ -9,11 +9,18 @@ import App from './App';
 import Footer from './Footer';
 import Header from './Header';
 
-if (window.location.href.search('/browse/') > -1) {
+const urlPath = window.location.pathname;
+
+if (urlPath.indexOf('/browse') > -1) {
```
mccalluc, line 14: Seems to overlap with Routes.jsx -- Are they both necessary?

john-conroy: Both are necessary so long as there are situations, now just the error pages, where we need to only attach the header/footer and not any content.

context/app/static/js/Routes.jsx
john-conroy: I'll create an issue and revisit.

john-conroy:
The `core_props` variable is repeated throughout `routes_main.py` which isn't ideal, but if it's fine for now I can open an issue and revisit once we're past the demo.

context/app/routes_main.py
```diff
@@ -31,7 +31,8 @@ def _get_client():
 
 @blueprint.route('/')
 def index():
-    return render_template('pages/index.html', types=types)
+    core_props = {'endpoints': {'esEndpoint': current_app.config['ELASTICSEARCH_ENDPOINT']}}
```
mccalluc, line 34: Prefer to spell things out, and generally keep names consistent.
```suggestion
    core_props = {'endpoints': {'elasticsearchEndpoint': current_app.config['ELASTICSEARCH_ENDPOINT']}}
```

context/app/routes_main.py
```diff
@@ -65,17 +66,19 @@ def details(type, uuid):
...
-    }
+    })
     return render_template(
-        template, type=type, uuid=uuid,
-        title_text='TODO: title_text',
-        flask_data=props
+        template,
+        type=type,
+        uuid=uuid,
+        flask_data=core_props
```
mccalluc, line 81: `core_props` could be moved out to the top level, and we don't need to mutate it.
```suggestion
        flask_data={
            'flashed_messages': flashed_messages,
            'entity': entity,
            'provenance': provenance,
            'vitessce_conf': client.get_vitessce_conf(),
            **core_props
        }
```

context/app/routes_main.py
```diff
@@ -95,10 +98,11 @@ def details_ext(type, uuid, ext):
 
 @blueprint.route('/search')
 def search():
+    core_props = {'endpoints': {'esEndpoint': current_app.config['ELASTICSEARCH_ENDPOINT']}}
```
mccalluc, line 101:
```suggestion
```
Move to top level

context/app/routes_main.py
john-conroy: I initially did that but I was getting a flask error similar to `RuntimeError: working outside of application context` .

context/app/static/js/Routes.jsx
john-conroy: Created #258

PR #257, opened 2020-05-12
Mccalluc/add facets

mccalluc:
My preference is actually that #251 be wrapped up, this PR be closed un-merged, and then a new one adds these facets in their new home: I don't want to introduce merge conflicts there for something as trivial as this.

Longer term, we shouldn't have hardcoded lists like this littered across the codebase. PR #246 would be the first step... and then? Could imagine python code generating and injecting JS globals with this information... Or maybe the webpack build could be configured to parse and include that yaml in the submodules?


mccalluc:
Replaced by #288. Closing.

PR #259, opened 2020-05-13
Mccalluc/constrained samples datasets links

### LINKED ISSUE #254: Show datasets for a given donor

### LINKED ISSUE #255: Show samples from a given donor

mccalluc:
This demonstrates how we could link down to samples and datasets. The best route might be to review my approach at a high level here, and then I will look at whether the merge conflicts are tractable, or I should make a new PR.

(This introduces a new facet in search which is just  ancestor UUIDs. I don't see an option right now to support a facet in the url without exposing it in the sidebar, but I don't think that should block.)

Fix #255, Fix #254.


mccalluc:
Replaced by #261. Closing.

PR #273, opened 2020-05-13
Vitessce Integration Pt 1

ilan-gold:
There is a lot hardcoded and/or wonky stuff here that I need to make issues for but it should "work."

Issues to follow up on:
- CSS from Vitessce is overriding the CSS from here in a few ways including layout and theme.  This only happens when using the CODEX data, uuid 9e04dbf8d911c2c875fc961d5b25286a, (i.e Spatial Vitessce config), but the controller has a Material UI theme.
- We only have one CODEX dataset and it has a hashed path for the offsets.  We need to follow up on this and the directory structure in general.
- The output for scRNASEQ is wrong, hence the commented out `FACTORS`.
- Make the little menu for layers scrollable while work is done for a more robust multi-dataset solution in Vitessce.

Once you have a chance to review/play around with this, we'll add whatever issues you find to the queue if they can't be resolved and then I'll open up issues for everything before merging.


ilan-gold:
I'll fix travis.  Seems to work locally.

ilan-gold:
https://github.com/gregberge/svgr/issues/448 Seems like the build issue is not just me.

context/app/api/client.py
```diff
@@ -165,45 +238,99 @@ def get_provenance(self, uuid):
...
+        return base_url + '?' + token_param
+
+    def _build_image_schema(self, rel_path, uuid):
+        schema = {}
+        schema['name'] = Path(rel_path).name
+        schema['type'] = 'ome-tiff'
+        schema['url'] = self._build_assets_url(rel_path, uuid)
+        schema['metadata'] = {}
+        offsets_path = Path(CODEX_OFFSETS_PATH) / \
+            Path(Path(rel_path).name.replace('ome.tiff', 'offsets.json'))
```
mccalluc, line 298:
```suggestion
            rel_path.replace('ome.tiff', 'offsets.json')
```
Assuming `rel_path` is a string.

context/app/api/client.py
```diff
@@ -165,45 +238,99 @@ def get_provenance(self, uuid):
...
+        schema['metadata']['omeTiffOffsetsUrl'] = self._build_assets_url(str(offsets_path), uuid)
+        return schema
+
+    def _build_image_layer_datauri(self, rel_path, uuid):
+        image_layer = {}
+        image_paths = [
+            str(rel_path).replace(
+                '#CODEX_TILE#', f'R{str(r).zfill(3)}_X{str(x).zfill(3)}_Y{str(y).zfill(3)}'
```
mccalluc, line 306: Is this structure something that part of standard? Or just for the demo? Or somewhere in between?

Name could be more clear... perhaps `path_template` for `rel_path`?

context/app/api/client.py
```diff
@@ -165,45 +238,99 @@ def get_provenance(self, uuid):
...
+        # Can there be more than one of these?  This seems like a fine default for now.
+        assay_type = entity['data_types'][0]
+        uuid = entity['uuid']
+        conf = ASSAY_CONF_LOOKUP[assay_type]['base_conf']
+        files = ASSAY_CONF_LOOKUP[assay_type]['files_conf']
+        layers = [self._build_layer_conf(file, uuid, assay_type) for file in files]
+
+        conf['layers'] = layers
+        conf['name'] = uuid
+        return conf
```
mccalluc, line 336: Can you pull all this new code into a file by itself, and maybe there's only one public function?

And can you add doctests on these functions? Shouldn't be hard to write, and it would help to understand what each part does.

test.sh
```diff
@@ -7,7 +7,7 @@ die() { set +v; echo "$*" 1>&2 ; sleep 1; exit 1; }
 
 server_up() {
   TRIES=0
-  MAX_TRIES=60
+  MAX_TRIES=500
```
mccalluc, line 10: If you're waiting 9 minutes for the server to come up, something is really wrong. (The 60 seems over-generous, actually.)

ilan-gold:
Travis issues resolved - bad `package-lock.json` + source map being too slow.

context/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM node:12.16.3-alpine as builder
 WORKDIR /work
 COPY . .
+RUN apk add --no-cache python git
```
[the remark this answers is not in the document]
mccalluc: Actually, shouldn't this like come before the `COPY . .`? The install layer will be relatively stable, our source code will be changing.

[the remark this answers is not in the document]
ilan-gold: HiGlass has multiple git dependencies: https://github.com/higlass/higlass/blob/230d8904d33a939f15039bdc8def9e4b56a4fc6b/package.json#L46-L97

context/app/api/client.py
```diff
@@ -165,45 +238,99 @@ def get_provenance(self, uuid):
...
+                            'mapping': 'UMAP',
+                            'view': {
+                                'zoom': 4,
+                                'target': [0, 0, 0]
+                            }
+                        },
+                        'x': 0, 'y': 0, 'w': 12, 'h': 2
+                    },
+                ]
+            }
```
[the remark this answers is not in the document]
ilan-gold: Yes.

test.sh
ilan-gold: The issue I was having was that the build step seems considerably longer now because of `vitessce`'s dependencies (all of the gl libraries, higlass etc.).  I'll tone this down though.

context/app/api/client.py
ilan-gold: I think it's not going to always be a template.  It is more frequently going to actually be a relative path (CODEX is the only one that I know of - the scRNASEQ is not a template but the actual relative path.  This is what the ES response calls it).

context/app/api/client.py
```diff
@@ -165,45 +238,99 @@ def get_provenance(self, uuid):
...
+        else:
+            return self._build_vitessce_conf(entity)
+
+    def _build_assets_url(self, rel_path, uuid):
+        base_url = urllib.parse.urljoin(
+            current_app.config['ASSETS_ENDPOINT'],
+            str(Path(uuid) / Path(rel_path))
```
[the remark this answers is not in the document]
ilan-gold: I liked it for dealing with slashes but I'll change.

context/app/api/client.py
ilan-gold: Path extracts the name.

PR #275, opened 2020-05-13
Sort by last modified. Fix #272

### LINKED ISSUE #272: Use modification date to when sorting results

mccalluc:
Tiffany, it looks like this right now... which may not be what we want for the long term? I don't know.
Screen Shot 2020-05-13 at 3 22 33 PM

If you don't see big problems, I'd suggest approving, and filing a follow-up issue for yourself, if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v0.0.16 - In progress
 ### Added
+- Sort by last_modified.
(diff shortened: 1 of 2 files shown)


mccalluc:
@tsliaw : Thanks! I'm not sure what the reverse sort would be: "Earliest Modified"? Can you file a new issue to resolve that.

Issue #263, opened 2020-05-13
Search facets


mccalluc:
### Donor:
- Center
- Creator
- Histograms for Age and BMI
### Sample:
- Organ
- Specimen Type
- Group
- Creator

(Donor fields below)
### Dataset
- How do we distinguish uploaded vs. processed?
- ??? Uncertain ???

(Sample and Donor fields below)

cc @tsliaw - I may or may not get started on this today.

mccalluc:
Histograms for Age and BMI just came in. Refinements can be new issues.

PR #288, opened 2020-05-14
Mccalluc/search facets

### LINKED ISSUE #205: Clean up display of specimen type facet

mccalluc:
These should reflect what Nils wants... organizationally, I think this is an ok scheme, separating enums, from the per-type config, from the JSX.

Towards #263.
Fix #205.


ilan-gold:
Or rather, why are some left out?

mccalluc:
@ilan: For datatypes, they're not actually giving us the enumerated values in the spec: Instead of doing multivalued fields the right way, they're returning a string containing a JSON array.

ilan-gold:
Ok.

PR #290, opened 2020-05-15
Fix sizing bug.

ilan-gold:
This fixes what I mentioned here: https://github.com/hubmapconsortium/vitessce/issues/556#issuecomment-628925071.  Not sure about the ingest tools submodule stuff.  @mccalluc Should I remove that?


ilan-gold:
@mccalluc it was causing the build to fail so I forced push to remove it.

ilan-gold:
It's not that it was too tall before, it's that the scatter plot was too short so I was correcting for that.  Here are screenshots with 6 * 75 = 450 pixel size Vitessce:

Screen Shot 2020-05-15 at 11 37 10 AM
Screen Shot 2020-05-15 at 11 36 48 AM

PR #298, opened 2020-05-15
Update scRNASEQ

ilan-gold:
This prepares us for the new scRNASEQ data that is coming with cluster assignments.  Since there will now be old data, we need to make sure that we only return a Vitessce configuration for the new data (hence the subset operation).  This should be the last update for a little while.  Marking this as a draft until we actually have the data (Joel says later tonight).


ilan-gold:
The data for UUID's Joel gave me look good.

context/app/api/vitessce.py
```diff
@@ -63,11 +64,13 @@
         "base_conf": SCATTERPLOT,
         "files_conf": [
             {
-                "rel_path": "dim_reduced_clustered/dim_reduced_clustered.json",
+                "rel_path": "cluster-marker-genes/output/cluster_marker_genes.cells.json",
```
mccalluc, line 67: pull `cluster-marker-genes/output/cluster_marker_genes` into variable.

context/app/api/vitessce.py
```diff
@@ -135,21 +141,30 @@ def _build_vitessce_conf(self):
...
-        ]
+        file_paths_expected = [file["rel_path"] for file in files]
+        file_paths_found = [file["rel_path"] for file in self.entity["files"]]
+        # Codex assay needs to be built up.
+        # We need a better way to handle it but for now, this is ok.
+        if self.assay_type != CODEX_ASSAY:
+            # We need to check that the files we expect actually exist.
+            # This is due to the volatility of the datasets.
+            if not set(file_paths_expected).issubset(set(file_paths_found)):
+                return {}
```
mccalluc, line 155: Maybe a warning to the log? I'm not sure where the boundary is between a negative that is entirely normal, and an expected state that should not be ignored silently.

context/app/api/vitessce.py
```diff
@@ -254,7 +269,7 @@ def _build_assets_url(self, rel_path, uuid):
         else:
             assets_endpoint = MOCK_URL
         base_url = urllib.parse.urljoin(
-            assets_endpoint, str(Path(uuid) / Path(rel_path))
+            assets_endpoint, str(Path(self.uuid) / Path(rel_path))
```
mccalluc, line 272:
```suggestion
            assets_endpoint, f'{self.uuid}/{rel_path}'
```
I think pathlib is great for filesystem paths, when there's a chance you'd do file-systemy things with it, but if you really just want to put a `/` between things, it's fine to do that, I think.

PR #304, opened 2020-05-17
John conroy/home page call to action

john-conroy:
Adding the requested call to action to the home page for the demo.


john-conroy:
Fixed and noted for the future.

PR #305, opened 2020-05-17
John conroy/fix prov table derived links

john-conroy:
Fix conditional logic in provenance table to stop creating links to derived donors.

### DIFF
--- a/context/app/static/js/components/Detail/ProvTable.jsx
+++ b/context/app/static/js/components/Detail/ProvTable.jsx
@@ -60,7 +60,7 @@ function ProvTable(props) {
             )
         }
             {
-          typesToSplit[i] === entity_type && type !== 'Donor'
+          typesToSplit[i] === entity_type && typesToSplit[i] !== 'Donor'
             ? (
               <ListItemLink href={`/search?ancestor_id[0]=${uuid}&entity_type[0]=${typesToSplit[i]}`}>
                 <ListItemText primary={`Derived ${typesToSplit[i]}s`} />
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/Detail/ProvTable.jsx
```diff
@@ -60,7 +60,7 @@ function ProvTable(props) {
             )
         }
             {
-          typesToSplit[i] === entity_type && type !== 'Donor'
+          typesToSplit[i] === entity_type && typesToSplit[i] !== 'Donor'
```
mccalluc, line 63: perhaps
```suggestion
          typesToSplit[i] === entity_type && entity_type !== 'Donor'
```
but whichever you think is more clear.

john-conroy: `entity_type` refers to the type for the detail page that is being viewed. The suggested change would cause a donor detail page to not render derived samples or datasets.

john-conroy: I could rename the variable to be more clear?

mccalluc: huh: Then I'm confused: From the first part, we know that that `typesToSplit[i] === entity_type`.... so in the second we should be able to substitute `entity_type` for `typesToSplit[i]` without changing the semantics? I may be missing something.

john-conroy: You're right, my mistake.

PR #307, opened 2020-05-18
Mccalluc/search cleanup

mccalluc:
- Just a single search directory.
- Detailed proptypes.
- pull code translations from yaml.
- translate results as well as filters.


context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -68,24 +73,8 @@ function makeTableComponent(resultFields, detailsUrlPrefix, idField) {
...
-    }],
-    hiddenFilterIds = [],
-    searchUrlPath = '_search',
-  } = props;
-  const resultFieldIds = (resultFields.map((field) => field.id)).concat(idField);
-  const searchkit = new SearchkitManager(apiUrl, { httpHeaders, searchUrlPath });
-
-  function MaskedSelectedFilters() {
+function makeMaskedSelectedFiltersComponent(hiddenFilterIds) {
+  return () => {
```
ilan-gold, line 77: Is there a cleaner way to do this rather than returning a nameless function?  It's tough to wrap your head around.

ilan-gold: Maybe just leave a comment.  Or name the wrapping function so it's clear that it's a component (unless this a paradigm you've seen before!  Then maybe just a note :) )

context/app/static/js/components/Search/utils.js
```diff
@@ -1,12 +1,32 @@
-export function field(id, name) {
-  return {
+import fromEntries from 'fromentries';
+import searchDefinitions from '../../../../../search-schema/data/definitions.yaml';
```
john-conroy, line 2: Is there a way to clean up paths to the submodules going forward instead of relative paths?

john-conroy, line 2: Is there a way to clean up paths to the submodules going forward instead of relative paths?

context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -68,24 +73,8 @@ function makeTableComponent(resultFields, detailsUrlPrefix, idField) {
...
-      defaultOption: true,
-    }],
-    hiddenFilterIds = [],
-    searchUrlPath = '_search',
-  } = props;
-  const resultFieldIds = (resultFields.map((field) => field.id)).concat(idField);
-  const searchkit = new SearchkitManager(apiUrl, { httpHeaders, searchUrlPath });
-
-  function MaskedSelectedFilters() {
+function makeMaskedSelectedFiltersComponent(hiddenFilterIds) {
```
john-conroy, line 76: PascalCase for components names.

context/app/static/js/components/Search/SearchWrapper.jsx
john-conroy: Can we make this a component?

mccalluc:
@ilan-gold , @john-conroy : You both rightly commented on `makeMaskedSelectedFiltersComponent`: I'll add a comment there, but there is probably/hopefully a better idiom. It itself is not a component, it's a function that *returns* a component function... Why? The list of fields to hide needs to be consulted, but is not a prop of the component itself, whose prop interface is fixed.

This will go in a comment. Thanks!

context/app/static/js/components/Search/utils.js
mccalluc: Looks like I could set up an alias in webpack. Filed #563. Thanks.

context/app/static/js/components/Search/SearchWrapper.jsx
mccalluc: I agree, it is confusing, but not sure about better approach. Will comment, and I've filed #564.

(This function-function was there before, it was just buried inside _another_ function... so I don't think this PR is making it any worse.)

mccalluc: (Never mind... it's easy. Sorry for the noise.)

mccalluc:
> Looks good - this doesn't have to go in the code anywhere, but where is this yaml coming from?

@ilan-gold : The yaml is coming from the `search-schema` submodule... This is one of the reasons I really wanted to try the submodules.

mccalluc:
(And thank you both for pushing on the weird function-function!)

PR #317, opened 2020-05-18
only validate in development mode

### LINKED ISSUE #271: Add option to turn off display of warnings for demo.
Not sure if a new config setting or url parameter is better. Competing needs: Some people just want to ignore them / some people think things won't get fixed unless there are big error messages.

mccalluc:
Started flask without `FLASK_ENV=development` and there are no warnings ... start it with the envvar, and there are warnings. Fix #271

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## v0.0.19 - In Progress
 
-...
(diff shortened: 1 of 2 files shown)


john-conroy:
Seems to work as desired.

PR #326, opened 2020-05-18
John conroy/fix sample organ logic

### LINKED ISSUE #320: Fix logic for displaying organ types in sample detail
Duplicate organ fields are displayed when both origin sample and organ at top level exist.

john-conroy:
Fixes #320 .

### DIFF
--- a/context/app/static/js/components/SampleDetail.jsx
+++ b/context/app/static/js/components/SampleDetail.jsx
@@ -27,7 +27,7 @@ function SummaryData(props) {
     <>
       {organ && organ.length
         ? (<SummaryItem>{organ}</SummaryItem>) : null}
-      {origin_sample.organ && origin_sample.organ.length
+      {origin_sample.organ && origin_sample.organ.length && !organ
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/SampleDetail.jsx
```diff
@@ -27,7 +27,7 @@ function SummaryData(props) {
     <>
       {organ && organ.length
         ? (<SummaryItem>{organ}</SummaryItem>) : null}
-      {origin_sample.organ && origin_sample.organ.length
+      {origin_sample.organ && origin_sample.organ.length && !organ
         ? (<SummaryItem>{origin_sample.organ}</SummaryItem>) : null}
```
mccalluc, line 31: (With all of this, do we worry about details now, or, more likely, come back later and try to have something more uniform to work against?)

- Consider pulling the complicated logic out of the JSX and defining an `organ_name` above.
- For parallelism, should it be `organ && organ.length && !origin_sample.organ` for the first case? Or there are both, does the first take precedence?

john-conroy: Will revisit, thanks!

PR #338, opened 2020-05-26
"Triage" template

mccalluc:


mccalluc:
(Going ahead and merging so I can reference the link, but still worth a look.)

PR #347, opened 2020-05-26
Build changelog from individual files to avoid merge conflicts

### LINKED ISSUE #332: Better changelog pattern with multiple devs?
When it was just me, doing the changelog by hand was fine... but now we get merge conflicts all the time. It's a headache. Is the a better practice we should follow? cc @ilan-gold / @john-conroy .

mccalluc:
Fix #332. Inspired by, but simpler than, the pattern John linked to. Explanation of changes is in README. Does this seem like good idea to people?

### DIFF
--- a/CHANGELOG-avoid-merge-conflicts.md
+++ b/CHANGELOG-avoid-merge-conflicts.md
@@ -1,0 +1,1 @@
+- Concatenate separate files to build changelog on release; This will help us minimize merge conflicts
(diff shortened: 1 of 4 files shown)


john-conroy:
Is there an easy way to require at least one new changelog file for each PR?

mccalluc:
travis is checking again. Thanks for the feedback! There may be hick-ups, but I think this will be manageable.

PR #352, opened 2020-05-27
Mccalluc/fix 500s hopefully

### LINKED ISSUE #335: portal returning "Internal Server Error" (500) for certain datasets
Problem: the portal UI returns a 500 error for certain datasets

Reproduce:
Access https://portal.test.hubmapconsortium.org/browse/dataset/662ef4c9807517df51554a6fa6ba33e3
(Dataset ID HBM657.ZRBM.357, LC-MS-untargeted submitted by Vanderbilt TMC, status new). 

Viewing the JSON for this dataset works: https://portal.test.hubmapconsortium.org/browse/dataset/662ef4c9807517df51554a6fa6ba33e3.json

I have found this problem with every single Vanderbilt TMC dataset (independently of the status) as well as several other data sets.

mccalluc:
Fix #335 : Move Prov API hits to client-side, so if that API fails, we just get one empty component, instead of a 500 for the whole page.

### DIFF
--- a/context/.eslintrc.yml
+++ b/context/.eslintrc.yml
@@ -10,6 +10,7 @@ settings:
   react:
     version: detect
 rules:
+  no-console: ["error", { allow: ["warn", "error"] }]
   react/jsx-filename-extension: [0]  # Imports don't work if I change extension.
   react/sort-comp: [0]  # Non-alphabetical groupings can make more sense.
   react/jsx-one-expression-per-line: [0]  # Makes punctuation after tab awkward.
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -121,49 +121,14 @@ class ApiClient():
             raise Exception(f'UUID not unique; got {len(hits)} matches')
         entity = hits[0]['_source']
 
+        if 'metadata' not in entity:
+            entity['metadata'] = {}
+
         # TODO: Reenable with new document structure.
         # entity['created'] = _format_timestamp(entity['provenance_create_timestamp'])
         # entity['modified'] = _format_timestamp(entity['provenance_modified_timestamp'])
         return entity
 
-    def get_provenance(self, uuid):
-        # TODO: When the API is fixed, only use this when is_mock.
-
-        if self.is_mock:
-            return {
-                'prefix': {
-                    'ex': 'http://example.com#',
-                    'prov': 'http://www.w3.org/ns/prov#',
-                    'hubmap': 'https://hubmapconsortium.org'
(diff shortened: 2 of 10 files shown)


context/.eslintrc.yml
```diff
@@ -10,6 +10,7 @@ settings:
   react:
     version: detect
 rules:
+  no-console: ["error", { allow: ["warn", "error"] }]
```
mccalluc, line 13: I would be in favor of making less of an effort to handle all the variants we see (since it's unclear now which are intended, and which are accidents)... but when there is something weird log an error close to where it happens.

context/app/api/client.py
```diff
@@ -121,49 +121,14 @@ def get_entity(self, uuid):
             raise Exception(f'UUID not unique; got {len(hits)} matches')
         entity = hits[0]['_source']
 
+        if 'metadata' not in entity:
+            entity['metadata'] = {}
```
mccalluc, line 125: I think John might have done handled this case on the JSX side, too. Should this go away?

context/app/static/js/components/Detail/ProvTabs.jsx
```diff
@@ -36,13 +37,38 @@ function TabPanel(props) {
...
+          console.error('Prov API failed', response);
+          return null;
+        }
+        return response.json();
+      })
+      .then(
+        (responseProvData) => {
+          setProvData(responseProvData);
+        },
+      );
```
john-conroy, line 69: I lean async/await over promise chains where possible. Do we care about being consistent with that?

john-conroy: We can figure this out later, but some sort of progress component while loading?

context/app/api/client.py
john-conroy: I think so.

context/.eslintrc.yml
john-conroy: I agree it makes sense for now.

john-conroy:
Looks good and we can figure out using a provider for endpoints.

context/app/static/js/components/Detail/ProvTabs.jsx
mccalluc: async implemented. Filed #365 for please wait. (It's typically below the fold, and loads fast, so I don't think this is pressing.)

context/app/api/client.py
mccalluc: removed.

PR #355, opened 2020-05-27
John conroy/dataset status

john-conroy:
Add status icon to dataset detail pages.


context/app/static/js/components/Detail/StatusIcon.jsx
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  if (['UNPUBLISHED', 'DEPRECATED'].includes(status)) {
+    return 'warning';
+  }
+
+  if (status === 'PUBLISHED') {
+    return 'success';
+  }
+
```
mccalluc, line 28:
```suggestion

  console.warn('Invalid status', status);
```
If people aren't seeing the validation warnings, we still want to communicate that there is problem.

context/app/static/js/components/Detail/StatusIcon.jsx
```diff
@@ -0,0 +1,45 @@
...
+  return null;
```
mccalluc, line 29:
```suggestion
  return 'error';
```
perhaps? It is an error, and not having to deal with the null downstream makes our lives easier.

context/app/static/js/components/Detail/StatusIcon.jsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function StatusIcon(props) {
+  const { status: irregularCaseStatus } = props;
+  const status = irregularCaseStatus.toUpperCase();
+  const color = getColor(status);
+
```
mccalluc, line 36:
```suggestion
```

PR #364, opened 2020-05-27
John conroy/detail table of contents

john-conroy:
Please give feedback on the table of contents functionality across entity types. How should we handle short pages (commonly donors) which don't need to be scrolled?

The details pages now exceed the intended maximum width. I think I'll fix it in a separate PR since this is an issue throughout the application


context/app/static/js/components/DatasetDetail.jsx
```diff
@@ -61,45 +48,46 @@ function DatasetDetail(props) {
...
+    uuid,
+    data_types,
+    origin_sample,
   } = assayMetadata;
+
+  const displayViz = 'name' in vitData;
+  const displayProtocol = (portal_uploaded_protocol_files || protocol_url);
+  const displayMetadataTable = (metadata && metadata.metadata);
+  const displayFiles = (files && files.length);
+  const displayDag = (metadata && (metadata.dag_provenance || metadata.dag_provenance_list));
```
mccalluc, line 68: Preference for boolean variables that sound like questions... Often `is...`, but in this case maybe `should...` would work.

context/app/static/js/components/DatasetDetail.jsx
```diff
@@ -61,45 +48,46 @@ function DatasetDetail(props) {
...
-          )
-          : null}
-      </SpacedContainer>
-    </FlexContainer>
+    <DetailLayout
+      displayViz={displayViz}
+      displayProtocol={displayProtocol}
+      displayMetadataTable={displayMetadataTable}
+      displayFiles={displayFiles}
+      displayDag={displayDag}
```
mccalluc, line 76: .... and similarly for prop names.

context/app/static/js/components/Detail/DetailLayout.jsx
```diff
@@ -0,0 +1,79 @@
...
+    displayProtocol,
+    displayMetadata,
+    displayMetadataTable,
+    displayFiles,
+    displayDag,
+    flashed_messages,
+    children,
+  } = props;
+
+  const updateSections = () => {
```
mccalluc, line 32: From the outside world, it's not really an update... may just
```suggestion
  const getSections = () => {
```

context/app/static/js/components/Detail/DetailLayout.jsx
```diff
@@ -0,0 +1,79 @@
...
+    const sections = new Map([
+      ['summary', { text: 'Summary', hash: 'summary' }],
```
mccalluc, line 34: Make a little function for this? If you really want dagProv to be a special case, ok, but in all the others, it would be hard to spot typos.

context/app/static/js/components/Detail/TableOfContents.jsx
```diff
@@ -0,0 +1,189 @@
...
+  headings.forEach((item) => {
+    itemsWithNode.push({
+      ...item,
+      node: document.getElementById(item.hash),
+    });
+  });
+  return itemsWithNode;
+}
+
+function useThrottledOnScroll(callback, delay) {
```
mccalluc, line 67: There's a lot of complexity here. At first look, I would prefer just to use the browser's built-in handling for anchor links... but if that's not good enough, add a comment explaining why. Thanks!

john-conroy: Nils wanted this functionality specifically.

context/app/static/js/components/DatasetDetail.jsx
```diff
@@ -60,50 +47,47 @@ function DatasetDetail(props) {
...
+    protocol_url,
+    portal_uploaded_protocol_files,
+    metadata,
+    files,
+    uuid,
+    data_types,
+    origin_sample,
   } = assayMetadata;
+
+  const shouldDisplaySection = {
```
mccalluc, line 64: Cool: I was thinking of a lot of separate `should`s, but this works well.

PR #376, opened 2020-06-01
John conroy/donor metadata units

john-conroy:


john-conroy:
I noticed that as well. Everything is taken directly from the elasticsearch response.

Issue #280, opened 2020-05-13
Document JS dev process


mccalluc:
@john-conroy : Can you add a section to the README about the dev process we should have for JS?
- Install prettier plugin for you IDE?
- I've just been `npm run dev-build`ing, but setting up a rebuild on change shouldn't be impossible?
- How tests can be organized. (Same directory / separate directory?)
- Other things?

ilan-gold:
Are we using prettier?  +1 For that, I like it a lot.

john-conroy:
We have yet to agree upon prettier. I am in favor of it, but in the meantime I've been using an eslint extension to format.

john-conroy:
Additional documentation was added.

PR #377, opened 2020-06-02
John conroy/prettier

john-conroy:
Towards #280 . Needed to install node v10.0+ in travis for prettier to work.


john-conroy:
It's just failing due to the new prettier errors. Chuck do you want to take a look at the changes before I go ahead and reformat everything?

context/.prettierrc.yml
```diff
@@ -0,0 +1,5 @@
+semi: true
+trailingComma: "all"
+singleQuote: true
+printWidth": 70
```
mccalluc, line 4:
```suggestion
printWidth: 70
```

mccalluc: Also, would you like this to be higher? Perhaps look at the reformatted code and see if it's doing too much wrapping?

john-conroy: Yeah, it was much too drastic. I increased it to 120.

john-conroy:
Adding prettier required some vscode workspace settings to get eslint to work properly. I think it makes sense to include those workspace specific settings in version control.

PR #385, opened 2020-06-02
ccf integration

mccalluc:
Fix #338. 
- @bherr2 : Should we change the text of the link from "CCF-UI" to "CCF-EUI"?
- or, conversely, should I change the path of the link itself?
- This is still opening in a new tab: If you have a different preference, that can change.


bherr2:
I think as you have it will be good (I'm checking with Katy to see her preference). When we have a reciprocol link setup, we can have it not open in a new window.

bherr2:
Ok, I talked with Katy and the preference is to have the link text be "CCF" and the link url be /ccf-eui. Other than that, I think we're good to go!

mccalluc:
@bherr2 / @ilan-gold : Approve if this looks good. Thanks!

PR #398, opened 2020-06-04
Mccalluc/assay type docs

mccalluc:
Link from the details pages to descriptions of all the assays, with an anchor link going to the right one... though for most datasets right now, the assay types do not match what Bill assures me we'll see.

@tsliaw : Some in-page information would be better, but right now the codes just aren't consistent enough to support that. Going to a page, and and just not having a matching anchor, is a more graceful fallback.

@cebriggs7135 : This is where these descriptions will live. I'll follow up with another PR in ingest-tools.

@shirey : FYI - This won't work well until we have the right codes. No rush, but if we should aim at something other than the TSV for release, please let me know.


john-conroy:
So the anchor tags in `assay.md` are a work in progress?

mccalluc:
Yeah... They match the TSV which should be official, but the actual data doesn't match that right now.

PR #401, opened 2020-06-04
Mccalluc/no cypress for build

### LINKED ISSUE #310: Shouldn't need to download Cypress.io for Docker build?
Maybe it's not marked as a dev dependency?

mccalluc:
<strike>Towards</strike> Fix #310. @john-conroy : How does this feel to you? I'm reloading cypress with every run of push.sh, but maybe we should just tolerate that. Did some googling to see if there was a way to distinguish dev dependencies from test, but didn't find anything. Another possibility that might be less confusing would be to use a subdirectory just for this purpose, instead of putting this package.json (which isn't the real package.jon) at the top level.

(Replaces #400, which I had meant to make a draft.)


john-conroy:
Not ideal to manage two package.json files, but not entirely against it if this is a huge pain point. Do we want to be able to test inside our docker containers?

test.sh
```diff
@@ -50,6 +50,5 @@ end pytest
 
 start docker
 ./docker.sh 5001
-server_up 5001
```
john-conroy, line 53: Did you determine this check to be redundant?

mccalluc: Yeah: Cypress has its own retry until it's up, so we don't need our own reinvented wheel here.

PR #405, opened 2020-06-05
Add docs from PK

mccalluc:
- This would be at `/about` ... Ok, or do you want something different?
- Should there be a link to this somewhere? Suggest speccing that in a separate issue, but could also be part of this PR: Your call.


ngehlenborg:
This is looking good. For now, I would create unique routes for each page underneath `/docs`. We will discuss where to link this from and how to create a navigation for these pages in a separate thread.

PR #408, opened 2020-06-05
Introduce separate envvars + add hubmapPortalUrl

### LINKED ISSUE #403: Pass "hubmapPortalUrl" to CCF-UI
So they can link back to us. (They are setting a base-uri, so I don't think it's as easy as passing `/`.)

mccalluc:
- CCF and Portal will hit separate indexes.
- Pass `hubmapPortalUrl` for CCF; Fix #403

### DIFF
--- a/CHANGELOG-ccf-root-url.md
+++ b/CHANGELOG-ccf-root-url.md
@@ -1,0 +1,1 @@
+- Provide root URL for CCF-EUI.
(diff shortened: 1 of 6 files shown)


mccalluc:
Talking with Bill this morning, and he is not in favor of giving us separate endpoints in docker compose configuration. I'll revisit this: I want to have the routing to two separate indexes at a fairly low level.

PR #410, opened 2020-06-06
John conroy/new react file structure

john-conroy:
Overhaul react file structure. I would have named the spec files `spec.js` instead of `Component.spec.js`, but eslint/vscode disagreed with that idea. Alternative to of adding an` index.js` file to each component directory we could simply rename the component implementation file `index.jsx`.


context/app/static/js/components/Detail/Attribution/index.js
```diff
@@ -0,0 +1,3 @@
+import Attribution from './Attribution';
+
+export default Attribution;
```
mccalluc, line 3: Thanks for doing this. It's most useful when there are multiple subsidiary components that can be all represented as separate files in a directory, with only one being exported. The benefit is hiding the details of the directory structure from the caller. Just want to be be sure the motivation makes sense, and it's not feeling arbitrary.

PR #425, opened 2020-06-09
Mccalluc/test ccf better

mccalluc:
Branch from #420 to use test improvements there.

- This would have caught the CCF problem.
- Expand the adhoc html->xml regexes.
- git rid of unit test (that is testing the test machinery itself) in preference to doc test.


mccalluc:
@john-conroy : Review?

PR #429, opened 2020-06-09
John conroy/further configure tests

john-conroy:
Delaying additional testing until home page is implemented per design. Working tests are blocked by #426 and #428.


context/app/static/js/components/Detail/Dataset/Dataset.spec.js
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable import/no-unresolved */
+/* eslint-disable camelcase */
+import React from 'react';
+import 'whatwg-fetch';
```
mccalluc, line 4: Add a comment about why this is needed?

context/jest.config.js
```diff
@@ -1,9 +1,16 @@
...
+    '^.+\\.css$': '<rootDir>/test-utils/__mocks__/style.js',
     '^components(.*)$': '<rootDir>/app/static/js/components$1',
     '^static(.*)$': '<rootDir>/app/static/$1',
-    '^test-utils.js': '<rootDir>/test-utils.js',
+    '^test-utils(.*)$': '<rootDir>/test-utils$1',
+    'search-schema-definitions$': '<rootDir>/search-schema/data/definitions.yaml',
+    'metadata-field-descriptions$': '<rootDir>/ingest-validation-tools/docs/field-descriptions.yaml',
+  },
+  transform: {
+    '^.+\\.[t|j]sx?$': 'babel-jest',
```
mccalluc, line 13: Are we already able to support typescript, or would that require something else?

context/test-utils/loaders/ymlLoader.js
```diff
@@ -0,0 +1,9 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
mccalluc, line 1: Are we doing something odd that this disable is needed here and elsewhere? Perhaps turn if off in test dirs?

context/jest.config.js
john-conroy: No that would require more configuration throughout the application. I can remove the `t` to avoid confusion?

context/app/static/js/components/Detail/Dataset/Dataset.spec.js
john-conroy: Good idea. I think I'm also going to move it into `setupTests.js`.

context/test-utils/loaders/ymlLoader.js
john-conroy: eslint doesn't recognize our webpack aliases/jest paths. I'll create an issue to think about how to fix this.

context/jest.config.js
mccalluc: unless we can actually handle it, let's remove it.

mccalluc:
@john-conroy : Looks good: Merge whenever. Thanks!

PR #469, opened 2020-06-18
Factors to cell-sets

keller-mark:
Depends on the ingest-pipeline running https://github.com/hubmapconsortium/ingest-pipeline/pull/85

Update: this is now ready for review.

This pull request replaces the `factors` component with the `cellSets` component. Note that the `factors` component depends on `factors.json` while `cellSets` uses a different schema and depends on `cell-sets.json`.

See discussion below for tradeoffs about the potential for `factors.json` to be available but not `cell-sets.json`


mccalluc:
@keller-mark : Do you want us to review this?

keller-mark:
I did not request any review because when I run locally, this pull request breaks the Visualization / Vitessce component for the `salmon_rnaseq_10x` datasets. It is expecting the `cell-sets.json` files to exist before creating the Vitessce view-config, but it looks like those files are not yet available.

The ingest-pipeline was [updated last week](https://github.com/hubmapconsortium/ingest-pipeline/pull/85) with the latest portal-containers, which should create the `cell-sets.json` files. Do you know if / when they will be generated?

keller-mark:
In the flask log in the console, it warns

```
[2020-06-23 09:31:20,818] INFO in vitessce: Files for assay "salmon_rnaseq_10x" uuid "{self.uuid}" not found as expected.
```

keller-mark:
Should I add some kind of condition, where it can fall back to `factors.json` if `cell-sets.json` does not yet exist?

mccalluc:
@keller-mark : I'm going to change this back to a draft to make it clear that it shouldn't be merged yet.
@ilan-gold : Can you take it out of draft when the backing data is available? Or if the error handling here should be more robust, can you get on the same page with Mark?

ilan-gold:
@keller-mark @mccalluc There should probably be a broader discussion of what to do about this sort hting.  I think falling back is a good idea @keller-mark but I think that there is also a broader problem of how to handle the constantly changing inputs.  I hardcoded a lot of this specifically because I did not know what stuff would look like going forward but perhaps now is the time to start to change that.  It's really a matter of how much we wish to support "old" runs that are still indexed.

mccalluc:
Here's my take:
- It only needs to gracefully handle the data we'll see on production... if you're not sure whether we'll see something on production or not, perhaps ping Bill?
- When it hits a case that it can't handle gracefully, it shouldn't be too awkward: The rest of the UI should still render.
- And if we can hit it in non-production environments, there should be a console warning: We have hit an error, it should never occur on production, and if TEST and DEV are ever cleaned up, we can get rid of the special case.

keller-mark:
In addition to this pull request, I made a second pull request with a fallback to factors in #504, as another option.

The way that this PR fails when `cell-sets.json` does not exist is that the `Visualization` component does not render because the available files do not contain the subset `{"cell-sets.json", "cells.json"}`, and there is an error in the console. The rest of the Details page does render.

The way that #504 fails is that it renders the visualization with the `factors` component rather than the `cellSets` component if `factors.json` is available but `cell-sets.json` is not.

I contacted Bill about what to expect in production which I have forwarded to @mccalluc: only one version of the data will be released to PROD, but he recommends being flexible since there will always be a chance that you get data from different versions of a pipeline, but that there should also be a way to know which pipeline version produced the data.

PR #477, opened 2020-06-21
No login required for CCF

### LINKED ISSUE #466: Pass CCF empty string if no nexus token
The team has committed to making public data available without login. Right now, there really is very-little-to-no public data in the various environments, but this is the plan.

@bherr2 : How do you want to the hand-off to CCF to work when there is no nexus token? Pass an empty string? Or something else?

mccalluc:
@bherr2 : When you've pushed a version to the CDN that should accommodate this, let me know here, and I'll run the portal locally to check, and then merge. Thanks!

Fix #466.

### DIFF
--- a/CHANGELOG-no-login-for-ccf.md
+++ b/CHANGELOG-no-login-for-ccf.md
@@ -1,0 +1,1 @@
+- No login required for CCF.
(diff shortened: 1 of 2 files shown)


bherr2:
I already have code in place to check when the nexus token is an empty string, but I haven't actually tested that it works. I can do some tests on Monday to be sure it works.

mccalluc:
And I just double checked here: Looks good. (The paucity of data is a concern, but I think that's orthogonal.) Thanks!

PR #479, opened 2020-06-21
Mccalluc/no login required

mccalluc:
@ngehlenborg : This is a start on removing the login requirements from all pages. It's not there yet, but I wanted to check-in and make extra sure this is the direction you want... If we had to add it back in next week that would be a hassle.


mccalluc:
@john-conroy : A simple login-logout works... but I should (1) fix travis (2) look for places where the client expects a nexus_token, and make sure we don't use an empty string when it's really missing.

PR #483, opened 2020-06-22
Mccalluc/clickable header links

### LINKED ISSUE #475: Make header counts clickable
Screen Shot 2020-06-20 at 2 26 20 PM

@ngehlenborg : I would expect to be able to click on these to go to the respective search pages... and maybe a static documentation page for "Centers"? Comment here if you are expecting these to work a different way.

mccalluc:
Fix #475
- @john-conroy : Am I following the conventions you'd like?
- @tsliaw : Does this seem like a reasonable direction for the UI? Would it be ok for "Donors" to go to the top level of the docs? I'd prefer not to create new documentation just so the link can go somewhere... but if this seems like one of the main questions people would have, there should be a way to answer it?

### DIFF
--- a/context/app/static/js/components/Home/DataSummaryItem/DataSummaryItem.jsx
+++ b/context/app/static/js/components/Home/DataSummaryItem/DataSummaryItem.jsx
@@ -6,7 +6,7 @@ function DataSummaryItem(props) {
   const { value, icon: Icon, label, href } = props;
 
   return (
-    <StyledLink href={href}>
+    <StyledLink href={href} underline="none">
       <Icon color="primary" style={{ fontSize: '36px' }} />
       <StyledTypography variant="h4">
         {value} {label}
--- a/context/app/static/js/components/Home/DataSummaryItem/DataSummaryItem.jsx
+++ b/context/app/static/js/components/Home/DataSummaryItem/DataSummaryItem.jsx
@@ -3,10 +3,10 @@ import PropTypes from 'prop-types';
 import { StyledTypography, StyledLink } from './style';
 
 function DataSummaryItem(props) {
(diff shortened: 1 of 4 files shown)


john-conroy:
Both the text and label are now primary purple. Is this what we want?

mccalluc:
@tsliaw : Can you make a call on whether putting the number and the text together is the right call? I think it looked weird to have them get there own individual link-styling on over, but happy to defer.

john-conroy:
Alternatively, we could either remove the underline text-decoration, have the link only include the label or some other design entirely since these are now links.

tsliaw:
So originally the design was: https://www.figma.com/proto/cEZdI2tOMByTBW8RQ2kuAm/Home-Page?node-id=38%3A3575&viewport=692%2C673%2C0.342952162027359&scaling=min-zoom
(hover over dataset to see).
I do want to stay away from the underline text-decoration, but if you think the individual link-styling looks odd, then I would prefer both the number and text be the light gray and be the purple during hover.

context/app/static/js/components/Home/DataSummaryItem/DataSummaryItem.jsx
```diff
@@ -1,20 +1,17 @@
...
-import { FlexRow, StyledTypography } from './style';
+import { StyledTypography, StyledLink } from './style';
 
 function DataSummaryItem(props) {
-  const { value, icon: Icon, label } = props;
+  const { value, icon: Icon, label, href } = props;
 
   return (
-    <FlexRow>
+    <StyledLink href={href}>
```
john-conroy, line 9:
```suggestion
    <StyledLink href={href} underline="none">
```
If we want to remove the underline on hover.

PR #490, opened 2020-06-22
Mccalluc/donor histogram facets

mccalluc:
- @john-conroy : Code review.
- @tsliaw : UI ok? Adding explicit inputs below would be easy... but feels like it would be clutter? Your call. http://docs.searchkit.co/v0.8.3/docs/components/ui/range-components.html

Screen Shot 2020-06-22 at 4 21 46 PM


context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -184,14 +184,19 @@ SearchWrapper.propTypes = {
...
+        'RangeFilter',
       ]).isRequired,
       props: PropTypes.exact({
+        // This is the union of two distinct prop lists.
         id: PropTypes.string.isRequired,
         title: PropTypes.string.isRequired,
         field: PropTypes.string.isRequired,
-        operator: PropTypes.string.isRequired,
-        size: PropTypes.number.isRequired,
+        operator: PropTypes.string,
```
john-conroy, line 194: Do you need `is.Required` with `exact()`? I believe you'll get a warning for a missing property either way.

mccalluc: I'm going to redo this to separate the types. FWIW, I don't think `exact` implies `is.Required`: Here's the example from the docs:
```
  // An object with warnings on extra properties
  optionalObjectWithStrictShape: PropTypes.exact({
    optionalProperty: PropTypes.string,
    requiredProperty: PropTypes.number.isRequired
  }),
```
👉 `exact` catches _extra_ props, not missing ones.

PR #501, opened 2020-06-23
John conroy/download metadata table

john-conroy:
I believe this is a situation where `React.memo()` makes sense.

MetadataTable

### DIFF
--- a/context/app/static/js/components/Detail/MetadataTable/MetadataTable.jsx
+++ b/context/app/static/js/components/Detail/MetadataTable/MetadataTable.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import Paper from '@material-ui/core/Paper';
 import Table from '@material-ui/core/Table';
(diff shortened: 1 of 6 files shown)


context/app/static/js/components/Detail/MetadataTable/MetadataTable.jsx
```diff
@@ -55,6 +88,7 @@ function MetadataTable(props) {
 
 MetadataTable.propTypes = {
   metadata: PropTypes.objectOf(PropTypes.string).isRequired,
+  display_doi: PropTypes.string.isRequired,
 };
 
-export default MetadataTable;
+export default React.memo(MetadataTable);
```
mccalluc, line 94: I see that additional computation is being done in this component, but I don't think there's anything that would prompt frequent re-renders that need to be optimized?

context/app/static/js/components/Detail/MetadataTable/MetadataTable.jsx
```diff
@@ -6,27 +6,60 @@ import TableBody from '@material-ui/core/TableBody';
...
-  { id: 'value', label: 'Value' },
-  { id: 'description', label: 'Description' },
-];
+function tableToDelimitedString(rows, colNames, d) {
+  const str = rows.reduce((acc, row) => {
+    const rowStr = Object.values(row).join(d);
+    return acc.concat('\n', rowStr);
+  }, colNames.join(d));
+  return str;
+}
```
mccalluc, line 23: Add unit tests for these two utilities?

PR #507, opened 2020-06-24
Mccalluc/globus link

mccalluc:
@tsliaw : It looks like this right now:
Screen Shot 2020-06-24 at 5 01 12 PM
If you can give some design direction, in whatever way is easiest for you.

@john-conroy : Code review. Would you like the component split off by itself, or it is small enough to live in the same file?

Both : Access denied will be a fairly common response when globus has genetic sequence data... but maybe showing the raw status code isn't the best choice?

Making this a draft for now, until I can actually get a 200 response. In contact with Bill.

### DIFF
--- a/CHANGELOG-globus.md
+++ b/CHANGELOG-globus.md
@@ -1,0 +1,1 @@
+- Link to globus file browser.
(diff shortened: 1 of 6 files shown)


mccalluc:
And Bill gave me a url that does give a 200: http://localhost:5001/browse/dataset/1009e3224eb8811a90aaf22aea019be2

Moving this out of Draft!

tsliaw:
Is there a dataset example where this url will be available?

tsliaw:
Updated file design with button for GLOBUS file browser.
file_design

Figma link: https://www.figma.com/file/awtd3Klx2hka54jyekr9pV/Detail-Pages?node-id=23%3A435

mccalluc:
> For the error code, if you want to show this, I think a hover tooltip might be appropriate.

I was just replacing it with text, but for stability, you're right, keeping the button, but disabled might make sense.
When the latest commit passes, I'll merge, and we can refine the look from there. Thanks, all!

PR #518, opened 2020-06-25
John conroy/basic collections

john-conroy:
Basic collections page to demonstrate api response. I didn't follow usual conventions for organizing components because they will likely be deleted shortly.

Screen Shot 2020-06-25 at 4 11 44 PM

### DIFF
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -141,3 +141,16 @@ def showcase_view(name):
         title='Showcase',
         flask_data=core_props
     )
+
+@blueprint.route('/collections')
+def collections():
+    if 'nexus_token' not in session:
+        abort(403)
+    core_props = {
+        'endpoints': _get_endpoints(),
+    }
+    return render_template(
+        'pages/base_react.html',
+        title='Collections',
+        flask_data=core_props
+    )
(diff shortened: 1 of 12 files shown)


context/app/routes_main.py
```diff
@@ -141,3 +141,16 @@ def showcase_view(name):
         title='Showcase',
         flask_data=core_props
     )
+
+@blueprint.route('/collections')
+def collections():
+    if 'nexus_token' not in session:
+        abort(403)
```
mccalluc, line 148:
```suggestion
```
Login shouldn't be necessary; but if it doesn't work right now without a token, let's get in touch with Bill and hear what the plan is.

context/app/static/js/components/Collections/Collections/Collections.jsx
```diff
@@ -0,0 +1,37 @@
...
+  const { entityEndpoint } = props;
+  const [collectionsData, setCollectionsData] = useState([]);
+  useEffect(() => {
+    async function getAllCollections() {
+      const response = await fetch(`${entityEndpoint}/collections`, {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${readCookie('nexus_token')}`,
+        },
```
mccalluc, line 16:
```suggestion
```
- Ditto about the nexus_token.
- `Content-Type` on GET seems odd?

john-conroy:
`/collections` works without a token `/collections/<uuid>` does not. So not needed here.

PR #521, opened 2020-06-26
Update vitessce to fix safari bugs.

ilan-gold:
@mccalluc Do you know what the error on Travis means?


ilan-gold:
Looks like it was removed some time back - sorry!

mccalluc:
We never got to a place where we could validate API responses against a schema without getting too much noise. Hope the removal doesn't disrupt you?

ilan-gold:
No worries! It failed, I did something else for a while, came back to it, and realized pretty immediately what had happened.

PR #524, opened 2020-06-26
markdown in React

mccalluc:
My initial motivation was to find a markdown library which handled the documentation as it stands... (In particular, handling tables that are missing their closing pipes.) I failed in that, but in the process, I did the work to bring the documentation into markdown. The one big improvement is that the rendering doesn't flash. This is probably the direction we want to go?
- It's simple enough that maybe I should get rid of the Markdown.jsx entirely?
- There are other options on NPM: This one is simple; I also tried `marked` which is just a markdown processor without react.


mccalluc:
@john-conroy : Can you give this a review, contingent on my making Travis happy, and fixing merge conflicts?

context/app/static/js/components/Routes.jsx
```diff
@@ -65,6 +66,14 @@ function Routes(props) {
       </Container>
     );
   }
+
+  if ('markdown' in flaskData) {
```
john-conroy, line 70: Actually should we also check the url path or is that difficult given the variety of paths for the documentation?

john-conroy, line 70:
```suggestion
  if ('markdown' in flaskData && urlPath.startsWith('/docs')) {
```

context/app/static/js/components/Routes.jsx
```diff
@@ -65,6 +66,14 @@ function Routes(props) {
...
+    return (
+      <Container maxWidth="lg">
+        <Markdown markdown={flaskData.markdown} />
```
john-conroy, line 73: Destructure at the top similar to other flaskData entries?

context/app/static/js/components/Routes.jsx
mccalluc: Right now, it could also be `VERSION`... and I feel like at some point there will be another addition at the top level, and tracking down why it doesn't render will be annoying.

context/app/static/js/components/Routes.jsx
mccalluc: Sure!

mccalluc:
@ilan-gold : This also moves the markdown rendering for Showcases to react, and gets rid of the dangerously_set_inner_html. I'm not really sure this is the best library, so if we get into and realize we need to swap, that's fine.

PR #560, opened 2020-07-01
Update donor.md

computationdoc:


computationdoc:
just testing process here with you Chuck - do we want documentation to be edited through pull requests - if so, I think this should be OK

pecan88:
just adding a @mccalluc call out so that he sees Jonathan's comment above.

PR #564, opened 2020-07-01
Image upload- metadata.tsv

cebriggs7135:


mccalluc:
No, sorry, we don't host images in git: I'd suggest asking PK what google drive folder they're using and put it there.

Also: It should have a descriptive filename.

PR #566, opened 2020-07-01
image upload- assay metadata levels

cebriggs7135:


mccalluc:
No, sorry, we don't host images in git: I'd suggest asking PK what google drive folder they're using and put it there.

Also: It should have a descriptive filename.

cebriggs7135:
@mccalluc I thought you said that the documents (images too?) linked to within an MD doc should be referred to as a path. So if I out these in Google drive, won't they have a url rather than a path?

PR #576, opened 2020-07-02
Mccalluc/errorboundaries

### LINKED ISSUE #505: Use Error Boundaries
We've never been able to get API responses to conform to a schema, and I don't know that we ever will. Given that, and errors like #331, and perhaps more as we get more users, we should make the frontend less brittle to unexpected input. I think [Error Boundaries](https://reactjs.org/docs/error-boundaries.html) would be useful here. I'm imagining we replace the higher level index.js code with an Error Boundary class around the original export.

@john-conroy : Do you have thoughts about this? Either how useful it would be, or it what level it would be best to set boundaries?

mccalluc:
Fix #505 

TODO: 
- padding on error page... but maybe we'll have something like a global solution, at some point?
- Make a cypress test.
- Ping a service on the server side to record the error, even if the user doesn't cooperate.
- Finish the error pages, so we can get rid of the extra layer of route matching.


mccalluc:
> Should we create design issues once we know our intentions for ErrorBoundaries?

You mean like making the error page itself look better? I wouldn't stop you, but I don't see it as a priority, either. I think, when the dust settles, we should make sure all our usages of Paper are getting the same padding from the same place, and not worry too much about individual cases.

PR #594, opened 2020-07-02
Fix ShowcaseDropdown bugs

### LINKED ISSUE #593: Showcase dropdown needs a higher z-index
Screen Shot 2020-07-02 at 12 32 06 PM

and while we are at it, the little arrow state should be reversed. I think when closed, the arrow should point down, and when open, it should point up.

keller-mark:
Fixes #593 by adding a z-index of `5` to the ShowcaseDropdown popper component. Changes the direction of the arrow icon.


keller-mark:
Ha no worries, you implemented it the other way in more places than not :)
Screen Shot 2020-07-02 at 1 36 58 PM

PR #597, opened 2020-07-02
John conroy/dataset file browser

john-conroy:
Still need to add prop types and tests for the `useEffect` hook in `FileBrowser`. We could probably benefit from converting the `useEffect` to `useMemo`  in `FileBrowser`, but @mccalluc wants to hold off until performance issues can be seen.

FileBrowser 1

FileBrowser 2

FileBrowser 5


context/app/static/js/components/Detail/FileBrowser/FileBrowser.jsx
```diff
@@ -0,0 +1,65 @@
...
+function FileBrowser(props) {
+  const { files } = props;
+  const [fileTree, setFileTree] = useState({});
+  useEffect(() => {
+    const treePath = {};
+    files.forEach((f) => {
+      const levels = f.rel_path.split('/');
+      const file = levels.pop();
+
+      const fileObj = { file, fullPath: f.rel_path, description: f.description, edam_term: f.edam_term, size: f.size };
```
ilan-gold, line 14: Spread syntax?

context/app/static/js/components/Detail/FileBrowser/FileBrowser.jsx
```diff
@@ -0,0 +1,65 @@
...
+import FileBrowserNode from '../FileBrowserNode';
+import { ScrollPaper } from './style';
+
+function FileBrowser(props) {
+  const { files } = props;
+  const [fileTree, setFileTree] = useState({});
+  useEffect(() => {
+    const treePath = {};
+    files.forEach((f) => {
+      const levels = f.rel_path.split('/');
```
ilan-gold, line 11: Can we use terminology about folders instead of levels?

context/app/static/js/components/Detail/FileBrowserNode/FileBrowserNode.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import FileBrowserDirectory from '../FileBrowserDirectory';
+import FileBrowserFile from '../FileBrowserFile';
+import { Column } from './style';
+
+function FileBrowserNode(props) {
```
ilan-gold, line 7: Perhaps comment that this is some sort of wrapper around the actual `File` and `Directory` components that one can see.

context/app/static/js/components/Detail/FileBrowser/FileBrowser.jsx
```diff
@@ -0,0 +1,65 @@
...
+      const file = levels.pop();
+
+      const fileObj = { file, fullPath: f.rel_path, description: f.description, edam_term: f.edam_term, size: f.size };
+
+      if (levels.length === 0) {
+        if ('files' in treePath) {
+          treePath.files.push(fileObj);
+        } else {
+          treePath.files = [fileObj];
```
ilan-gold, line 20: Instantiating `treePath.files = []` above would make this a little clearer since you could just `.push`, I think.

context/app/static/js/components/Detail/FileBrowser/FileBrowser.jsx
```diff
@@ -0,0 +1,65 @@
+import React, { useState, useEffect } from 'react';
+import FileBrowserNode from '../FileBrowserNode';
+import { ScrollPaper } from './style';
+
+function FileBrowser(props) {
+  const { files } = props;
+  const [fileTree, setFileTree] = useState({});
+  useEffect(() => {
```
ilan-gold, line 8: Maybe add a comment about what `treePath` could look like?

context/app/static/js/components/Detail/FileBrowser/FileBrowser.jsx
ilan-gold: Or, I guess, be consistent about camel-case

context/app/static/js/components/Detail/Dataset/context.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const DatasetContext = React.createContext({});
```
keller-mark, line 3: Can you add a comment about what can/should be stored in this context?
```suggestion
// Stores the `assetsEndpoint` and `uuid` values.
const DatasetContext = React.createContext({});
```

context/app/static/js/components/Detail/Dataset/Dataset.jsx
```diff
@@ -1,14 +1,18 @@
...
 import Attribution from '../Attribution';
 import Protocol from '../Protocol';
 import MetadataTable from '../MetadataTable';
-import FileTable from '../FileTable';
 import Visualization from '../Visualization';
 import DetailLayout from '../DetailLayout';
+import Files from '../Files';
+
+// TODO use this context for components other than FileBrowser
+import DatasetContext from './context';
```
keller-mark, line 15: It seems like this context should be located at a common place relative to where it is used, for example `Detail/contexts.js`

context/app/static/js/components/Detail/FileBrowser/FileBrowser.jsx
keller-mark, line 8: Minor thing but if possible, it would be nice to extract this logic in the useEffect to a utility function stored in a separate js file.

john-conroy:
I actually previously had some logic that if you're at the bottom of the page, the selected section is set to the last, but it had its own drawbacks (most of which I forget), but its definitely worth further investigation.

context/app/static/js/components/Detail/FileBrowser/FileBrowser.jsx
john-conroy: Agreed, I'll fix it.

ilan-gold:
Ok no need to get in the weeds now with it then.  I can make an issue.

john-conroy:
Thanks, and thank you both for the quick reviews.

PR #600, opened 2020-07-02
SPRM Vitessce Integration

ilan-gold:
This adds SPRM in - there are no cell sets now but the hope is that there will be.  Here's a working UUID (on test): fa84d7f5b2ee6875f70e1b6ce919c1fe


context/app/api/vitessce.py
```diff
@@ -65,30 +59,61 @@
...
+            {"rel_path": f"{SCRNASEQ_BASE_PATH}.cells.json", "type": "CELLS"},
+            {"rel_path": f"{SCRNASEQ_BASE_PATH}.cell-sets.json", "type": "CELL-SETS"},
         ],
     },
     "codex_cytokit": {
         "base_conf": IMAGING,
-        "view": {"zoom": -1.5, "target": [600, 600, 0], },
+        "view": {"zoom": -1.5, "target": [600, 600, 0]},
         "files_conf": [
             # Hardcoded for now only one tile.
```
mccalluc, line 70: No longer true?

context/app/api/vitessce.py
```diff
@@ -65,30 +59,61 @@
...
         ],
     },
 }
 
 CODEX_ASSAY = "codex_cytokit"
 
 IMAGE_ASSAYS = [CODEX_ASSAY]
+TILED_ASSAYS = [CODEX_ASSAY]
+
+TILE_REGEX = ".*(R(\\d+)_X(\\d+)_Y(\\d+)).*"
```
mccalluc, line 100:
```suggestion
TILE_REGEX = r"R\d+_X\d+_Y\d+"
```
- You almost always want raw strings with regexes in python.
- Are you using the inner capturing sub expressions?
- Use `search` instead of `match`, if you're only interested in the capture.

context/app/api/vitessce.py
```diff
@@ -65,30 +59,61 @@
...
+def _get_tiles(files):
+    return list(
+        set(
+            [
+                re.match(TILE_REGEX, file).group(1)
+                for file in files
+                if re.match(TILE_REGEX, file)
+            ]
+        )
+    )
```
mccalluc, line 114: Avoid doing the same operation twice.
```
>>> animals = ['dog', 'cat', 'mouse']
>>> import re
>>> [match[0] for match in set(re.search(r'o.*', animal) for animal in animals) if match]
['ouse', 'og']
```

context/app/static/js/components/Detail/Dataset/Dataset.jsx
```diff
@@ -61,7 +61,7 @@ function DatasetDetail(props) {
   } = assayMetadata;
 
   const shouldDisplaySection = {
-    visualization: 'name' in vitData,
+    visualization: 'name' in vitData || (vitData[0] && 'name' in vitData[0]),
```
mccalluc, line 64: It scares me a bit if sometimes we'll get an object, and sometimes there's an array of objects.

context/app/static/js/components/Detail/Visualization/Visualization.jsx
```diff
@@ -11,19 +18,34 @@ import {
...
   StyledFooterText,
+  SelectionButton,
 } from './style';
 import 'vitessce/dist/es/production/static/css/index.css';
 
+function getTileOrDatasetString(vitData) {
+  // If the name of a configuration includes the tiled name, return 'Tile'; else return 'Dataset.'
+  if (Array.isArray(vitData)) {
+    const firstEntry = vitData[0];
+    const regex = new RegExp('.*(R(\\d+)_X(\\d+)_Y(\\d+)).*');
```
mccalluc, line 34: Ditto!

context/app/api/vitessce.py
ilan-gold: Cool! I was wondering about this!

context/app/static/js/components/Detail/Dataset/Dataset.jsx
ilan-gold: Do you have any other recommendations?  I saw it a few times when looking up how to handle it.

PR #606, opened 2020-07-03
Cell type annotation showcase

keller-mark:
This adds a new showcase http://localhost:5001/showcase/satija


keller-mark:
Oh right! Thanks for catching that

PR #608, opened 2020-07-03
Mccalluc/clean up md

mccalluc:
Remove stub pages and fill in links, as discussed.

(As you can imagine, I am not excited about maintaining a parallel list of assay types. Your call for how we handle this now, and in the future.)


context/app/markdown/docs/metadata.md
```diff
@@ -2,38 +2,73 @@
 
 HuBMAP metadata documentation and submission specifications
 
-## Donor
```
ngehlenborg, line 5: This section should remain until it is clear that this won't be delivered.

context/app/markdown/docs/metadata.md
```diff
@@ -2,38 +2,73 @@
 
 HuBMAP metadata documentation and submission specifications
```
mccalluc, line 4:
```suggestion

## Donor

Donor Specification

TODO: Donor specification pending from Jonathan Silverstein.

```

context/app/markdown/docs/metadata.md
mccalluc: Reverted. Thanks!

mccalluc:
Brought back the donor section, with a TODO. The separate donor.md is still being removed, but it can be brought back easily.

PR #645, opened 2020-07-06
Mccalluc/search for existence

mccalluc:
I have often found it hard to find examples of documents which possessed a particular attribute I wanted to exercise: Right now, I want to find datasets that have files so I can work on the modal... but I'd imagine that outside users also might be interested in this filter, so they could find data they could immediately download.

Screen Shot 2020-07-06 at 2 47 18 PM

- Ok to expose this, or does it need to be a development-mode only feature?
- If it will be exposed, is this text ok, or does it need to change?
- If we add additional ones, they will also have their own bold heading... Is that ok, or do we need to find some way to group them?


mccalluc:
I'll close it then, but keep it in my tool box. Hope that's ok.

PR #648, opened 2020-07-06
pass down hasAgreedToDUA

mccalluc:
Work in progress; Draft PR, but feedback appreciated before going deeper.
- I notice there's a react context here: Would it be good to add DUA approval to that?
- File organization: I would incline towards moving all the components that are only used in the FileBrowser under it... or, if a flat directory structure is better, perhaps rename ConditionalLink to FileBrowserConditionalLink?

TODOs:
- Set cookie on approval, and read cookie on initial load.
- CSS: Currently does not display pointer cursor until after DUA is approaved.
- Use Material-UI modal.
- PropTypes.


john-conroy:
- Moving DUA approval to context sounds good.
- The `Detail` domain within components is too large now. I think moving at least `FileBrowser`, `Prov`, `Visualization` and their related components into their own directories within `components` could make sense, but I'm not entirely sure and open to a different structure if you have any thoughts.

mccalluc:
I think this could be merged now, particularly if we want to be sure to have it in the next release, so I'll take it out of draft, but there is still a lot that should be done:
- Behavior after click: Right now, nothing happens after click... another click is needed to actually download.
- Prop types
- Make the UI match the mockup.
- Testing

context/app/static/js/components/Detail/FileBrowserDUA/FileBrowserDUA.jsx
```diff
@@ -0,0 +1,72 @@
...
+            an acknowledgement of HuBMAP. Suggested language for such an acknowledgment is: “The results here are in
+            whole or part based upon data generated by the NIH Human BioMolecular Atlas Program (HuBMAP):{' '}
+            <a href="https://hubmapconsortium.org">https://hubmapconsortium.org</a>.”
+          </p>
+
+          <h2>Data Sharing Policy</h2>
+          <p>
+            The HuBMAP Data Sharing Policy can be found at{' '}
+            <a href="https://hubmapconsortium.org/policies/">https://hubmapconsortium.org/policies/</a>.
+          </p>
```
john-conroy, line 46: We should probably use Typography for the text elements.

PR #659, opened 2020-07-08
John conroy/prov table tiles

john-conroy:
Need to add tests, prop-types and add the bottom section included in the design to each tile. The table is responsive, however it doesn't look great currently when all three columns cannot be in a single row.


context/app/static/js/components/Detail/EntityTile/style.js
```diff
@@ -0,0 +1,94 @@
...
+  text-overflow: ellipsis;
+`;
+
+const StyledDivider = styled(Divider)`
+  background-color: ${(props) => props.theme.palette.text.primary};
+  margin: 0px ${(props) => props.theme.spacing(0.5)}px;
+`;
+
+const TextSection = styled.div`
+  min-width: 0;
```
mccalluc, line 79: Dumb question, but what does this do? Is negative width a thing?

context/app/static/js/components/Detail/ProvTable/ProvTable.jsx
```diff
@@ -34,34 +22,28 @@ function ProvTable(props) {
...
-                  <ListItemText primary={item['hubmap:displayDOI']} />
-                </ListItemLink>
+              type.map((item, j) => (
+                <ProvTableTile
+                  key={item.uuid}
+                  uuid={item.uuid}
+                  id={item.display_doi}
+                  entity_type={item.entity_type}
+                  isCurrentEntity={uuid === item.uuid}
+                  isNotSibling={j > 0 ? type[j - 1].specimen_type !== item.specimen_type : false}
```
mccalluc, line 36: Ternary not needed?
```suggestion
                  isNotSibling={j > 0 && type[j - 1].specimen_type !== item.specimen_type}
```
... and prefer not to have negative variables?
```suggestion
                  isSibling={!(j > 0 && type[j - 1].specimen_type !== item.specimen_type)}
```
... and then simplify?
```suggestion
                  isSibling={j === 0 || type[j - 1].specimen_type === item.specimen_type)}
```

Not a blocker for merge.

context/app/static/js/hooks/useEntityData.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
mccalluc, line 1: I think this does the same thing as the [LookupEntity](https://github.com/hubmapconsortium/portal-ui/pull/610/files#diff-606a7a99a80c265ec783d37535884a59) I introduced over the weekend... Let's just keep one? No strong preference between the two.

john-conroy: I'd prefer to use the hook.

context/app/static/js/components/Detail/EntityTile/style.js
john-conroy: Sorry I mean to include a comment. This is a weird fix to make the `text-overflow: ellipsis;`  work

PR #663, opened 2020-07-08
Add files via upload

pecan88:
This is Jonathan's donor metadata.

I uploaded the file and GitHub converted it.  If there are any issues with rendering, let me know and I will send you the file / text directly @mccalluc

### DIFF
--- a/context/app/markdown/docs/donor.md
+++ b/context/app/markdown/docs/donor.md
@@ -1,0 +1,197 @@
+# donor
+
+## Table of contents
+<details><summary>Background</summary>
+
+[`data_acquisition`](#data_acquisition)<br>
+[`data_transformation`](#data_transformation)<br>
+</details>
+
+<details><summary>Format</summary>
+
+[`donor_id`](#donor_id)<br>
+[`source_name`](#source_name)<br>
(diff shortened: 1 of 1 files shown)


computationdoc:
Chuck - the content was provided to be consistent with everything else for users to understand what the donor metadata is - so of course it is a copy/paste of the samples documentation - that is the point - indeed, yes, it is a different process of actual ingest and process validation (which the documentation - this file - explains - that it is hand abstracted) - the sample DOCUMENTATION, does not state (though its in the code): 
"the repo has machinery for making sure that sample files are in sync with the validation rules"

The documentation is simply a way of showing the users what's what.

SO, if you want the content in a different format for this for some reason (e.g. because it seems to imply something is happening that isn't, which is what appears to bother you), then change the format. e.g. you could eliminate the "tables" of "validation rules" and just have the text description for each one, fine. Instead if you are saying that you want someone to process validation on the donor metadata, that's not going to happen this release, but could in future. Third, you could add a warning in the documentation file that these are not "machine validated" documentation.

Please advise what approach is acceptable to you. I suggest merging it as is and adding in to it whatever caution suits you for now.

Thanks, JCS.

PR #668, opened 2020-07-08
John conroy/fix prov table duplication

### LINKED ISSUE #667: Duplicate entity tiles for the current entity in the provenance table are being created on each additional render

john-conroy:
Fixes #667


context/app/static/js/components/Detail/ProvTable/ProvTable.jsx
```diff
@@ -9,9 +9,9 @@ import ProvTableDerivedLink from '../ProvTableDerivedLink';
 function ProvTable(props) {
   const { uuid, entity_type, typesToSplit, ancestors, assayMetadata } = props;
 
-  ancestors.push(assayMetadata);
+  const entities = [...ancestors, assayMetadata];
```
mccalluc, line 12:
```suggestion
  // Make a new list rather modifying old one in place: Caused duplication in UI.
  const entities = [...ancestors, assayMetadata];
```

PR #670, opened 2020-07-08
Add donor facets

mccalluc:
Towards #671 , but does not provide any organization for the facet list.

### DIFF
--- a/context/app/static/js/components/Search/config.js
+++ b/context/app/static/js/components/Search/config.js
@@ -1,15 +1,22 @@
 // eslint-disable-next-line import/named
 import { filter, rangeFilter, field } from './utils';
 
+function makeDonorMetadataFilters(isDonor) {
+  const pathPrefix = isDonor ? '' : 'donor.';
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/Search/config.js
```diff
@@ -27,7 +34,7 @@ export const sampleConfig = {
     filter('mapped_specimen_type', 'Specimen Type'),
     filter('donor.group_name', 'Group'),
     filter('created_by_user_displayname', 'Creator'),
-  ],
+  ].concat(makeDonorMetadataFilters(false)),
```
mccalluc, line 37: @ngehlenborg : This and the one below add the four donor filters.

ngehlenborg: Is this for sample only or also for datasets? We want it for both, I don't remember if I said that on Slack.

mccalluc: Both: should have been more clear.

ngehlenborg: Thanks!

PR #673, opened 2020-07-08
John conroy/fix prov table order

### LINKED ISSUE #669: Provenance table tiles are not in order

john-conroy:
Temporary fix until a more permanent solution can be found. Fixes #669


john-conroy:
Yes, I understand that these are db timestamps. Sorting by `created_timestamp` did fix the few provenance tables with out of order entities without negatively affecting those that were previously in-order. I don't think it's infallible, but not sure what else we can do in the short-term.

PR #683, opened 2020-07-09
Update af.md

cebriggs7135:


cebriggs7135:
@mccalluc I wasn't aware that I was reverting anything. Apologies.

mccalluc:
The new lines are still being removed. I'll merge to get this in, but more tweaks may be required.

Issue #722, opened 2020-07-10
Still referencing mixif


mccalluc:
Find ref to https://github.com/hubmapconsortium/ingest-validation-tools/tree/master/docs/mixif and correct

cebriggs7135:
Changed Github link in the .md doc from ....mixif to .....mxif.

mccalluc:
Please do not close issues manually: Instead, add a note like "Fixes #722". Sometimes there are problems with PRs, and we don't want to think work is done when it isn't.

PR #734, opened 2020-07-10
Update mxif.md

cebriggs7135:
Fixed Github link to the metadata specs (github spec name was previously changed from mixif to mxif)


mccalluc:
Fix #722.

PR #749, opened 2020-07-13
Add status component to Vitessce view configs

keller-mark:
This pull request adds the `"status"` component to both vitessce view config templates, to allow data loading/parsing error messages to appear in the UI.

While doing this, I also added the `cellSetSizes` component for the salmon_rnaseq_10x view config.


ilan-gold:
I agree with Chuck here.  I am not sure there is a "production" use case.

PR #766, opened 2020-07-14
John conroy/split bundle vendors

john-conroy:
Splits all shared node modules into own entry chunk named `vendors.bundle.js`. The main bundle has been renamed `main.bundle.js`.

### DIFF
--- a/context/build-utils/webpack.common.js
+++ b/context/build-utils/webpack.common.js
@@ -6,6 +6,7 @@ const config = {
   entry: { bundle: './app/static/js/index.jsx' },
   output: {
     path: resolve('./app/static/public'),
+    publicPath: './static/public/',
     filename: '[name].js',
   },
   resolve: {
--- a/context/build-utils/webpack.common.js
+++ b/context/build-utils/webpack.common.js
@@ -6,7 +6,7 @@ const config = {
   entry: { main: './app/static/js/index.jsx' },
   output: {
     path: resolve('./app/static/public'),
-    publicPath: '../static/public/',
+    publicPath: `${resolve('/static/public/')}/`,
     filename: '[name].bundle.js',
   },
   resolve: {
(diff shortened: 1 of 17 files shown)


context/build-utils/webpack.common.js
```diff
@@ -3,11 +3,11 @@ const { resolve } = require('path');
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
 
 const config = {
-  entry: './app/static/js/index.jsx',
+  entry: { main: './app/static/js/index.jsx' },
   output: {
     path: resolve('./app/static/public'),
-    filename: 'bundle.js',
-    publicPath: resolve('../app/static/public'),
+    publicPath: `${resolve('/static/public/')}/`,
```
mccalluc, line 9: The final "/" probably is required? Would it just be an error if you gave a directory name without the final "/"?

context/build-utils/webpack.common.js
```diff
@@ -86,6 +86,17 @@ const config = {
     ],
   },
   plugins: [new CleanWebpackPlugin()],
+  optimization: {
+    splitChunks: {
+      cacheGroups: {
+        commons: {
+          test: /[\\/]node_modules[\\/]/,
```
mccalluc, line 93:
```suggestion
          test: RegExp('/node_modules/'),
```
- I don't think we need to ensure MS-Windows compatibility?
- Longer syntax is more readable here.

context/build-utils/webpack.common.js
john-conroy: Yeah, not including the additional / was producing an error for whatever reason.

context/build-utils/webpack.common.js
john-conroy: Sure, I just included a suggested snippet.

john-conroy:
@ilan-gold sounds good, I'll do so in the future.

PR #738, opened 2020-07-13
John conroy/font update

john-conroy:
Fixes #382 . Fixes #386 . Fixes #441 .

PR #768, opened 2020-07-14
John conroy/lazy vitessce

john-conroy:
Dynamically import vitessce visualization component and display circular progress indicator as fallback. At least now first-time visitors won't have to load vitessce when visiting the homepage.

Also fixes issues with visualization title font sizing introduced by #738 and increased bottom margin to prevent overlap. Sorry for the bad code suggestion in one of your earlier PRs, @ilan-gold.

### DIFF
--- a/context/app/static/js/components/Detail/Visualization/Visualization.jsx
+++ b/context/app/static/js/components/Detail/Visualization/Visualization.jsx
@@ -60,7 +60,9 @@ function Visualization(props) {
   return (
     <StyledSectionContainer id="visualization">
       <StyledHeader>
-        <StyledHeaderText>Visualization</StyledHeaderText>
+        <StyledHeaderText variant="h4" component="h2">
+          Visualization
+        </StyledHeaderText>
         <StyledHeaderRight>
           <VisualizationThemeSwitch theme={vitessceTheme} onChange={(e, theme) => setVitessceTheme(theme)} />
           <ExpandButton size="small" onClick={handleExpand} variant="contained" disableElevation>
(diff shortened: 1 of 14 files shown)


context/app/static/js/components/Detail/Visualization/Visualization.jsx
```diff
@@ -60,7 +60,9 @@ function Visualization(props) {
   return (
     <StyledSectionContainer id="visualization">
       <StyledHeader>
-        <StyledHeaderText>Visualization</StyledHeaderText>
+        <StyledHeaderText variant="h4" component="h2">
```
mccalluc, line 63: There was a PR yesterday, I think, that got rid of a lot of these attribute usages, and I'm reluctant to go in the other direction. I've wondered if we could just have an `<H1>`, `<H2>`, that would wrap `<h1>` with whatever material UI details we need.

context/app/static/js/components/Detail/VisualizationWrapper/VisualizationWrapper.jsx
```diff
@@ -0,0 +1,32 @@
...
+      fallback={
+        <StyledSectionContainer id="visualization">
+          <StyledHeader>
+            <StyledHeaderText variant="h4" component="h2">
+              Visualization
+            </StyledHeaderText>
+          </StyledHeader>
+          <VisualizationBackground>
+            <CircularProgress />
+          </VisualizationBackground>
```
mccalluc, line 23: Good, so the please-wait will use as much height as the final result, right?

john-conroy: Yes.

context/app/static/js/components/Detail/Visualization/Visualization.jsx
john-conroy: I think I fixed that, let me know if the change addresses your concerns.

PR #769, opened 2020-07-14
Search for items with or without metadata

### LINKED ISSUE #757: count samples that actually have metadata, report that to Nils

mccalluc:
Revive ideas from #645, and add a docs page that helps you find items with or without metadata. (Possible to make a custom report like the one in slack, but that would be more work.)

Fix #757, I think.


mccalluc:
Thanks. May make a top-level `/dev-tools` for better organization.

PR #780, opened 2020-07-14
Mccalluc/dataset icon

mccalluc:


context/app/static/js/shared-styles/icons/index.js
```diff
@@ -2,6 +2,11 @@ import styled from 'styled-components';
 import BubbleChartIcon from '@material-ui/icons/BubbleChart';
 import PersonIcon from '@material-ui/icons/Person';
 import TableChartIcon from '@material-ui/icons/TableChart';
+import AccountBalanceIcon from '@material-ui/icons/AccountBalance';
```
tsliaw, line 5: Can you make sure all the icons are the rounded icons? We decided to use the rounded icon library from Materials. So for example,
`import AccountBalanceIcon from '@material-ui/icons/AccountBalanceRounded'`

mccalluc: Will do! (Also, if there's just a small typo in a PR, click the "±" in the toolbar to suggest edits: The author can accept the changes from gitup without going back to the editor.)

PR #790, opened 2020-07-15
Mccalluc/no more quick start

mccalluc:
- Speed up travis by removing Docker build. (I could be talked out of this, but it seems safe.)
- Remove old, duplicate quick-start.sh.
- What was the intention of `trap "kill 0" EXIT` in dev-start.sh? I could be confused, but the `0` is the PID, which on travis is the main process, so when the `grep TODO || die` caused an EXIT, instead of just ending that script, it ended the entire Travis run.
- Add some more output in Travis to help with debugging.


README.md
```diff
@@ -48,9 +48,9 @@ After checking out the project, cd-ing into it, and setting up a Python3.7 virtu
     - (Note: Error with @hubmap-prov-vis dependency. Manually add
     the schema.json until [this issue is resolved](https://github.com/hubmapconsortium/portal-ui/issues/139).)
 - in the root project directory
-    - `./quick-start.sh`,
-    - update `context/instance/app.conf` with the Globus client ID and client secret, and the Elasticsearch endpoint,
-    - `./dev-start.sh` to run the webpack dev and flask servers and visit [localhost:5001](http://localhost:5001).
+    - Run `./dev-start.sh`, which will fail with a warning.
+    - Update `context/instance/app.conf` with the Globus client ID and client secret. (Get the those codes from another developer.)
```
john-conroy:
```suggestion
    - Update `context/instance/app.conf` with the Globus client ID and client secret. (Get those codes from another developer.)
```

dev-start.sh
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 set -o errexit
-trap "kill 0" EXIT
```
john-conroy, line 3: I defer to you with bash. You were able to end everything cleanly with a SIGINT?

README.md
mccalluc: I've simplified this section, asking the reader to just get the conf first.

mccalluc:
> I'm a little weary of using ./dev-start.sh alone

Ok, will re-add docker.sh. Good to be safe.

And I added the `trap` back in, but changed what gets killed. Learned that `kill 0` kills the entire process group... which may include the parent.

It looks like the new approach is cleaning everything up:

```
$ ps aux | grep 'webpack\|python'
chuck            90751   4.9  0.2  4322336  34788 s004  S+    4:41PM   0:03.06 /opt/anaconda3/envs/portal-ui-37/bin/python -m flask run
chuck            90782   0.0  0.0  4286732    752 s001  S+    4:44PM   0:00.00 grep webpack\|python
chuck            90757   0.0  4.2  5315396 704952 s004  S+    4:41PM   0:56.53 node /Users/chuck/github/hubmap/portal-ui/context/node_modules/.bin/webpack-dev-server --config build-utils/webpack.dev.js --hot
chuck            90748   0.0  0.2  4316948  34652 s004  S+    4:41PM   0:00.41 python -m flask run

# Then ctrl-C in the other terminal...

$ ps aux | grep 'webpack\|python'
chuck            90797   0.0  0.0  4280588    756 s001  S+    4:44PM   0:00.00 grep webpack\|python
```

Issue #793, opened 2020-07-15
Search api not returning descendant entity counts


john-conroy:
hubmapconsortium/search-api#85 was meant to introduce descendant entity counts, but they are not appearing in the search api response.

john-conroy:
Need to reindex elasticsearch.

PR #804, opened 2020-07-16
John conroy/entity tile bottom

john-conroy:
Adds bottom section to the entity tile. I can't think of a better name other than 'bottom'. I'm reluctant to call it a footer.

Missing map key is added in #810.


context/app/static/js/components/Detail/EntityTile/EntityTile.jsx
```diff
@@ -50,6 +56,18 @@ function EntityTile(props) {
               )}
             </TextSection>
           </FixedWidthFlex>
+          <BottomSection $isCurrentEntity={isCurrentEntity}>
+            {Object.entries(descendantCounts).map(([k, v]) => (
+              <>
+                {k === 'Dataset' ? <BottomDatasetIcon /> : <BottomSampleIcon />}
```
mccalluc, line 62:
```suggestion
                {k === 'Dataset' && <BottomDatasetIcon />}
                {k === 'Sample' && <BottomSampleIcon />}
```
Same commend in a related PR?

context/app/static/js/components/Detail/ProvTableTile/ProvTableTile.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useMemo } from 'react';
```
mccalluc, line 1: I wouldn't use memo... I don't think we're doing any re-rendering in a tight loop, so it's a bit of extra complexity for us to worry about, without gaining much.

context/app/static/js/components/Detail/ProvTableTile/ProvTableTile.jsx
```diff
@@ -12,6 +12,22 @@ function ProvTableTile(props) {
   // mapped fields are not included in ancestor object
   const entityData = useEntityData(uuid, elasticsearchEndpoint);
 
+  const types = entity_type === 'Donor' ? ['Sample', 'Dataset'] : ['Dataset'];
+
+  const descendantCounts = useMemo(() => {
```
mccalluc, line 17: Should be able to use `descendant_counts` in the ES response instead of calculating it here.

context/package.json
```diff
@@ -7,6 +7,7 @@
     "@material-ui/core": "^4.9.4",
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.45",
+    "date-fns": "^2.14.0",
```
mccalluc, line 10: Slight preference to use the YYYY-MM-DD format straight from the index... Unless there's been a directive to go with a particular format, in which case need to make sure it gets applied everywhere.

context/app/static/js/components/Detail/ProvTableTile/ProvTableTile.jsx
john-conroy: `descendant_counts` weren't showing / #793

context/app/static/js/components/Detail/ProvTableTile/ProvTableTile.jsx
john-conroy: I needed to use `useMemo` here, `useEffect` was causing it to get caught re-rendering.

context/package.json
john-conroy: I'd like to keep this parallel to the design. I can file an issue to update once the design is.

context/app/static/js/components/Detail/ProvTableTile/ProvTableTile.jsx
john-conroy: Needed may be the wrong word. `useMemo` solved the issue and I figured it would be better to use it than seek some other solution since this will be removed once we get `descendant_counts` from the api.

mccalluc:
@john-conroy : Nils would like to see this, so file an issue to pull the counts from the index, when they're available, and merge this as is, I think?

PR #805, opened 2020-07-16
John conroy/split entity tile

john-conroy:
Splits `EntityTile` into multiple components and pulls them into a directory within components. This is how I plan to continue to break up the 'Detail' domain.

Any thoughts whether we should continue to prefix component names with the name of the domain such as `entity-tile/EntityTileTop` or just use `entity-tile/Top` (or a better name)?


context/app/static/js/components/entity-tile/EntityTileBottom/EntityTileBottom.jsx
```diff
@@ -0,0 +1,25 @@
...
+import { FixedWidthFlex, StyledDivider, StyledTypography, StyledDatasetIcon, StyledSampleIcon } from './style';
+
+function EntityTileBottom(props) {
+  const { isCurrentEntity, entityData, descendantCounts } = props;
+
+  return (
+    <FixedWidthFlex $isCurrentEntity={isCurrentEntity}>
+      {Object.entries(descendantCounts).map(([k, v]) => (
+        <>
+          {k === 'Dataset' ? <StyledDatasetIcon /> : <StyledSampleIcon />}
```
mccalluc, line 13:
```suggestion
          {k === 'Dataset' && <StyledDatasetIcon />}
          {k === 'Sample' && <StyledSampleIcon />}
```
Slightly paranoid, probably.

PR #806, opened 2020-07-16
John conroy/fix entity tile style

john-conroy:
Fixes and simplifies styles for entity tile components.

### DIFF
--- a/context/app/static/js/components/entity-tile/EntityTileTop/style.js
+++ b/context/app/static/js/components/entity-tile/EntityTileTop/style.js
@@ -1,7 +1,6 @@
 import styled, { css } from 'styled-components';
 
 import { DatasetIcon, SampleIcon, DonorIcon } from 'shared-styles/icons';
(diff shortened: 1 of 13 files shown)


context/app/static/js/components/entity-tile/EntityTileTop/style.js
```diff
@@ -25,8 +24,24 @@ const StyledDonorIcon = styled(DonorIcon)`
...
+  ${(props) =>
+    props.$invertColors &&
+    css`
+      background-color: ${props.theme.palette.primary.main};
+      color: #ffffff;
+
+      svg {
+        color: #ffffff;
+      }
+    `}
```
mccalluc, line 44: `EntityTileBottom/style.js` also has this: Can you factor it out to a place they could share from?

john-conroy: Similar but the colors are swapped.

PR #808, opened 2020-07-16
Create user-agreement.md

pecan88:
@computationdoc  7 @schwenk102  - this is a _draft_ page for handling linking to the User Agreement from the Globus InfoLink fields as part of use cases for accessing via protected data group and for consortium access to non-published datasets.

@computationdoc  and I will work on refining with the Policy WG chair prior to release.


pecan88:
Yes -
@ngehlenborg  - we should use the most recently agreed-upon text which would obviate the request above.  We can close I believe and follow the Portal-displayed templates agreed upon.

mccalluc:
>  We can close

Closing PR and deleting branch.

PR #833, opened 2020-07-19
update ingest-validation-tools

mccalluc:
This issue https://github.com/hubmapconsortium/ingest-validation-tools/issues/340 suggested that we'd like the same ability to show the descriptions for sample metadata fields as for datasets. Updating the submodule to makes it available here: Samples aren't handled any differently than Datasets in this regard. 

... but, right now, I can't find any samples with metadata. ... Wait to merge till we've seen it work.


mccalluc:
Submodule update was handled with latest deployment.

PR #843, opened 2020-07-20
Image Pyramids

ilan-gold:
This should give us a generic view configuration for image pyramids that will allow users (especially Ivan) to spot check runs immediately.  This also cleans up the CODEX stuff a bit to match more what we have now for pyramids.

Here are some uuids to check that the view config works:

AF: ec27c2b6b5f177c0ca6387120d8736b0 (should work)
CODEX: 53f19f543429a03af5a6c0e333f593de (getting a 403 and not sure why, but the configuration looks right)
PAS: 2ce183e38a31ff1d836675d3b75654b0 (part of the image is cutoff, a known bug, and the colors are wrong which is because we don't have rgb handling yet in vitessce)

### DIFF
--- a/context/app/api/vitessce.py
+++ b/context/app/api/vitessce.py
@@ -7,13 +7,6 @@ import copy
 
 from flask import current_app
 
-# Hardcoded CODEX offsets and tile path.
-CODEX_OFFSETS_PATH = "output_offsets"
-CODEX_TILE_PATH = "output/extract/expressions/ome-tiff"
-CODDEX_SPRM_PATH = "output_json"
-
-SCRNASEQ_BASE_PATH = "cluster-marker-genes/output/cluster_marker_genes"
-
 SCATTERPLOT = {
     "layers": [],
     "name": "NAME",
@@ -34,7 +27,7 @@ SCATTERPLOT = {
     ],
 }
 
-IMAGING = {
+TILED_SPRM_IMAGING = {
     "name": "NAME",
     "layers": [],
     "staticLayout": [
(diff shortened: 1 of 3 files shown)


mccalluc:
@ilan-gold : If you can include the environment when citing UUIDs... or just use a link? Thanks! Reviewing now...

ilan-gold:
> @ilan-gold : If you can include the environment when citing UUIDs... or just use a link? Thanks! Reviewing now...

Sorry - it is on test.

context/app/api/vitessce.py
```diff
@@ -268,12 +327,15 @@ def _build_image_schema(self, rel_path):
...
+        schema["url"] = self._build_assets_url(image_rel_path)
         schema["metadata"] = {}
-        offsets_path = Path(CODEX_OFFSETS_PATH) / Path(
-            Path(rel_path).name.replace("ome.tiff", "offsets.json")
+        imaging_paths = IMAGING_PATHS[self.assay_type]
+        offsets_path = (
+            image_rel_path.replace(imaging_paths["image"], imaging_paths["offsets"])
+            .replace("ome.tiff", "offsets.json")
+            .replace("ome.tif", "offsets.json")
         )
```
mccalluc, line 339:
```suggestion
        offsets_path = re.sub(
            r'ome\.tiff?$', 'offsets.json',
            image_rel_path.replace(imaging_paths["image"], imaging_paths["offsets"]))
```

ilan-gold:
@mccalluc I can't check that CODEX still works because I'm getting 403's.  Does it work for you?  It looks like everything works but I want to be sure.

ilan-gold:
I was able to confirm this works on one dataset.

PR #845, opened 2020-07-20
Mccalluc/pull in prov vis

mccalluc:


context/schema.json
```diff
@@ -0,0 +1,40 @@
+{
```
john-conroy, line 1: This is a copy of the schema now included in the ProvVis directory? Do we need this here? If so, could we make the file name more descriptive if it's not much trouble? Do we have other components which require this schema?

context/app/static/js/components/Detail/ProvVis/index.js
```diff
@@ -0,0 +1,21 @@
...
+  ReactDOM.render(
+    <ProvVis
+      prov={prov}
+      getNameForActivity={getNameForActivity}
+      getNameForEntity={getNameForEntity}
+      renderDetailPane={renderDetailPane}
+    />,
+    element,
+  );
+}
```
john-conroy, line 18: Do we still need to use `ReactDOM.render`? Could we refactor to make this more declarative?

context/app/static/js/components/Detail/ProvVis/tests/ProvData.test.js
```diff
@@ -0,0 +1,171 @@
+import expect from 'expect';
```
john-conroy, line 1: Rename `tests` directory `__tests__`?

john-conroy: We are also using `Name.spec.js ` to name tests elsewhere. I don't have a strong preference on which pattern should be used, but we should be consistent.

context/app/static/js/components/Detail/ProvVis/tests/ProvData.test.js
```diff
@@ -0,0 +1,171 @@
...
+
+import ProvData, { _makeCwlInput, _makeCwlOutput, _expand } from '../src/ProvData';
```
john-conroy, line 3: Does` ../src/ProvData` exist?

mccalluc:
>     * We are moving towards breaking up `components/Detail`. We could put the files inside `components/prov-vis` while keeping `ProvGraph` inside `components/Detail`.

Will do.

>     * We should dynamically import these components to separate them from the main bundle. I can file an issue to do that later.

Thanks!

Working through the other comments...

context/schema.json
mccalluc: Not sure why this ended up here. Deleting, and giving a fuller name for the other copy.

context/app/static/js/components/Detail/ProvVis/index.js
mccalluc: I'll make a branch off of this PR, but I wanted to just replicate the existing behavior for now, in case doing it the right way turns out to be hard for some reason.

context/app/static/js/components/Detail/ProvVis/tests/ProvData.test.js
mccalluc: will do.

context/app/static/js/components/Detail/ProvVis/tests/ProvData.test.js
mccalluc: will fix.

PR #852, opened 2020-07-20
John conroy/prov table dataset chevron

### LINKED ISSUE #835: chevron between derived datasets are missing in provenance table
When a sequence of derived datasets is shown, a chevron should be inserted between the dataset tiles in the provenance table.

image

(Source: https://portal.hubmapconsortium.org/browse/dataset/8adbc2501d124943c07f5c755b563ef0)

john-conroy:
Fixes #835

### DIFF
--- a/context/app/static/js/components/Detail/ProvTableTile/ProvTableTile.jsx
+++ b/context/app/static/js/components/Detail/ProvTableTile/ProvTableTile.jsx
@@ -15,7 +15,7 @@ function ProvTableTile(props) {
 
   return (
     <>
-      {!isFirstTile && ((entity_type === 'Sample' && !isSampleSibling) || entity_type === 'Dataset') && <DownIcon />}
+      {!isFirstTile && !isSampleSibling && entity_type !== 'Donor' && <DownIcon />}
       {entityData && (
         <EntityTile
           uuid={uuid}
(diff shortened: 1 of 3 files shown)


context/app/static/js/components/Detail/ProvTableTile/ProvTableTile.jsx
```diff
@@ -1,20 +1,21 @@
...
+  const { uuid, entity_type, id, isCurrentEntity, isSampleSibling, isFirstTile } = props;
   const { elasticsearchEndpoint } = useContext(DetailContext);
 
   // mapped fields are not included in ancestor object
   const entityData = useEntityData(uuid, elasticsearchEndpoint);
 
   return (
     <>
-      {entity_type === 'Sample' && isNotSibling && <DownIcon />}
+      {!isFirstTile && ((entity_type === 'Sample' && !isSampleSibling) || entity_type === 'Dataset') && <DownIcon />}
```
mccalluc, line 18:
```suggestion
      {!isFirstTile && !isSampleSibling && <DownIcon />}
```
`isSampleSibling` already checks that that tyle is `Sample` so can this be simplified?

john-conroy: We don't want Donors to display the icon and while we don't expect more than one donor, I'd prefer to be overly strict.

john-conroy: I could use `!isFirstTile && !isSampleSibling && entity_type !== 'Donor`, I believe.

PR #856, opened 2020-07-21
Mccalluc/simplify client py

### LINKED ISSUE #561: Fix lots of copy-and-paste in client.py
Essentially the same code for making an http request and possibly aborting now exists in three copies:
- GET
- POST
- GET, with longer timeout?

Tidy it up.

mccalluc:
Fix #561


mccalluc:
> Looks good. Do you think we lose any clarity by not explicitly distinguishing request methods in the function name or parameters?

I feel like it's ok... I'm more concerned about changing the error-handling on one, and forgetting to update the orther.

Issue #792, opened 2020-07-15
Get `npm test` working, and add to travis


mccalluc:
@john-conroy : I hadn't realized we weren't actually running this from travis right now... but you said you'd look at JS tests today, so you may already be on this.

In the end we want to make sure test pass locally and on travis.

```
$ npm test

> portal-ui@ test /Users/chuck/github/hubmap/portal-ui/context
> jest

 PASS  app/static/js/helpers/__tests__/tableToDelimitedString.spec.js
 FAIL  app/static/js/components/Detail/Attribution/Attribution.spec.js
  ● Test suite failed to run

    /Users/chuck/github/hubmap/portal-ui/context/node_modules/typeface-inter/Inter (web)/Inter.var.woff2:1
    wOF2


    SyntaxError: Invalid or unexpected token
        at compileFunction (<anonymous>)

      1 | import { createGlobalStyle } from 'styled-components';
      2 |
    > 3 | import InterVariable from 'typeface-inter/Inter (web)/Inter.var.woff2';
        | ^
      4 |
      5 | export default createGlobalStyle`
      6 |     @supports (font-variation-settings: normal) {

      at Runtime._execModule (node_modules/jest-runtime/build/index.js:1166:56)
      at Object.<anonymous> (app/static/js/fonts.js:3:1)
      at Object.<anonymous> (app/static/js/components/Providers.jsx:8:1)
      at Object.<anonymous> (test-utils/functions.js:5:1)
      at Object.<anonymous> (app/static/js/components/Detail/Attribution/Attribution.spec.js:3:1)

 FAIL  app/static/js/components/Detail/Dataset/Dataset.spec.js
```

john-conroy:
I think I've mentioned before, but perhaps I didn't outline it clearly in an issue. We need to [fix prov-vis](https://github.com/hubmapconsortium/prov-vis/issues/40) before this is possible. There are issues running tests in the portal due to the version of prov-vis we're using not including the schema.json, but upgrading to the version of prov-vis which does include schema.json breaks a lot of the graphs for other reasons.

john-conroy:
Fixed in #871

PR #859, opened 2020-07-21
Mccalluc/npm test

mccalluc:
Turn JS tests on. The prov vis tests I ported over did not run out of the box... So I tried some global search and replace, which helped, but in the end commented out some tests.

I think getting the JS tests back in CI is important... perhaps it would be better just to delete the prov tests that aren't running than to disable them?


mccalluc:
Fix #792.

test.sh
```diff
@@ -64,6 +64,13 @@ pytest
 end pytest
 
 
+start npm-test
+cd context
+npm test
```
john-conroy, line 69:
```suggestion
npm run test
```

mccalluc:
@john-conroy : `node_modules/jest/bin/jest.js test` passes locally, but `npm run test` doesn't: Errors look like:
```
    SyntaxError: Invalid or unexpected token
        at compileFunction (<anonymous>)
      1 | import { createGlobalStyle } from 'styled-components';
      2 | 
    > 3 | import InterVariable from 'typeface-inter/Inter (web)/Inter.var.woff2';
```
Something different about the transpiling process between those two commands? If it's not obvious to you, I would argue for just changing to `jest.js test` for now, to have something in place, and then loop back later to understand why the preferred method doesn't work.

john-conroy:
We probably need a mock. So use `jest.js` test for now and I can triage this later.

PR #866, opened 2020-07-22
CCF no longer opens in new tab

### LINKED ISSUE #864: No longer open CCF-EUI in a separate window
The CCF EUI is now setup to match the styling of the portal and the HuBMAP logo links back to the portal (and also uses same style of login/logout), so it really doesn't need to open in a separate window anymore.

I feel like the changes we have made make the EUI part of the portal now (even though it uses a different code base). If there are other specific minor changes to make for this to be acceptable, let me know.

mccalluc:
Fixes #864 from @bherr2. Nils: From IU's perspective, they'd be happy with this change. I think down the road, you'd prefer that CCF not be full-page -- This doesn't do that, but it doesn't preclude it either.


bherr2:
Yeah, I agree that eventually we'll provide you guys a version that sits more fully within the portal. We have plans for that in Y3, but it is a bigger change than we can handle right now. But for now not opening in a new window would be beneficial I think.

mccalluc:
(Travis error is random segfault from Cypress.)

PR #878, opened 2020-07-23
John conroy/json icon

john-conroy:
Dataset

Screen Shot 2020-07-23 at 2 39 27 PM

Donor

Screen Shot 2020-07-23 at 2 39 56 PM

### DIFF
--- a/CHANGELOG-json-icon.md
+++ b/CHANGELOG-json-icon.md
@@ -1,0 +1,1 @@
+- Replace link to JSON on entity detail pages with icon.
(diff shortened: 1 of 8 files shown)


tsliaw:
I don't know about the pretty-JSON at the moment, but about the icon, I did ideally want the JSON or {} icon since those are the standard for JSON icons and I agree that the current icon is not very indicative of JSON, but those are not available within the material-ui library. If you think it is fine to bring icons outside the library for the JSON, I'll change that later in the mockup after this merge.

john-conroy:
Filed #886

PR #899, opened 2020-07-27
Update donor.md

schwenk102:
Formatting on Portal page https://portal.hubmapconsortium.org/docs/donor#more is not working correctly for the end table and for the links to further down the page.


mccalluc:
The change you make here (adding whitespace) has no relation to the problems you describe.

PR #908, opened 2020-07-27
John conroy/different dua messages

### LINKED ISSUE #443: Implement click-through agreement
@ngehlenborg will deliver text, and specify the logic.

Questions:
- Is this a modal? A bottom of the page gdpr-style click-through? A full page?
- When is this presented? On welcome page? On details?
- I think I've heard that this is about downloading data... Is viewing imagery a "download" in this sense, or do we only mean looking at a list of files and "downloading" there?
- Is it ok if this is only a UI layer, and not actually enforced by the assets API?

john-conroy:
Fixes #443 . 

We only expect the three values handled in `getDUAText`, but do we want to be more cautious than I am here and handle unexpected values?


CHANGELOG-different-dua-messages.md
```diff
@@ -0,0 +1 @@
+- Handle DUA by data access leevel.
```
ngehlenborg, line 1:
```suggestion
- Handle DUA by data access level.
```

context/app/static/js/components/Detail/FileBrowserDUA/FileBrowserDUA.jsx
```diff
@@ -21,28 +27,36 @@ function FileBrowserDUA(props) {
...
+          <p>{appropriateUse}</p>
 
           <h2>Acknowledgement</h2>
           <p>
             Investigators using HuBMAP data in publications or presentations are requested to cite The Human Body at
-            Cellular Resolution: the NIH Human BioMolecular Atlas Program (doi:10.1038/s41586-019-1629-x) and to include
-            an acknowledgement of HuBMAP. Suggested language for such an acknowledgment is: “The results here are in
-            whole or part based upon data generated by the NIH Human BioMolecular Atlas Program (HuBMAP):{' '}
-            <a href="https://hubmapconsortium.org">https://hubmapconsortium.org</a>.”
+            Cellular Resolution: the NIH Human BioMolecular Atlas Program (doi:
```
mccalluc, line 40: As it's the name of an article should "The Human Body at Cellular Resolution: the NIH Human BioMolecular Atlas Program" be quoted?

context/app/static/js/components/Detail/FileBrowserDUA/FileBrowserDUA.jsx
```diff
@@ -21,28 +27,36 @@ function FileBrowserDUA(props) {
...
+              https://hubmapconsortium.org
+            </LightBlueLink>
+            .”
           </p>
 
           <h2>Data Sharing Policy</h2>
           <p>
             The HuBMAP Data Sharing Policy can be found at{' '}
-            <a href="https://hubmapconsortium.org/policies/">https://hubmapconsortium.org/policies/</a>.
+            <LightBlueLink href="https://hubmapconsortium.org/policies/" target="_blank" rel="noopener noreferrer">
```
mccalluc, line 56: Late in the game to ask this, but what is wrong with plain `<a>`?

context/app/static/js/components/Detail/FileBrowserDUA/style.js
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components';
+
+const ObliqueSpan = styled.span`
```
mccalluc, line 3: Similarly, what would be wrong with `<i>`?

john-conroy:
I'll loop back to address the questions. I also need to file an issue to handle the italic/oblique font inside the different text sections.

context/app/static/js/components/Detail/FileBrowserDUA/FileBrowserDUA.jsx
john-conroy: Need to use MUI components if we want font styles from the palette by default.

PR #913, opened 2020-07-27
Mccalluc/docs as submodule

mccalluc:
Right now, `push.sh` updates all the submodules _after_ the image is taken, to make sure we don't fall behind, but to also make sure we're not putting changes into the image we haven't been able to see in our local environments.

This is relatively low risk, and it means that they don't need me reviewing PRs, and I don't need to review PRs.

It doesn't address image asset management, differences in markdown dialects, or ToC generation... but doesn't preclude improvements there, either.

I've moved documentation issues to the new repo.


mccalluc:
I've moved all `documentation` issues to the new repo, and removed that tag here.

PR #917, opened 2020-07-28
Mccalluc/try different md engine

### LINKED ISSUE #900: Sub headers in markdown should get anchor links
Particularly for Jonathan's donor data, having links within the page is important. Perhaps a different markdown library would offer this by default?

mccalluc:
Swap out markdown engines. The new one has a tiny user base, but it
- Is a thin wrapper around `marked` with no extra dependencies.
- Has been updated more recently than other wrappers.
- `marked` itself is very popular, but is not a react component.
- `marked` puts an ID attribute on titles, so the sub-sections in the donor metadata page work. (Primary motivation.)

In addition: `Preview` and `Markdown` pages now get the same styling rules.

In the long run, I think we don't want components referencing other component's internals... but I'm ok doing it for a little while: Identify the parts that are being reused, and then pull them up and out.

Fix #900.


mccalluc:
Good catch... I had thought styles were coming through, but now they definitely are not... Investigating...

mccalluc:
I'm also noticing npm errors, possibly because of problems in the dependency?
```
./node_modules/@jnbelo/react-marked/lib/MarkedViewer.js
Module Warning (from ./node_modules/source-map-loader/index.js):
(Emitted value instead of an instance of Error) Cannot find source file '../src/MarkedViewer.tsx': Error: Can't resolve '../src/MarkedViewer.tsx' in '/Users/chuck/github/hubmap/portal-ui/context/node_modules/@jnbelo/react-marked/lib' client:135
./node_modules/@jnbelo/react-marked/lib/index.js
Module Warning (from ./node_modules/source-map-loader/index.js):
(Emitted value instead of an instance of Error) Cannot find source file '../src/index.ts': Error: Can't resolve '../src/index.ts' in '/Users/chuck/github/hubmap/portal-ui/context/node_modules/@jnbelo/react-marked/lib'
```

ilan-gold:
Sorry I am just seeing the question at the end.  I think that would be fine to do.

john-conroy:
Agreed.

mccalluc:
Replaced with new PR.

PR #918, opened 2020-07-28
John conroy/dua before globus

### LINKED ISSUE #915: User must confirm DUA before accessing Globus files

john-conroy:
Fixes #915.

I opted to pass DUA variables by context instead of props, context providers which aren't updated shouldn't negatively impact performance and are cleaner than passing props to far away children.

A link without an href isn't keyboard accessible. Reading around many believe using a void or similar href is bad practice. I opted to use a button although the component names are now a little confusing. Any thoughts on the subject?

Example with file browser and globus link: http://localhost:5001/browse/dataset/63a35af025a3662cb368eb3999fe2948

### DIFF
--- a/context/app/static/js/components/Detail/FilesConditionalLink/style.js
+++ b/context/app/static/js/components/Detail/FilesConditionalLink/style.js
@@ -1,0 +1,8 @@
+import styled from 'styled-components';
+import { LightBlueLink } from 'shared-styles/Links';
+
+const AlignedLink = styled(LightBlueLink)`
+  vertical-align: baseline;
+`;
+
+export { AlignedLink };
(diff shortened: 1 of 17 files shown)


context/app/static/js/components/Detail/FilesConditionalLink/style.js
```diff
@@ -1,8 +1,8 @@
 import styled from 'styled-components';
 import { LightBlueLink } from 'shared-styles/Links';
 
-const StyledLink = styled(LightBlueLink)`
-  cursor: pointer;
+const AlignedLink = styled(LightBlueLink)`
+  vertical-align: baseline;
```
mccalluc, line 5: I can't recall now why `pointer` was added, but if it's no longer needed, great!

john-conroy: You needed pointer at the time because the element was an anchor without an href.

PR #920, opened 2020-07-28
Tweak results table so more clear the whole thing is the result, not separate links

### LINKED ISSUE #701: Table links in Donor, Sample, Dataset take you to the same place
## Feedback

__Help Desk Issue Tracker URL__: https://on.spiceworks.com/tickets/open/593/activity

__Message__:
*See Ticketbove for referenced image*

On the tables across Donor, Sample and Datasets tabs, each of the links for the same data set take you to the same dataset page, regardless if you pick ID, Group, specimen type, organ, last modified etc…   This seem redundant and overkill. At first I thought that when clicking the “group” hyperlink, it would have taken me a to a description of that group and all their datasets, or likewise clicking the hyperlink in the organ section would have take me to all the “thymus” data for example. If I wanted to get to the actual data set, that’s when I would have clicked on the ID itself

mccalluc:
Fix #701, and perhaps another one I can't find now: I have this memory that someone was really frustrated that the different "links" all went to the same place.  This is an ad-hoc solution, and may be replaced.

Screen Shot 2020-07-27 at 10 47 42 PM

### DIFF
--- a/context/app/static/js/components/Search/SearchWrapper.jsx
+++ b/context/app/static/js/components/Search/SearchWrapper.jsx
@@ -51,7 +51,7 @@ function makeTableComponent(resultFields, detailsUrlPrefix, idField) {
     const { hits } = props;
     /* eslint-disable no-underscore-dangle */
     return (
-      <table className="sk-table sk-table-striped" style={{ width: '100%' }}>
+      <table className="sk-table sk-table-striped">
         <thead>
           <tr>
             {resultFields.map((field) => (
@@ -64,9 +64,7 @@ function makeTableComponent(resultFields, detailsUrlPrefix, idField) {
             <tr key={hit._id}>
               {resultFields.map((field) => (
                 <td key={field.id}>
-                  <a href={detailsUrlPrefix + hit._source[idField]} style={{ display: 'block' }}>
-                    {getByPath(hit._source, field)}
-                  </a>
+                  <a href={detailsUrlPrefix + hit._source[idField]}>{getByPath(hit._source, field)}</a>
                 </td>
               ))}
             </tr>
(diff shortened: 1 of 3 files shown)


context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -51,7 +51,7 @@ function makeTableComponent(resultFields, detailsUrlPrefix, idField) {
     const { hits } = props;
     /* eslint-disable no-underscore-dangle */
     return (
-      <table className="sk-table sk-table-striped" style={{ width: '100%' }}>
```
ngehlenborg, line 54: Is the width setting being removed for a particular reason?

ngehlenborg:
Ok, we can discuss color scheme separately.

context/app/static/js/components/Search/SearchWrapper.jsx
mccalluc: It just got moved to the SCCS file to tidy things up.

mccalluc:
I'm not sure how best to handle this: Because we're already darkening every other row by default, add an extra 8% on hover makes it almost but not quite the same color as its neighbor rows, while hovering on the other rows is very dark. Perhaps get rid of the even-odd coloring?

Screen Shot 2020-07-28 at 3 35 24 PM

Without row striping and and with tr border, it looks like this:

Screen Shot 2020-07-28 at 3 52 09 PM

mccalluc:
(Got a gyp error on travis: Restarting build.)

mccalluc:
@tsliaw : Good catch! Fixed now. Will ping Nils on Slack.

Issue #914, opened 2020-07-27
check appropriate use statement text contrast for WCAG compliance


ngehlenborg:
Confirm that lightness contrast of text in this view is sufficient for at least WCAG AA (e.g. via https://webaim.org/resources/contrastchecker/):

<img width="669" alt="image" src="https://user-images.githubusercontent.com/1216518/88600876-45bfc680-d03d-11ea-9e04-3235591c9b54.png">

Please change font color to make it compliant if it is not yet.

PR #921, opened 2020-07-28
John conroy/dua typography

john-conroy:
`DialogContentText` is just a paragraph element with some default styles so we were getting a number of DOM nesting errors because we were placing headers inside of a paragraph.  Switched to using typography components to get font styles from the theme. I also opted to use our main text color instead of the secondary gray from our palette because the new font weights heightened the contrast issues discussed in #914 .

Screen Shot 2020-07-28 at 9 52 35 AM

### DIFF
--- a/context/app/static/js/components/Detail/FileBrowserDUA/FileBrowserDUA.jsx
+++ b/context/app/static/js/components/Detail/FileBrowserDUA/FileBrowserDUA.jsx
@@ -3,13 +3,13 @@ import Button from '@material-ui/core/Button';
 import Dialog from '@material-ui/core/Dialog';
 import DialogActions from '@material-ui/core/DialogActions';
 import DialogContent from '@material-ui/core/DialogContent';
-import DialogContentText from '@material-ui/core/DialogContentText';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Checkbox from '@material-ui/core/Checkbox';
+import Typography from '@material-ui/core/Typography';
 
 import { LightBlueLink } from 'shared-styles/Links';
 import { getDUAText } from './utils';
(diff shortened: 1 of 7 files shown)


john-conroy:
@tsliaw see #443

context/app/static/js/components/Detail/FileBrowserDUA/FileBrowserDUA.jsx
```diff
@@ -26,16 +26,20 @@ function FileBrowserDUA(props) {
...
-          <h1 id="alert-dialog-title">
+        <StyledDiv id="alert-dialog-description">
+          <StyledHeader id="alert-dialog-title" variant="h2" component="h1">
             {'HuBMAP '} <ObliqueSpan>{`${title} Data`}</ObliqueSpan> {' Usage'}
-          </h1>
+          </StyledHeader>
 
-          <h2>Appropriate Use</h2>
-          <p>{appropriateUse}</p>
+          <StyledHeader variant="h4" component="h2">
```
mccalluc, line 34: This has been confusing me... if `<h1>` and `<h2>` don't give us the control we need, could we define `<H1>` and `<H2>`? I don't know how we know which variant is used.

john-conroy: Variant determines the styles used from the theme and the component (if a component prop isn't defined). Component overrides the variant component and is the element used in the DOM.

john-conroy: `Typography` allows us to use the font styles defined in theme easily. This is just the MUI way of doing things, we can find an alternative if it doesn't suit us?

PR #930, opened 2020-07-29
Update ccf-eui base href

bherr2:
We have changed our method of delivering CCF EUI releases to use a tagged versioning system. This PR just updates the base href it needs to point to.


mccalluc:
@bherr2 : I'll go ahead and merge. If that should be the default going forward, make a note in each PR, and we'll handle it.

bherr2:
> @bherr2 : I'll go ahead and merge. If that should be the default going forward, make a note in each PR, and we'll handle it.

Ok, thanks. Will do!

PR #939, opened 2020-07-29
Add snackbar to show warning messages emitted from Vitessce

keller-mark:
This pull request adds a close-able `<Snackbar/>` component for displaying warning messages from Vitessce, such as `Error HTTP status fetching cells.` or `Error while validating cells.`  These messages are emitted from the `onWarn` callback on the `<Vitessce/>` component. Right now these warnings only appear in the console.

This PR can be tested by temporarily changing line 411 of vitessce.py https://github.com/hubmapconsortium/portal-ui/blob/971e0ee9e82d674fb9acc6a54d5dd6ed6f0fca9a/context/app/api/vitessce.py#L411

to `base_url = urllib.parse.urljoin(assets_endpoint, f"{self.uuid}/{rel_path}_BAD_URL")`

(I also saw that there is now an ErrorBoundary component. I tried throwing the warning as an error, but it did not seem to trigger that error handler, but maybe I was doing something wrong. Not sure if that would be possible and preferred over this UI.)

Screen Shot 2020-07-29 at 6 37 41 PM

### DIFF
--- a/context/app/static/js/components/Detail/Visualization/Visualization.jsx
+++ b/context/app/static/js/components/Detail/Visualization/Visualization.jsx
@@ -33,8 +33,7 @@ function Visualization(props) {
 
   const [isExpanded, setIsExpanded] = useState(false);
   const [isEscSnackbarOpen, setIsEscSnackbarOpen] = useState(false);
-  const [isWarnSnackbarOpen, setIsWarnSnackbarOpen] = useState(false);
-  const [vitessceWarning, setVitessceWarning] = useState();
+  const [vitessceWarnings, setVitessceWarnings] = useState([]);
   const [vitessceTheme, setVitessceTheme] = useState('light');
   const [vitessceSelection, setVitessceSelection] = useState(0);
   const [open, toggle] = useReducer((v) => !v, false);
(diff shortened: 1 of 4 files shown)


keller-mark:
Not sure what is going on here with the tests - they pass locally and 3/4 times on travis 
Screen Shot 2020-07-29 at 7 10 28 PM

mccalluc:
huh: Restarted that travis build once, but it still seems to have the same problem: Not connecting to the portal server. I'd try restarting it a couple more times, and/or extending the time limit for it to connect.

mccalluc:
> I also saw that there is now an ErrorBoundary component.

I know there's the caveat that Error Boundaries [don't catch errors inside event handlers](https://reactjs.org/docs/error-boundaries.html#how-about-event-handlers)... but that might not be related at all to what you're describing. In any case, it's absolutely a good thing to have lower lever error handling too: If there's a problem in Vitessce we don't want the whole window to the cleared: Thanks for doing this!

context/app/static/js/components/Detail/Visualization/Visualization.jsx
```diff
@@ -97,20 +107,35 @@ function Visualization(props) {
...
+            <Alert severity="warning" variant="filled" onClose={() => setIsWarnSnackbarOpen(false)}>
+              {vitessceWarning}
+            </Alert>
+          </WarnSnackbar>
           <Vitessce
             config={vitData[vitessceSelection] || vitData}
             theme={vitessceTheme}
             height={isExpanded ? null : vitessceFixedHeight}
+            onWarn={(warning) => {
+              setVitessceWarning(warning);
```
mccalluc, line 136: I'm wondering about the case where multiple warnings come in together... perhaps they should be added to an array, and closing the snackbar clears and resets the array?

keller-mark: I have added the array, and closing snackbar removes a particular warning from the array so that the next warning can be shown if there is more than one

keller-mark:
Thanks, I have changed the styling and variable names from "warning" to "error" since the messages do contain the word "Error"

PR #943, opened 2020-07-30
John conroy/update footer

### LINKED ISSUE #919: Make no underline a default for links
Better not to repeatedly specify `underline="none"` throughout.

### LINKED ISSUE #936: Implement Footer
See #904 for design

john-conroy:
Fix #936. Fix #919. 

- Footer is now below the fold for collections page.
- No link exists for 'Citing HuBMAP' in footer.

Screen Shot 2020-07-30 at 4 09 29 PM

### DIFF
--- a/CHANGELOG-update-footer.md
+++ b/CHANGELOG-update-footer.md
@@ -1,0 +1,1 @@
+- Update footer per 07/29/2020 design.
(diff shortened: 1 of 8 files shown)


john-conroy:
Do we want to think about representing the links with different functionalities differently? So someone doesn't click a link with the expectation of visiting another website/page and instead downloads a file/opens their mail application?

tsliaw:
Typically if it's an email link, I would like to have the email logo next to it, but because I expect to have some sort of "Contact Us" page in the future to replace that link, I think what we currently have is fine for the footer. But until then actually, is is possible to have the email logo to the right of contact in the blue link color? Feel free to do this in a different pr though. I'll file an issue for this so I can make sure there's no other places in the portal with this issue.

PR #953, opened 2020-07-31
No more currently selected filters at top

### LINKED ISSUE #951: No tag-box for currently selected facets
Tiffany reports that they are not part of the design.

mccalluc:
Fix #951 
Screen Shot 2020-07-31 at 1 15 14 PM

@tsliaw : I won't block, but I am really uncomfortable with this. It's going to be so easy for the user to loose track of what filters are applied. If there is some replacement that you and Nils are envisaging, I would like to get that in now rather than leaving it in this state.

### DIFF
--- a/CHANGELOG-no-selected-filters-at-top.md
+++ b/CHANGELOG-no-selected-filters-at-top.md
@@ -1,0 +1,1 @@
+- Don't show the currently selected filters across the top.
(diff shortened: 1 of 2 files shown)


tsliaw:
@ngehlenborg I know we discussed getting rid of the filter tags at the top, so I wanted to make sure this I understood correctly since there's nothing in the current mockup that is going to replace this.

PR #962, opened 2020-08-03
John conroy/display access level

### LINKED ISSUE #959: Add access level information to dataset detail pages
Dataset detail pages should include "Access Level" information (Public/Consortium/Protected). 

For now, I propose to add this to the top of the page between status and the JSON download link, e.g. "QA | Public Access | <json icon>".

See also #946.

john-conroy:
Fix #959.

### DIFF
--- a/context/app/static/js/components/Detail/Summary/Summary.jsx
+++ b/context/app/static/js/components/Detail/Summary/Summary.jsx
@@ -19,6 +19,7 @@ function Summary(props) {
     status,
     children,
     collectionName,
+    data_access_level,
   } = props;
 
   const createdDate = create_timestamp ? new Date(create_timestamp).toDateString() : 'Undefined';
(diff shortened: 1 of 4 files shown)


context/app/static/js/components/Detail/Summary/Summary.jsx
```diff
@@ -63,6 +64,7 @@ Summary.propTypes = {
   uuid: PropTypes.string.isRequired,
   description: PropTypes.string,
   status: PropTypes.string,
+  data_access_level: PropTypes.string,
```
mccalluc, line 67: Is there a reason not to make this required?

context/app/static/js/components/Detail/SummaryData/SummaryData.jsx
```diff
@@ -1,21 +1,23 @@
...
   return (
     <Flex>
       {children && <FlexCenterAlign>{children}</FlexCenterAlign>}
       <FlexRight>
-        {entity_type === 'Dataset' && status && status.length > 0 && (
+        {entity_type === 'Dataset' && (
           <FlexCenterAlign>
             <StatusIcon status={status} />
             <SummaryItem>{status}</SummaryItem>
+            <SummaryItem>{`${capitalizeString(data_access_level)} Access`}</SummaryItem>
```
mccalluc, line 20: After the next reindex, there will be an uppercased version of this, but we'll simplify this later.

context/app/static/js/components/Detail/Summary/Summary.jsx
john-conroy: We're only passing the `data_access_level` prop in datasets.

PR #968, opened 2020-08-03
Only Show Imaging in SPRM When There are no Segmentations

### LINKED ISSUE #963: SPRM Tile does not Load
On prod 31d0eca8591c5d8592992df452f629be does not load because it expects to have more than just imaging data but there are no segmentations.

There are two issues here:
1.  Foremost, we should not be telling `vitessce` to render things that are not there, like cell segmentations.
2. `vitessce-grid` is not updating properly when the proper view config is passed in.

ilan-gold:
Fixes #963: see the issue for testing this PR.  Beyond that, no other Vitessce changes should be noticeable.

### DIFF
--- a/context/app/api/vitessce.py
+++ b/context/app/api/vitessce.py
@@ -61,6 +61,23 @@ TILED_SPRM_IMAGING = {
     ],
 }
 
+TILED_SPRM_IMAGING_ONLY = {
+    "name": "NAME",
+    "layers": [],
+    "staticLayout": [
+        {"component": "layerController", "x": 0, "y": 0, "w": 3, "h": 4},
+        {"component": "description", "x": 0, "y": 4, "w": 3, "h": 2},
+        {
+            "component": "spatial",
+            "props": {"view": {}},
+            "x": 3,
+            "y": 0,
+            "w": 9,
+            "h": 6,
+        },
+    ],
+}
+
 IMAGING_ONLY = {
     "name": "NAME",
     "layers": [],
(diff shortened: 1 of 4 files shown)


context/app/api/vitessce.py
```diff
@@ -285,8 +302,13 @@ def _build_vitessce_conf(self):
...
             for tile in sorted(found_tiles):
-                new_conf = copy.deepcopy(conf)
-                layers = [self._build_layer_conf(file, tile) for file in files]
+                # If there are no cell segmentations, remove the non-imaging parts of the assay.
+                if f"{CODDEX_SPRM_PATH}/{tile}.cell-sets.json" not in file_paths_found:
+                    new_conf = copy.deepcopy(TILED_SPRM_IMAGING_ONLY)
+                    layers = [self._build_layer_conf(files[0], tile)]
+                else:
+                    new_conf = copy.deepcopy(conf)
+                    layers = [self._build_layer_conf(file, tile) for file in files]
```
mccalluc, line 311:
```suggestion
                if f"{CODDEX_SPRM_PATH}/{tile}.cell-sets.json" not in file_paths_found:
                    new_conf = copy.deepcopy(TILED_SPRM_IMAGING_ONLY)
                    files_for_layers = [files[0]]
                else:
                    new_conf = copy.deepcopy(conf)
                    files_for_layers = files
                layers = [self._build_layer_conf(file, tile) for file in files_for_layers]
```
slight preference to pull this out...

PR #969, opened 2020-08-03
John conroy/setup google analytics - track uuids as pages

john-conroy:
Initializes google analytics and tracks page views. Search pages are tracked with `entity_type[0]` param, other params are not tracked with page views. I could use your opinion on which of the following you'd prefer (or neither).

**Track uuids as pages** (this PR)

Entity detail pages are tracked with uuids.

**Track uuids as events** (#972)

Entity detail pages are tracked without uuids and are instead tracked separately not as page views, but as an event.


john-conroy:
Travis is failing because we're exceeding max log length. Filed #974/

PR #972, opened 2020-08-03
John conroy/setup google analytics track uuids as events

john-conroy:
Initializes google analytics and tracks page views. Search pages are tracked with entity_type[0] param, other params are not tracked with page views. I could use your opinion on which of the following you'd prefer (or neither).

Track uuids as events (this PR)

Entity detail pages are tracked without uuids and are instead tracked separately not as page views, but as an event.

~~Track uuids as pages (#969)~~

~~Entity detail pages are tracked with uuids.~~

Talked to Nils and he prefers tracking uuids as events.


context/app/static/js/components/Routes/useSendPageView.js
```diff
@@ -0,0 +1,21 @@
+import { useEffect } from 'react';
+import ReactGA from 'react-ga';
+
+function useSendPageView(path) {
+  useEffect(() => {
+    if (path.startsWith('/browse')) {
+      const pathWithoutUUID = path.match(/\/\w+\/\w+/)[0];
```
mccalluc, line 7:
```suggestion
      const pathWithoutUUID = path.match(/\/browse\/\w+/)[0];
```

context/app/static/js/components/App.jsx
```diff
@@ -7,6 +9,8 @@ import { Header } from './Header';
 
 function App(props) {
   const { flaskData } = props;
+  ReactGA.initialize('UA-133341631-3');
```
mccalluc, line 12: Did we get the access we want? In the future, might want to pull this into app.conf, but probably not right now.

john-conroy:
@mccalluc Looking at [google's reference docs ](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference) the sample rate is only 1% by default for measuring site speed. The default sample rate for tracking users is 100%.

context/app/static/js/components/App.jsx
ngehlenborg: I agree with that but it is indeed not urgent. We did get the access that we asked for.

PR #983, opened 2020-08-04
John conroy/pre commit lint

### LINKED ISSUE #973: Add pre-push hook to prettier write js files
Travis is failing from prettier more often since we moved eslint/prettier check outside of the webpack build. We'd like to catch prettier errors sooner.

john-conroy:
Fixes #973 . Opted for a pre-commit prettier write then lint check only on staged files. We could switch it to lint fix if we're comfortable with that.

Bypass pre-commit hook by using `HUSKY_SKIP_HOOKS=1 git <command> ...`.


context/package.json
```diff
@@ -82,5 +84,13 @@
     "test:coverage": "jest --coverage",
     "lint": "eslint -c .eslintrc.yml --ext .js,.jsx app/static/js/",
     "lint:fix": "eslint -c .eslintrc.yml --ext .js,.jsx --fix app/static/js/"
+  },
+  "lint-staged": {
+    "app/static/js/**/*.{js,jsx}": [  "prettier --write", "eslint -c .eslintrc.yml"]
```
mccalluc, line 89: Would the changes from `prettier` be a subset of those from `eslint --fix`? Why not the latter?

john-conroy: If you're comfortable with `eslint --fix` that would be the only command needed.

john-conroy: @mccalluc I'm going to merge, but let me know if you're comfortable using `eslint --fix` and I'll make another PR.

Issue #982, opened 2020-08-04
Handle multiple data_types when creating Vitessce config


mccalluc:
```
         # Can there be more than one of these?  This seems like a fine default for now.
        self.assay_type = entity["data_types"][0]
```

There can be more than one: We will be seeing `image_pyramid` in addition to other codes.

_Originally posted by @mccalluc in https://github.com/hubmapconsortium/search-api/issues/123#issuecomment-668766850_

mccalluc:
Not needed: We've concluded that it will be single-valued, so we will have more control over how the names are built.

PR #985, opened 2020-08-04
Start to handle image_pyramid

mccalluc:
Towards #982... Not sure what the behavior should be if it fails? @ilan-gold : Should we be concerned about this?

### DIFF
--- a/context/app/api/vitessce.py
+++ b/context/app/api/vitessce.py
@@ -234,9 +234,8 @@ class Vitessce:
           >> vitessce = Vitessce(entity, nexus_token)
 
         """
-
-        # Can there be more than one of these?  This seems like a fine default for now.
-        self.assay_type = entity["data_types"][0]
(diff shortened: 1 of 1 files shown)


ilan-gold:
This make sense but I use the `image_pyramid` type a  bunch.  An entire configuration is based on it.  I think the solution is to just use the list as is, as a data structure.

mccalluc:
@ilan-gold : I'm concerned about relying on a particular ordering in the `data_types` list... It's been stable, but I think it's the sort of thing that they could change without realizing there's a downstream dependency.

Would you like me to continue this work, trying to take into account your concerns? Or would you like to handle it?

ilan-gold:
@mccalluc It's up to you.  I'm not too worried, but I agree things could change without us knowing so changing this should be a priority.  I can look at this first-thing after all this Viv stuff is wrapped up - should be tomorrow or Wednesday.

ilan-gold:
I think you can handle this if done carefully, and I can try to support you going through it if you want to take it on.

Issue #977, opened 2020-08-04
IMC image pyramids fail to load in Vitessce 0.2.2 in Safari and Chrome


ngehlenborg:
See https://portal.test.hubmapconsortium.org/browse/dataset/b81ca02627280283e791c82d8e3f8e09.

It loads fine in Firefox but in Safari and Chrome the process indicator never disappears and this error message appears in the console after a while:

<img width="1374" alt="image" src="https://user-images.githubusercontent.com/1216518/89295549-2e7d6c00-d62f-11ea-839f-e4df5cdf130c.png">

ilan-gold:
See https://github.com/hubmapconsortium/vitessce-image-viewer/issues/236

PR #987, opened 2020-08-04
Upgrade and handle IMC.

### LINKED ISSUE #978: make IMC image pyramids fit the viewport better upon load
See https://portal.test.hubmapconsortium.org/browse/dataset/b81ca02627280283e791c82d8e3f8e09 (on Firefox until #977 has been fixed).

Currently upon loading:
image

Should ideally look like this:
image

ilan-gold:
Fixes #978 and #977 

I know you opened #982 and #985 @mccalluc, which goes somewhat against what I have here, so I will handle that tomorrow.  There are some deeper fixes that I could put into it.  Thanks for being on top of this.

Using the PROD conf I sent yesterday, you should be able to see:
http://localhost:5001/browse/dataset/912889b51fed4ac4fdf2d4d3c5550f72
http://localhost:5001/browse/dataset/46a69f99d8793f9ddf8ee4a1dba5d0e3
for example.

Nothing else should change.

### DIFF
--- a/context/app/api/vitessce.py
+++ b/context/app/api/vitessce.py
@@ -470,6 +470,8 @@ class Vitessce:
 
         # IMS images are named under this convention, "IMS_XXXMode".
         # Some of the non IMS images contain the substring "IMS," so that is insufficient.
+        # IMC also needs to be handled specially.
+        # TODO: Clean this up by handling multiple assay types (#982).
         if any(
             [
                 "IMS_PosMode" in file["rel_path"]
@@ -481,4 +483,3 @@ class Vitessce:
             conf["staticLayout"][-1]["props"]["view"]["zoom"] = -2
             conf["staticLayout"][-1]["props"]["view"]["target"] = [1000, 1000, 0]
         return conf
-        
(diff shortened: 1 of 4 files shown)


context/app/api/vitessce.py
```diff
@@ -470,9 +470,13 @@ def _replace_view(self, conf):
 
         # IMS images are named under this convention, "IMS_XXXMode".
         # Some of the non IMS images contain the substring "IMS," so that is insufficient.
+        # IMC also needs to be handled specially.
+        # TODO: Clean this up by handling multiple assay types (#982).
         if any(
             [
```
mccalluc, line 476:
```suggestion
```
Don't make a list if a generator suffices.

context/app/api/vitessce.py
```diff
@@ -470,9 +470,13 @@ def _replace_view(self, conf):
...
-                "IMS_PosMode" in file["rel_path"] or "IMS_NegMode" in file["rel_path"]
+                "IMS_PosMode" in file["rel_path"]
+                or "IMS_NegMode" in file["rel_path"]
+                or "IMC" in self.entity["data_types"]
                 for file in self.entity["files"]
             ]
```
mccalluc, line 481:
```suggestion
```

PR #988, opened 2020-08-04
HBM redirect. Fix #844

### LINKED ISSUE #844: Redirect from "HBM###.ABCD.###"
Users often supply the HBM IDs: Redirect from those to UUIDs.

mccalluc:


context/app/routes_main.py
```diff
@@ -83,6 +83,13 @@ def collection_details_ext(uuid, ext):
     return collection
 
 
+@blueprint.route('/browse/<type>/HBM<hbm_suffix>')
```
ngehlenborg, line 86: Is there a way to do this without knowing the type? That would greatly simplify things for everyone.

context/app/api/client.py
```diff
@@ -56,7 +56,17 @@ def get_entity(self, uuid):
                 'description': 'Mock Entity'
             }
 
-        query = {'query': {'ids': {'values': [uuid]}}}
+        if uuid is not None and hbm_id is not None:
+            raise Exception('Either UUID or HBM ID should be provided')
```
john-conroy, line 60: We want to raise an exception when both values exist? Or do we want to raise an exception when neither values exist?

context/app/routes_main.py
mccalluc: Sure... I think I set it up this way for consistency with the others, but not necessary. (If you provide the wrong one, it will still redirect.)

context/app/api/client.py
mccalluc: If neither is provided, you'll get an error downstream... If both are provided, the function is being misused, and that's what this error is for. Rephrase error message? `Only UUID or HBM ID should be provided, not both`?

PR #1002, opened 2020-08-05
Remove access group facet

### LINKED ISSUE #999: "Access Group" facet needs to be removed from Donor and Sample search
"Access Group" facet needs to be removed from Donor and Sample search.

mccalluc:
Fix #999. @ngehlenborg : Datasets (and datasets only) have the distinct `mapped_data_access_level` facet: This is correct?


ngehlenborg:
@mccalluc Yes, I believe only datasets have a `data_access_level` defined.

PR #1003, opened 2020-08-05
John conroy/detail summary data tests

john-conroy:
Adds basic tests to `SummaryData` component for entity detail pages.

### DIFF
--- a/context/app/static/js/components/Detail/SummaryData/SummaryData.spec.js
+++ b/context/app/static/js/components/Detail/SummaryData/SummaryData.spec.js
@@ -5,12 +5,12 @@ import SummaryData from './SummaryData';
 
 test('json button exists and has href', () => {
   // const { entity_type, uuid, status, mapped_data_access_level, children } = props;
-  const { getByTestId } = render(
+  const { getByRole } = render(
     <SummaryData entity_type="Fake" uuid="fakeuuid" status="QA" mapped_data_access_level="Public" />,
   );
-  expect(getByTestId('summary-data-json-button')).toBeInTheDocument();
-  expect(getByTestId('summary-data-json-button')).not.toBeEmptyDOMElement();
-  expect(getByTestId('summary-data-json-button')).toHaveAttribute('href', `/browse/fake/fakeuuid.json`);
+  expect(getByRole('link')).toBeInTheDocument();
+  expect(getByRole('link')).not.toBeEmptyDOMElement();
+  expect(getByRole('link')).toHaveAttribute('href', `/browse/fake/fakeuuid.json`);
 });
 
 test('dataset displays properly', () => {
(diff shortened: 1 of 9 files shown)


context/app/static/js/components/Detail/SummaryData/SummaryData.spec.js
```diff
@@ -0,0 +1,55 @@
...
+  expect(queryByTestId('status-svg-icon')).toBeNull();
+});
+
+test('children display when provided', () => {
+  const { getByText, getByTestId } = render(
+    <SummaryData entity_type="fake" uuid="fakeuuid" status="QA" mapped_data_access_level="Public">
+      <>
+        <div>child 1</div>
+        <div>child 2</div>
+      </>
```
mccalluc, line 37:
```suggestion
      <div>child 1</div>
      <div>child 2</div>
```
There is a parent, so do we need the empty tags?

context/app/static/js/components/Detail/SummaryData/SummaryData.spec.js
```diff
@@ -0,0 +1,55 @@
...
+import { render } from 'test-utils/functions';
+import SummaryData from './SummaryData';
+
+test('json button exists and has href', () => {
+  const { getByRole } = render(
+    <SummaryData entity_type="Fake" uuid="fakeuuid" status="QA" mapped_data_access_level="Public" />,
+  );
+  expect(getByRole('link')).toBeInTheDocument();
+  expect(getByRole('link')).not.toBeEmptyDOMElement();
+  expect(getByRole('link')).toHaveAttribute('href', `/browse/fake/fakeuuid.json`);
```
mccalluc, line 12:
```suggestion
  expect(getByRole('link')).toHaveAttribute('href', `/browse/fake/fakeuuid.json`);
```
Would the error on failure be too cryptic without these?

context/app/static/js/components/Detail/SummaryData/SummaryData.spec.js
```diff
@@ -0,0 +1,55 @@
...
+});
+
+test('children do not display when undefined', () => {
+  const { queryByText, queryByTestId } = render(
+    <SummaryData entity_type="fake" uuid="fakeuuid" status="QA" mapped_data_access_level="Public" />,
+  );
+  expect(queryByTestId('summary-data-parent')).toBeNull();
+
+  const textToTest = ['child 1', 'child 2'];
+  textToTest.forEach((text) => expect(queryByText(text)).toBeNull());
```
mccalluc, line 54:
```suggestion
```
Maybe there's a way to check that it's empty? There's nothing special about 'child 1'... Could just as well say 'arbitrary-string-we-haven't-mentioned-anywhere'.

john-conroy: @mccalluc I am checking its emptiness directly above. While there isn't anything special about 'child 1', I am checking that text is being displayed. I guess I don't understand the issue?

john-conroy: @mccalluc Sorry, I misunderstood the test on which you commented. You are correct, I removed the assertion.

PR #1004, opened 2020-08-05
Show search matches

mccalluc:
@tsliaw : If you already have a design can you point me to it, or can you suggest something? Right now it looks like:
Screen Shot 2020-08-05 at 6 50 57 PM

- Perhaps bold the matches instead of italics?
- Make it span the whole row? Or add it as new column?
- (I need to make sure there's not a blank row when there is no match.)
- Highlighting matches in the row we already have would be harder... but not impossible?
- I think I could control the size of the window around the match, if that would be helpful?

After fixing the design, I'll resubmit this for a code review.

([Elasticsearch reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html) for more details.)

(Doesn't seem to be an issue for this? Please reference if there is one.)


tsliaw:
There isn't a current existing design for this. I want to ask if this feature is necessary because most designs showing where the results matched (or tile when that is implemented) will clutter the search page. If this feature is desired, then the best design would be to bold matches within the row (when applicable). If the search matches something that's within the detail page and not as a row entry, then nothing should be bolded. Let me know if that doesn't make sense and I can make a quick mockup of what I mean.

mccalluc:
@ngehlenborg : I've done a few things to distinguish the context rows:
- Grey instead of black text.
- Padding on left on right.
- I tried making the border dotted, but there were artifacts, so in the PR now I've taken off the bottom border.

Any other ideas? Is this sufficient for now?

----

With border:

Screen Shot 2020-08-10 at 4 57 23 PM

Without (what is in the PR now):

Screen Shot 2020-08-10 at 5 06 35 PM

ngehlenborg:
Getting there. Font color and indentation are good, no line is fine, too, but let's change the spacing to show the relationship between match text and table row more clearly:

image

With annotation:
image


(this is based on the Gestalt law of proximity: https://en.wikipedia.org/wiki/Principles_of_grouping#Proximity)

mccalluc:
(I thought I had it, but then it looked horrible on Chrome. This reduces the internal padding to half what it was. I'd either like to merge this much, or take a break and see if there's a better way to approach this.)

mccalluc:
@john-conroy : UI now matches Nils' request. Can you give this a code review? (I've added some tests that are in the neighborhood... does this seems like a reasonable step towards reducing the debt, or do you think we should be more aggressive with each PR?)

context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -101,20 +108,36 @@ function makeTableBodyComponent(resultFields, detailsUrlPrefix, idField) {
   return function ResultsTableBody(props) {
     const { hits } = props;
     /* eslint-disable no-underscore-dangle */
+    /* eslint-disable react/no-danger */
+    /* eslint-disable jsx-a11y/control-has-associated-label */
```
john-conroy, line 112:
```suggestion
    /* eslint-disable no-underscore-dangle react/no-danger jsx-a11y/control-has-associated-label */
```

john-conroy: Pretty sure you can combine these to a single line.

context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -101,20 +108,36 @@ function makeTableBodyComponent(resultFields, detailsUrlPrefix, idField) {
...
+                </StyledTableCell>
+              </StyledTableRow>
+            )}
+          </StyledTableBody>
         ))}
-      </TableBody>
+      </>
     );
-    /* eslint-enable no-underscore-dangle */
+    /* eslint-enable */
```
john-conroy, line 140:
```suggestion
    /* eslint-enable no-underscore-dangle react/no-danger jsx-a11y/control-has-associated-label */
```
Do we need to specify the rules for `eslint enable`?

context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -264,3 +290,4 @@ SearchWrapper.defaultProps = {
 };
 
 export default SearchWrapper;
+export { getByPath, getOrder }; // For tests
```
john-conroy, line 293: This file is pretty long. Would it make sense to organize into multiple files?

context/app/static/js/components/Search/__tests__/SearchWrapper.spec.js
```diff
@@ -0,0 +1,34 @@
+import { getByPath, getOrder } from '../SearchWrapper';
+
```
john-conroy, line 2: Testing utility functions is great. We should also test the `SearchWrapper` component as well in a later PR.

context/app/static/js/components/Search/__tests__/SearchWrapper.spec.js
```diff
@@ -0,0 +1,34 @@
...
+describe('SearchWrapper helper functions', () => {
+  describe('getByPath', () => {
```
john-conroy, line 4: Nesting describe can become confusing and cluttered. I think separating tests into files instead of using `describe` is more clear, but this is an extension of my preference to have more files instead of longer files which you may not share.  I've also been using `test` instead of `it` when not inside of `describe`. I am fine with using both, but wanted to point that out in case you prefer to be consistent.

context/app/static/js/components/Search/style.js
```diff
@@ -29,33 +30,77 @@ const StyledHeaderCell = styled(HeaderCell)`
...
   // NOTE: If we want to darken on hover, we need to give an explicit background to all rows.
   // What looks white is actually transparent and brightness() has no effect.
   background-color: #fff;
 
   :hover {
     filter: brightness(96%);
   }
+
+  // Material would apply this on TD, but we override, so there is no internal border above the highlight.
+  border: 1px solid rgba(224, 224, 224, 1);
```
john-conroy, line 43: Get color from theme?

context/app/static/js/components/Search/style.js
```diff
@@ -29,33 +30,77 @@ const StyledHeaderCell = styled(HeaderCell)`
...
+
+  border-left: none;
+  border-right: none;
+`;
+
+const interPadding = `${16 * 0.6}px`;
+const sidePadding = '4em';
```
john-conroy, line 50: Different units will response to changes differently.

context/app/static/js/components/Search/style.js
```diff
@@ -29,33 +30,77 @@ const StyledHeaderCell = styled(HeaderCell)`
...
+  &.before-highlight td {
+    padding-bottom: 0px;
+  }
+
+  &.highlight td {
+    padding-top: ${interPadding};
+    padding-left: ${sidePadding};
+    padding-right: ${sidePadding};
+    & a {
+      color: rgba(0, 0, 0, 0.54);
```
john-conroy, line 64: Get color from theme?

context/app/static/js/components/Search/style.js
```diff
@@ -29,33 +30,77 @@ const StyledHeaderCell = styled(HeaderCell)`
...
     margin: -100%;
     padding: 100%;
     color: rgb(0, 0, 0);
     overflow-wrap: break-word;
   }
 
-  &:first-child a {
+  // So just one entry in the row looks like a link.
+  :first-child a {
     color: #3781d1;
```
john-conroy, line 84: Get color from theme?

context/app/static/js/components/Search/style.js
```diff
@@ -29,33 +30,77 @@ const StyledHeaderCell = styled(HeaderCell)`
...
 `;
 
-const StyledTableRow = styled(TableRow)`
+const StyledTableBody = styled(TableBody)`
   // NOTE: If we want to darken on hover, we need to give an explicit background to all rows.
   // What looks white is actually transparent and brightness() has no effect.
   background-color: #fff;
 
   :hover {
     filter: brightness(96%);
```
john-conroy, line 39: Should be available in the theme.

context/app/static/js/components/Search/SearchWrapper.jsx
mccalluc: sure!

context/app/static/js/components/Search/__tests__/SearchWrapper.spec.js
mccalluc: no strong preference... I'll split this up.

context/app/static/js/components/Search/__tests__/SearchWrapper.spec.js
mccalluc: Let me file an issue for that... Getting into mocking the api responses was more than I wanted to tackle right now. https://github.com/hubmapconsortium/portal-ui/issues/1038

context/app/static/js/components/Search/style.js
mccalluc: good call.

PR #1036, opened 2020-08-12
Mccalluc/maintenace mode

mccalluc:
There is no design work here, but I want to see if this workflow makes sense to both of you. We know that having a consistent look matters to Nils, and we know this look may change over time. I think making a one-off HTML or CRA would be a pain in the medium to long run, and I think this will be easier, but I could be wrong.

Thoughts?

### DIFF
--- a/make-maintenance-tarball.sh
+++ b/make-maintenance-tarball.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+set -o errexit
+
+red=`tput setaf 1`
+green=`tput setaf 2`
+reset=`tput sgr0`
+die() { set +v; echo "$red$*$reset" 1>&2 ; exit 1; }
+
+WGET_DEST='/tmp/hubmap-portal-maintenance'
+wget --recursive http://localhost:5001/ --directory-prefix "$WGET_DEST" || die 'Server not running on port 5001.'
+grep '"maintenance_mode": true' "$WGET_DEST"/localhost:5001/index.html || die 'Set "MAINTENANCE_MODE = True" in app.conf and restart.'
+
+TAR_GZ='/tmp/hubmap-portal-maintenance.tar.gz'
+cd "$WGET_DEST"
+rm -rf hubmap-portal-maintenance
+mv localhost:5001 hubmap-portal-maintenance
+tar -czvf "$TAR_GZ" hubmap-portal-maintenance
+echo "${green}Send $TAR_GZ to Bill.$reset" 
--- a/CHANGELOG-maintenance.md
+++ b/CHANGELOG-maintenance.md
@@ -1,0 +1,1 @@
+- Script to generate static html for server maintenance.
--- a/README.md
+++ b/README.md
@@ -101,6 +101,14 @@ Then, to redeploy `dev` and `test`:
 ./redeploy.sh $USERNAME test
 ```
 
+## Maintenance static page
+Periodically, even on production, our container will need to be down for maintenance.
+The IEC can put up a static "Under Construction" page if we provide them with static HTML bundle.
+To generate this bundle:
+- Add `MAINTENANCE_MODE = True` in `app.conf`.
+- Restart the dev server.
+- Run `./make-maintenance-tarball.sh`.
+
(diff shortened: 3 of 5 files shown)


mccalluc:
And now it's styled, too:
Screen Shot 2020-08-12 at 6 48 06 PM

context/app/static/js/components/Maintenance/style.js
```diff
@@ -0,0 +1,16 @@
+import Container from '@material-ui/core/Container';
+import styled from 'styled-components';
+
+const Background = styled.div`
+  background-color: ${(props) => props.theme.palette.warning.main};
+  margin-top: -16px; // Override the Header's margin.
```
john-conroy, line 6: I'd prefer not to use negative margin here. I'd like to use a separate `MaintenanceHeader`. I also think we can remove the default `margin-bottom` from the Header and incorporate it in a wrapper component for a route which will allow you to specify whether the route's content should be have a width constraint, top margin, or bottom margin.

context/app/static/js/components/Maintenance/style.js
```diff
@@ -0,0 +1,16 @@
...
+`;
+
+const StyledContainer = styled(Container)`
+  position: absolute;
```
john-conroy, line 10: I don't think we need to use absolute positioning. We could just use flexbox.

context/app/static/js/index.jsx
```diff
@@ -22,7 +23,10 @@ const availableRoutes = [
   '/client-side-error',
 ];
 
-if (urlPath === '/' || availableRoutes.some(isRoute)) {
+// eslint-disable-next-line no-undef
+if ('maintenance_mode' in flaskData && flaskData.maintenance_mode) {
+  ReactDOM.render(<Maintenance />, document.getElementById('react-content'));
+} else if (urlPath === '/' || availableRoutes.some(isRoute)) {
   ReactDOM.render(
```
john-conroy, line 30: With the upcoming error pages, I was hoping to remove the conditionals here.

context/app/static/js/components/Header/Header/Header.jsx
```diff
@@ -1,4 +1,5 @@
 import React, { useRef } from 'react';
+import PropTypes from 'prop-types';
 import Container from '@material-ui/core/Container';
 import Toolbar from '@material-ui/core/Toolbar';
```
john-conroy, line 4: Let's just make a separate `MaintenanceHeader`?

context/app/static/js/components/Footer/Footer.jsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
```
john-conroy, line 1: Separate `MaintenanceFooter`?

make-maintenance-tarball.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -o errexit
+
+red=`tput setaf 1`
+green=`tput setaf 2`
+reset=`tput sgr0`
+die() { set +v; echo "$red$*$reset" 1>&2 ; exit 1; }
+
+WGET_DEST='/tmp/hubmap-portal-maintenance'
+wget --recursive http://localhost:5001/ --directory-prefix "$WGET_DEST" || die 'Server not running on port 5001.'
```
john-conroy, line 10: Why are we running the server to create the tarball?

mccalluc:
> I think it wouldn't be too much effort (happy to do it) to create and maintain a separate App.jsx,index.jsx, and maybe a maintenance webpack config.

@john-conroy : Cool -- Sure, if you've got a clear sense of direction, please go ahead... either branch from this, or make something new, and copy and past over whatever can be re-used.

> Why are we running the server to create the tarball?

For me, running the server we have and using wget to capture everything that needs to be captured felt simplest... but I think you're right that having an entirely separate entry point to the build would probably be better.

> Separate `MaintenanceFooter`?

I could be thinking about this wrong, but to me, it feels like we'd want to keep these in sync, and the handful of `&&` on links that are shouldn't be shown doesn't seem like too much clutter.

PR #1037, opened 2020-08-12
John conroy/file browser tests

### LINKED ISSUE #602: Add tests and prop types for FileBrowser and related components

john-conroy:
Fixes #602.

There's a lot of room for refactoring to avoid repeating shared test props and context providers, but I'd like to get your feedback before doing so.

I added a new hook, `useAbortableEffect`, and included where I found it in comments. While testing the `GlobusLink` component I was getting the following warning:

> Warning: Can’t perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.

The error results from the test ending before the async operation could finish and one of the set state functions was being called on the unmounted `GlobusLink` component. While we haven't observed it in our application, this could also occur if the user changes pages or similar before the async operation can resolve. We should probably use this hook for all async `useEffect` hooks which update the component's state. If this makes sense to you, I can file an issue for to update existing hooks.

Sorry the PR is so large. I opted to keep it as a single PR because my progress was not very linear.


context/app/static/js/components/Detail/Files/Files.spec.js
```diff
@@ -0,0 +1,107 @@
...
+const entityEndpoint = 'fakeendpoint';
+
+const globusUrlResponse = {
+  url: 'fakeglobusurl',
+};
+
+const server = setupServer(
+  rest.get(`/${entityEndpoint}/entities/dataset/globus-url/${uuid}`, (req, res, ctx) => {
+    return res(ctx.json(globusUrlResponse), ctx.status(200));
+  }),
```
mccalluc, line 25: Cool: I'll try to do something like this when I get to the search testing.

context/app/static/js/components/Detail/GlobusLinkMessage/GlobusLinkMessage.spec.js
```diff
@@ -0,0 +1,56 @@
...
+      { exact: false },
+    ),
+  ).toBeInTheDocument();
+});
+
+const messages = {
+  401: 'Unauthorized access to the Globus Research Management System (bad or expired token). If you believe this to be an error, please contact',
+  403: 'Access to files on the Globus Research Management system are restricted. You may not have access to these files because the Consortium is still curating data and/or the data is protected data that requires you to be a member of the Consortium “Protected Data Group”. Such protected data will be available via dbGaP in the future. If you believe this to be an error, please contact',
+  404: 'Files are not available through the Globus Research Management system. If you believe this to be an error, please contact',
+  500: 'Unexpected server or other error. Report error to',
```
mccalluc, line 45:
```suggestion
  401: 'Unauthorized access to the Globus Research Management System',
  403: 'Access to files on the Globus Research Management system are restricted',
  404: 'Files are not available through the Globus Research Management system',
  500: 'Unexpected server or other error',
```
Prefer shorter, but your call.

context/app/static/js/hooks/useAbortableEffect.js
```diff
@@ -0,0 +1,24 @@
+// https://www.debuggr.io/react-update-unmounted-component/
+// https://github.com/reactjs/rfcs/issues/137
+import { useEffect } from 'react';
+
+function useAbortableEffect(effect, dependencies) {
```
mccalluc, line 5: We should use this for any effect around an async? I think either ilan or mark wrote a eslint rule for something special we'd like to enforce, and this might be another case: it didn't seem that complicated. (But that's certainly not a priority.)

context/app/static/js/components/Detail/FileBrowserDUA/FileBrowserDUA.spec.js
```diff
@@ -0,0 +1,110 @@
...
+});
+
+test('displays correct text for consortium access level', () => {
+  render(
+    <FileBrowserDUA isOpen handleAgree={handleAgree} handleClose={handleClose} mapped_data_access_level="Consortium" />,
+  );
+
+  expect(screen.getByText(/Controlled/)).toBeInTheDocument();
+  expect(
+    screen.getByText(/You are attempting to download data accessible only to HuBMAP Members./),
```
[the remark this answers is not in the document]
john-conroy: We can avoid using the regex match, but will still need to use a substring match.

`screen.getByText('You are attempting to download data accessible only to HuBMAP Members.', { exact: false })`

context/app/static/js/components/Detail/FileBrowserFile/FileBrowserFile.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+import FileBrowserFile from './FileBrowserFile';
+import DetailContext from '../context';
+import FilesContext from '../Files/context';
+
+const fakeOpenDUA = jest.fn();
+
+const assetsEndpoint = 'fakeendpoint';
+const uuid = 'fakeuuid';
+const token = 'faketoken';
```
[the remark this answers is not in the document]
john-conroy: Test strings should be kebab case unless we're testing for a specific structure?

context/app/static/js/hooks/useAbortableEffect.js
john-conroy: Filed #1054

PR #1040, opened 2020-08-13
seqFish Vitessce Display

ilan-gold:
This updates the view configuration for seqFish so that hybridization cycles with a given position can be overlaid, as is suggested by the data.  To test go to http://localhost:5001/browse/dataset/38d0c2dc36f512ac54a5a62318033ecb on TEST

### DIFF
--- a/context/app/api/test_vitessce.py
+++ b/context/app/api/test_vitessce.py
@@ -84,7 +84,7 @@ def test_build_layer_conf_empty():
     assert conf == {}
 
 
-def test_group_by_parent_path():
+def test_group_by_file_name():
     data = ['foo/bar.sh', 'zap/bar.sh', 'jazz/bar.js', 'jazz/not_bar.js']
     grouped = _group_by_file_name(data)
     # Grouped by file name.
--- a/context/app/api/test_vitessce.py
+++ b/context/app/api/test_vitessce.py
@@ -83,8 +83,9 @@ def test_build_layer_conf_empty():
     conf = vitessce.conf
     assert conf == {}
 
+
 def test_group_by_parent_path():
     data = ['foo/bar.sh', 'zap/bar.sh', 'jazz/bar.js', 'jazz/not_bar.js']
-    grouped = _group_by_parent_path(data)
+    grouped = _group_by_file_name(data)
     # Grouped by file name.
     assert [['jazz/bar.js'], ['foo/bar.sh', 'zap/bar.sh'], ['jazz/not_bar.js']] == grouped
--- a/context/app/api/test_vitessce.py
+++ b/context/app/api/test_vitessce.py
@@ -1,4 +1,4 @@
-from .vitessce import Vitessce
+from .vitessce import Vitessce, _group_by_file_name
 
 TEST_ENTITY_CODEX = {
     "data_types": ["codex_cytokit"],
@@ -82,3 +82,9 @@ def test_build_layer_conf_empty():
     vitessce._build_vitessce_conf()
     conf = vitessce.conf
     assert conf == {}
+
+def test_group_by_parent_path():
+    data = ['foo/bar.sh', 'zap/bar.sh', 'jazz/bar.js', 'jazz/not_bar.js']
+    grouped = _group_by_parent_path(data)
+    # Grouped by file name.
+    assert [['jazz/bar.js'], ['foo/bar.sh', 'zap/bar.sh'], ['jazz/not_bar.js']] == grouped
--- a/context/app/api/vitessce.py
+++ b/context/app/api/vitessce.py
@@ -116,6 +116,9 @@ IMAGE_PYRAMID_PATH = "ometiff-pyramids"
 CODEX_TILE_PATH = "output/extract/expressions/ome-tiff"
 CODDEX_SPRM_PATH = "output_json"
 
+SEQFISH_HYB_CYCLE = r"HybCycle_\d+"
+SEQFISH_NAME = r"MMStack_Pos\d+\.ome\.tiff?"
+
 # Hardcoded CODEX offsets and tile path.
 IMAGING_PATHS = {
     CODEX_CYTOKIT: {"offsets": OFFSETS_PATH, "image": CODEX_TILE_PATH, },
@@ -123,6 +126,7 @@ IMAGING_PATHS = {
 }
 
 TILE_REGEX = r"R\d+_X\d+_Y\d+"
+SEQFISH_PATH = f"{IMAGE_PYRAMID_PATH}/{SEQFISH_HYB_CYCLE}/{SEQFISH_NAME}"
 
 SCRNA_SEQ_CONFIG = {
     "base_conf": SCATTERPLOT,
@@ -288,7 +292,6 @@ class Vitessce:
         files = copy.deepcopy(ASSAY_CONF_LOOKUP[self.assay_type]["files_conf"])
         file_paths_expected = [file["rel_path"] for file in files]
         file_paths_found = [file["rel_path"] for file in self.entity["files"]]
-        print(file_paths_found)
         conf = copy.deepcopy(ASSAY_CONF_LOOKUP[self.assay_type]["base_conf"])
         # Codex and other tiled assays needs to be built up based on their input tiles.
         if self.assay_type not in IMAGE_ASSAYS:
@@ -335,11 +338,14 @@ class Vitessce:
             # each conf to be all images across a given PosM, each image named HybCycle_N.
             # This is unique to seqFish.
             if "seqFish" in self.entity["data_types"]:
+                is_valid_directory = all([re.match(SEQFISH_PATH, file) for file in found_images])
+                if not is_valid_directory:
+                    raise NameError("Directory structure for seqFish is incorrect")
                 # Get all files grouped by PosN names.
                 images_by_pos = _group_by_file_name(found_images)
                 # Get Hybridization per paths grouped by Pos.
                 hyb_cycles_per_pos = [
-                    sorted([Path(image).parts[-2] for image in images])
+                    sorted([re.search(SEQFISH_HYB_CYCLE, image)[0] for image in images])
                     for images in images_by_pos
                 ]
                 confs = []
@@ -352,9 +358,7 @@ class Vitessce:
                         )
                     ]
                     new_conf["layers"] = layers
-                    # The stem is the name of the ome.tif file (since it has "PosN")
-                    # but the stem includes "ome" as well so we split on the ".".
-                    new_conf["name"] = Path(images[0]).stem.split(".")[0]
+                    new_conf["name"] = re.search(SEQFISH_NAME, images[0])[0].split(".")[0]
                     confs.append(self._replace_view(new_conf))
                 self.conf = confs
                 return confs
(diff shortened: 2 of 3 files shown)


context/app/api/vitessce.py
```diff
@@ -320,8 +320,40 @@ def _build_vitessce_conf(self):
...
+            # each conf to be all images across a given PosM, each image named HybCycle_N.
+            # This is unique to seqFish.
+            if "seqFish" in self.entity["data_types"]:
+                # Get all PosN names.
+                pos_name_set = set([Path(image).name for image in found_images])
+                # Get all the paths grouped by Pos.
+                images_by_pos = [
+                    [image for image in found_images if Path(image).name == name]
+                    for name in sorted(pos_name_set)
+                ]
```
mccalluc, line 333: Consider using [`itertools.groupby`](https://docs.python.org/3/library/itertools.html#itertools.groupby)

context/app/api/vitessce.py
```diff
@@ -320,8 +320,40 @@ def _build_vitessce_conf(self):
...
+                # Build up a conf for each Pos.
+                for i, images in enumerate(images_by_pos):
+                    new_conf = copy.deepcopy(conf)
+                    layers = [
+                        self._build_multi_file_image_layer_conf(
+                            images, hyb_cycles_per_pos[i]
+                        )
+                    ]
+                    new_conf["layers"] = layers
+                    # The stem is the name of the ome.tif file (since it has pos) but stem g.
```
mccalluc, line 349: Comment unclear to me?

context/app/api/vitessce.py
```diff
@@ -320,8 +320,40 @@ def _build_vitessce_conf(self):
...
+                    new_conf["name"] = Path(images[0]).stem.split(".")[0]
+                    new_conf = self._replace_view(new_conf)
+                    confs += [new_conf]
```
mccalluc, line 352:
```suggestion
                    confs.append(self._replace_view(new_conf))
```
- Reassigning variables can be a source of confusion.
- `append`.

context/app/api/vitessce.py
```diff
@@ -399,14 +431,15 @@ def _build_image_layer_datauri(self, rel_paths):
 
         image_layer = {}
         image_layer["images"] = [
-            self._build_image_schema(rel_path) for rel_path in rel_paths
+            self._build_image_schema(rel_path, names[ind] if names else "")
+            for ind, rel_path in enumerate(rel_paths)
```
mccalluc, line 435:
```suggestion
            self._build_image_schema(rel_path, names[i] if names else None)
            for i, rel_path in enumerate(rel_paths)
```

context/app/api/vitessce.py
```diff
@@ -399,14 +431,15 @@ def _build_image_layer_datauri(self, rel_paths):
...
         ]
         image_layer["schemaVersion"] = "0.0.2"
         return DataURI.make(
             "text/plain", charset="us-ascii", base64=True, data=json.dumps(image_layer)
         )
 
-    def _build_image_schema(self, image_rel_path):
+    def _build_image_schema(self, image_rel_path, name=""):
```
mccalluc, line 442:
```suggestion
    def _build_image_schema(self, image_rel_path, name=None):
```

context/app/api/vitessce.py
```diff
@@ -421,7 +454,7 @@ def _build_image_schema(self, image_rel_path):
         """
 
         schema = {}
-        schema["name"] = Path(image_rel_path).name
+        schema["name"] = name if name else Path(image_rel_path).name
```
mccalluc, line 457:
```suggestion
        schema["name"] = name if name is not None else Path(image_rel_path).name
```
Let's be more explicit, rather than relying on the falsiness of `""`.

context/app/api/vitessce.py
ilan-gold: Used.

context/app/api/vitessce.py
ilan-gold: Fixed.

ilan-gold:
@mccalluc Can I merge? It says I need an approval beforehand (or I can "override" and merge anyway)

context/app/api/test_vitessce.py
```diff
@@ -82,3 +82,10 @@ def test_build_layer_conf_empty():
     vitessce._build_vitessce_conf()
     conf = vitessce.conf
     assert conf == {}
+
+
+def test_group_by_parent_path():
```
mccalluc, line 87:
```suggestion
def test_group_by_file_name():
```

context/app/api/vitessce.py
```diff
@@ -222,9 +223,19 @@ def _get_matches(files, regex):
...
+
+def _get_path_name(file):
+    return Path(file).name
+
+
+def _group_by_file_name(files):
+    files_grouped_by_name = []
+    sorted_files = sorted(files, key=_get_path_name)
+    for _, g in groupby(sorted_files, _get_path_name):
+        files_grouped_by_name.append(list(g))
```
mccalluc, line 237:
```suggestion
    sorted_files = sorted(files, key=_get_path_name)
    return groupby(sorted_files, _get_path_name).values()
```
Or a comprehension if needed.

context/app/api/vitessce.py
```diff
@@ -320,8 +333,35 @@ def _build_vitessce_conf(self):
...
+            # each conf to be all images across a given PosM, each image named HybCycle_N.
+            # This is unique to seqFish.
+            if "seqFish" in self.entity["data_types"]:
+                # Get all files grouped by PosN names.
+                images_by_pos = _group_by_file_name(found_images)
+                # Get Hybridization per paths grouped by Pos.
+                hyb_cycles_per_pos = [
+                    sorted([Path(image).parts[-2] for image in images])
+                    for images in images_by_pos
+                ]
```
mccalluc, line 346: This is one of the places where we're really expecting a very particular structure, and I'd be more comfortable if there were validation up-front that that structure holds. I also think we should have some kind of testing about the viewconfs that are produced.

ilan-gold: Would a regex work?  It bypasses the "structure" a bit but it makes sure that what I need is there.

Something like
```python
re.search(r"HybCycle_[0-9]", 'output-pyramids/HybCycle_4/MMStack_Pos25.ome.tiff')[0]
# HybCycle_4
```

ilan-gold: I mean more generally, would regexes work as a way of validating?  We could also do something more like a `match` with the directory structure.

ilan-gold: @mccalluc Do you have an opinion here about what sort of validation is done? Or is this more of a "something is better than nothing" situation?

mccalluc: @ilan-gold : Regexes could work. I'm really not sure what part is fixed and which is variable, but perhaps checking that every path matches:`output-pyramids/HybCycle_\d+/MMStack_Pos\d+\.ome\.tiff`? Let me look at the latest changes...

context/app/api/vitessce.py
```diff
@@ -115,26 +116,30 @@
 CODEX_TILE_PATH = "output/extract/expressions/ome-tiff"
 CODDEX_SPRM_PATH = "output_json"
 
+SEQFISH_HYB_CYCLE = r"HybCycle_\d+"
+SEQFISH_NAME = r"MMStack_Pos\d+\.ome\.tiff?"
```
mccalluc, line 120: Unless there's another organizing principle, define these right before they are used. Consider `_RE`  in name.

context/app/api/vitessce.py
```diff
@@ -115,26 +116,30 @@
...
+
 # Hardcoded CODEX offsets and tile path.
 IMAGING_PATHS = {
     CODEX_CYTOKIT: {"offsets": OFFSETS_PATH, "image": CODEX_TILE_PATH, },
     IMAGE_PYRAMID: {"offsets": OFFSETS_PATH, "image": IMAGE_PYRAMID_PATH, },
 }
 
 TILE_REGEX = r"R\d+_X\d+_Y\d+"
+SEQFISH_PATH = f"{IMAGE_PYRAMID_PATH}/{SEQFISH_HYB_CYCLE}/{SEQFISH_NAME}"
```
mccalluc, line 129: Consider `_RE` in name.

context/app/api/vitessce.py
```diff
@@ -320,8 +334,36 @@ def _build_vitessce_conf(self):
...
             no_ims_separate_images = _exclude_matches(found_images, r"/separate/")
-            layer = self._build_multi_file_image_layer_conf(no_ims_separate_images)
-            conf["layers"] = [layer]
+            # File paths are like path/to/HybCycle_N/MMStack_PosM.ome.tif and we want
+            # each conf to be all images across a given PosM, each image named HybCycle_N.
+            # This is unique to seqFish.
+            if "seqFish" in self.entity["data_types"]:
+                is_valid_directory = all([re.match(SEQFISH_PATH, file) for file in found_images])
+                if not is_valid_directory:
+                    raise NameError("Directory structure for seqFish is incorrect")
```
mccalluc, line 343: Thanks! If we ever actually hit this, in the process of debugging will probably want to add more information to error message: Might collect just those paths which are not valid, and include the RE, so the user could start to understand what went wrong without going back to the code.

```suggestion
                    raise Exception("Directory structure for seqFish is incorrect")
```
I would only raise [Exception](https://docs.python.org/3/library/exceptions.html#Exception), or a subclass we make. I don't think we want to re-use the other built in exceptions with our own semantics.

context/app/api/vitessce.py
```diff
@@ -320,8 +334,36 @@ def _build_vitessce_conf(self):
             found_images = _get_matches(file_paths_found, files[0]["rel_path"])
             # Do not show IMS images that are in the "/separate/" folder.
             no_ims_separate_images = _exclude_matches(found_images, r"/separate/")
-            layer = self._build_multi_file_image_layer_conf(no_ims_separate_images)
-            conf["layers"] = [layer]
+            # File paths are like path/to/HybCycle_N/MMStack_PosM.ome.tif and we want
+            # each conf to be all images across a given PosM, each image named HybCycle_N.
+            # This is unique to seqFish.
+            if "seqFish" in self.entity["data_types"]:
+                is_valid_directory = all([re.match(SEQFISH_PATH, file) for file in found_images])
```
mccalluc, line 341:
```suggestion
                is_valid_directory = all([re.fullmatch(SEQFISH_PATH, file) for file in found_images])
```
`match` allows unmatched characters at the end.

context/app/api/vitessce.py
```diff
@@ -320,8 +334,36 @@ def _build_vitessce_conf(self):
             found_images = _get_matches(file_paths_found, files[0]["rel_path"])
             # Do not show IMS images that are in the "/separate/" folder.
             no_ims_separate_images = _exclude_matches(found_images, r"/separate/")
-            layer = self._build_multi_file_image_layer_conf(no_ims_separate_images)
-            conf["layers"] = [layer]
+            # File paths are like path/to/HybCycle_N/MMStack_PosM.ome.tif and we want
+            # each conf to be all images across a given PosM, each image named HybCycle_N.
+            # This is unique to seqFish.
+            if "seqFish" in self.entity["data_types"]:
```
mccalluc, line 340: Since the body of the `if` is so much longer than what follows, good to either pull out into helper function, or negate the sense and use a guard clause, so the nesting doesn't get as deep.

ilan-gold: Move to its own function.  Thanks

Issue #1046, opened 2020-08-14
Preview Page Finetuning


tsliaw:
<img width="1426" alt="Screen Shot 2020-08-14 at 12 12 19 PM" src="https://user-images.githubusercontent.com/16737715/90284839-83876200-de27-11ea-820b-9ec1a5891408.png">
<img width="1432" alt="Screen Shot 2020-08-14 at 12 10 39 PM" src="https://user-images.githubusercontent.com/16737715/90284842-84b88f00-de27-11ea-8a7e-2e0328f3fa9e.png">
<img width="1437" alt="Screen Shot 2020-08-14 at 12 09 13 PM" src="https://user-images.githubusercontent.com/16737715/90284870-926e1480-de27-11ea-9319-4b4c6758a6e3.png">

tsliaw:
<img width="1530" alt="Screen Shot 2020-08-14 at 12 15 54 PM" src="https://user-images.githubusercontent.com/16737715/90285031-eb3dad00-de27-11ea-81b1-fa26663bea25.png">

tsliaw:
Things to Finetune:

- Change "Preview" subtitle to Subtitle1 Font (Match entity pages)
- Fix heading fonts 
- Fix "Fullscreen icon" to match brightness icons (Square shaped, not rectangle)
- Fix spacing between attribution and content to match visualization section (Tiffany will check mockup to make sure)

PR #1059, opened 2020-08-17
Mccalluc/marked redo

### LINKED ISSUE #900: Sub headers in markdown should get anchor links

### LINKED ISSUE #901: Wide tables in markdown should scroll rather than spill over
At the bottom of the donor metadata page there is a particularly wide table that they would like to scroll within the page, rather than spill over.

### LINKED ISSUE #937: update styles for documentation and preview Markdown
Update that CSS to match our style guide more closely. In particular the bold headers should be replaced to better match the styles on the other pages. I also believe that the font size for the main text might be a bit too small.

mccalluc:
Replaces #917. Towards #1046.

Fix #901, fix #900, fix #937.


mccalluc:
@ilan-gold : Huh: http://localhost:5001/docs works for me... Restart the dev server to make sure the npm dependencies are installed? (Though preview shouldn't have worked if they weren't...) Or do you not have the docs submodule pulled down? (Though `docs.md` itself is not from the submodule, so that doesn't seem like the problem.) Does `/VERSION` work?

So... not really sure, but I'd like to know that it works for you before merging.

mccalluc:
@tsliaw : pushed a change to use `subtitle1`. (Off hand, I'm really not sure if it's supposed to be a subtitle, or something else.) Does this look right?
Screen Shot 2020-08-18 at 12 17 16 PM

tsliaw:
Using subtitle is correct and the styling is correct now, but can you make it the theme purple color?

ilan-gold:
@mccalluc I can see VERSION but I think you misread my comment - I was saying all of the links within that page are not working, like http://localhost:5001/docs/about.  Does this work for you?   Everything seems to be installed OK

ilan-gold:
Which submodule is supposed to contain the docs @mccalluc?  All I see is `ingest-validation-tools` in my repo (and on Github) in terms of submodules.

ilan-gold:
Ah I see it @mccalluc my apologies.  It was not being pulled down.

mccalluc:
- @tsliaw : Green checkmark if the purple is good, and there's nothing else that needs to be fixed?
- @ilan-gold : If you have any process-level suggestions that would make it less confusing, that'd be great... but sometimes things are irreducibly confusing. :(

Thanks!

ilan-gold:
@mccalluc Is it possible to somehow automate the update of sub-modules?  Or did I do something wrong?  Could we add something to `dev-start.sh` to automatically update?

mccalluc:
We could add something to dev start, but did you see this section in the README?:
> Git does not update submodules on pull by default... but you can make it the default:
> ```
> git config --global submodule.recurse true # Run this once...
> git pull                                   # Now pulls submodules!
> ```
Would that do what you want?

PR #1064, opened 2020-08-18
Replace gender with sex

mccalluc:
This complements the indexing changes in search-api: https://github.com/hubmapconsortium/search-api/pull/144

- @john-conroy : Code review... just a search and replace, so hopefully I didn't find a way to mess this up.
- @ngehlenborg : When should this be merged? I think it's ok as long as the production environment has `sex`... but if the other environments have not caught up, the facet will not be available there. I think that's ok, but wanted to confirm.

(Separately from this, we want a tabular display of Donor metadata, but I don't think that more involved change should be crammed into this PR.)

### DIFF
--- a/CHANGELOG-sex-instead-of-gender.md
+++ b/CHANGELOG-sex-instead-of-gender.md
@@ -1,0 +1,1 @@
+- Replace "gender" with "sex" in the donor facets.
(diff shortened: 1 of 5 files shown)


john-conroy:
@ngehlenborg Chuck decided he'd like to implement the tabular donor metadata.

mccalluc:
oops:
> I think it's ok as long as the production environment has sex

That is not actually the case yet... but hopefully it will be indexed there by Thursday. Until it is, the new `sex` facet will be empty.

PR #1069, opened 2020-08-18
Mccalluc/donor metadata table

mccalluc:
- @john-conroy : Since it's being reused, it makes sense to move MetadataTable out of Detail, right?
- @tsliaw : We are getting rid of the old metadata section, right? I wasn't 100% sure about this. I'll work on providing descriptions... I'll keep this PR as a draft till they are available... but if you think that's not needed, say so.

Screen Shot 2020-08-18 at 4 59 38 PM


ngehlenborg:
Have you tried it with a donor with the complete metadata? We should no longer display "Gender" and in this case race should be "Black or African American".

It is ok to leave the descriptions empty for now.

john-conroy:
As we discussed in standup, keep `MetadataTable` in detail as its only used on entity detail pages.

mccalluc:
@tsliaw : If you think it's ok, can you give this a green checkmark?
- duplication of `gender` and `sex` is a data issue, which I'll address in `search-api`.
- Weird `kidney_donor_profile_index` is also a data issue... The fix won't be in the front-end code.
- now concatenating with `, ` (including space)
- Nils writes: "leaving keys as is (with underscore) is desired"

mccalluc:
Verbal confirmation from Nils.

PR #1095, opened 2020-08-21
Mccalluc/value and unit

mccalluc:
Should https://github.com/hubmapconsortium/search-api/pull/155 be merged, and the donor metadata reindexed, this would be the corresponding change to the search page. Draft until the direction of that PR is clear.


mccalluc:
https://github.com/hubmapconsortium/search-api/pull/156 was merged, so we should make corresponding updates to search config here.

PR #1111, opened 2020-08-24
John conroy/webpack bundle caching

### LINKED ISSUE #1049: Need to add version-number or hash on bundle.js
When the site is redeployed, the browser will still use the old javascript until it expires from the cache. We need to make sure the built js filename changes with each release, at least. Either grab the version number, or include a hash?

@john-conroy : Do you have opinions about the right way to do this?

john-conroy:
Fix #1049. 

- Modifies code from [flask-static-digest](https://github.com/nickjj/flask-static-digest). Instead using webpack to generate the hashed filenames and manifest.
- Only hashes files and generates manifest in production. Uses a new method `static_url_for` to look for hashed filenames from the manifest if the manifest exists, otherwise it behaves like the normal `url_for` for the most part.

### DIFF
--- a/context/app/flask_static_digest.py
+++ b/context/app/flask_static_digest.py
@@ -1,8 +1,8 @@
 import json
 import os
 
-
 from flask import url_for as flask_url_for
+
 
 class FlaskStaticDigest(object):
     def __init__(self, app=None):
@@ -12,7 +12,7 @@ class FlaskStaticDigest(object):
             self.init_app(app)
 
     def init_app(self, app):
-        self.manifest_path = os.path.join(app.static_folder,"public/","manifest.json")
+        self.manifest_path = os.path.join(app.static_folder, "public/", "manifest.json")
         self.has_manifest = os.path.exists(self.manifest_path)
 
         self.manifest = {}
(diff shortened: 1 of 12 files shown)


context/app/flask_static_digest.py
```diff
@@ -0,0 +1,52 @@
...
+class FlaskStaticDigest(object):
+    def __init__(self, app=None):
+        self.app = app
+
+        if app is not None:
+            self.init_app(app)
+
+    def init_app(self, app):
+        self.manifest_path = os.path.join(app.static_folder, "public/", "manifest.json")
+        self.has_manifest = os.path.exists(self.manifest_path)
```
mccalluc, line 16:
```suggestion
        self.manifest_path = Path(app.static_folder) / 'public' / 'manifest.json'
        self.has_manifest = self.manifest_path.exists()
```
Prefer using pathlib; References to `manifest_path` would also need to change.

context/app/flask_static_digest.py
```diff
@@ -0,0 +1,52 @@
...
+        :param values: Arguments of the URL rule
+        :return: Static file path.
+        """
+        filename = values.get("filename")
+        new_filename = {}
+
+        if not self.has_manifest:
+            temp_filename = filename
+        else:
+            temp_filename = self.manifest.get(filename).split('/')[-1]
```
mccalluc, line 47:
```suggestion
            temp_filename = Path(self.manifest[filename]).basename()
```

john-conroy:
Filed #1118

PR #1114, opened 2020-08-25
John conroy/route

john-conroy:
Add a wrapper component for a route and remove margin bottom from the header component.

Screen Shot 2020-08-25 at 10 37 21 AM

### DIFF
--- a/context/app/static/js/components/Routes/Route.jsx
+++ b/context/app/static/js/components/Routes/Route.jsx
@@ -4,27 +4,23 @@ import styled from 'styled-components';
 import Container from '@material-ui/core/Container';
 
 const StyledContainer = styled(Container)`
-  margin-top: ${(props) => (props.$disableMarginTop ? 0 : props.theme.spacing(2))}px;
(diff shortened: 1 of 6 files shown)


context/app/static/js/components/Routes/Route.jsx
```diff
@@ -0,0 +1,31 @@
...
+  mt: PropTypes.bool,
+  constrainWidth: PropTypes.bool,
+};
+
+Route.defaultProps = {
+  mt: false,
+  constrainWidth: false,
+};
+
+export default Route;
```
mccalluc, line 31: I'm not sure that `Route` is the clearest name for this... I think of a route as something like a (path, action) pair, and neither of those are here, though maybe the children are the action?

john-conroy: - Naming the component `Route` is similar to patterns I've seen people use with `react-router`.
- I thought it would make sense to be more explicit about when something is being constrained vs. not, but I can make the switch if you think it's worthwhile.
- Error pages do not use `mt` and we will likely have additional pages which don't in the future.

PR #1117, opened 2020-08-25
John conroy/mapped bar chart labels

### LINKED ISSUE #1072: Use mapped labels for assay types & use combined assay types for homepage bar chart
1. Bar chart labels should use the mapped labels, rather than the raw ids.
1. If multiple labels are assigned to a single dataset, it should be counted as a single assay type (see image pyramid).

image

john-conroy:
Fix #1072 

Screen Shot 2020-08-25 at 4 43 09 PM

### DIFF
--- a/context/app/static/js/components/Home/BarChart/BarChart.jsx
+++ b/context/app/static/js/components/Home/BarChart/BarChart.jsx
@@ -1,29 +1,9 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Vega as ReactVega } from 'react-vega';
-/* eslint-disable import/no-unresolved */
+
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/Home/BarChart/BarChart.jsx
```diff
@@ -131,6 +107,18 @@ function BarChart(props) {
...
+
+  useEffect(
+    () =>
+      setSpec((prevState) => {
+        const temp = prevState.encoding;
+        temp.x.axis.labelAngle = labelAngle;
+        return { ...prevState, ...temp };
+      }),
+    [labelAngle],
+  );
```
mccalluc, line 120: I'd prefer to leave this out, and just be more conservative for all screen sizes, but ... never mind, we talked about this in the call.

PR #1122, opened 2020-08-27
John conroy/eslint ignore public

john-conroy:
The public dir for the maintenance is inside `app/static/js` which we lint. Add `eslintignore` file and ignore all public directories to be safe.


john-conroy:
Yeah this was the culprit.

PR #1128, opened 2020-08-28
John conroy/disable prov table for snare

john-conroy:
Disables the provenance table for datasets which include `snare_lab` or `TMT-LC-MS` in its `data_types` array. It's working as expected for the `TMT-LC-MS` dataset in test, but there isn't a `snare_lab` example to check.

### DIFF
--- a/context/app/static/js/components/Detail/ProvTabs/ProvTabs.jsx
+++ b/context/app/static/js/components/Detail/ProvTabs/ProvTabs.jsx
@@ -6,7 +6,7 @@ import { StyledTab, StyledTabs, StyledTabPanel } from './style';
 import ProvGraph from '../ProvGraph';
 import ProvTable from '../ProvTable';
 import ProvAnalysisDetails from '../ProvAnalysisDetails';
-import { checkDataTypesForValues } from './utils';
+import { hasDataTypes } from './utils';
 
 function ProvTabs(props) {
   const { uuid, assayMetadata, provData } = props;
@@ -17,11 +17,7 @@ function ProvTabs(props) {
     setOpen(newValue);
   };
 
-  const shouldDisplayTable = !checkDataTypesForValues(data_types, [
-    'sc_rna_seq_snare_lab',
-    'sc_atac_seq_snare_lab',
-    'TMT-LC-MS',
-  ]);
+  const shouldDisplayTable = !hasDataTypes(data_types, ['sc_rna_seq_snare_lab', 'sc_atac_seq_snare_lab', 'TMT-LC-MS']);
   const shouldDisplayDag = entity_type === 'Dataset' && metadata && 'dag_provenance_list' in metadata;
 
   const graphIndex = shouldDisplayTable ? 1 : 0;
@@ -36,10 +32,10 @@ function ProvTabs(props) {
         aria-label="Detail View Tabs"
         TabIndicatorProps={{ style: { backgroundColor: '#9CB965' } }}
       >
-        {shouldDisplayTable && <StyledTab label="Table" id="tab-table" aria-controls="tabpanel-0" />}
-        <StyledTab label="Graph" id="tab-graph" aria-controls={`tabpanel-${graphIndex}`} />
+        {shouldDisplayTable && <StyledTab label="Table" id="tab-table" aria-controls="tabpanel-table" />}
+        <StyledTab label="Graph" id="tab-graph" aria-controls="tabpanel-graph" />
         {shouldDisplayDag && (
-          <StyledTab label="Analysis Details" id="tab-analysis-details" aria-controls={`tabpanel-${dagIndex}`} />
+          <StyledTab label="Analysis Details" id="tab-analysis-details" aria-controls="tabpanel-analysis-details" />
         )}
       </StyledTabs>
       {shouldDisplayTable && (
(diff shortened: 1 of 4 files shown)


context/app/static/js/components/Detail/ProvTabs/utils.js
```diff
@@ -0,0 +1,11 @@
+export function checkDataTypesForValues(dataTypes, valuesToCheck) {
+  return (
+    dataTypes.filter(
+      (type) => valuesToCheck.filter((value) => type.toLowerCase().includes(value.toLowerCase())).length > 0,
```
mccalluc, line 4:
```suggestion
      (type) => valuesToCheck.filter((value) => type === value).length > 0,
```
In some contexts, "be liberal in what you accept" is the right approach... but here I think we want to be conservative... otherwise we case-normalize in some places and not in others, and instead of cleaning up the data, people think something is wrong with the UI when a bug crops up.

mccalluc: Also, use [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) instead of checking list length.

context/app/static/js/components/Detail/ProvTabs/ProvTabs.jsx
```diff
@@ -6,18 +6,27 @@ import { StyledTab, StyledTabs, StyledTabPanel } from './style';
...
   const [open, setOpen] = React.useState(0);
   const handleChange = (event, newValue) => {
     setOpen(newValue);
   };
 
+  const shouldDisplayTable = !(
+    entity_type === 'Dataset' &&
+    data_types &&
+    checkDataTypesForValues(data_types, ['snare_lab', 'TMT-LC-MS'])
+  );
```
mccalluc, line 24:
```suggestion
  const shouldDisplayTable = !checkDataTypesForValues(data_types, ['snare_lab', 'TMT-LC-MS']);
```
Handle null data_types inside the function? Don't need to explicitly check entity type?

context/app/static/js/components/Detail/ProvTabs/utils.js
```diff
@@ -0,0 +1,11 @@
...
+    ).length > 0
+  );
+}
+
+export function getTabIndex(defaultIndex, shouldDisplayTable) {
+  return shouldDisplayTable ? defaultIndex : defaultIndex - 1;
+}
```
mccalluc, line 11: I think it'd be more clear just to inline this?

context/app/static/js/components/Detail/ProvTabs/ProvTabs.jsx
```diff
@@ -27,30 +36,34 @@ function ProvTabs(props) {
...
         TabIndicatorProps={{ style: { backgroundColor: '#9CB965' } }}
       >
-        <StyledTab label="Table" id="tab-0" aria-controls="tabpanel-0" />
-        <StyledTab label="Graph" id="tab-1" aria-controls="tabpanel-1" />
-        {shouldDisplayDag && <StyledTab label="Analysis Details" id="tab-2" aria-controls="tabpanel-2" />}
+        {shouldDisplayTable && <StyledTab label="Table" id="tab-0" aria-controls="tabpanel-0" />}
+        <StyledTab label="Graph" id={`tab-${graphIndex}`} aria-controls={`tabpanel-${graphIndex}`} />
+        {shouldDisplayDag && (
+          <StyledTab label="Analysis Details" id={`tab-${dagIndex}`} aria-controls={`tabpanel-${dagIndex}`} />
+        )}
```
mccalluc, line 43: I could be mistaken, but I don't think the IDs are required to have any particular structure, so instead of numbering them, just use something like `tab-graph` or `tab-dag`?

context/app/static/js/components/Detail/ProvTabs/ProvTabs.jsx
```diff
@@ -27,30 +36,34 @@ function ProvTabs(props) {
         aria-label="Detail View Tabs"
         TabIndicatorProps={{ style: { backgroundColor: '#9CB965' } }}
       >
-        <StyledTab label="Table" id="tab-0" aria-controls="tabpanel-0" />
-        <StyledTab label="Graph" id="tab-1" aria-controls="tabpanel-1" />
-        {shouldDisplayDag && <StyledTab label="Analysis Details" id="tab-2" aria-controls="tabpanel-2" />}
+        {shouldDisplayTable && <StyledTab label="Table" id="tab-table" aria-controls="tabpanel-0" />}
+        <StyledTab label="Graph" id="tab-graph" aria-controls={`tabpanel-${graphIndex}`} />
```
mccalluc, line 40: Still doing the integer indexes in `aria-controls`: Is that necessary?

context/app/static/js/components/Detail/ProvTabs/utils.js
```diff
@@ -0,0 +1,7 @@
+export function checkDataTypesForValues(dataTypes, valuesToCheck) {
```
mccalluc, line 1: Consider renaming, perhaps `hasDataTypes`? 

`check` is ambiguous, and what the return value should be gets muddled if we're looking for a good thing or a bad thing.

context/app/static/js/components/Detail/ProvTabs/ProvTabs.jsx
john-conroy: I honestly don't know. I think this was pulled from a material ui example and thought it best to keep since we're in a time crunch.

john-conroy:
Filed #1137

PR #1153, opened 2020-08-31
John conroy/fix search requests

john-conroy:
Can we just use `useEntityData` hook? They seem similar. I've also looked into using `react-query` which may simplify our data fetching.


john-conroy:
I think consolidating our fetch methods would be a start. We shouldn't be maintaining similar code across the repo.

PR #1166, opened 2020-09-01
Mccalluc/folding facets

### LINKED ISSUE #755: Fold up facets (group + creator in particular)

mccalluc:
Screen Shot 2020-09-01 at 6 28 51 PM

Marking as draft for now: I want to add tests. Otherwise, how does it look? One CSS issue I haven't resolved: Where does the space come from between facet divs, and why do the new divs not get it? Will look a bit more at that tomorrow, but may solve a separate issue for the spacing.

Fix #755.

### DIFF
--- a/CHANGELOG-folding-facets.md
+++ b/CHANGELOG-folding-facets.md
@@ -1,0 +1,1 @@
+- Make "Group" and "Creator" folded by default.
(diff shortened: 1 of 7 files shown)


tsliaw:
Link to design: https://www.figma.com/file/pkIZbpJOAQnylo1zjyhaU6/Search-Page?node-id=1%3A2908
If you press the "play" icon in the upper right hand corner of the Figma, you should be able to see how the folding facets work if you interact with it. (Only the donor search page mockup)

mccalluc:
Accordion https://material-ui.com/components/accordion/ will be useful.

PR #1172, opened 2020-09-03
John conroy/fix donor metadata anchor

### LINKED ISSUE #1158: Missing metadata anchor (at least for donors)
For example: https://portal.stage.hubmapconsortium.org/browse/donor/3439871e0617807d2711465a249af9a5

The other links under `Sections` work, but not `Metadata`.

john-conroy:
Fix #1158. Is it worth refactoring some of the repeated code in the entity detail pages?

### DIFF
--- a/context/app/static/js/components/Detail/DetailLayout/utils.js
+++ b/context/app/static/js/components/Detail/DetailLayout/utils.js
@@ -1,0 +1,12 @@
+import { capitalizeString } from 'js/helpers/functions';
+
+function getSectionFromString(s) {
+  return [s, { text: capitalizeString(s), hash: s }];
+}
+
+function getSections(sectionOrder) {
+  // Array order reflects order of table of contents.
+  return sectionOrder.map((s) => getSectionFromString(s));
+}
+
+export { getSections };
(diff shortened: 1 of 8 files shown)


context/app/static/js/components/Detail/utils.js
```diff
@@ -0,0 +1,5 @@
+export function getSectionOrder(possibleSections, optionalSectionsToInclude) {
+  return possibleSections.filter(
+    (section) => !(section in optionalSectionsToInclude) || optionalSectionsToInclude[section] === true,
```
mccalluc, line 3:
```suggestion
    (section) => !(section in optionalSectionsToInclude) || optionalSectionsToInclude[section],
```

context/app/static/js/components/Detail/DetailLayout/utils.js
```diff
@@ -0,0 +1,12 @@
+import { capitalizeString } from 'js/helpers/functions';
+
+function getSectionFromString(s) {
+  return [s, { text: capitalizeString(s), hash: s }];
```
mccalluc, line 4: I don't think this is a change in this PR, but if it's not complicated, could the lists we give to `getSectionOrder` have the intended capitalization? In the past I've been boxed in when I started like this, and then there was some name that needed non-standard capitalization.

If it's an easy change in this PR, great, or file an issue if you think it's worthwhile.

mccalluc (approved):

PR #1174, opened 2020-09-03
John conroy/fix sample ccf link

### LINKED ISSUE #1159: Link to CCF should not include host
https://portal.stage.hubmapconsortium.org/browse/sample/037a19b6501577e99f6e1d405731330b#tissue

> The spatial coordinates of this sample have been registered and it can be found in the *Common Coordinate Framework Exploration User Interface*.

The link goes to https://portal.hubmapconsortium.org/ccf-eui ... the hostname should not be included.

john-conroy:
Fix #1159. We can remove `target="_blank" rel="noopener noreferrer"`, right?

### DIFF
--- a/CHANGELOG-fix-sample-ccf-link.md
+++ b/CHANGELOG-fix-sample-ccf-link.md
@@ -1,0 +1,1 @@
+- Remove host from sample tissue ccf link.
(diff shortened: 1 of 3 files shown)


mccalluc (approved):
Unless you've heard otherwise, I think keeping the `_blank` on CCF is a good thing: It takes 10-20s to load, so putting it in a new tab makes it a little less disruptive?

PR #1177, opened 2020-09-03
John conroy/flask token and endpoints app provider

### LINKED ISSUE #1110: "We were unable to retrieve provenance information for this donor."
I believe the root cause is a stale token in the cookie.

```
curl 'https://entity-api.stage.hubmapconsortium.org/entities/e0176e0fc05285de9c79eafed45dad7f/provenance' -H 'Authorization: Bearer [token]'
```
returns:
```
Bad Request

An error occurred: <class 'hubmap_commons.exceptions.HTTPException'>No valid authorization token found.<traceback object at 0x7f6bb6d87588>
```

We need to stop using that cookie.

### LINKED ISSUE #1152: Provenance tries to use out-of-date cookie, causing error
I think Provenance is still using the cookie... one consequence is that the cookie token may have expired, and even if you're logged out, it's still being used, and the Prov table and graph fail.

Fix: Use the token from the Flask session everywhere.

john-conroy:
Fix #1152. Fix #1110 .

- Create app context provider which includes endpoints and flask nexus token.
- Use flask nexus token instead of cookie nexus token for all requests.


mccalluc (commented):
With these changes, can we clean up routes_auth.py
```
    response.set_cookie(
        key='nexus_token',
        value=nexus_token)
```
Or is there still JS that's looking for the cookie?

john-conroy:
Removed cookie from `routes_auth.py`

PR #1186, opened 2020-09-04
John conroy/fix prov data hook

john-conroy:
- Fix headers in useProvData hook.

### DIFF
--- a/context/app/static/js/hooks/useProvData.js
+++ b/context/app/static/js/hooks/useProvData.js
@@ -10,7 +10,7 @@ function useProvData(uuid, entityEndpoint, nexusToken) {
     async function getAndSetProvData() {
       const requestInit = getAuthHeader(nexusToken);
 
-      const response = await fetch(`${entityEndpoint}/entities/${uuid}/provenance`, requestInit);
+      const response = await fetch(`${entityEndpoint}/entities/${uuid}/provenance`, { headers: requestInit });
 
       if (!response.ok) {
         console.error('Prov API failed', response);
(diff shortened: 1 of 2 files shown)


context/app/static/js/hooks/useProvData.js
```diff
@@ -10,7 +10,7 @@ function useProvData(uuid, entityEndpoint, nexusToken) {
     async function getAndSetProvData() {
       const requestInit = getAuthHeader(nexusToken);
 
-      const response = await fetch(`${entityEndpoint}/entities/${uuid}/provenance`, requestInit);
+      const response = await fetch(`${entityEndpoint}/entities/${uuid}/provenance`, { headers: requestInit });
```
mccalluc, line 13:
```suggestion
      const response = await fetch(`${entityEndpoint}/entities/${uuid}/provenance`, { headers });
```

context/app/static/js/hooks/useProvData.js
```diff
@@ -10,7 +10,7 @@ function useProvData(uuid, entityEndpoint, nexusToken) {
     async function getAndSetProvData() {
       const requestInit = getAuthHeader(nexusToken);
```
mccalluc, line 11:
```suggestion
      const headers = getAuthHeader(nexusToken);
```

mccalluc (changes requested):
Could you rename the `requestInit`? That name suggests it's an object up a level, with a lot more options. https://developer.mozilla.org/en-US/docs/Web/API/Request/Request

mccalluc (approved):

PR #1188, opened 2020-09-08
John conroy/user email header

### LINKED ISSUE #1030: Implement new login/logout header
Assigned to John, but feel free to reassign since I was not sure who to assign to. See design in #903

### LINKED ISSUE #1178: Fix header design (blending into components)
Header currently blends into components that are also the primary color. 

Material Design recommends:
- Adding elevation
- Translucency to header while scrolling (more for iOS development)

john-conroy:
Fix #1030. Fix #1178.

Should I check the status code for the get request or is this sufficient?

### DIFF
--- a/context/app/routes_auth.py
+++ b/context/app/routes_auth.py
@@ -86,8 +86,9 @@ def login():
     auth_token_object = tokens.by_resource_server['auth.globus.org']
     auth_token = auth_token_object['access_token']
 
-    user_info_request_headers={'Authorization': 'Bearer ' + auth_token}
-    user_info = requests.get('https://auth.globus.org/v2/oauth2/userinfo', headers=user_info_request_headers).json()
+    user_info_request_headers = {'Authorization': 'Bearer ' + auth_token}
+    user_info = requests.get('https://auth.globus.org/v2/oauth2/userinfo',
+                             headers=user_info_request_headers).json()
     user_email = user_info['email'] if 'email' in user_info else 'User'
 
     if not has_hubmap_group(nexus_token):
--- a/context/app/routes_auth.py
+++ b/context/app/routes_auth.py
@@ -89,7 +89,7 @@ def login():
     user_info_request_headers = {'Authorization': 'Bearer ' + auth_token}
     user_info = requests.get('https://auth.globus.org/v2/oauth2/userinfo',
                              headers=user_info_request_headers).json()
-    user_email = user_info['email'] if 'email' in user_info else 'User'
+    user_email = user_info['email'] if 'email' in user_info else ''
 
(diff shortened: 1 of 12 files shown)


tsliaw (approved):

context/app/routes_auth.py
```diff
@@ -83,13 +83,22 @@ def login():
     token_object = tokens.by_resource_server['nexus.api.globus.org']
     nexus_token = token_object['access_token']
 
+    auth_token_object = tokens.by_resource_server['auth.globus.org']
+    auth_token = auth_token_object['access_token']
+
+    user_info_request_headers = {'Authorization': 'Bearer ' + auth_token}
+    user_info = requests.get('https://auth.globus.org/v2/oauth2/userinfo',
+                             headers=user_info_request_headers).json()
+    user_email = user_info['email'] if 'email' in user_info else 'User'
```
mccalluc, line 92: A little uncomfortable here, and with main.py: 
- I think showing "User" when we don't have an user is a UI choice and I'd like to keep it out of the python.
- If the design changes, we'd need update both of these.
- Muddled semantics: `User` isn't an email, and we would never construct something like `mailto:User`.

context/app/static/js/components/Header/LoginButton/LoginButton.spec.js
```diff
@@ -11,7 +12,8 @@ test('should be login button when not authenticated', () => {
 });
 
 test('should be logout button when authenticated', () => {
-  render(<LoginButton isAuthenticated />);
-  expect(screen.getByText('logout')).toBeInTheDocument();
-  expect(screen.getByRole('link')).toHaveAttribute('href', '/logout');
+  render(<LoginButton isAuthenticated user_email="fake@fake.fake" />);
+  expect(screen.getByText('fake@fake.fake')).toBeInTheDocument();
+  userEvent.click(screen.getByText('fake@fake.fake'));
+  expect(screen.getByText('Log Out')).toHaveAttribute('href', '/logout');
```
mccalluc, line 18:
```suggestion
  // In drop-down:
  expect(screen.getByText('Log Out')).toHaveAttribute('href', '/logout');
```

mccalluc (changes requested):
I think the special value `User` should be moved up from the python to the JS. Does that seem ok?

mccalluc:
> Should I check the status code for the get request or is this sufficient?

If it's non 200, in all likelihood it will be an html body and the `.json()` will fail, and we'll get a 500... and that seems fine to me. If we learn the API is really flakey _and_ we need to be more reliable than our APIs, then we can worry about special error handling.

mccalluc (approved):
See note in comment. Approved!

PR #1197, opened 2020-09-09
version bump

mccalluc:
I think I just need to remember to do this, and there's nothing that needs to be more automatic here?


john-conroy (approved):
So the person creating the image will increment the version? Should this be documented in the README?

mccalluc:
Only at the end of each milestone, so they're (roughly) in sync. The problem is just remembering to do it... not sure if adding it to the README would have changed anything.

PR #1204, opened 2020-09-10
John conroy/style error boundary

### LINKED ISSUE #1181: Style react error boundary
https://portal.test.hubmapconsortium.org/search?entity_type[0]=asdf

Style it like the other error pages

john-conroy:
Fix #1181.

Screen Shot 2020-09-09 at 6 05 14 PM

### DIFF
--- a/context/app/static/js/pages/Error/utils.js
+++ b/context/app/static/js/pages/Error/utils.js
@@ -1,4 +1,4 @@
-export function getErrorTitleAndSubtitle(errorCode, isMaintenancePage) {
+export function getErrorTitleAndSubtitle(errorCode, isMaintenancePage, isErrorBoundary) {
   const errorTitle = {
     400: 'Bad Request',
     401: 'Unauthorized',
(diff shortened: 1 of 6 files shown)


tsliaw (approved):

context/app/static/js/pages/Error/utils.js
```diff
@@ -12,6 +12,10 @@ export function getErrorTitleAndSubtitle(errorCode, isMaintenancePage) {
     return { title: 'Portal Maintenance', subtitle: 'Portal unavailable for scheduled maintenance.' };
   }
 
+  if (isErrorBoundary) {
+    return { title: 'Error', subtitle: `URL: ${window.location.href}` };
+  }
+
   const expectedTitle = errorTitle?.[errorCode];
```
mccalluc, line 19: I didn't know `?.` was available now. Thanks!

mccalluc (approved):
Approved... But I wonder if we might want to handle different types of errors differently. I'm thinking of the bad entity types on search: Perhaps throw a UserError we'll create, and in the UI, make it clear that this is the user's problem, not ours. Like the distinction between a 4xx and 5xx.

This is fine to merge, but make an issue from my note above, if you think it'd be a good distinction to have?

PR #1206, opened 2020-09-10
John conroy/fix metadata table short circuit

### LINKED ISSUE #1203: Dataset metadata table not being displayed when metadata exists
I need to update the short circuit in Dataset's return.

john-conroy:
Fix #1203


mccalluc (approved):
If it is confusing, would we want to bring the keys and the component names in line with each other? but that doesn't need to block this. (And there may be reasons why it's not feasible.)

PR #1207, opened 2020-09-10
John conroy/redirect to previous page on login

### LINKED ISSUE #354: Redirect to previous page after login
Right now:
- You click on a link for a details page.
- You get an access-denied message.
- You login
- You end up on the welcome page.

Todo:
- after globus login, should redirect back to original page.

john-conroy:
Fix #354.

@mccalluc this is working, can you take a look to see if this aligns with what you were imagining?


context/app/routes_auth.py
```diff
@@ -99,8 +99,12 @@ def login():
         nexus_token=nexus_token,
         is_authenticated=True,
         user_email=user_email)
+
+    previous_path_and_search = unquote(request.cookies.get('urlPathAndSearchBeforeLogin'))
```
mccalluc, line 103: Is this just the path and search? or the whole url? I can't remember why the `_external=True` was needed below, but should make sure we don't trip over the same issue. (Looking at the `git blame` might refresh my memory, if needed.)

context/app/static/js/components/Routes/Routes.jsx
```diff
@@ -11,13 +11,16 @@ import Preview from '../Preview';
 import { Collections } from '../Collections';
 import Markdown from '../Markdown';
 import useSendPageView from './useSendPageView';
+import useSetUrlPathAndSearchBeforeLogin from './useSetUrlPathAndSearchBeforeLogin';
 
 function Routes(props) {
   const { flaskData } = props;
   const { entity, vitessce_conf, title, markdown, collection, errorCode } = flaskData;
   const urlPath = window.location.pathname;
+  const pathAndSearch = (window.location.pathname + window.location.search).substr(1);
```
mccalluc, line 20: Why not capture the whole URL?

mccalluc: Yeah, and then the `external=True` becomes moot. And if there's a fragment in the URL, good to capture it, too.

context/app/static/js/components/Routes/useSetUrlPathAndSearchBeforeLogin.js
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from 'react';
+import Cookies from 'universal-cookie';
+
+function useSetPathAndSearchBeforeLogin(path) {
+  useEffect(() => {
+    const cookies = new Cookies();
+    if (path !== '/login' && path !== '/logout') {
```
mccalluc, line 7: On the javascript side, will the path ever be /login or /logout? I had thought those were handled by python.

mccalluc (commented):
Can you just capture the whole URL? Otherwise, great!

john-conroy:
I originally used the url, but it wasn't working. I'll take another look.

mccalluc (approved):

PR #1210, opened 2020-09-11
John conroy/fix vitessce theme switch

### LINKED ISSUE #1208: Clicking Vitessce light/dark theme button twice removes theme
I am guessing this is because the `onChange` callback is called with `null` or `undefined` and then the theme is neither "light" nor "dark".

With no theme, the buttons and other styles look ugly

Screen Shot 2020-09-11 at 2 15 45 PM

john-conroy:
Fix #1208. Let me know if there is a more sensible solution and I'd be glad to implement it.

### DIFF
--- a/CHANGELOG-fix-vitessce-theme-switch.md
+++ b/CHANGELOG-fix-vitessce-theme-switch.md
@@ -1,0 +1,1 @@
+- Fix bug where clicking a visualization theme switch button twice removes theme.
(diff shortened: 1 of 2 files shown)


keller-mark (approved):
I like this solution. I tested locally and looks good.

keller-mark:
Thanks!

PR #1216, opened 2020-09-14
John conroy/fix analysis details link ext bug

### LINKED ISSUE #1215: Handle dag provenance origin fields which do not have .git extension in provenance analysis details
From slack:

> future versions will not include the '.git' extension but some data already present in the PROD database does include the '.git'.  In any case, both URLs are valid, so I think it's best to accept both forms.

john-conroy:
Towards #1215. As noted in the issue, I will wait to close it until the slack thread is concluded. Does this work or would you prefer a regex solution?

Update: Fix #1215.

### DIFF
--- a/context/app/static/js/components/Detail/ProvAnalysisDetailsLink/ProvAnalysisDetailsLink.jsx
+++ b/context/app/static/js/components/Detail/ProvAnalysisDetailsLink/ProvAnalysisDetailsLink.jsx
@@ -6,7 +6,7 @@ import { CwlIcon, FlexLightBlueLink, PrimaryTextDivider, StyledListItem } from '
 function ProvAnalysisDetailsLink(props) {
   const { data } = props;
 
-  const trimmedOrigin = data.origin.endsWith('.git') ? data.origin.split('.').slice(0, -1).join('.') : data.origin;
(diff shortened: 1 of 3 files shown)


john-conroy:
Fix #1215.

context/app/static/js/components/Detail/ProvAnalysisDetailsLink/ProvAnalysisDetailsLink.jsx
```diff
@@ -6,7 +6,7 @@ import { CwlIcon, FlexLightBlueLink, PrimaryTextDivider, StyledListItem } from '
 function ProvAnalysisDetailsLink(props) {
   const { data } = props;
 
-  const trimmedOrigin = data.origin.split('.').slice(0, -1).join('.');
+  const trimmedOrigin = data.origin.endsWith('.git') ? data.origin.split('.').slice(0, -1).join('.') : data.origin;
```
mccalluc, line 9:
```suggestion
  const trimmedOrigin = data.origin.replace(/\.git$/, '');
```
I think this would be more clear.

mccalluc (changes requested):
Prefer regex, but everything else looks good. Thanks!

PR #1220, opened 2020-09-15
John conroy/fix error page html titles

### LINKED ISSUE #1213: Error pages are missing tite
Error page titles are all " | HuBMAP".

They should be "<Error Type> | HuBMAP", e.g., "Page Not Found | HuBMAP".

image

john-conroy:
Fix #1213. This should fix error page html titles for all flask registered error routes.

### DIFF
--- a/context/app/main.py
+++ b/context/app/main.py
@@ -56,7 +56,7 @@ def create_app(testing=False):
     app.register_error_handler(404, not_found)
     app.register_error_handler(504, gateway_timeout)
 
-    @ app.context_processor
+    @app.context_processor
     def inject_template_globals():
         return {
             'is_authenticated': session.get('is_authenticated'),
--- a/context/app/main.py
+++ b/context/app/main.py
@@ -5,8 +5,18 @@ from .flask_static_digest import FlaskStaticDigest
 flask_static_digest = FlaskStaticDigest()
 
 
+def get_error_title(code):
+    error_titles = {
+        400: 'Bad Request',
+        401: 'Unauthorized',
+        404: 'Page Not Found',
+        504: 'Gateway Timeout',
+    }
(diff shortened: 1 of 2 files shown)


mccalluc (commented):
Sorry to be picky, but could we add a second argument to `render_react_error`, and specify the messages where it's called from `main.py`, for instance:
```diff
def bad_request(e):
    '''A 400 means the request to the API failed.'''
-    return render_react_error(400)
+    return render_react_error(400, 'Bad Request')
```

We want to avoid having to keep two parallel lists in sync (the functions, and the dict).

john-conroy:
Not picky, thanks for the suggestion. It was just one of those moments where you overlook the obvious solution.

context/app/main.py
```diff
@@ -54,15 +56,15 @@ def create_app(testing=False):
     app.register_error_handler(404, not_found)
     app.register_error_handler(504, gateway_timeout)
 
-    @app.context_processor
+    @ app.context_processor
```
mccalluc, line 59:
```suggestion
    @app.context_processor
```
I think no space is more usual?

context/app/main.py
```diff
@@ -54,15 +56,15 @@ def create_app(testing=False):
...
     def inject_template_globals():
         return {
             'is_authenticated': session.get('is_authenticated'),
             'nexus_token': session.get('nexus_token'),
             'user_email': session.get('user_email')
         }
 
-    @app.before_request
+    @ app.before_request
```
mccalluc, line 67:
```suggestion
    @app.before_request
```

mccalluc (commented):
Thanks for the tweaks! Probably revert adding the spaces after `@` and then merge.

PR #1238, opened 2020-09-22
John conroy/move pages

john-conroy:
Just moving pages from components directory to pages. I'll fix directory capitalization in a later PR.


mccalluc (approved):
To confirm, top level pages are in one place, and the components used are in a separate place? And it doesn't make sense to have components under the page because many components are used across pages, and we want to be consistent for all components?

And when should `__tests__` be used? Only in certain kinds of parent directories?

(With the master merge I may be a bit confused about what's actually going on; Sorry!)

PR #1248, opened 2020-09-25
John conroy/search id link

john-conroy:
Link to detail page in each row's HuBMAP ID cell not the entire row in search table. The row and highlight should look the same, but will no longer link to the detail page.

Screen Shot 2020-09-25 at 4 09 13 PM

Screen Shot 2020-09-25 at 4 08 59 PM

### DIFF
--- a/context/app/static/js/components/Search/ResultsTable.jsx
+++ b/context/app/static/js/components/Search/ResultsTable.jsx
@@ -5,6 +5,7 @@ import Table from '@material-ui/core/Table';
 
 import { SortingSelector, Hits } from 'searchkit'; // eslint-disable-line import/no-duplicates
 
+import { LightBlueLink } from 'js/shared-styles/Links';
 import { StyledTableBody, StyledTableRow, StyledTableCell } from './style';
 
 import SortingTableHead from './SortingTableHead';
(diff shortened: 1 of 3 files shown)


tsliaw:
I suggest that you should take out the hover effect since that makes it look like the whole row is clickable when it's not anymore.

context/app/static/js/components/Search/ResultsTable.jsx
```diff
@@ -38,20 +39,23 @@ function makeTableBodyComponent(resultFields, detailsUrlPrefix, idField) {
       <>
         {hits.map((hit) => (
           <StyledTableBody key={hit._id}>
-            <StyledTableRow
-              className={'highlight' in hit && 'before-highlight'}
-              onClick={() => window.location.assign(detailsUrlPrefix + hit._source[idField])}
-              role="row link"
-            >
+            <StyledTableRow className={'highlight' in hit && 'before-highlight'}>
```
mccalluc, line 42:
```suggestion
            <StyledTableRow
              className={'highlight' in hit && 'before-highlight'}
              onClick={(event) => {
                if (
                  !window.getSelection().toString() && // If text selected, user doesn't want to navigate.
                  event.target.localName !== 'a' // Let <a> handle itself, for command-clicks.
                ) {
                  window.location.assign(detailsUrlPrefix + hit._source[idField]);
                }
              }}
              role="row link"
            >
```
Maybe?

mccalluc (commented):
Consider the diff below, and put this on the agenda for the UI meeting? Or I'm happy to defer to Tiffany.

context/app/static/js/components/Search/ResultsTable.jsx
john-conroy: Per ui meeting, we're fine with just the ID link.

john-conroy:
We've opted to keep the hover effect.

PR #1253, opened 2020-09-25
Mccalluc/doc maintenance build

mccalluc:
Since I don't want to assume how folks manage their checkouts, and since we'll be doing this infrequently, this seems better than a script... but I could be swayed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ cd ${YOUR_HUBMAP_REPOS}/gateway
 git checkout master; git pull
 git checkout -b update-portal-ui-maintenance
 cp ${YOUR_HUBMAP_REPOS}/portal-ui/context/app/static/js/maintenance/public/* \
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -80,6 +80,27 @@ Then, to redeploy `dev`, `test`, and `stage`:
...
+<details>
+
+```
+cd context/
+npm run build:maintain
+cd ${YOUR_HUBMAP_REPOS}/gateway
+git checkout master; git pull
+git checkout -b update-portal-ui-maintenance
+cp ${YOUR_HUBMAP_REPOS}/portal-ui/context/app/static/js/maintenance/public/* \
+   nginx/html/portal-ui/maintenance/
```
mccalluc, line 97:
```suggestion
   nginx/html/portal-ui-maintenance/
```

john-conroy (approved):
Works for me, thanks.

PR #1254, opened 2020-09-27
dynamic robots.txt: Fix #1224

### LINKED ISSUE #1224: dynamic robots.txt
I think the other environments are only visible if the client IP is on the allow list, but with the new deploy machinery, this might be loosened. In any case, I think it would be good to have a dynamic robots.txt that preventing indexing of non-production versions of the site.

mccalluc:


john-conroy (approved):
We're doing this to reduce crawling and remove other portal environments from Google index?

mccalluc:
Right... it's not actually a problem right now... but we want Bill to set up a second instance that points at the PROD APIs, and I think that might not be behind the IP block list, and so it might be spidered. (I'm not so concerned about the other instances which are up right now, because they shouldn't be visible to robots, because of the IP list.)

PR #1264, opened 2020-09-30
Mccalluc/log doc validation errors

mccalluc:
The next time we get a reindex, following https://github.com/hubmapconsortium/search-api/pull/188, there will be information about schema validation errors. To me, just logging this at the earliest possible point seems like the least obtrusive way of making this information available. (I'll make another PR adding another boolean facet to dev-search: Until I've seen them work, they'll remain drafts.)

### DIFF
--- a/context/app/static/js/index.jsx
+++ b/context/app/static/js/index.jsx
@@ -3,6 +3,12 @@ import ReactDOM from 'react-dom';
 
 import App from './components/App';
 
+// eslint-disable-next-line no-undef
+const { validation_errors } = flaskData.entity.mapper_metadata;
+if (validation_errors && validation_errors.length) {
+  console.warn('Schema validation errors', validation_errors);
+}
+
 ReactDOM.render(
   // eslint-disable-next-line no-undef
   <App flaskData={flaskData} />,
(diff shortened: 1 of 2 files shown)


mccalluc:
(Keeping as draft until we can see this work against a real index.)

mccalluc:
Validation errors are now in the production index, and this is doing what it should do. There are a *lot* of validation errors, but they are bundled together so the console isn't too noisy. Fixing them is a separate issue.

context/app/static/js/index.jsx
```diff
@@ -3,6 +3,12 @@ import ReactDOM from 'react-dom';
 
 import App from './components/App';
 
+// eslint-disable-next-line no-undef
+const validation_errors = flaskData.entity?.mapper_metadata?.validation_errors;
+if (validation_errors && validation_errors.length) {
```
john-conroy, line 8:
```suggestion
if (validation_errors) {
```
I don't think we have to worry about a falsey value here since we're not displaying anything, but we can keep it if you want to be safe.

john-conroy (approved):

context/app/static/js/index.jsx
john-conroy: (and it's also not a short circuit)

john-conroy:
Disregard previous comments. Looks good.

PR #1280, opened 2020-10-05
John conroy/fix first header render

### LINKED ISSUE #1266: Header initially renders without elevation and quickly updates on entity detail pages.

### LINKED ISSUE #1269: Entity subheader quickly appears then disappears on render
> The subheader is looking great, in particular with the animation. I noticed that it pops up for a short moment before disappearing when I load the page (on TEST). Is that a hard problem to resolve?

john-conroy:
Fix #1266, Fix #1269. Let me know if anything is unclear and you feel like we'd benefit from a comment or two. This should be the last batch of fixes for the subheader.

### DIFF
--- a/context/app/static/js/components/Detail/entityHeader/EntityHeaderItem/EntityHeaderItem.jsx
+++ b/context/app/static/js/components/Detail/entityHeader/EntityHeaderItem/EntityHeaderItem.jsx
@@ -14,7 +14,7 @@ function EntityHeaderItem({ text }) {
 }
 
 EntityHeaderItem.propTypes = {
-  text: PropTypes.string,
+  text: PropTypes.oneOfType(PropTypes.string, PropTypes.arrayOf(PropTypes.string)),
 };
 
 EntityHeaderItem.defaultProps = {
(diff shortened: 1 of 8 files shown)


context/app/static/js/components/Detail/entityHeader/EntityHeaderItem/EntityHeaderItem.jsx
```diff
@@ -14,7 +14,7 @@ function EntityHeaderItem({ text }) {
 }
 
 EntityHeaderItem.propTypes = {
-  text: PropTypes.string,
+  text: PropTypes.oneOfType(PropTypes.string, PropTypes.arrayOf(PropTypes.string)),
```
mccalluc, line 17: Not sure exactly how this is used, but could we say it's always an array, and perhaps the length is just one? Since both array and string can be subscripted, and have a length, I'd really like to avoid unioning these types.

mccalluc (commented):
One comment below. Since this is a bug-fix, perhaps go ahead and merge it, and maybe tweak it for the next release?

john-conroy:
@mccalluc  gave approval to merge over slack.

PR #1287, opened 2020-10-06
Fix "Invalid argument supplied to oneOfType"

mccalluc:


john-conroy (approved):
Is this a temporary fix just to get rid of the warning? We wanted to remove this per yesterday's slack conversation.

mccalluc:
If you'll be getting rid of it entirely, great!... Until then, I'd still like to avoid spurious errors: That was really just a syntax problem.

PR #1290, opened 2020-10-07
John conroy/standardize icon buttons

john-conroy:
Fix bug regarding visualization collapse button placement in safari. All icon buttons now have a uniform height, width and svg icon size.

Screen Shot 2020-10-07 at 1 32 52 PM

Screen Shot 2020-10-07 at 2 40 33 PM

### DIFF
--- a/context/app/static/js/components/Detail/entityHeader/EntityHeader/EntityHeader.jsx
+++ b/context/app/static/js/components/Detail/entityHeader/EntityHeader/EntityHeader.jsx
@@ -15,7 +15,7 @@ const visualizationSelector = (state) => ({
   vizIsFullscreen: state.vizIsFullscreen,
 });
 
-const entityHeaderHeight = 35;
+const entityHeaderHeight = 40;
 
 function Header() {
   const {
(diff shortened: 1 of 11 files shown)


context/app/static/js/components/Detail/entityHeader/EntityHeader/EntityHeader.jsx
```diff
@@ -15,7 +15,7 @@ const visualizationSelector = (state) => ({
   vizIsFullscreen: state.vizIsFullscreen,
 });
 
-const entityHeaderHeight = 35;
+const entityHeaderHeight = 40;
```
mccalluc, line 18: Consider exporting this from style.js, if it should be kept in sync.

context/app/static/js/shared-styles/buttons/index.js
```diff
@@ -5,27 +5,34 @@ import ToggleButton from '@material-ui/lab/ToggleButton';
 
 const WhiteBackgroundButtonCSS = css`
   background-color: #fff;
-  padding: 10px;
+  height: 40px;
+  width: 40px;
```
mccalluc, line 9: ... and should these also be kept in sync?

mccalluc (commented):
There's several places were 40 comes up... should we define this once and then reference this from other locations?

tsliaw (approved):

mccalluc (approved):

PR #1292, opened 2020-10-09
Mccalluc/machine readable metadata

mccalluc:
@tsliaw : This is one more channel through which we'll be presenting HuBMAP, so I'd like you to think if this makes sense.

### DIFF
--- a/context/app/static/js/index.jsx
+++ b/context/app/static/js/index.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 import App from './components/App';
+import { setJsonLD } from './schema.org';
 
 ReactDOM.render(
   // eslint-disable-next-line no-undef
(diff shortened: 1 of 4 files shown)


context/app/static/js/schema.org.js
```diff
@@ -0,0 +1,43 @@
+function getDatasetLD(entity) {
+  // Based on https://developers.google.com/search/docs/data-types/dataset#guidelines
+  const donor = entity.donor.mapped_metadata;
+  const shortDonorString = `${donor.sex || '(Unknown sex)'}, ${donor.age_value || '(Unknown age)'} ${
+    donor.age_unit || '(Unknown age unit)'
+  } old`;
+  const heightString = `${donor.height_value || ''}${donor.height_unit || ''}`;
+  const weightString = `${donor.weight_value || ''}${donor.weight_unit || ''}`;
+  const longDonorString = `${heightString}, ${weightString}, ${donor.race || ''} ${shortDonorString}`;
```
john-conroy, line 9: Could this have a lot of extra spaces if heightString, weightString or race are undefined?

context/app/static/js/index.jsx
```diff
@@ -2,9 +2,16 @@ import React from 'react';
...
+import { setJsonLD } from './schema.org';
 
 ReactDOM.render(
   // eslint-disable-next-line no-undef
   <App flaskData={flaskData} />,
   document.getElementById('react-content'),
 );
+
+/* eslint-disable no-undef */
+if (flaskData?.entity?.entity_type === 'Dataset') {
```
john-conroy, line 14: Could we do this from the dataset page? Would that make sense? More so a question than a suggestion.

john-conroy (approved):
Does it matter of any of this changes down the road? Do we have a responsibility to keep this schema consistent?

context/app/static/js/schema.org.js
mccalluc: yes, it could, but my sense is that these will be provided for every donor, so I don't want to make the code more complicated than it needs to be. I'm happy with this... but if you think it could be less fragile, happy to do so.

context/app/static/js/index.jsx
mccalluc: My sense is that react code shouldn't have DOM-altering side effects, so I didn't want to push this farther down... So, I'm happy with this, but happy to change, too, if you think that would be better.

PR #1300, opened 2020-10-15
Mccalluc/api status

### LINKED ISSUE #1187: Add service status page and link in footer
We should make the portal UI version and the software versions of other services accessible through the footer.

mccalluc:
@tsliaw : This is blocked by https://github.com/hubmapconsortium/gateway/issues/58, and I need to tidy up the code, but this could plausibly be the UI:

Screen Shot 2020-10-14 at 9 58 00 PM

Thoughts about
-  the text at the top?
-  handing of missing values?
-  notes field?
- ordering?

and whatever else you think needs commenting on. Thanks!

fix #1187


tsliaw (commented):
Looks good for the most part. 
Comments:
- Add link to page in footer
- When there isn't a Github repo, it says "none", while for other fields like Version Number, it says "na". I would standardize one to use if possible. 
- I don't think the Notes section currently is very understandable, but you should ask Nils for input about how to display those information. 
- For the text on the top explaining what the service page is about, you should ask Nils for input/comments. But looks good to me.

mccalluc:
(Still waiting for Cors to be fixed, so still a draft.)

@ngehlenborg : Over-all UI ok? And what should it say up top? Currently:
> HuBMAP is powered by a number of APIs. The Portal depends directly on only search-api, entity-api, and assets. Other services use the others listed below. 

@john-conroy : Code organization... does the `ServiceStatus` vs. `ServiceStatusTable` seem right? Should these
```
import SectionHeader from 'js/components/Detail/SectionHeader';
import SectionContainer from 'js/components/Detail/SectionContainer';
import Description from 'js/components/Detail/Description';
```
be pulled up into `js/shared-styles`?

Screen Shot 2020-10-16 at 3 35 00 PM

ngehlenborg:
I would include the portal ui service as well. While the status is obviously always up when that page is visible, it is still useful for linking to Github and having the version numbers and build info (assuming that we are aligning our service info with the one that the IEC is using. Other than that is it looking great.

ngehlenborg (changes requested):
Please include portal UI service in the list as well.

context/app/static/js/pages/Services/Services.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import Paper from '@material-ui/core/Paper';
+
+import SectionHeader from 'js/components/Detail/SectionHeader';
+import SectionContainer from 'js/components/Detail/SectionContainer';
```
john-conroy, line 5: Yes, I think they can be moved into `shared-styles`.

context/app/static/js/components/ServiceStatusTable/ServiceStatusTable.jsx
```diff
@@ -0,0 +1,92 @@
...
+    version,
+    isUp,
+    note: noteFunc(response),
+  };
+}
+
+function ServiceStatusTable() {
+  const gatewayStatus = useGatewayStatus();
+
+  const apiStatuses = gatewayStatus
```
john-conroy, line 34: This operation could be synced with changes to `gatewayStatus` so it doesn't run on each render, but I think it's lightweight enough that it doesn't matter.

context/app/static/js/components/ServiceStatusTable/ServiceStatusTable.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+import { StyledExternalLinkIcon } from 'js/components/files/GlobusLinkMessage/style';
+
+import { HeaderCell } from 'js/shared-styles/Table';
+import { LightBlueLink } from 'js/shared-styles/Links';
+
+import StatusIcon from './StatusIcon';
+import { useGatewayStatus } from './hooks';
+
+function buildServiceStatus(name, response, noteFunc) {
```
john-conroy, line 16: Maybe pull into a utils file?

john-conroy (approved):
Nothing blocking. From the screenshot, the status icon and text looked a little misaligned, but maybe you've fixed this?

context/app/static/js/pages/Services/Services.jsx
mccalluc: Done. Touched a lot of other files in the process, but I think the end result makes sense.

context/app/static/js/components/ServiceStatusTable/ServiceStatusTable.jsx
mccalluc: done.

mccalluc:
> having the version numbers and build info (assuming that we are aligning our service info with the one that the IEC is using. 

@nils: That will take a little more work, but I've filed #1310 for that and we can come back to it.

mccalluc:
(I think I've satisfied everyone's concerns so I'll go ahead and merge.)

PR #1303, opened 2020-10-16
John conroy/search tile view

john-conroy:
Adds a tile view to SearchPages. I'd suggest running this locally to see if everything works to your expectation. As we discussed in standup, there is room for confusion with how sorting is handled in the table versus the tiles so I put an item on the UI agenda to discuss the issue.

There are a few follow up PRs to review as well.
- #1308 Makes spacing in the sidebar more compact and updates accordion font color/size.
- #1309 Changes the components displayed when no results are returned or an error is experienced when searching.
- #1311 Decreases the size of the sort arrow icons and slightly repositions them.
- #1314 Fixes dev search page following the changes to search.
- #1319 Fixes the tiles sort dropdown so that it doesn't display the selected option in its available options.

Areas to focus review: 
- Utility functions in `ResultsTiles/utils.js` handle shaping descendant counts and setting default values.
- `SearchViewSwitch`, I opted for a store for ease of use, but I could have used local state which would be passed throughout the components as props.
- The updated `resultsFields` in `Search/config.js`  handle the different fields in table/tile view and how they are combined in `SearchWrapper`. I think it's fine to return data which isn't used by the table in the ES hits, but since the table depends on `resultsFields` to determine what's shown we couldn't use just a single array. Maybe there is a better way to organize this?
- `TilesSortDropdown` for general logic/organization.

`ResultsTable` and `SortingTableHead` are just reorganizations so there shouldn't be changes there.

Screen Shot 2020-10-21 at 3 35 05 PM
Screen Shot 2020-10-21 at 3 34 48 PM


context/app/static/js/components/Search/Results/Results.jsx
```diff
@@ -0,0 +1,48 @@
...
+import PropTypes from 'prop-types';
+import { ViewSwitcherHits } from 'searchkit'; // eslint-disable-line import/no-duplicates
+
+import ResultsTable from '../ResultsTable';
+import ResultsTiles from '../ResultsTiles';
+
+function Results(props) {
+  const { sortOptions, hitsPerPage, resultFields, detailsUrlPrefix, idField, resultFieldIds, type } = props;
+
+  return (
```
mccalluc, line 11:
```suggestion
  // Both the table and tiles are rendered, no matter the view.
  // Conditionally rendering one or the other is complicated because ...
  return (
```
If you can remind our future selves why it's being done this way. Thanks!

context/app/static/js/components/Search/config.js
```diff
@@ -18,20 +19,28 @@ function makeDonorMetadataFilters(isDonor) {
...
+      field('display_doi', 'Donor'),
+      field('group_name', 'Group'),
+      field(`mapped_metadata.${ageField}`, 'Age'),
+      field(`mapped_metadata.${bmiField}`, 'BMI'),
+      field('mapped_metadata.sex', 'Sex'),
+      field('mapped_metadata.race', 'Race'),
+      field('mapped_last_modified_timestamp', 'Last Modified'),
+    ],
+    tile: [...sharedTileFields, field(`mapped_metadata.${ageUnitField}`, 'Age Unit')],
+  },
```
mccalluc, line 43: Would it be simpler just always get a superset of fields that will work for both UIs? I don't think over-fetching is a big concern, and I wouldn't want to add complexity to the code just to make the requests a little smaller.

But not a big deal: Your call.

context/app/static/js/pages/search/Search.jsx
```diff
@@ -53,7 +53,9 @@ function Search(props) {
     // Search results fields to display in table:
     resultFields,
     // Default hitsPerPage is 10:
-    hitsPerPage: 20,
+    hitsPerPage: 18,
```
mccalluc, line 56: Huh: The tile view seems to be one short of a multiple of three... maybe the results are getting truncated, somehow?

Screen Shot 2020-10-23 at 1 50 28 PM

mccalluc (commented):
Some comments in line, plus:
- Add some height to the pull down: it's weird to have to scroll just to see the lower half of the last option. (I'm on FF.)
Screen Shot 2020-10-23 at 1 32 31 PM


- The sort by "Dataset" option is weird... in the table it's clear what it does, even if there's no reason for a user to use it... but at the top of the list, it may be confusing. Maybe remove "dataset" sort from both views? -- Nothing that should go in this PR, but give it some throught, and add a UI meeting item, if you think it's worthwhile.

Make a couple tweaks and merge when you think it's ready.

context/app/static/js/components/Search/config.js
john-conroy: I think the way it is setup we are over-fetching which I agree isn't a big concern. The reason why these are kept separately is to be able to pass the specific fields needed to `ResultsTable` and `SortingTableHead`. Maybe I should rename the variables to make more sense?

john-conroy: So we could keep a resultsFields superset and also a tableFields variable to inform `ResultsTable` and `SortingTableHead`?

context/app/static/js/pages/search/Search.jsx
john-conroy: Can you send me the link? I can't reproduce. Was this on the first page?

context/app/static/js/components/Search/Results/Results.jsx
john-conroy: Only one of the sort components must stay mounted (the sort dropdown). I'll add a comment to reflect that.

context/app/static/js/components/Search/config.js
john-conroy: Per our Zoom conversation, the way this is now handled is fine, but worth monitoring if things get more complex in the future.

context/app/static/js/pages/search/Search.jsx
john-conroy: We can't reproduce this issue.

PR #1314, opened 2020-10-21
John conroy/fix dev search

john-conroy:
Fixes dev search page following the changes to search.

### DIFF
--- a/context/app/static/js/pages/search/DevSearch.jsx
+++ b/context/app/static/js/pages/search/DevSearch.jsx
@@ -5,6 +5,7 @@ import { getAuthHeader } from 'js/helpers/functions';
 import { AppContext } from 'js/components/Providers';
 import { field, listFilter, checkboxFilter } from 'js/components/Search/utils';
 import SearchWrapper from 'js/components/Search/SearchWrapper';
+import DevResults from 'js/components/Search/DevResults';
 import { SearchHeader } from './style';
 
 function DevSearch() {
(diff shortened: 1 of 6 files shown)


context/app/static/js/pages/search/DevSearch.jsx
```diff
@@ -22,12 +23,15 @@ function DevSearch() {
...
-      field('mapper_metadata.size', 'Doc Size'),
-    ],
+    resultFields: {
+      table: [
+        field('entity_type', 'type'),
+        field('display_doi', 'ID'),
+        field('mapped_last_modified_timestamp', 'Last Modified'),
+        field('mapper_metadata.size', 'Doc Size'),
+      ],
+      tile: [],
```
mccalluc, line 33: Ditto from the earlier PR, wonder if we can just keep this flatter and request a superset.

context/app/static/js/stores/useSearchViewStore.js
```diff
@@ -1,7 +1,9 @@
 import create from 'zustand';
 
 const useSearchViewStore = create((set) => ({
-  searchView: new URLSearchParams(window.location.search).get('view') || 'table',
+  searchView:
+    (window.location.pathname.startsWith('/search') && new URLSearchParams(window.location.search).get('view')) ||
+    'table',
```
mccalluc, line 6: I'm not sure this is needed: `new URLSearchParams(window.location.search).get('view') || 'table'` will evaluate to `'table'` on the /dev-search.

mccalluc (changes requested):
If it's an option, I think I would prefer a dev-search that had a tile view with some missing data to this. I worry about the little configuration details that are repeated in Results.jsx and DevResults.jsx, and I want to avoid creating new code to support dev-search. Maybe talk about this?

context/app/static/js/stores/useSearchViewStore.js
john-conroy: Unless a user enters `&view=tile` in the url. Which we might not care about?

context/app/static/js/pages/search/DevSearch.jsx
john-conroy: We agreed this is fine for now, more in #1303.

context/app/static/js/stores/useSearchViewStore.js
john-conroy: Per our Zoom conversation, we shouldn't worry about users modifying the url and things breaking. Prefer to keep code simpler.

PR #1317, opened 2020-10-22
John conroy/update home page title

### LINKED ISSUE #1315: Change title of portal homepage
The title (HTML <title> element) of the portal homepage should be "HuBMAP Data Portal". Currently, the title is "Welcome | HuBMAP", which doesn't look great in Google search results.

image

john-conroy:
Fix #1315.

### DIFF
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -43,7 +43,7 @@ def index():
         types=types,
         flask_data=core_props,
         title='HuBMAP Data Portal',
-        isHomePage=True
+        is_home_page=True
     )
 
 
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -42,7 +42,8 @@ def index():
         'pages/base_react.html',
         types=types,
         flask_data=core_props,
-        title='Welcome'
+        title='HuBMAP Data Portal',
+        isHomePage=True
     )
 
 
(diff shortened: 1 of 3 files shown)


context/app/routes_main.py
```diff
@@ -42,7 +42,8 @@ def index():
         'pages/base_react.html',
         types=types,
         flask_data=core_props,
-        title='Welcome'
+        title='HuBMAP Data Portal',
+        isHomePage=True
```
mccalluc, line 46:
```suggestion
        is_home_page=True
```

context/app/templates/pages/base_react.html
```diff
@@ -11,7 +11,7 @@
   
     {% block extra_stylesheets %}{% endblock %}
 
-    <title>{% block title_text %}{{ title }}{% endblock %} | HuBMAP</title>
+    <title>{% block title_text %}{{ title }}{% endblock %}{{' | HuBMAP' if not isHomePage}}</title>
```
mccalluc, line 14:
```suggestion
    <title>{% block title_text %}{{ title }}{% endblock %}{{' | HuBMAP' if not is_home_page}}</title>
```

mccalluc (commented):
Looks good... I think we decided that we would be prefer python underscore naming conventions leaking into JS, rather than the reverse? But do what you think best.

context/app/routes_main.py
john-conroy: The casing was just reflex and not intentional.

PR #1318, opened 2020-10-22
John conroy/disable prov table snare rna

### LINKED ISSUE #1294: Disable Prov table for SNARE RNA
The table isn't accurate with the provenance we have from this type: Shouldn't be displayed. Example (which might not exist in the future): https://portal.dev.hubmapconsortium.org/browse/dataset/7593d3e07e7eacb9d07aeedc3996fdfb

john-conroy:
Fix #1294. This was the value in the `data_types` array for the example provided in the issue. Would we want to be more broad and disable the table for any data types which include the words `snare` and `rna` or similar?

### DIFF
--- a/CHANGELOG-disable-prov-table-snare-rna.md
+++ b/CHANGELOG-disable-prov-table-snare-rna.md
@@ -1,0 +1,1 @@
+- Disable provenance table for salmon_rnaseq_snareseq data types.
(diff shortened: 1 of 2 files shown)


mccalluc (approved):
My first impulse is to prefer the exact matches... but I'm really not sure about the larger story... Phil blood does plan to change the IDs, and we don't want to re-break.

I'd say merge as is, but put an item on the UI agenda about how fuzzy our matches should be?

PR #1319, opened 2020-10-23
John conroy/fix and test tiles sort dropdown

john-conroy:
Fixes the dropdown ~~so that it doesn't display the selected option in its available options~~ and adds a few tests.

### DIFF
--- a/context/app/static/js/shared-styles/dropdowns/index.js
+++ b/context/app/static/js/shared-styles/dropdowns/index.js
@@ -23,7 +23,7 @@ function DropdownSelectItem(props) {
   return (
     <FlexMenuItem className={className} {...rest}>
       {isSelected && <CheckIcon color="primary" />}
-      <StyledSpan isSelected={isSelected}> {children}</StyledSpan>
+      <StyledSpan isSelected={isSelected}>{children}</StyledSpan>
     </FlexMenuItem>
   );
 }
--- a/context/app/static/js/shared-styles/dropdowns/index.js
+++ b/context/app/static/js/shared-styles/dropdowns/index.js
@@ -1,0 +1,31 @@
+import React from 'react';
+import styled from 'styled-components';
+import MenuItem from '@material-ui/core/MenuItem';
+import CheckRoundedIcon from '@material-ui/icons/CheckRounded';
+
+const iconSize = '1.25rem';
+
+const FlexMenuItem = styled(MenuItem)`
+  display: flex;
+`;
+
+const CheckIcon = styled(CheckRoundedIcon)`
+  font-size: ${iconSize};
+  margin-right: ${(props) => props.theme.spacing(1)}px;
+`;
+
+const StyledSpan = styled.span`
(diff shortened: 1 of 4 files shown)


mccalluc (changes requested):
I would defer to to @tsliaw on this, and I may be thinking about it wrong, but I'm expecting it to work like a pulldown menu, where the items on the list are stable. It's different than a select, where, when you click on it, all the items, including the current one are included.

I might be thinking about it wrong.

john-conroy:
Updated to @tsliaw new design. If this is approved, I will file issues to update the dropdown for the visualization tile selector and also evaluate the component's accessibility.

Screen Shot 2020-10-26 at 3 35 37 PM

Do we want to change the type label for the `display_doi` in the table and tile sort components to 'ID'? I think I would prefer that to switching it for the tiles only. So in this case instead of 'Dataset' it would say 'ID' or maybe 'HuBMAP ID'.

mccalluc:
@john-conroy :
> Do we want to change the type label for the display_doi in the table and tile sort components to 'ID'? I think I would prefer that to switching it for the tiles only. So in this case instead of 'Dataset' it would say 'ID' or maybe 'HuBMAP ID'.

Yeah, that sounds like a good approach to me. Doesn't need to block this, though.

context/app/static/js/shared-styles/dropdowns/index.js
```diff
@@ -0,0 +1,31 @@
...
+const StyledSpan = styled.span`
+  ${(props) => !props.isSelected && `margin-left: calc(${iconSize} + ${props.theme.spacing(1)}px)`};
+`;
+
+function DropdownSelectItem(props) {
+  const { children, isSelected, className, ...rest } = props;
+  return (
+    <FlexMenuItem className={className} {...rest}>
+      {isSelected && <CheckIcon color="primary" />}
+      <StyledSpan isSelected={isSelected}> {children}</StyledSpan>
```
mccalluc, line 26:
```suggestion
      <StyledSpan isSelected={isSelected}>{children}</StyledSpan>
```
?

context/app/static/js/shared-styles/dropdowns/index.js
```diff
@@ -0,0 +1,31 @@
...
+  display: flex;
+`;
+
+const CheckIcon = styled(CheckRoundedIcon)`
+  font-size: ${iconSize};
+  margin-right: ${(props) => props.theme.spacing(1)}px;
+`;
+
+const StyledSpan = styled.span`
+  ${(props) => !props.isSelected && `margin-left: calc(${iconSize} + ${props.theme.spacing(1)}px)`};
```
mccalluc, line 18: This is getting a little hairy. Any chance there's a class or something you could use to move the logic into CSS?

mccalluc (approved):
Thanks! One line in the style is getting complicated, but I don't know if there's an obvious simplification.

context/app/static/js/shared-styles/dropdowns/index.js
john-conroy: Hairy as in it's difficult to read/interpret?

tsliaw (approved):

PR #1326, opened 2020-10-27
Facet schema violation errors

mccalluc:
Under /dev-search, we can now see where validation errors are coming from:
Screen Shot 2020-10-27 at 11 05 47 AM

I recognize that we're not on the same page about some aspects of the architecture, but I think we do agree that there should be clean, usable, reliable APIs, and one part of that is documenting the API responses. One step on that path is moving the current entity-api responses closer to the schema... or the schema closer to the api. These facets show what part of the documents are in violation, and what part of the schema they are violating. I hope this will be useful.

### DIFF
--- a/CHANGELOG-dev-search-validation-errors.md
+++ b/CHANGELOG-dev-search-validation-errors.md
@@ -1,0 +1,2 @@
+- Facet the validation errors, to understand what part of the schema is being violated,
+  and what part of the documents are in violation.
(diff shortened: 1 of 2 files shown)


mccalluc:
@john-conroy : This is trivial: Initially put Bill as a reviewer, just as a heads-up, but he doesn't feel able to approve. Thanks.

john-conroy (approved):

PR #1329, opened 2020-10-27
John conroy/dropdown accessibility

john-conroy:
The dropdown now autofocuses on the selected item when open and can be traversed by keydown/up. Aria attributes taken from w3.


john-conroy:
Closing, didn't make it more clear to review separately and code is included in #1330.

PR #1330, opened 2020-10-28
John conroy/shared dropdown listbox component

john-conroy:
@mccalluc As we discussed, let me know if this seems on track. Passing in the child components and their props this way isn't the cleanest solution but we have challenges with the portal and needing to make minor adjustments to styles. 

@tsliaw do you want to take a look at both the sort dropdown in search's tile view and the visualization dropdown to make sure they work as you expect.

9db61adfc017670a196ea9b3ca1852a0 has a visualization with the dropdown.

### DIFF
--- a/context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
+++ b/context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
@@ -8,23 +8,23 @@ import { getSortPairs } from '../utils';
 
 const searchViewStoreSelector = (state) => state.searchView;
 
-function getSelectedItemLabel(items, selectedItems) {
+function getSelectedItemIndex(items, selectedItems) {
   if (selectedItems.length > 1) {
     console.warn('Expected only a single sort, not:', selectedItems);
   }
   const selectedItem = selectedItems.length ? selectedItems[0] : undefined;
-  const match = items.filter((item) => item.key === selectedItem);
-  return match.length ? match[0].label : '';
+  return items.findIndex((item) => [item[0].key, item[1].key].includes(selectedItem)) || 0;
 }
 
 function TilesSortDropdown(props) {
   const { items, toggleItem, selectedItems } = props;
-  const selectedItemLabel = getSelectedItemLabel(items, selectedItems);
+  const pairs = getSortPairs(items);
+  const selectedItemIndex = getSelectedItemIndex(pairs, selectedItems);
 
   const searchView = useSearchViewStore(searchViewStoreSelector);
 
-  const pairs = getSortPairs(items);
-  function selectSortItem(pair) {
+  function selectSortItem(itemAndIndex) {
+    const pair = itemAndIndex.option;
     // Sort everything in ascending order except for last modified
     const item = pair[0].field === 'mapped_last_modified_timestamp.keyword' ? pair[0] : pair[1];
     toggleItem(item.key);
@@ -34,7 +34,7 @@ function TilesSortDropdown(props) {
     <DropdownListbox
       buttonComponent={StyledButton}
       optionComponent={StyledDropdownListboxOption}
-      selectedItemLabel={selectedItemLabel}
+      selectedOptionIndex={selectedItemIndex}
       buttonProps={{
         $searchView: searchView,
       }}
--- a/context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
+++ b/context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
@@ -48,18 +48,22 @@ function TilesSortDropdown(props) {
         variant="contained"
         color="primary"
         $searchView={searchView}
+        aria-haspopup="true"
       >
         {selectedItemLabel} {isOpen ? <ArrowDropUpIcon /> : <ArrowDropDownIcon />}
       </SelectionButton>
       <Popper open={isOpen} anchorEl={anchorRef.current} placement="bottom-start" style={{ zIndex: 50 }}>
         <Paper>
           <ClickAwayListener onClickAway={() => setIsOpen(false)}>
-            <MenuList id="preview-options">
+            <MenuList id="preview-options" role="listbox">
               {pairs.map((pair) => (
                 <StyledDropdownSelectItem
                   onClick={() => selectSortItem(pair)}
                   key={pair[0].field}
                   isSelected={pair[0].label === selectedItemLabel}
+                  autoFocus={pair[0].label === selectedItemLabel}
+                  selected={pair[0].label === selectedItemLabel}
+                  aria-selected={pair[0].label === selectedItemLabel}
                 >
                   {pair[0].label}
                 </StyledDropdownSelectItem>
--- a/context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
+++ b/context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
@@ -41,6 +41,7 @@ function TilesSortDropdown(props) {
       options={pairs}
       selectOnClick={selectSortItem}
       getOptionLabel={(pair) => pair[0].label}
+      id="search-tiles-sort"
     />
   );
 }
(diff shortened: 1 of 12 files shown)


context/app/static/js/shared-styles/dropdowns/DropdownListbox/DropdownListbox.jsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+DropdownListbox.propTypes = {
+  selectedOptionIndex: PropTypes.number.isRequired,
+  buttonComponent: PropTypes.elementType.isRequired,
+  optionComponent: PropTypes.elementType.isRequired,
+  /* eslint-disable react/forbid-prop-types */
+  buttonProps: PropTypes.object,
+  options: PropTypes.array.isRequired,
+  /* eslint-enable react/forbid-prop-types */
```
mccalluc, line 73: How do you feel about these eslint rules overall? Maybe it's fine: it still catches someone passing entirely the wrong thing... or is it good to have something nudge us to complete it more fully?

context/app/static/js/shared-styles/dropdowns/DropdownListbox/style.js
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components';
+import Popper from '@material-ui/core/Popper';
+import Paper from '@material-ui/core/Paper';
+
+const StyledPopper = styled(Popper)`
+  z-index: 50;
```
mccalluc, line 6: Do you think it would be good to have a central definition of z-index values? Or do you know of other good practices? I'm imaging a file with a list of integer consts, so as we're adding a new rule, it's clear what it's above or below.

This might be overkill.

context/app/static/js/shared-styles/dropdowns/DropdownListboxOption/style.js
```diff
@@ -0,0 +1,21 @@
...
+  display: flex;
+`;
+
+const CheckIcon = styled(CheckRoundedIcon)`
+  font-size: ${iconSize};
+  margin-right: ${(props) => props.theme.spacing(1)}px;
+`;
+
+const StyledSpan = styled.span`
+  ${(props) => !props.isSelected && `padding-left: calc(${iconSize} + ${props.theme.spacing(1)}px)`};
```
mccalluc, line 17: Sorry, I feel like I commented on something like this before: Do you think it would be possible to pull this out, perhaps add a class so more of the logic can be handled by CSS? It's just a pretty complicated line. No worries if nothing better strikes you.

mccalluc (changes requested):
I think the big picture is fine, but some of the details should be clarified, particularly `getSelectedItemIndex`. (That said, I know that I tend to review more on a line by line basis, and I might not be thinking hard enough about the overall architecture... if you have concerns about overall direction, I think it'd be fine to take some time to think it over again. Let me know if I can be helpful.)

context/app/static/js/shared-styles/dropdowns/DropdownListboxOption/style.js
john-conroy: Using a class does nothing but change the selector, `padding-left: calc(${iconSize} + ${props.theme.spacing(1)}px)` won't change, so does using `.selected-class { ... }` instead of `${(props) => !props.isSelected &&` make it much more clear?

context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
```diff
@@ -1,78 +1,48 @@
...
+import DropdownListbox from 'js/shared-styles/dropdowns/DropdownListbox';
 import useSearchViewStore from 'js/stores/useSearchViewStore';
-import { SelectionButton, StyledDropdownSelectItem } from './style';
+import { StyledButton, StyledDropdownListboxOption } from './style';
 import { getSortPairs } from '../utils';
 
 const searchViewStoreSelector = (state) => state.searchView;
 
-function getSelectedItemLabel(items, selectedItems) {
+function getSelectedItemIndex(items, selectedItems) {
```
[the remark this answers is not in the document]
john-conroy: So is the question what does the function do? If that's the case, I can add a comment. To clarify here, Searchkit provides lists of items and selectedItems as props, `getSelectedItemIndex` returns the index of the selected item in the list of items. Or is the question something else?

context/app/static/js/shared-styles/dropdowns/DropdownListbox/DropdownListbox.jsx
john-conroy: I think it does make sense to be more specific where possible and eslint nudges us in that direction, but in situations like this where it can be an array of many different types it is simplest to just ignore it. I'm open to doing something different if you have a stronger preference.

context/app/static/js/shared-styles/dropdowns/DropdownListbox/style.js
john-conroy: I think that makes sense either putting the z-index values in the theme or a separate file which can be imported in the style.js files.

context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
```diff
@@ -1,78 +1,48 @@
...
   if (selectedItems.length > 1) {
     console.warn('Expected only a single sort, not:', selectedItems);
   }
   const selectedItem = selectedItems.length ? selectedItems[0] : undefined;
-  const match = items.filter((item) => item.key === selectedItem);
-  return match.length ? match[0].label : '';
+  return items.findIndex((item) => [item[0].key, item[1].key].includes(selectedItem)) || 0;
```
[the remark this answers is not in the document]
john-conroy: There shouldn't ever be a case where the selected item isn't found. How would you suggest we handle possible errors? Add a console.warn?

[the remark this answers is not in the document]
mccalluc: I guess I'm stuck on the `|| 0`... What is it there for? Reading [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex), `findIndex` always returns a number, so it would only be false-y if it were already returning 0?

context/app/static/js/shared-styles/dropdowns/DropdownListbox/DropdownListbox.jsx
mccalluc: Fine as-is, just wanted to check. Thanks!

context/app/static/js/shared-styles/dropdowns/DropdownListboxOption/style.js
mccalluc: A little more clear? I guess I'm just worried by they wrapped contexts here: The "padding-left" inside a boolean short circuit inside a function inside an interpolated string... but I might be coming with this with a set of experiences where more logic was put into the CSS, because we weren't building it in the code.

If you think this is readable for a contemporary audience, I won't argue it further.

context/app/static/js/components/Search/TilesSortDropdown/TilesSortDropdown.jsx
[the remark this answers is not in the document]
mccalluc: Renaming the parameter may be enough: On the first review, it was hard to understand what was going on, because it wasn't clear that the inputs were different.

context/app/static/js/shared-styles/dropdowns/DropdownListbox/style.js
mccalluc: Yeah: Something like that. Doesn't need to be this PR.

john-conroy: Filed #1340

context/app/static/js/shared-styles/dropdowns/DropdownListboxOption/style.js
john-conroy: Props are primarily used for styled components, so i'd prefer to keep it this way. I do think it's a little cluttered, but I think this is the CSS in JS approach.

mccalluc (approved):

tsliaw (approved):

PR #1337, opened 2020-10-30
(apologetic) version bump

mccalluc:


john-conroy (approved):

john-conroy:
Is there a way for this to return to being less manual?

mccalluc:
> Is there a way for this to return to being less manual?

@john-conroy : This is just the minor version number, with each sprint. Something like
```
$ expr $(date +%s) / 86400 / 14 - 1320
6
```
could give us the number for the current sprint going forward (assuming 2 week sprints, and next starts every other thursday.) I'll make a PR.

PR #1338, opened 2020-10-30
Mccalluc/start versioning

mccalluc:
@john-conroy : Still in flux, so I'll make this a draft, but you could start on details page work. Attributes may change slightly, so if you'd like to hold off, that's fine, too.

(Looking at this, I wonder if using `addDefaultQuery` would be / would have been a better alternative to the hidden facets?)


mccalluc:
IEC will not be able to deliver versioning any time soon. Closing.

PR #1341, opened 2020-11-02
Mccalluc/npm versioning

mccalluc:
The real test will be seeing if the script works from day to day, but let me know if you see problems.


mccalluc:
@john-conroy : I can reproduce the travis error locally:
```
ERROR in ./app/static/js/components/ServiceStatusTable/ServiceStatusTable.jsx
Module not found: Error: Can't resolve 'context/package' in '/work/app/static/js/components/ServiceStatusTable'
```
I had introduced these changes in webpack.common.js:
```
  resolve: {
    extensions: [..., '.json'],
    alias: {
      ...
      context: resolve(__dirname, '../../context'),
    },
  },
```
It works on dev, but not for the prod build. Does anything strike you?

john-conroy:
Could be an issue with how we're copying things over in docker?

mccalluc:
@john-conroy : D'oh: Of course the `context` directory itself isn't going to work in Docker... we just copy the stuff inside it. Thanks!

john-conroy (approved):
Alternatively, you could use webpack's [DefinePlugin](https://webpack.js.org/plugins/define-plugin/) to make `VERSION` available throughout. I don't have a strong preference, but it's probably better to keep what is here and not pollute global namespace?

mccalluc:
> Alternatively, you could use webpack's [DefinePlugin](https://webpack.js.org/plugins/define-plugin/) to make `VERSION` available throughout. I don't have a strong preference, but it's probably better to keep what is here and not pollute global namespace?

I think I'll keep the status quo, but thanks for pointing that out... I don't have the webpack tools at my fingertips like I should.

PR #1344, opened 2020-11-03
Update redeploy

mccalluc:


yuanzhou:
Assumptions for this script change to work:
1. we only have `portal-ui-prod` container running along with the regular `portal-ui` container on STAGE.
2. there are two files needed under `/home/centos/hubmap/portal-ui-prod` : `compose/hubmap.yaml` and `context/instance/app.conf`, so the two containers would have the same file structure needed for docker-compose.

yuanzhou (approved):

yuanzhou (changes requested):
"stage" is supposed to be there too? https://github.com/hubmapconsortium/portal-ui/blob/mccalluc/update-redeploy/redeploy.sh#L6

yuanzhou:
@mccalluc I created another PR #1345 with a better approach to handle this `portal-ui-prod` redeployment on STAGE. We can close this PR.

PR #1345, opened 2020-11-04
Portal prod.stage zhou

yuanzhou:
This is to replace PR #1344 (causes strange 502 error, probably due to orphan container) by handling the `portal-ui-prod` container in the same project.

- changes only apply to STAGE redeployment
- there are two docker compose yaml files now: `hubmap.yml` (only one `portal-ui` container for DEV/TEST/PROD) and `hubmap.stage.yml` (one `portal-ui` and one `portal-ui-prod` containers for STAGE only)
- the `portal-ui-prod` container will require `context/instance/app.prod.conf` to make calls against the PROD API services

@mccalluc you may need to tweak the `test.sh`. Feel free to tweak the `redeploy.sh` too.

I've tested the redeploy script with a modified script without the ssh section on STAGE and it works well. Please give it a try by running this script from your local.

### DIFF
--- a/compose/hubmap.stage.yml
+++ b/compose/hubmap.stage.yml
@@ -1,0 +1,26 @@
+version: "3.7"
+
+services:
+
+  portal-ui-prod:
+    hostname: portal-ui-prod
+    container_name: portal-ui-prod
+    image: hubmap/portal-ui:latest
+
+    # Mount the app config in container to keep it outside the image.
+    volumes:
+      - "../context/instance/portal-prod.stage-app.conf:/app/instance/app.conf"
+
+    # Avoid accidentally creating zombie processes
+    init: true
+
+    # Specifying a restart policy to avoid downtime
+    restart: always
+
+    networks:
+      - gateway_hubmap
+
+networks:
+  # This is the network created by gateway to enable communicaton between multiple docker-compose projects
+  gateway_hubmap:
+    external: true
--- a/compose/hubmap.stage.yml
+++ b/compose/hubmap.stage.yml
@@ -9,7 +9,7 @@ services:
 
     # Mount the app config in container to keep it outside the image.
     volumes:
-      - "../context/instance/portal-prod.stage-app.conf:/app/instance/app.conf"
+      - "../context/instance/app.stage.conf:/app/instance/app.conf"
 
     # Avoid accidentally creating zombie processes
     init: true
--- a/compose/hubmap.stage.yml
+++ b/compose/hubmap.stage.yml
@@ -9,7 +9,7 @@ services:
 
     # Mount the app config in container to keep it outside the image.
     volumes:
-      - "../context/instance/app.stage.conf:/app/instance/app.conf"
+      - "../context/instance/app.prod.conf:/app/instance/app.conf"
 
     # Avoid accidentally creating zombie processes
     init: true
--- a/compose/hubmap.stage.yml
+++ b/compose/hubmap.stage.yml
@@ -2,6 +2,24 @@ version: "3.7"
 
 services:
 
+  portal-ui:
+    hostname: portal-ui
+    container_name: portal-ui
+    image: hubmap/portal-ui:latest
+
+    # Mount the app config in container to keep it outside the image.
+    volumes:
+      - "../context/instance/app.conf:/app/instance/app.conf"
+
+    # Avoid accidentally creating zombie processes
+    init: true
+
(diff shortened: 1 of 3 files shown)


compose/hubmap.stage.yml
```diff
@@ -0,0 +1,44 @@
...
+      - "../context/instance/app.prod.conf:/app/instance/app.conf"
+
+    # Avoid accidentally creating zombie processes
+    init: true
+
+    # Specifying a restart policy to avoid downtime
+    restart: always
+
+    networks:
+      - gateway_hubmap
```
mccalluc, line 39:
```suggestion
  portal-ui: &default
    hostname: portal-ui
    container_name: portal-ui
    image: hubmap/portal-ui:latest

    # Mount the app config in container to keep it outside the image.
    volumes:
      - "../context/instance/app.conf:/app/instance/app.conf"

    # Avoid accidentally creating zombie processes
    init: true

    # Specifying a restart policy to avoid downtime
    restart: always

    networks:
      - gateway_hubmap

  portal-ui-prod:
    <<: *default
    hostname: portal-ui-prod
    container_name: portal-ui-prod

    # Mount the app config in container to keep it outside the image.
    volumes:
      - "../context/instance/app.prod.conf:/app/instance/app.conf"
```
Perhaps use yaml syntax to highlight just the parts that are different?

redeploy.sh
```diff
@@ -18,15 +18,29 @@ cd /home/centos/hubmap/portal-ui/compose
...
 
 echo 'stopping...'
-docker-compose -f hubmap.yml down
+
+# Also stop and remove the portal-ui-pord container on STAGE only
+if [ "$2" = "stage" ]; then
+    docker-compose -f hubmap.stage.yml down
+else
+    docker-compose -f hubmap.yml down
+fi
```
mccalluc, line 27: I could be missing something, but I think you want `hubmap.stage.yml down` in addition to `hubmap.yml down` if appropriate, and it seems that only one or the other is being run... 
but I'm wondering if it'd be cleaner just to try it in every environment...
```suggestion
docker-compose -f hubmap.yml down

docker-compose -f hubmap.stage.yml down || echo 'No stage config: will continue...'
```

redeploy.sh
```diff
@@ -18,15 +18,29 @@ cd /home/centos/hubmap/portal-ui/compose
...
 
 echo 'starting...'
-docker-compose -f hubmap.yml up -d
+
+# Also start the portal-ui-pord container on STAGE only
+if [ "$2" = "stage" ]; then
+    docker-compose -f hubmap.stage.yml up -d
+else
+    docker-compose -f hubmap.yml up -d
+fi
```
mccalluc, line 42: Ditto:
```suggestion
docker-compose -f hubmap.yml up -d

docker-compose -f hubmap.stage.yml up -d || || echo 'No stage config: will continue...'
```

redeploy.sh
```diff
@@ -35,3 +49,8 @@ docker exec -it portal-ui cat package.json
 EOF
 
 echo "Visit --> http://portal.$TARGET.hubmapconsortium.org/"
+
+if [ "$2" = "stage" ]; then
```
mccalluc, line 53:
```suggestion
if [ "$TARGET" = "stage" ]; then
```

mccalluc (changes requested):
I could be misreading it, but with your if-thens, it seems like only one of the containers will be restarted in the stage environment.

(The yaml tweak is your call: For really long configs it's handy, but here it might just be confusing.)

yuanzhou:
@mccalluc the `hubmap.stage.yml` is used for STAGE redeploy and it contains both the `portal-ui` and `portal-ui-prod` services. On DEV/TEST/PROD, we use `hubmap.yml` which only has the `portal-ui` service. 
````
# Also start the portal-ui-pord container on STAGE only
if [ "$2" = "stage" ]; then
    docker-compose -f hubmap.stage.yml up -d
else
    docker-compose -f hubmap.yml up -d
fi
````

**Comment by mccalluc** [comment]
@yuanzhou :

> @mccalluc the `hubmap.stage.yml` is used for STAGE redeploy and it contains both the `portal-ui` and `portal-ui-prod` services. On DEV/TEST/PROD, we use `hubmap.yml` which only has the `portal-ui` service.

Got it. Thanks!

I'll try to fix it from here. There are still problems:
```
<< 'EOF'
...
if [ "$2" = "stage" ];
```
The `$2` is effectively in single-quotes: it will not be interpolated.

PR #1349, opened 2020-11-05
John conroy/route based code splitting

### LINKED ISSUE #1347: Implement route based code splitting
To enhance performance, we should further divide the bundle. Routes seem to be a good place to do this.

john-conroy:
Fix #1347. We needed to further split up the bundles (and still need to do more splitting). I know we don't want too many loading spinners, but I can't perceive much of a difference between this and what is currently deployed.

@mccalluc I'd run this locally in addition to looking at the code changes.

@tsliaw For simplicity, I ended up not pushing the footer below the fold for any of the routes. It does smooth out the layout shift, but I'll take a look at that in a later PR.

### DIFF
--- a/CHANGELOG-route-based-code-splitting.md
+++ b/CHANGELOG-route-based-code-splitting.md
@@ -1,0 +1,1 @@
+- Split js bundle by route.
--- a/context/app/static/js/components/App.jsx
+++ b/context/app/static/js/components/App.jsx
@@ -16,9 +16,7 @@ function App(props) {
     // eslint-disable-next-line no-undef
     <Providers endpoints={endpoints} nexusToken={nexus_token}>
       <Header />
-      <div className="main-content">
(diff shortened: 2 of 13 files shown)


mccalluc (changes requested):
Code is fine, but I think we need to really minimize the flicker as things come and go on page load. I'd defer to Tiffany and Nils on the details, but here are my thoughts:
- Consider a css animation on the opacity for the spinner... perhaps transparent for 0.5s, and then take 1s to ramp to 100%... just so you don't see it flash for a moment.
- Find a way not to show the footer initially, only to have it pushed below the fold an instant later. Perhaps set the height of the placeholder to 100% of window height?
- Consider moving the `Suspense` deeper. Maybe keep page titles outside the code splitting?
- Probably more trouble than it's worth... but having larger elements like Paper stable on the screen, and then filling in the content, would minimize the sense of disruption... but it would be more work.

Thanks for tackling this!

john-conroy:
Merging following changes and conversation with @mccalluc

PR #1352, opened 2020-11-10
Mccalluc/fix versioning

mccalluc:


mccalluc:
Some tests fail on the changelog check... I think that's because my work last night didn't get merged, so the logic is off.

mccalluc:
Going ahead with merge to have a basis for testing version bumping. Not ideal.

PR #1353, opened 2020-11-10
Mccalluc/deactivate prov vis

### LINKED ISSUE #1333: Deactivate prov vis for bulkATACseq
Currently in QA on PROD:
```
HBM395.NRTL.659
HBM279.JRTJ.535
HBM583.CJZM.893
HBM742.TXRC.975
HBM555.CGHX.875
HBM256.TSRN.268
HBM643.KKCR.667
HBM488.SDDC.876
```

mccalluc:
Fix #1333. I couldn't find the IDs listed in the original issue, but there are other assays of the same type. I think this is correct?

### DIFF
--- a/CHANGELOG-bulk-atac-provvis.md
+++ b/CHANGELOG-bulk-atac-provvis.md
@@ -1,0 +1,1 @@
+- Hide Prov table for bulk ATACseq.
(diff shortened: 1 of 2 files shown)


john-conroy (changes requested):
All of the bulk ATACseq datasets I reviewed have linear provenance graphs. I'm not sure this is the correct data type.

mccalluc:
Good call. None of these IDs exist... maybe we weren't looking at prod? I can't remember any more details, and it doesn't feel like there's anything else to go on. Re-open if there's more information.

PR #1354, opened 2020-11-10
Tweak table height

### LINKED ISSUE #1263: Show that Creators list on Contact form is scrollable
There is some confusion with the Creators list on the Collection form.  When the list is longer than the display box it isn't obvious that there are more entries that can be scrolled to because there is no visible scroll bar and the last entry in the list perfectly aligns with the bottom of the box.

mccalluc:
Fix #1263. @tsliaw - We had tabled this, but I still find myself wondering where the rest of the metadata is. I believe that a half-obscured row is a clear way to suggest that it can be scrolled. Need Nils' approval before merging, but what's your first impression?
Screen Shot 2020-11-10 at 1 39 52 PM

### DIFF
--- a/CHANGELOG-table-height.md
+++ b/CHANGELOG-table-height.md
@@ -1,0 +1,1 @@
+- Change table-height to land in the middle of a row.
--- a/context/app/static/js/components/Detail/CollectionCreatorsTable/CollectionCreatorsTable.jsx
+++ b/context/app/static/js/components/Detail/CollectionCreatorsTable/CollectionCreatorsTable.jsx
@@ -26,7 +26,7 @@ function CollectionCreatorsTable(props) {
     <SectionContainer id="datasets-table">
       <SectionHeader>Creators</SectionHeader>
       <Paper>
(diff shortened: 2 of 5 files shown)


tsliaw:
I think that a half-obscured row should help indicate that the table is scrollable

mccalluc:
@tsliaw :
>  Is there way for when a user hover over the table to show the scrollbar?

That seems to be hard. There are techniques that seemed to work at some point, but when I run the [demo](http://jsfiddle.net/simurai/UsvLN/) now, they are ineffective. I would suggest not looking at CSS for a solution.

Do you feel comfortable approving this? Or can you put it on the UI agenda if you think Nils should be involved? Thanks!

mccalluc:
@tsliaw : Ping. Slight preference for merging and seeing how that feels, but if you think it's worth meeting time, that's ok too. Your call?

tsliaw (approved):

tsliaw:
Sorry just saw your comment, I think this is a good enough fix for now

PR #1356, opened 2020-11-12
John conroy/qa files

### LINKED ISSUE #1009: identify and display QC reports for raw and processed data
We propose to use the manifest files that are currently used to annotate the pipeline outputs to identify QC reports among the pipeline-generated files and do the same for uploaded data sets by introducing manifests for them as well. This will be also need to be addressed in the context of validating the file listing/directory structure for uploaded datasets (somewhat related to https://github.com/hubmapconsortium/ingest-validation-tools/issues/346).

Once we can identify QC reports among the data files associated with a dataset and these reports being available through the Assets API, we can list them in a separate section of the dataset detail pages. 

Solution for #1006. Supersedes #23.

john-conroy:
Fix #1009. Arrow icons to be added in #1375. Tests to be added in #1376. May be worthwhile to run this locally.

@tsliaw the 'QA' chip in each row is a little farther to the right than in your mockup, but this is how the table positions it by default and leaves a little more room for longer file/dir names.

Screen Shot 2020-11-20 at 3 40 01 PM

Screen Shot 2020-11-20 at 3 39 45 PM

### DIFF
--- a/context/app/static/js/components/files/FileBrowser/FileBrowser.jsx
+++ b/context/app/static/js/components/files/FileBrowser/FileBrowser.jsx
@@ -38,7 +38,7 @@ function FileBrowser(props) {
           clickable
           onClick={toggleDisplayOnlyQaQc}
           color={displayOnlyQaQc ? 'primary' : undefined}
-          icon={displayOnlyQaQc ? <DoneIcon /> : null}
+          icon={displayOnlyQaQc && <DoneIcon />}
           component="button"
           disabled={Object.keys(fileTrees.qa).length === 0}
         />
--- a/context/app/static/js/components/files/FileBrowser/FileBrowser.jsx
+++ b/context/app/static/js/components/files/FileBrowser/FileBrowser.jsx
@@ -4,14 +4,12 @@ import Chip from '@material-ui/core/Chip';
 import DoneIcon from '@material-ui/icons/Done';
 import Paper from '@material-ui/core/Paper';
 import Table from '@material-ui/core/Table';
-import TableContainer from '@material-ui/core/TableContainer';
 import TableRow from '@material-ui/core/TableRow';
-import TableHead from '@material-ui/core/TableHead';
 
 import useFilesStore from 'js/stores/useFilesStore';
 import { relativeFilePathsToTree } from './utils';
 import FileBrowserNode from '../FileBrowserNode';
-import { ChipWrapper, ScrollTableBody } from './style';
+import { ChipWrapper, ScrollTableBody, StyledTableContainer, HiddenTableHead } from './style';
 
 const filesStoreSelector = (state) => ({
   displayOnlyQaQc: state.displayOnlyQaQc,
@@ -31,7 +29,7 @@ function FileBrowser(props) {
   }, [files]);
 
   return (
-    <TableContainer component={Paper} style={{ maxHeight: '600px' }}>
+    <StyledTableContainer component={Paper}>
       <ChipWrapper>
         <Chip
           label="Show QA Files Only"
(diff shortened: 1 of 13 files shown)


tsliaw (approved):

context/app/static/js/components/files/FileBrowser/FileBrowser.jsx
```diff
@@ -1,34 +1,72 @@
...
 
 FileBrowser.propTypes = {
   files: PropTypes.arrayOf(
-    PropTypes.exact({
+    PropTypes.shape({
       rel_path: PropTypes.string,
       edam_term: PropTypes.string,
       description: PropTypes.string,
       size: PropTypes.number,
       type: PropTypes.string,
```
mccalluc, line 68:
```suggestion
      rel_path: PropTypes.string.isRequired,
      edam_term: PropTypes.string.isRequired,
      description: PropTypes.string.isRequired,
      size: PropTypes.number.isRequired,
      type: PropTypes.string.isRequired,
```
Would this work? no worries if not.

context/app/static/js/components/files/FileBrowser/style.js
```diff
@@ -1,9 +1,27 @@
...
-import Paper from '@material-ui/core/Paper';
+import TableBody from '@material-ui/core/TableBody';
+import TableContainer from '@material-ui/core/TableContainer';
+import TableHead from '@material-ui/core/TableHead';
 
-const ScrollPaper = styled(Paper)`
-  max-height: 600px;
+const ChipWrapper = styled.div`
+  border-bottom: 1px solid ${(props) => props.theme.palette.divider};
+  padding: 12px 15px 12px 15px;
```
mccalluc, line 8:
```suggestion
  padding: 12px 15px;
```
Since the the pair match?

context/app/static/js/components/files/FileBrowser/style.js
```diff
@@ -1,9 +1,27 @@
...
+`;
+
+const ScrollTableBody = styled(TableBody)`
   overflow-y: auto;
+  max-height: 600px;
+`;
+
+const StyledTableContainer = styled(TableContainer)`
+  max-height: 600px;
+`;
```
mccalluc, line 21: Slight preference to factor out the 600... but since the lines are right next to each other, probably gratuitous.

context/app/static/js/components/files/FileBrowserDirectory/FileBrowserDirectory.jsx
```diff
@@ -13,20 +16,35 @@ function FileBrowserDirectory(props) {
...
   };
 
+  const displayOnlyQaQc = useFilesStore(filesStoreSelector);
+
+  useEffect(() => {
+    if (displayOnlyQaQc === true) {
+      setIsExpanded(true);
+      return;
+    }
+    setIsExpanded(false);
```
mccalluc, line 26: or...
```suggestion
    setIsExpanded(displayOnlyQaQc);
```
or...
```suggestion
    setIsExpanded(Boolean(displayOnlyQaQc));
```

If there's some other case you're considering that would break these, please add a comment to explain.

mccalluc (commented):
Some of the boolean logic seems like it could be simpler... if there are cases you're catching with the code as it is, please spell those out more explicitly.

context/app/static/js/components/files/FileBrowser/FileBrowser.jsx
```diff
@@ -1,34 +1,72 @@
...
-      <FileBrowserNode fileSubTree={fileTree} depth={0} />
-    </ScrollPaper>
+    <StyledTableContainer component={Paper}>
+      <ChipWrapper>
+        <Chip
+          label="Show QA Files Only"
+          clickable
+          onClick={toggleDisplayOnlyQaQc}
+          color={displayOnlyQaQc ? 'primary' : undefined}
+          icon={displayOnlyQaQc ? <DoneIcon /> : null}
```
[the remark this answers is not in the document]
john-conroy: These are mui props which expect certain types, perhaps a short circuit could be used for the icon prop.

context/app/static/js/components/files/FileBrowser/FileBrowser.jsx
john-conroy: Added to applicable files, thanks.

context/app/static/js/components/files/FileBrowser/style.js
john-conroy: Turns out one of these was redundant and I removed it, thanks.

PR #1361, opened 2020-11-13
Remove one CSS rule to fix pagination position

### LINKED ISSUE #1350: Move search pagination to bottom of table, not bottom of facet list

mccalluc:
Fix #1350. Works on Chrome and FF.
Screen Shot 2020-11-13 at 11 48 04 AM


john-conroy (approved):

john-conroy:
We may not need the `flex-direction: column`, I have a bad habit of doing that when it's not needed.

PR #1362, opened 2020-11-13
Update CCF analytics key

mccalluc:
The long term fix is to have CCF-EUI delivered as a webcomponent, or something like that, so I don't think it's worth investing time on machinery that would keep this in sync. It shouldn't be changing going forward, in any case.


john-conroy:
I think before we merge this, I want to ensure we can create separate views for each application so we have the option of looking at each separately. I'll take a look next week, thanks!

john-conroy (approved):
We should be able to create separate views if we ever find the need.

PR #1368, opened 2020-11-17
John conroy/return collections list

### LINKED ISSUE #1262: Add contacts list to Collection form
Add a list of contacts to the Collection form.  These are returned in the new array attribute **contacts** from the existing enity-api calls (PR pending on dev/test/stage for this functionality).  The **contacts** attribute is formatted exactly the same as the exiting **creators** attribute.

john-conroy:
Fix #1262. Adds collections back to the header, adds contacts table and dates to collection page, and gets collection data from the search-api instead of the entity api.

Screen Shot 2020-11-18 at 3 55 20 PM

I filed issues in search-api because we are still not getting mapped values from es or the organ type for each dataset included in a collection. I also filed #1369 to fix spacing between the description and dates on the collection page.


tsliaw:
Screen Shot 2020-11-19 at 11 58 55 AM
Can you leave a little more space between the description and the dates in the summary section? They look like they're running into each other. Also do you think we should hide the doi.org until we actually get doi's?

tsliaw (approved):

context/app/config.py
```diff
@@ -1,5 +1,6 @@
 types = {
     'donor': {'display_name': 'Donors', 'in_header': True},
     'sample': {'display_name': 'Samples', 'in_header': True},
-    'dataset': {'display_name': 'Datasets', 'in_header': True}
+    'dataset': {'display_name': 'Datasets', 'in_header': True},
+    'collection': {'display_name': 'Collections', 'in_header': True}
```
mccalluc, line 5: This file is vestigial. Filed #1374 to clean up.

context/app/routes_main.py
```diff
@@ -72,38 +72,6 @@ def ccf_eui():
...
-
-
-@blueprint.route('/browse/HBM<hbm_suffix>')
-def hbm_redirect(hbm_suffix):
-    client = _get_client()
-    entity = client.get_entity(hbm_id=f'HBM{hbm_suffix}')
-    return redirect(
-        url_for('routes.details', type=entity['entity_type'].lower(), uuid=entity['uuid']))
-
-
```
mccalluc, line 106: Keep this one... Not much used, but at one point it was asked for.

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
```diff
@@ -29,6 +29,9 @@ function HeaderContent({ anchorRef }) {
                 {`${type}s`}
               </HeaderButton>
             ))}
+            <HeaderButton component={Link} href="/collections">
+              Collections
```
mccalluc, line 33:
```suggestion
              {/* TODO: Move this into the loop above when search works. */}
              Collections
```

context/app/static/js/components/Header/Menu/Menu.jsx
```diff
@@ -29,6 +29,7 @@ function Menu(props) {
             {['Donor', 'Sample', 'Dataset'].map((type) => (
```
mccalluc, line 29:
```suggestion
            {/* NOTE: Changes here should be in sync with HeaderContent.jsx. */}
            {['Donor', 'Sample', 'Dataset'].map((type) => (
```

... and similarly in the other file. I had started to do this in my branch, but forgot that there were two places to change.

context/app/static/js/hooks/useCollectionsData.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { getAuthHeader } from 'js/helpers/functions';
+
+function useCollectionsData(elasticsearchEndpoint, nexusToken) {
```
mccalluc, line 4: Is there a reason we can't have this request from the server side? As with the other entity pages, I want to make sure we don't have soft 404s.

mccalluc (commented):
I think we want to have the fetch for the details page coming from the server side... but if there's a reason we can't do that yet, not a big deal. That's the only big thing.

context/app/routes_main.py
john-conroy: Whoops, not sure how that got in there, thanks!

context/app/static/js/hooks/useCollectionsData.js
john-conroy: This is for all of the collections data not a specific entity. I added it here because eventually we will eventually be moving towards adding collections as a search page and it's easier to put this in temporarily and remove it down the road then passing everything in through flask. Does that make sense @mccalluc?

john-conroy: A single collection entity is coming from the server side.

john-conroy:
@tsliaw that should be resolved in #1372.

context/app/config.py
john-conroy: Thanks!

context/app/static/js/hooks/useCollectionsData.js
mccalluc: That sounds good. Thanks for explaining!

PR #1371, opened 2020-11-19
Upgrade Vitessce to 1.1.0 (Shareable Links + Scale bar for Microscopy)

### LINKED ISSUE #1005: add scale bar to imaging data displays
Scale bars should be activated in Vitessce for all image data.

### LINKED ISSUE #1065: Save State in Vitessce
Just want to create something here to track https://github.com/hubmapconsortium/vitessce/issues/241

### LINKED ISSUE #1259: Implement Vitessce Shareable Link
See #1250 for design. You should currently have view access to the Figma, but let me know if you need edit access.

ilan-gold:
This fixes #1259, fixes #1005, and fixes #1065.

I think only one of your reviews should be enough, but just wanted to give a heads up so you could test it out.  I think the length of the URL should not be a problem, regardless of the browser.


keller-mark (changes requested):
Looks good. I think the final Cell Type Annotation preview view config is missing `'version': '0.1.0'` (HBM396.RPRR.624)

ilan-gold:
@keller-mark Good catch.  Just fixed it.  Also addressed Nils' comments

keller-mark (approved):

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -45,11 +52,33 @@ function Visualization(props) {
...
+    setVitessceConfig,
+    onCopyUrlWarning,
+    onCopyUrlSnackbarOpen,
+    setOnCopyUrlSnackbarOpen,
   } = useVisualizationStore(visualizationStoreSelector);
 
-  const [vitessceErrors, setVitessceErrors] = useState([]);
-  const [vitessceSelection, setVitessceSelection] = useState(0);
+  // Get the vitessce configuration from the url if available and set the initial selection if it is a multi-dataset.
+  const [initializedVitData, initialSelection] = useMemo(() => {
```
john-conroy, line 62: Possibly pull this out to a hooks or utils file?

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+    <>
+      <SecondaryBackgroundTooltip title="Share Visualization">
+        <WhiteBackgroundIconButton ref={anchorRef} onClick={toggle}>
+          <ShareIcon color="primary" />
+        </WhiteBackgroundIconButton>
+      </SecondaryBackgroundTooltip>
+      <Popper open={open} anchorEl={anchorRef.current} placement="bottom-start" style={{ zIndex: 50 }}>
+        <Paper style={{ maxHeight: 200, overflow: 'auto' }}>
+          <ClickAwayListener onClickAway={toggle}>
+            <MenuList id="preview-options">
```
john-conroy, line 76: This isn't keyboard accessible. I think there are a few props you can pass to the mui menu components to handle this.

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+  return (
+    <>
+      <SecondaryBackgroundTooltip title="Share Visualization">
+        <WhiteBackgroundIconButton ref={anchorRef} onClick={toggle}>
+          <ShareIcon color="primary" />
+        </WhiteBackgroundIconButton>
+      </SecondaryBackgroundTooltip>
+      <Popper open={open} anchorEl={anchorRef.current} placement="bottom-start" style={{ zIndex: 50 }}>
+        <Paper style={{ maxHeight: 200, overflow: 'auto' }}>
```
john-conroy, line 74: We should probably not use in-line styles, but not a blocking issue.

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+        setOnCopyUrlWarning(DEFAULT_LONG_URL_WARNING);
+      },
+    })}`;
+    setOnCopyUrlSnackbarOpen(true);
+    dummy.setAttribute('value', url);
+    dummy.select();
+    document.execCommand('copy');
+    document.body.removeChild(dummy);
+  };
+  const emailUrl = (conf) => {
```
john-conroy, line 52: Rename `setEmailUrl` or `encodeEmailUrl` or similar?

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+  setOnCopyUrlSnackbarOpen: state.setOnCopyUrlSnackbarOpen,
+});
+function VisualizationShareButton() {
+  const [open, toggle] = useReducer((v) => !v, false);
+  const anchorRef = useRef(null);
+  const { vitessceConfig, setOnCopyUrlWarning, setOnCopyUrlSnackbarOpen } = useVisualizationStore(
+    visualizationStoreSelector,
+  );
+
+  const copyToClipBoard = (conf) => {
```
john-conroy, line 37: Possibly pull these functions out to a different file?

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -45,11 +52,33 @@ function Visualization(props) {
...
+    } else {
+      initializedVitDataFromUrl = vitessceURLConf || vitData;
+    }
+    setVitessceConfig(initializedVitDataFromUrl);
+    return [initializedVitDataFromUrl, initialSelectionFromUrl];
+  }, [vitData, setVitessceConfig]);
 
+  const [vitessceErrors, setVitessceErrors] = useState([]);
+  const [vitessceSelection, setVitessceSelection] = useState(initialSelection);
+  const handleVitessceConfigDebounced = useCallback(debounce(setVitessceConfig, 250, { trailing: true }), [
```
john-conroy, line 79: Was use callback needed here? Is it because of the vitessce internals?

john-conroy (approved):

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
john-conroy: I know we kept in-line styles on similar popper components, this is more so a note to myself.

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
ilan-gold: I think if you're going with the "verb first" convention, that's what I did here which I can see might not be clear.  Maybe `createEmailWithUrl` since this function actually opens your email app.

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
ilan-gold: Yes, this is why I did so here.

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -45,11 +52,33 @@ function Visualization(props) {
...
+    onCopyUrlSnackbarOpen,
+    setOnCopyUrlSnackbarOpen,
   } = useVisualizationStore(visualizationStoreSelector);
 
-  const [vitessceErrors, setVitessceErrors] = useState([]);
-  const [vitessceSelection, setVitessceSelection] = useState(0);
+  // Get the vitessce configuration from the url if available and set the initial selection if it is a multi-dataset.
+  const [initializedVitData, initialSelection] = useMemo(() => {
+    const queryString = window.location.href.split('#')[1];
+    const vitessceURLConf = queryString?.length > 0 ? decodeURLParamsToConf(queryString) : null;
```
mccalluc, line 64:
```suggestion
    const fragment = window.location.hash.substr(1);
    const vitessceURLConf = fragment && decodeURLParamsToConf(fragment);
```

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -45,11 +52,33 @@ function Visualization(props) {
...
+    const initialSelectionFromUrl =
+      Array.isArray(vitData) && Math.max(0, vitData.map(({ name }) => name).indexOf(vitessceURLConf?.name));
+    let initializedVitDataFromUrl = vitData;
+    if (Array.isArray(vitData)) {
+      initializedVitDataFromUrl[initialSelectionFromUrl] = vitessceURLConf || vitData[initialSelectionFromUrl];
+    } else {
+      initializedVitDataFromUrl = vitessceURLConf || vitData;
+    }
```
mccalluc, line 72: This whole chunk can probably be simpler... hard to understand what it's doing right now.

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
mccalluc, line 79: add a comment? what goes wrong without the debounce?

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -112,6 +142,17 @@ function Visualization(props) {
...
+          <VitessceInfoSnackbar
+            anchorOrigin={{
+              vertical: 'top',
+              horizontal: 'center',
+            }}
+            open={onCopyUrlSnackbarOpen}
+            autoHideDuration={4000}
+            $isWarning={onCopyUrlWarning}
+            onClose={() => setOnCopyUrlSnackbarOpen(false)}
+            message={`Shareable URL has been copied to clipboard. ${onCopyUrlWarning}`}
```
mccalluc, line 154:
```suggestion
            message={`Shareable URL copied to clipboard. ${onCopyUrlWarning}`}
```

context/app/static/js/components/Detail/visualization/Visualization/style.js
```diff
@@ -45,10 +45,11 @@ const SelectionButton = styled(Button)`
   border-radius: 3px;
 `;
 
-const EscSnackbar = styled(Snackbar)`
+const VitessceInfoSnackbar = styled(Snackbar)`
   top: ${totalHeightOffset + 10}px;
   & > div {
-    background-color: dimgray;
+    background-color: ${(props) => (props.$isWarning ? props.theme.palette.warning.main : 'dimgray')};
```
mccalluc, line 51: Is there an existing palette color to use? if not, let's add this to the global palette... Might loop in @tsliaw .

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+import { SecondaryBackgroundTooltip } from 'js/shared-styles/tooltips';
+import useVisualizationStore from 'js/stores/useVisualizationStore';
+import { WhiteBackgroundIconButton } from 'js/shared-styles/buttons';
+
+import { StyledLinkIcon, StyledTypography, StyledEmailIcon } from './style';
+import 'vitessce/dist/es/production/static/css/index.css';
+
+const DEFAULT_LONG_URL_WARNING =
+  'Warning: this is a long URL which may be incompatible or load slowly with some browsers.';
```
mccalluc, line 21:
```suggestion
  'Warning: Long URLs may not work on some browsers.';
```

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+const DEFAULT_EMAIL_MESSAGE = 'Here is an interesting dataset I found in the HuBMAP Data Portal:';
+
+const visualizationStoreSelector = (state) => ({
+  vitessceConfig: state.vitessceConfig,
+  setOnCopyUrlWarning: state.setOnCopyUrlWarning,
+  setOnCopyUrlSnackbarOpen: state.setOnCopyUrlSnackbarOpen,
+});
+function VisualizationShareButton() {
```
mccalluc, line 30:
```suggestion

function VisualizationShareButton() {
```

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const copyToClipBoard = (conf) => {
+    const dummy = document.createElement('input');
+    document.body.appendChild(dummy);
+    const url = `${window.location.href.split('#')[0]}#${encodeConfInUrl({
+      conf,
+      onOverMaximumUrlLength: () => {
+        setOnCopyUrlWarning(DEFAULT_LONG_URL_WARNING);
+      },
+    })}`;
```
mccalluc, line 45:
```suggestion
    const encodedConf = encodeConfInUrl({
      conf,
      onOverMaximumUrlLength: () => {
        setOnCopyUrlWarning(DEFAULT_LONG_URL_WARNING);
      },
    });
    const url = `${window.location.href.split('#')[0]}#${encodedConf}`;
```

Too much nesting gets hard to read.

mccalluc: Actually, pull this out as a utility function, since it's copy and pasted below.

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+    let longUrlWarning = '';
+    const url = `${window.location.href.split('#')[0]}#${encodeConfInUrl({
+      conf,
+      onOverMaximumUrlLength: () => {
+        longUrlWarning = DEFAULT_LONG_URL_WARNING;
+      },
+    })}`;
+    // We need to encode the URL so its parameters do not conflict with mailto's.
+    const encodedUrl = encodeURIComponent(url);
+    const mailtoLink = `mailto:?body=${longUrlWarning}%0D%0A%0D%0A${DEFAULT_EMAIL_MESSAGE} ${encodedUrl}`;
```
mccalluc, line 62:
```suggestion
    const mailtoLink = `mailto:?body=${longUrlWarning}${encodeURI('\r\n\r\n')}${DEFAULT_EMAIL_MESSAGE} ${encodedUrl}`;
```

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
mccalluc, line 74: move to style.js?

mccalluc (changes requested):
Lots of little comments. Also, could you add some unit tests? The useMemo logic is kind of complicated... I think it can be simplified, but a test would help too.

Would it be feasible to add either react component tests, or something in Cypress?

john-conroy:
From standup: Does it make sense to only set the config when it's needed instead of using the on change?

We don't think it's a big issue, and the pathway to doing so might introduce more issues than it is worth.

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
john-conroy: From standup: Does it make sense to move the useCallback into vitessce? Possibly, but probably fine as is.

ilan-gold: Not needed (or rather, not performant).  Has been removed.

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
ilan-gold: Discussed offline - will be resolved in a separate PR.

john-conroy: Some of our other menus which follow this pattern aren't keyboard accessible either, so this isn't an issue specific to this PR. I'll just file an issue and do this later. Thanks!

context/app/static/js/components/Detail/visualization/VisualizationShareButton/VisualizationShareButton.jsx
ilan-gold: Will be resolved in future PR.

context/app/static/js/components/Detail/visualization/Visualization/hooks.js
```diff
@@ -0,0 +1,29 @@
...
+      // If these is a url conf and the we have a multidataset, use the url conf to find the initial selection of the multi-dataset.
+      const initialSelectionFromUrl =
+        isMultiDataset && Math.max(0, vitData.map(({ name }) => name).indexOf(vitessceURLConf?.name));
+      let initializedVitDataFromUrl = vitData;
+      // Set the initialized  view config to be either the url configuration or the default that came from the server.
+      if (isMultiDataset) {
+        initializedVitDataFromUrl[initialSelectionFromUrl] = vitessceURLConf || vitData[initialSelectionFromUrl];
+      } else {
+        initializedVitDataFromUrl = vitessceURLConf || vitData;
+      }
```
mccalluc, line 22: Would it make sense to separate the `isMultiDataset` case, rather than having repeated checks? Not sure.

context/app/static/js/components/Detail/visualization/VisualizationShareButton/constants.js
```diff
@@ -0,0 +1,5 @@
+const DEFAULT_LONG_URL_WARNING = 'Warning: Long URLs may not work on some browsers.';
+
+const DEFAULT_EMAIL_MESSAGE = 'Here is an interesting dataset I found in the HuBMAP Data Portal:';
+
+export { DEFAULT_LONG_URL_WARNING, DEFAULT_EMAIL_MESSAGE };
```
mccalluc, line 5: Not sure we gain that much by splitting this out.

mccalluc (commented):
Did a little bit more manual QA, and didn't notice any problems. Realize that I haven't used the multidataset functionality... that's in the Showcases?

I have a stylistic preference for boolean short circuits over `?:` ternaries, particularly when one branch is boolean-ish. Not a big deal, but it can make it a little bit harder to read... Like, if the "else" is `null`, someone reading the code from the top has to find the end of the "then" to understand what's going on.

I'm a little worried about the performance for people with slightly older browsers (FF, in particular)... particularly because it's a degradation, and not an outright error, it can be hard for the user to know what the problem is, and they'll either leave in frustration, or inn the best case, we get a really unhelpful bug report. Maybe something like the terms-of-use click-through would be appropriate? ... but maybe that could be a separate PR?

mccalluc (commented):
Sorry, meant the last review to be a "comment"... I think there should be a few more tweaks, and I think there should be some consideration of messaging for folks with older browsers, but I don't need to re-review before you merge. Thanks!

ilan-gold:
@mccalluc We should discuss the FF use case during the UI meeting today.  Otherwise I'll address these comments.  Additionally, the CODEX datasets are all multi-dataset.

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -75,73 +93,90 @@ function Visualization(props) {
...
+              <ErrorSnackbar
+                anchorOrigin={{
+                  vertical: 'bottom',
+                  horizontal: 'right',
+                }}
+                open
+                key={vitessceErrors[0]}
+              >
+                <Alert severity="error" onClose={() => removeError(vitessceErrors[0])}>
+                  {vitessceErrors[0]}
```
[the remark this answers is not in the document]
ilan-gold: This is a "false dif" so this is the same as before.  Not sure if @john-conroy or @keller-mark implmented this....

context/app/static/js/components/Detail/visualization/VisualizationShareButton/constants.js
ilan-gold: I'm in general a fan of "constants" files especially because the number is now going to go up again if we show a message for FF.

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -75,73 +93,90 @@ function Visualization(props) {
...
+                </Alert>
+              </ErrorSnackbar>
+            )}
+            <Vitessce
+              config={vitessceConfig[vitessceSelection] || vitessceConfig}
+              theme={vizTheme}
+              onConfigChange={handleVitessceConfigDebounced}
+              height={vizIsFullscreen ? null : vitessceFixedHeight}
```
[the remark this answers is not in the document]
ilan-gold: I think there is probably some sort of internal check for truthiness here so I am going to leave this.  I am in general a fan of ternaries because they are clear about what they do.  Especially in jsx, you can end up rendering things unintentionally with && or || if you are not careful.  This change would have broken full-screen mode.

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -75,73 +93,90 @@ function Visualization(props) {
...
-            <DropdownListbox
-              buttonComponent={SelectionButton}
-              optionComponent={DropdownListboxOption}
-              selectedOptionIndex={vitessceSelection}
-              options={vitData}
-              selectOnClick={setSelectionAndClearErrors}
-              getOptionLabel={(v) => v.name}
-              id="visualization-data"
+    vitessceConfig &&
+    (isMultiDataset ? Number.isInteger(vitessceSelection) : true) && (
```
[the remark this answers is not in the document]
ilan-gold: I'll add a comment - it's meant to prevent multi datasets from rendering before the selection has been initialized.

ilan-gold:
@mccalluc Could you look at the [warning for FF](https://github.com/hubmapconsortium/portal-ui/pull/1371/files#diff-e1a15accd7a4a60ccb99335eb4deda2d28b5c9ceceeeeda2512832abaa38c718R38-R40)?

context/app/static/js/components/Detail/visualization/Visualization/Visualization.jsx
```diff
@@ -18,21 +21,34 @@ import {
...
+import { useVitessceConfig } from './hooks';
 import 'vitessce/dist/es/production/static/css/index.css';
 
+function sniffBrowser() {
+  const { browser } = Bowser.parse(window.navigator.userAgent);
+  return browser.name;
+}
+
+const FIREFOX_WARNING =
+  'Visualizations may be slow on Firefox.  If there is latency, please check if you are on the latest version - upgrading Firefox may alleviate the problem.';
```
mccalluc, line 39:
```suggestion
  'On older versions of Firefox, zooming in on visualizations is very slow; Upgrading to the latest version will fix the problem.';
```
- "may be" -> "is"
- Be more more precise about what interactions are slow.
- A little shorter.

Thanks!

mccalluc (commented):
Thanks! This is how I might phrase it, but your call.

PR #1376, opened 2020-11-20
John conroy/qa files tests

john-conroy:
Adds tests for the updated file browser with qa files.

### DIFF
--- a/context/app/static/js/components/files/FileBrowser/FileBrowser.spec.js
+++ b/context/app/static/js/components/files/FileBrowser/FileBrowser.spec.js
@@ -130,7 +130,7 @@ test('Displays correct files before and after display only QA files chip is clic
 
   userEvent.click(screen.getByRole('button', { name: 'Show QA Files Only' }));
 
-  // returns to all dirs closes
+  // returns to all dirs closed
   expectArrayOfStringsToExist(textInDocumentBeforeQAFilesOnly);
   expectArrayOfStringsToNotExist(textNotInDocumentBeforeQAFilesOnly);
 });
--- a/context/app/static/js/components/files/FileBrowser/FileBrowser.spec.js
+++ b/context/app/static/js/components/files/FileBrowser/FileBrowser.spec.js
@@ -69,3 +69,65 @@ test('displays files and directories', () => {
   const textInDocumentAfterOpenDirectory = [...textInDocumentBeforeOpenDirectory, 'fake3.txt', 'path2'];
   textInDocumentAfterOpenDirectory.forEach((text) => expect(screen.getByText(text)).toBeInTheDocument());
 });
+
+test('Displays correct files before and after display only QA files chip is clicked', () => {
+  const sharedEntries = {
+    edam_term: 'faketerm',
+    description: 'fakedescription',
+    size: 1000,
+    type: 'faketype',
+  };
+
+  const testFiles = [
+    {
+      rel_path: 'path1/path2/fake1.txt',
+      ...sharedEntries,
+      is_qa_qc: true,
+    },
+    {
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/files/FileBrowser/FileBrowser.spec.js
```diff
@@ -59,13 +62,75 @@ test('displays files and directories', () => {
...
+  expectArrayOfStringsToExist(textInDocumentAfterQAFilesOnly);
+
+  const textNotInDocumentAfterQAFilesOnly = ['path3', 'fake4.txt', 'fake2.txt', 'fake5.txt'];
+  expectArrayOfStringsToNotExist(textNotInDocumentAfterQAFilesOnly);
+
+  userEvent.click(screen.getByRole('button', { name: 'Show QA Files Only' }));
+
+  // returns to all dirs closes
+  expectArrayOfStringsToExist(textInDocumentBeforeQAFilesOnly);
+  expectArrayOfStringsToNotExist(textNotInDocumentBeforeQAFilesOnly);
```
mccalluc, line 135: Cool. I've sometimes put more semantics into the mocks... ie, `file-to-hide-unless-button-pressed.txt` or something like that. If assertions fail in the future, it can tell immediately what went wrong... but there's the risk of paying too much attention to the names, and not the criteria that actually matter.

Just a thought. Your call.

context/app/static/js/components/files/FileBrowser/FileBrowser.spec.js
```diff
@@ -59,13 +62,75 @@ test('displays files and directories', () => {
...
+  // after QA files chip clicked
+  const textInDocumentAfterQAFilesOnly = ['path1', 'path2', 'fake1.txt', 'fake3.txt'];
+  expectArrayOfStringsToExist(textInDocumentAfterQAFilesOnly);
+
+  const textNotInDocumentAfterQAFilesOnly = ['path3', 'fake4.txt', 'fake2.txt', 'fake5.txt'];
+  expectArrayOfStringsToNotExist(textNotInDocumentAfterQAFilesOnly);
+
+  userEvent.click(screen.getByRole('button', { name: 'Show QA Files Only' }));
+
+  // returns to all dirs closes
```
mccalluc, line 133:
```suggestion
  // returns to all dirs closed
```

context/app/static/js/components/files/FileBrowserFile/FileBrowserFile.spec.js
```diff
@@ -65,3 +65,44 @@ test('has correct left margin', () => {
...
+  const depth = 0;
+
+  render(
+    <FilesProviders>
+      <FileBrowserFile fileObj={fileObj} depth={depth} />
+    </FilesProviders>,
+  );
+
+  expect(screen.queryByText('QA')).toBeNull();
+});
```
mccalluc, line 108: Since it's easy, maybe add a test for the case where `is_qa_qc` is false... Not that we expect that to happen, but if it did, we know what behavior we want.

mccalluc (approved):
A couple possible tweaks, but also fine to merge as-is. Your call.

context/app/static/js/components/files/FileBrowser/FileBrowser.spec.js
john-conroy: I'll try to do so in the future, but I'll leave it as is here unless you think it makes sense to incorporate now.

context/app/static/js/components/files/FileBrowserFile/FileBrowserFile.spec.js
john-conroy: Done, thanks.

Issue #1348, opened 2020-11-05
Use long term caching for bundles


john-conroy:
We can set long term caching for our bundles since we're changing the file name/hash for each release.

https://developers.google.com/web/fundamentals/performance/webpack/use-long-term-caching

mccalluc:
(With that PR, caching 5 minutes... we'll deploy that, make sure it works, and then up it to 4 days, or so.)

mccalluc:
@john-conroy : In browser devtools, I've confirmed that js is coming from cache... but, somewhat surprisingly, I do not see `Cache-control` or `Expires`... this means, apparently, that we fall back to [browser heuristics](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#heuristic_freshness_checking), based on checking current time and the Last-Modified.

I don't see any sign that the settings we tried to [configure NGINX](https://github.com/hubmapconsortium/portal-ui/pull/1377/files) with are coming through. 

Perhaps it's being stripped by the gateway? I'll ask zhao.
My sense is that the caching we have now may be sufficient, though not ideal... but if the config we think should be doing the caching isn't being used, it should be removed.

Keeping open, for now.

mccalluc:
running docker.sh, I don't see cache headers, so this is on our side:
```
$ curl -I http://localhost:5001/static/public/17.a4e2be9ff53fd34c0827.js
HTTP/1.1 200 OK
Server: nginx/1.17.10
Date: Fri, 15 Jan 2021 22:21:33 GMT
Content-Type: application/javascript
Content-Length: 24162
Last-Modified: Fri, 15 Jan 2021 22:17:31 GMT
Connection: keep-alive
Vary: Accept-Encoding
ETag: "6002147b-5e62"
Accept-Ranges: bytes
```

PR #1377, opened 2020-11-23
Caching

mccalluc:
Towards #1348 ... For now, out of an abundance of caution, just caching for 5 minutes, but this can be increased, perhaps even before merging. (Since we will be pushing out new versions every couple days, I'd suggest we not cache for more than a week... anything more is just wasting disk space on the client.)

Checked in devtools, and works locally. In FF, in devtools, there's a checkbox that disables caching by default, but it worked when turned off. In Chrome, still shows as 200 response, and there's another column that shows its coming from cache.

### DIFF
--- a/CHANGELOG-caching.md
+++ b/CHANGELOG-caching.md
@@ -1,0 +1,2 @@
+- Tweak NGINX settings to allow caching on client.
+- Rearrange Dockerfile to improve layer caching.
--- a/context/Dockerfile
+++ b/context/Dockerfile
@@ -8,19 +8,28 @@ COPY package-lock.json .
 
 RUN npm ci
 
-# Most of our changes are in the JS, so don't worry about blowing away cache.
-COPY . .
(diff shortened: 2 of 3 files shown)


john-conroy (changes requested):
I think we should set explicit `no-cache` or `epoch` for text/html. Should we set default for all other types to `no-cache`? Most resources suggest setting `max-cache` for our built assets, but if we keep to our current release cadence two weeks makes sense.

mccalluc:
@john-conroy :

> I think we should set explicit no-cache or epoch for text/html

I could be confused, but the patterns are based on URL, not mime type. Maybe there's a way to do that, too, but it feels like a bit much.

> Most resources suggest setting max-cache for our built assets

Also confused about this... I'm not seeing that as a setting, or a header. Does the `expires 5m` do what we need? Let's talk tomorrow rather than starting a thread.

mccalluc:
talked with John. All good!

PR #1381, opened 2020-11-24
John conroy/docker memory warning

john-conroy:
Adds the message we discussed to the dockerfile.


context/Dockerfile
```diff
@@ -14,7 +14,7 @@ COPY app app
 COPY ingest-validation-tools ingest-validation-tools
 
 ENV NODE_OPTIONS="--max-old-space-size=4096"
-RUN npm run build
+RUN npm run build || die 'Make sure Docker has at least 4GB'
```
mccalluc, line 17: Sorry: `die` isn't a built-in... we define it in test.sh.... Instead, perhaps add error message in `docker.sh` here:
```
docker build --tag $IMAGE_NAME $CONTEXT
```
(`test.sh` calls `docker.sh`.)

mccalluc (changes requested):
Thanks... but needs to be in a different file, I think.

mccalluc:
@john-conroy : Ping! Should this be followed up on? Though it's the sort of "Well, this worked for me one time" message that might not that useful in the long run. Whatever you think best.

PR #1382, opened 2020-11-24
John conroy/compression

### LINKED ISSUE #711: Compress webpack bundle
We should be compressing our bundles into a gzip. The change will require some nginx [configuration](http://nginx.org/en/docs/http/ngx_http_gzip_static_module.html).

john-conroy:
Fix #711. Compresses webpack assets with gzip and updates the nginx config to serve gzipped files. We should ultimately use both brotli and gzip with preference for brotli where browser compatible, but that requires installing a nginx module. 

I used [this article](https://itnext.io/nginx-create-react-app-gzip-tripple-your-lighthouse-performance-score-in-5-minutes-627465c3f445) as a resource. It explains some of the settings. We could possibly trim the types or add a `gzip_min_length` setting. We could also probably remove the ie6 setting since our application wouldn't work there anyways.

You can test this by running the docker container and looking for `content-encoding: gzip` in the response headers for our webpack assets.

### DIFF
--- a/context/portal.nginx.conf
+++ b/context/portal.nginx.conf
@@ -1,13 +1,8 @@
-gzip on;
+gzip_static on;
 gzip_disable "msie6";
 
 gzip_vary on;
 gzip_proxied any;
-gzip_comp_level 6;
-gzip_buffers 16 8k;
-gzip_http_version 1.1;
-gzip_min_length 0;
-gzip_types text/plain application/javascript text/css application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript application/vnd.ms-fontobject application/x-font-ttf font/opentype;
 
 server {
     # "~": Case-sensitive regex matching
--- a/context/portal.nginx.conf
+++ b/context/portal.nginx.conf
@@ -1,3 +1,14 @@
+gzip on;
+gzip_disable "msie6";
+
+gzip_vary on;
+gzip_proxied any;
+gzip_comp_level 6;
+gzip_buffers 16 8k;
+gzip_http_version 1.1;
+gzip_min_length 0;
+gzip_types text/plain application/javascript text/css application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript application/vnd.ms-fontobject application/x-font-ttf font/opentype;
+
 server {
     # "~": Case-sensitive regex matching
     # http://nginx.org/en/docs/http/ngx_http_core_module.html#location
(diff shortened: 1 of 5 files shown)


context/portal.nginx.conf
```diff
@@ -1,3 +1,14 @@
+gzip on;
```
mccalluc, line 1:
```suggestion
# Compression mostly from https://itnext.io/627465c3f445:
gzip on;
```
Credit where credit is due.

context/portal.nginx.conf
```diff
@@ -1,3 +1,14 @@
...
+gzip_disable "msie6";
+
+gzip_vary on;
+gzip_proxied any;
+gzip_comp_level 6;
```
mccalluc, line 6: I haven't read deeply, but is it re-compressing on every request? There are also settings for server side caching, so it might be good to look at that, so the work only needs to be done once per file.

context/portal.nginx.conf
```diff
@@ -1,3 +1,14 @@
...
+gzip_buffers 16 8k;
```
mccalluc, line 7: Any reason not to go with the default?

context/portal.nginx.conf
```diff
@@ -1,3 +1,14 @@
...
+gzip_http_version 1.1;
```
mccalluc, line 8: Ditto: prefer not to over-specify things, unless there's a reason?

context/portal.nginx.conf
```diff
@@ -1,3 +1,14 @@
...
+gzip_min_length 0;
+gzip_types text/plain application/javascript text/css application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript application/vnd.ms-fontobject application/x-font-ttf font/opentype;
```
mccalluc, line 10: Maybe just start with JS and see how that works?

mccalluc (commented):
I haven't run it locally -- If you'd really like me too, I can. I'd have a preference for going with defaults, if possible, and starting with a smaller set of MIME types... but this is perfectly reasonable, too.

Add a note about where it came from, and then merge as-is, if that's your preference, or trim it down. Your call. Thanks!

context/portal.nginx.conf
john-conroy: I think I need to add the `gzip_static: on` and it will defer to using existing zipped assets where they exist.

john-conroy:
@mccalluc I made a change so we solely rely on webpack for compression. I feel like this is what you were moving towards?

PR #1383, opened 2020-11-24
John conroy/try GitHub actions

john-conroy:
A work in progress, but I think this would be the general idea. We may have to modify `test.sh` and `portal-ci.yml` to make this work. Right now I'm just making `test.sh` flexible enough to run both github actions and travis.

I was getting the error below and had to remove the `tput` code from `docker.sh`.
```
tput: No value for $TERM and no -T specified
Error: Process completed with exit code 2.
```

### DIFF
--- a/.github/workflows/portal-ci.yml
+++ b/.github/workflows/portal-ci.yml
@@ -22,9 +22,9 @@ jobs:
         key: ${{ runner.os }}-pip-${{ hashFiles('context/requirements.txt') }}
 
     - name: Set up Node 12.16.3
-        uses: actions/setup-node@v1
-        with:
-          node-version: '12.16.3'
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.16.3'
 
     - name: Set up git config to make .dev-start happy
       run: git config --global submodule.recurse true
--- a/.github/workflows/portal-ci.yml
+++ b/.github/workflows/portal-ci.yml
@@ -32,5 +32,5 @@ jobs:
     - name: Run test script
       run: ./test.sh
       env:
-        ACTIONS_BRANCH: ${{ github.ref }}
-        ACTIONS: true
+        GH_ACTIONS_BRANCH: ${{ github.ref }}
+        GH_ACTIONS: true
--- a/.github/workflows/portal-ci.yml
+++ b/.github/workflows/portal-ci.yml
@@ -1,0 +1,33 @@
+name: Portal CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.7'
+        architecture: 'x64'
+
+    - uses: actions/cache@v2
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('context/requirements.txt') }}
+
+    - name: Set up Node 12.16.3
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.16.3'
+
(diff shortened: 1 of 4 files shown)


.github/workflows/portal-ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.7'
+        architecture: 'x64'
+
+    - uses: actions/cache@v2
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('context/requirements.txt') }}
```
mccalluc, line 24: I have this memory of reading that caching pip in CI doesn't help much... There isn't that much repeated computation that's being saved... but maybe the github cache is significantly closer or faster than pypi?

.github/workflows/portal-ci.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: Set up Node 12.16.3
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.16.3'
+
+    - name: Set up git config to make .dev-start happy
+      run: git config --global submodule.recurse true
```
mccalluc, line 32: Huh... `submodules: 'recursive'` above isn't sufficient?

docker.sh
```diff
@@ -24,13 +24,8 @@ docker run -d \
   --mount type=bind,source="$(pwd)"/"$CONF_PATH",target=/app/instance/app.conf \
   $IMAGE_NAME
 
-green=`tput setaf 2`
-reset=`tput sgr0`
```
mccalluc, line 28: Consider 
```
green=`tput setaf 2 || echo ''`
reset=`tput sgr0 || echo ''`
```
... though it'd be nice to understand why it won't work, but don't let that block the PR.

mccalluc (commented):
Nifty. Before merging, would like to understand the recursive checkout, and keep the tput during local development.

Not the highest priority, but is there an easy way to do output folding on github?

.github/workflows/portal-ci.yml
john-conroy: Isn't it necessary since `dev-start` has the below?

```
[ "`git config --get submodule.recurse`" == 'true' ] \
  || die "To update submodules automatically: git config --global submodule.recurse true";
```

docker.sh
john-conroy: Thanks, I'll add it back in.

.github/workflows/portal-ci.yml
john-conroy: I just did this to mirror what we have in travis, I can remove it for simplicity?

mccalluc: Not a strong preference... if it's in, and it works, fine to leave it, I think.

.github/workflows/portal-ci.yml
mccalluc: Got it. I think you're right: What you're doing is the simplest way to do this.

.github/workflows/portal-ci.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('context/requirements.txt') }}
+
+    - name: Set up Node 12.16.3
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.16.3'
+
+    - name: Set up git config to make .dev-start happy
```
mccalluc, line 31:
```suggestion
    - name: Set up git config to make dev-start.sh happy
```

mccalluc (approved):
Looks good... One final oddity: Travis isn't running, but I don't think it's been disabled? I guess if Github CI is turned on, it doesn't notify Travis? In that case, yes, let's drop the travis.yml... We can always get it from the history, if we need it.

PR #1388, opened 2020-11-29
centralize z-indexes

### LINKED ISSUE #1340: Manage z-index values in a single location
From #1330 

> Do you think it would be good to have a central definition of z-index values? Or do you know of other good practices? I'm imaging a file with a list of integer consts, so as we're adding a new rule, it's clear what it's above or below.

> We can either put the z-index values in the theme or a separate file which can be imported in the style.js files.

mccalluc:
Fix #1340.
- Not sure about the best names... Generally chose something based on the directory it's in, but they could probably be improved.
- Have not confirmed that each of these works in the browser.

### DIFF
--- a/CHANGELOG-z-index.md
+++ b/CHANGELOG-z-index.md
@@ -1,0 +1,1 @@
+- Centralize our CSS z-indexes, so we have global view of what overlays what.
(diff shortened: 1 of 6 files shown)


john-conroy (approved):
Only looked at the code diff. Are these values things we would consider to be a part of a 'theme' or should they go in a separate file? Not a strong preference. Thanks.

mccalluc:
I guess not a "theme" per-se, but it seemed like an ok place to put css configuration stuff. May revisit.

PR #1389, opened 2020-11-29
Mccalluc/colors to theme

### LINKED ISSUE #1052: Move E0E0E0 to theme (and any other details hiding in others style.js files)
The search style.js references `rgb(224,224,224)` ... All colors should be moved to the theme, here and elsewhere.

mccalluc:
Fix #1052. @tsliaw, can you give your opinion on this? What we want to avoid is having inconsistencies arise across components, because different folks chose slightly different greys, for instance... but it does add complexity, and if you really want to make a change in only one place that becomes harder. More precise questions:

- Maybe references to `#fff` were best left as they were? It was clear, concise, simple.
- Do you think any of the current colors should be combined? Greys with different alpha values.
- Design consistency is the primary goal of this... Do think it will make your life easier?

### DIFF
--- a/CHANGELOG-centralize-colors.md
+++ b/CHANGELOG-centralize-colors.md
@@ -1,0 +1,1 @@
+- Centralize color definitions in the theme.
--- a/context/app/static/js/components/Collections/Panel/style.js
+++ b/context/app/static/js/components/Collections/Panel/style.js
@@ -13,7 +13,7 @@ const PanelWrapper = styled.div`
   flex-direction: column;
 
   &:hover {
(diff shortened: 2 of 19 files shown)


tsliaw:
I think replacing all the #fff colors with references to 'white' should be fine, especially if we move towards something like dark theme. As for the greys, I'd eventually want to make the greys to be more consistent across the board so I think this PR is a move towards the right direction, but there are going to be some different greys for different components (hover state, dividers).

tsliaw (approved):

PR #1397, opened 2020-11-30
John conroy/proxy http version for gzip

john-conroy:
Gzip should work once this change is made to the real nginx configuration for deployment. 

`gzip_http_version` sets the minimum HTTP version of a request required to compress a response to 1.1 by default. `proxy_http_version` defaults to 1.0 so by upgrading the version, the portal-ui will serve gzipped files when possible. Alternatively we could set `gzip_http_version` to 1.0.

### DIFF
--- a/compose/nginx.conf
+++ b/compose/nginx.conf
@@ -8,6 +8,7 @@ http {
             proxy_pass        http://portal-ui;
             proxy_set_header  Host $http_host;
             proxy_set_header  X-Forwarded-Proto $scheme;
+            proxy_http_version  1.1;
 
             # If proxy_set_header is missing,
             # the redirect url passed to Globus will use "portal-ui" instead.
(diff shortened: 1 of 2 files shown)


compose/nginx.conf
```diff
@@ -8,6 +8,7 @@ http {
             proxy_pass        http://portal-ui;
             proxy_set_header  Host $http_host;
             proxy_set_header  X-Forwarded-Proto $scheme;
+            proxy_http_version  1.1;
```
mccalluc, line 11: Can you move this below the comment for `proxy_set_header`, and add another comment? Something like:
```
gzip_http_version sets the minimum HTTP version of a request required to compress a response to 1.1 by default.
proxy_http_version needs to be >= that, but is 1.0 by default.
```

mccalluc (commented):
Add a comment in the config for our future-selves and merge. In your court to work this out with Zhou, I think, but let me know if I can be useful? Thanks for figuring this out!

john-conroy:
Waiting to merge until the parallel change is merged into the gateway repo.

PR #1400, opened 2020-12-01
John conroy/remove dev server from ci

### LINKED ISSUE #1387: 1000s of chokidar errors on Github CI?
Like this, though the details vary:
```
Error from chokidar (/home/runner/work/portal-ui/portal-ui/context/node_modules/@material-ui/icons):
Error: ENOSPC: System limit for number of file watchers reached, watch 
'/home/runner/work/portal-ui/portal-ui/context/node_modules/@material-ui/icons/package.json'
```

I haven't run the test suite locally in a long while, and maybe this was the case on Travis, and isn't related to Github CI?

john-conroy:
Fix #1387. Do we need to everything between lines 34-46 of `test.sh` or can we just run ./copy-app-conf.sh?

### DIFF
--- a/copy-app-conf.sh
+++ b/copy-app-conf.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
-CONTEXT=context
-
-APP_INSTANCE="$CONTEXT/instance"
+APP_INSTANCE="context/instance"
 APP_CONF="$APP_INSTANCE/app.conf"
 
 if [ ! -e $APP_CONF ]; then
--- a/copy-app-conf.sh
+++ b/copy-app-conf.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+CONTEXT=context
+
+APP_INSTANCE="$CONTEXT/instance"
+APP_CONF="$APP_INSTANCE/app.conf"
+
+if [ ! -e $APP_CONF ]; then
+  mkdir -p "$APP_INSTANCE"
+  cp example-app.conf "$APP_CONF"
+else
+  echo "Using existing $APP_CONF."
+fi
--- a/test.sh
+++ b/test.sh
@@ -62,6 +62,7 @@ end pytest
 
 start npm-test
 cd context
+npm ci
 npm run test
 cd -
 end npm-test
--- a/test.sh
+++ b/test.sh
@@ -31,6 +31,21 @@ fi
 end changelog
 
 
+start copy-app-conf
+if [ ! -z "$TRAVIS" ] || [ ! -z "$GH_ACTIONS" ]; then
+  echo 'Running on ci...'
+  ./copy-app-conf.sh || (
+    echo 'app.conf before:'
+    cat context/instance/app.conf
+    echo 'Rewrite conf...'
+    sed -i 's/TODO/FAKE/' context/instance/app.conf
+    echo 'app.conf after:'
+    cat context/instance/app.conf
+  )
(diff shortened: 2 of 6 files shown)


test.sh
```diff
@@ -43,12 +43,7 @@ if [ ! -z "$TRAVIS" ] || [ ! -z "$GH_ACTIONS" ]; then
     cat context/instance/app.conf
   )
 fi
-set -m; ./dev-start.sh & set +m  # Without job control, I had trouble killing parent and children.
-PID=$!
-server_up 5001  # Not really needed: Cypress will wait for response.
-end-to-end/test.sh
-kill -TERM -$PID  # Kill dev server processes
-end dev-start
+end copy-app-conf
```
john-conroy, line 46: Do we need to keep all of this or can we just run `./copy-app-conf.sh`

john-conroy:
Passing on travis, but failing on Github Actions. Going to set this to draft.

john-conroy:
Fixed and ready for review.

copy-app-conf.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+CONTEXT=context
+
+APP_INSTANCE="$CONTEXT/instance"
```
mccalluc, line 4:
```suggestion
APP_INSTANCE="context/instance"
```
In the original `$CONTEXT` is reused; Here it isn't.

test.sh
```diff
@@ -43,15 +43,11 @@ if [ ! -z "$TRAVIS" ] || [ ! -z "$GH_ACTIONS" ]; then
     cat context/instance/app.conf
   )
 fi
-set -m; ./dev-start.sh & set +m  # Without job control, I had trouble killing parent and children.
-PID=$!
-server_up 5001  # Not really needed: Cypress will wait for response.
-end-to-end/test.sh
```
mccalluc, line 49: Doesn't `end-to-end/test.sh` still need to be called somewhere?

test.sh
```diff
@@ -43,15 +43,11 @@ if [ ! -z "$TRAVIS" ] || [ ! -z "$GH_ACTIONS" ]; then
...
-kill -TERM -$PID  # Kill dev server processes
-end dev-start
+end copy-app-conf
 
 
 start flake8
+pip install -r context/requirements.txt > /dev/null
 # Unit tests require dev dependencies beyond requirements.txt.
 pip install -r context/requirements-dev.txt > /dev/null
```
mccalluc, line 52: If we're getting rid of the weird `dev-start.sh` on CI, installs should probably just move outside this script.

mccalluc (changes requested):
Don't we still need to call `end-to-end/test.sh`? Sorry if I'm missing something. Other comments minor.

test.sh
john-conroy: Isn't it being called on line 75?

test.sh
john-conroy: I can move the installs to the yml files for Github Actions and Travis. It would also then require you to run the installs before running this script locally. Should I wait until we decide between the two options?

john-conroy: From stand-up: we'll move the installs out of test.sh. Those running `test.sh` locally can be entrusted to make installs.

test.sh
mccalluc: Yep! Sorry to be confused.

mccalluc (approved):

PR #1419, opened 2020-12-14
Mount /var/log

mccalluc:
@yuanzhou : I think you'll need to create the directory for the mount. Maybe you have some convention about how these should be arranged?

### DIFF
--- a/compose/hubmap.yml
+++ b/compose/hubmap.yml
@@ -7,9 +7,11 @@ services:
     container_name: portal-ui
     image: hubmap/portal-ui:latest
 
-    # Mount the app config in container to keep it outside the image.
     volumes:
+      # Mount the app config in container to keep it outside the container.
       - "../context/instance/app.conf:/app/instance/app.conf"
+      # Mount the logs so they are visible outside the container.
+      - "/var/log/portal-ui/:/var/log/"
 
     # Avoid accidentally creating zombie processes
     init: true
--- a/compose/hubmap.yml
+++ b/compose/hubmap.yml
@@ -11,7 +11,7 @@ services:
       # Mount the app config in container to keep it outside the container.
       - "../context/instance/app.conf:/app/instance/app.conf"
       # Mount the logs so they are visible outside the container.
-      - "/var/log/portal-ui/:/var/log/"
+      - "../log/:/var/log/portal-ui/"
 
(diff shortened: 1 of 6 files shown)


compose/hubmap.yml
```diff
@@ -7,9 +7,11 @@ services:
     container_name: portal-ui
     image: hubmap/portal-ui:latest
 
-    # Mount the app config in container to keep it outside the image.
     volumes:
+      # Mount the app config in container to keep it outside the container.
       - "../context/instance/app.conf:/app/instance/app.conf"
+      # Mount the logs so they are visible outside the container.
+      - "/var/log/portal-ui/:/var/log/"
```
mccalluc, line 14:
```suggestion
      - "/home/centos/hubmap/portal-ui/log/:/var/log/"
```
@yuanzhou 
> if you just mount (using the relative path) the nginx logging (access log and error log) to a sub-directory under portal-ui project root directory, that should work well across all deployment environments.

Is this what you'd prefer? I think this is really deployment configuration, and you can take the lead.

yuanzhou:
@mccalluc I made some additional tweaks on top of your changes. But this won't work because by default, the nginx image used in portal-ui is configured to send nginx access log and error log to the Docker log collector. This is done by linking them to stdout and stderr. So just mounting won't work. Based on my tweaks: https://github.com/hubmapconsortium/portal-ui/pull/1419/commits/fcc77246a7bc4bc76e53fdcf559202d58713389e can you modify the nginx conf file used by portal-ui to save the access log and error log files to `/var/log/portal-ui/`? Rather than using the default: 
`access_log  /var/log/nginx/access.log` and `error_log  /var/log/nginx/error.log`.

mccalluc:
@yuanzhou : I am somewhat concerned that this is heading in the wrong direction. Writing to stdout is the standard way for Docker containers to log, and I'm sorry I didn't realize IEC wasn't working within the standard earlier. As it stands, I've removed the symlinks from the nginx logs... But Python logs are still going to stdout.

mccalluc:
This is the wrong direction. We shouldn't fight against the standard ways of working with logs in docker: Inside the containers, logs go to STDOUT; Outside, it's is accessible with `docker logs`... If you really need to know the file, `docker inspect --format='{{.LogPath}}' portal-ui`.

PR #1421, opened 2020-12-15
double uwsgi buffer

### LINKED ISSUE #1420: 502s
In the docker logs we see this set of three lines repeated, with small variations:

Error from uwsgi:
```
invalid request block size: 4143 (max 4096)...skip
```

Internal error from nginx:
```
recv() failed (104: Connection reset by peer) while reading response header from upstream, 
client: 172.18.0.6, server: , request: \"GET /favicon.ico HTTP/1.1\", 
upstream: \"uwsgi://unix:///tmp/uwsgi.sock:\", 
host: \"portal.hubmapconsortium.org\", 
referrer: \"https://portal.hubmapconsortium.org/search?entity_type[0]=Donor\"\n"
``` 

Error response from nginx:
```
"GET /favicon.ico HTTP/1.1" 502 560
```

mccalluc:
Fix #1420... hopefully. Could make the buffer even larger. Could try to understand why the requests are large, too.

### DIFF
--- a/CHANGELOG-uwsgi-buffer.md
+++ b/CHANGELOG-uwsgi-buffer.md
@@ -1,0 +1,1 @@
+- Double UWSGI buffer size.
--- a/context/uwsgi.ini
+++ b/context/uwsgi.ini
@@ -1,3 +1,6 @@
 [uwsgi]
 module = app.main
 callable = app
+
+# Twice default:
+buffer-size = 8192


john-conroy (approved):
So this sets the max size for the request headers, correct?

mccalluc:
@john-conroy : Not from the outside world, but on the internal requests to uwsgi... which is not http? So I'm not sure what's in those headers, or why it's more than 4K... but I also think the 4K default is pretty arbitrary. We ask them to look at the production logs after this is deployed to confirm.

Another possible follow-up is to figure out how to log those headers for debugging.

PR #1425, opened 2020-12-16
minor version math

mccalluc:
We need some kind of tweak between now and the next release... this might be making it more complicated than it needs to be.... could go back to doing it by hand, if this is getting to be too much.

### DIFF
--- a/push.sh
+++ b/push.sh
@@ -7,7 +7,18 @@ git diff --quiet || die 'Uncommitted changes: Stash or commit'
 git checkout master
 git pull
 
-EXPECTED_MINOR=v0.`expr $(date +%s) / 86400 / 14 - 1320`
+get_minor_version() {
+  REF_MINOR=$1
+  REF_DATE=$2
+  REF_EPOCH_DAY=`expr $(date -j -f "%d %b %Y" "$REF_DATE" +%s) / 86400`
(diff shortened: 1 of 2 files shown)


push.sh
```diff
@@ -7,7 +7,18 @@ git diff --quiet || die 'Uncommitted changes: Stash or commit'
...
 git pull
 
-EXPECTED_MINOR=v0.`expr $(date +%s) / 86400 / 14 - 1320`
+get_minor_version() {
+  REF_MINOR=$1
+  REF_DATE=$2
+  REF_EPOCH_DAY=`expr $(date -j -f "%d %b %Y" "$REF_DATE" +%s) / 86400`
+  NOW_EPOCH_DAY=`expr $(date +%s) / 86400`
+  DAYS_PAST_REF=`expr $NOW_EPOCH_DAY - $REF_EPOCH_DAY`
+  echo v0.`expr $REF_MINOR + $DAYS_PAST_REF / 14`
```
john-conroy, line 16: Probably missing something, but why days past ref / 14?

mccalluc: Divide by 14 because we're incrementing the minor version every two weeks.

john-conroy (approved):

PR #1427, opened 2020-12-16
Disable schema validation warnings

### LINKED ISSUE #1424: Get rid of schema validation
I strongly believe there should be a schema which describes our document strucuture, but until there is, this is just noise. Comment out, with a TODO and an issue to turn it back on.

mccalluc:
Fix #1424

### DIFF
--- a/context/app/static/js/index.jsx
+++ b/context/app/static/js/index.jsx
@@ -4,11 +4,12 @@ import ReactDOM from 'react-dom';
 import App from './components/App';
 import { setJsonLD } from './schema.org';
 
+/* TODO: Re-enable. https://github.com/hubmapconsortium/portal-ui/issues/1426 */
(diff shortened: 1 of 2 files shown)


john-conroy (approved):
Single line comment up top and block below, maybe?

context/app/static/js/index.jsx
```diff
@@ -4,11 +4,12 @@ import ReactDOM from 'react-dom';
 import App from './components/App';
 import { setJsonLD } from './schema.org';
 
+/* TODO: Re-enable. https://github.com/hubmapconsortium/portal-ui/issues/1426 */
```
mccalluc, line 7:
```suggestion
// TODO: Re-enable. https://github.com/hubmapconsortium/portal-ui/issues/1426
```

PR #1436, opened 2021-01-06
John conroy/increment changelog version

john-conroy:
Increments changelog to match dockerhub version.


mccalluc (commented):
Can you add a note like this to the changelog?
```
## v0.9.4

- Skipped because of deployment snafu.
```
Thanks!

PR #1447, opened 2021-01-12
Mccalluc/search tags

### LINKED ISSUE #1408: Add search tags back to search page
Design in #1133

mccalluc:
Fix #1408. Tiffany, design review; John, code review.

Draft: I'm not sure about the the best way to get `CancelIcon` instead of the "x":
- Maybe there's a way to pass in a component it should use instead of the default... but looking at the source code on their end, I don't see it.
- Could clone their code, and add the tweak we need. :(
- or change the design. :(
- Maybe, use CSS to try to create a black circle, and use CSS content to inject a multiply-x?
- or, get the SVG file for `CancelIcon`, add it to our assets, and reference it from CSS.

Last option seems like the best?

Screen Shot 2021-01-12 at 11 35 41 AM


tsliaw:
Other than the CancelIcon, everything else looks good. There does need to be a "Clear All Filters" button at the end of the search tags. Screenshot of design below
Screen Shot 2021-01-12 at 1 19 33 PM

mccalluc:
(Forgot about the clear-all button. Changing back to draft. Thanks!)

john-conroy:
You can create your own component. See [selected filters in searchkit docs](http://searchkit.github.io/searchkit/stable/components/navigation/selected-filters.html).

mccalluc:
One headache: Our reset shouldn't actually clear all facets... we want to keep the entity type. I'll try a simple fix that is just a link... Tweaking the behavior of the existing reset button seems hard. Something like:
```
class TweakedResetSearchAccessor extends ResetSearchAccessor {
  performReset() {
    this.performReset();
    // TODO: Add entity_type back to query.
  }
}

class TweakedResetFilters extends ResetFilters {
  constructor() {
    super();
    // TODO:
    // _this.props is undefined
    this.accessor = new TweakedResetSearchAccessor(this.props.options);
  }
}
```

mccalluc:
@john-conroy / @tsliaw : Thanks! Button looks better now. The clear-all may be tricky to do well. Could that be filed as a separate issue?

Screen Shot 2021-01-15 at 4 53 39 PM

context/app/static/js/components/Search/SearchBarLayout/SearchBarLayout.jsx
```diff
@@ -1,21 +1,37 @@
...
+
+function SelectedFilter(props) {
+  const { bemBlocks, labelKey, labelValue, removeFilter, filterId } = props;
+  return (
+    <div className={bemBlocks.option().mix(bemBlocks.container('item')).mix(`selected-filter--${filterId}`)}>
+      <div className={bemBlocks.option('name')}>
+        {labelKey}: {labelValue}
+      </div>
+      <StyledCancelIcon onClick={removeFilter} />
+    </div>
```
john-conroy, line 18: Seems a bit odd to extend the seachkit styles in Search.scss and then have to add in all of the bemBlocks code. It could be simpler to just write our own styled components/css and not rely on searchkit styles, especially since we'll need to migrate to the new searchkit or find an alternative.

context/app/static/js/components/Search/SearchBarLayout/style.js
```diff
@@ -11,4 +12,11 @@ const StyledDiv = styled.div`
   align-items: center;
 `;
 
-export { Flex, StyledDiv };
+const StyledCancelIcon = styled(CancelIcon)`
+  height: 0.8em;
```
john-conroy, line 16: We're using `rem` for icons elsewhere.

john-conroy (changes requested):

context/app/static/js/components/Search/SearchBarLayout/SearchBarLayout.jsx
```diff
@@ -1,21 +1,37 @@
...
 import PropTypes from 'prop-types';
-import { SearchBox, SortingSelector, ViewSwitcherToggle } from 'searchkit';
+import { SearchBox, SelectedFilters, SortingSelector, ViewSwitcherToggle } from 'searchkit';
 
 import SearchViewSwitch from '../SearchViewSwitch';
 import TilesSortDropdown from '../TilesSortDropdown';
-import { Flex, StyledDiv } from './style';
+import { Flex, StyledDiv, StyledCancelIcon } from './style';
+
+function SelectedFilter(props) {
```
john-conroy, line 10: Could we pull this out of SearchBarLayout?

context/app/static/js/components/Search/SearchBarLayout/SearchBarLayout.jsx
mccalluc: good call! fixing...

context/app/static/js/components/Search/SearchBarLayout/style.js
mccalluc: thanks. fixing...

context/app/static/js/components/Search/SearchBarLayout/SearchBarLayout.jsx
mccalluc: yep: good call.

context/app/static/js/components/Search/SelectedFilter/style.js
```diff
@@ -0,0 +1,27 @@
...
+`;
+
+const SelectedFilterDiv = styled.div`
+  background-color: white;
+  border: 1px solid ${(props) => props.theme.palette.hoverShadow.main};
+  border-radius: 8px;
+  margin: 16px 16px 0 0;
+  display: flex;
+  padding: 10px 0 10px 10px;
+  font-size: 14px;
```
john-conroy, line 18: Use `Typography` for the text inside the div to keep things consistent and use theme values by default?

context/app/static/js/components/Search/Search.scss
```diff
@@ -7,6 +7,15 @@ $secondary-gray: #636363;
     margin: 16px 0px 32px 0px;
 }
 
+.sk-selected-filters-option__name {
+    font-weight: 200 !important;
+    color: #000;
+}
```
john-conroy, line 13: Is this still needed?

context/app/static/js/components/Search/SelectedFilter/style.js
```diff
@@ -0,0 +1,27 @@
...
+  line-height: 20px;
+`;
+
+const SelectedFilterName = styled.div`
```
john-conroy, line 22: Can we also use a Typography component here? We would probably have to add something with `font-weight: 200` to the theme unless it will only be used here. Also worth noting that the default typography color isn't #000 elsewhere.

context/app/static/js/components/Search/SelectedFilter/style.js
```diff
@@ -0,0 +1,27 @@
+import styled from 'styled-components';
+import CancelIcon from '@material-ui/icons/CancelRounded';
+
+const StyledCancelIcon = styled(CancelIcon)`
+  height: 1.2rem;
```
john-conroy, line 5: I think icons are sized using `font-size` in MUI. If this works or has a separate purpose not a big issue.

john-conroy (changes requested):

tsliaw:
Icon for search tags look good now and feel free to do the "clear_all" button in a separate PR. However, is the font for the tags body2? Because inspecting the page, it is currently 14px, but body2 should 12.8px. 

Also do you think you can decrease the spacing between the next line of search tags slightly? Maybe about half of the current spacing. It's a bit too much spacing at the moment.

context/app/static/js/components/Search/SelectedFilter/style.js
mccalluc: no good reason. fixed.

mccalluc:
@tsliaw 
> Also do you think you can decrease the spacing between the next line of search tags slightly? Maybe about half of the current spacing. It's a bit too much spacing at the moment

Good call. Fixed:
Screen Shot 2021-01-22 at 2 18 23 PM

context/app/static/js/components/Search/Search.scss
mccalluc: fixed

mccalluc:
(Pinged tiffany about 200 vs 300. I think everything else is good.)

tsliaw (approved):

context/app/static/js/components/Search/Search.scss
```diff
@@ -7,6 +7,15 @@ $secondary-gray: #636363;
...
 
+.sk-selected-filters {
+    // Together with 8px margin on SelectedFilterDiv, gives 16px between elements
+    margin-top: 8px;
+}
+
+.sk-selected-filters-option__remove-action {
+    border: none;
+}
+
```
john-conroy, line 18: Are these styles for the element that contains the selected filters which is only accessible through extending the searchkit styles?

john-conroy (approved):

context/app/static/js/components/Search/Search.scss
mccalluc: Good catch: the `remove-action` isn't being used at all now, but `.sk-selected-filters` I think is still the best of doing that: We've overriden `SelectedFilter`, singular, but are still using the SK parent `SelectedFilters` plural.

PR #1448, opened 2021-01-12
branch.sh script

mccalluc:
Maybe this saves a little bit of data entry? Or maybe it's gratuitous bash/perl kludging? Is this useful?

One quirk: since it's using `whoami`, the branch names are different, unless your local name matches your github name.


john-conroy:
It does seem gratuitous, merge if you feel it is truly needed, but I don't think it is.

PR #1451, opened 2021-01-12
Link to collections DOI URL; not sure about design

### LINKED ISSUE #1449: doi link on collections
For example: https://portal.hubmapconsortium.org/browse/collection/6a6efd0c1a2681dc7d2faab8e4ab0bca

Just goes to doi.org: Should go to the url for this particular item.
Screen Shot 2021-01-12 at 1 42 38 PM

mccalluc:
Not sure about design here.

Screen Shot 2021-01-12 at 2 49 55 PM

Fix #1449

### DIFF
--- a/CHANGELOG-collections-doi.md
+++ b/CHANGELOG-collections-doi.md
@@ -1,0 +1,1 @@
+- Link to collections DOI.
--- a/context/app/static/js/pages/Collection/Collection.jsx
+++ b/context/app/static/js/pages/Collection/Collection.jsx
@@ -13,6 +13,7 @@ function Collection(props) {
     uuid,
     entity_type,
     display_doi,
+    doi_url,
     title,
     description,
     create_timestamp,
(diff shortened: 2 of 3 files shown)


tsliaw:
This looks good, can you match the url to match the design here of adding "doi" to the beginning and the "OpenInNewRoundedIcon" at the end. Feel free to merge after the change.

tsliaw (approved):

john-conroy (commented):
The icon is not vertically aligned with the text. There isn't a clean solution to the issue.

john-conroy:
Bumping the `font-size: 1.1rem;` and adding `vertical-align:text-bottom;` works but I wish we could find a more universal solution.

mccalluc:
@john-conroy : "universal solution" means an approach that would get these inline icons right on the first pass? With your tweaks it looks like this:
Screen Shot 2021-01-19 at 9 47 41 AM

A think a challenge we have is that the drawing doesn't use all of the SVG's space, so aligning on the baseline doesn't work: It comes out too high.

Is this acceptable for this particular case? Do you want to file an issue for the more general problem?

john-conroy (approved):

PR #1453, opened 2021-01-12
Log config during docker startup

### LINKED ISSUE #1430: Log non-sensitive configuration details on startup
Zhou thinks he saw our staging instance hitting production APIs... which would be strange... let's add a bit of logging at startup so we can confirm that instances are hitting what we think they should hit.

mccalluc:
Fix  #1430

Using docker.sh locally, and `docker logs`, confirmed that the config is in there.
- @john-conroy : code review
- @yuanzhou : Is this what you need? Is there anything else you'd like recorded?

I think the security on the logs is the same as the security on the config itself, but we could use `grep -v SECRET` instead of `cat`, if anyone would prefer?

### DIFF
--- a/CHANGELOG-docker-config-logging.md
+++ b/CHANGELOG-docker-config-logging.md
@@ -1,0 +1,1 @@
+- Log config during docker startup.
--- a/context/prestart.sh
+++ b/context/prestart.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Only designed to be called from the base Docker image on startup:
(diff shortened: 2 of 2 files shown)


john-conroy (approved):

yuanzhou (approved):

yuanzhou:
This is good! All I need to know is the portal.stage and portal-prod.stage are using the corresponding config files. And if we still encounter the behavior in #1430 while the config files are being used correctly, then it must be something either on the docker end or nginx related routing for the proxy. And it'll be on my court. Thanks!

PR #1455, opened 2021-01-12
emulate travis [skip ci]

mccalluc:
Suggestion taken from: https://github.community/t/github-actions-does-not-respect-skip-ci/17325/9.

... though, unlike Travis, we still get a green checkmark, even when it skips, so may this isn't the safest idea? I'm ambivalent.

### DIFF
--- a/.github/workflows/portal-ci.yml
+++ b/.github/workflows/portal-ci.yml
@@ -5,6 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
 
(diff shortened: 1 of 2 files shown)


john-conroy:
Any motivation for this PR?

mccalluc:
Not a good one... coming from a mindset where compute is scarce... with the green checkmark even if tests haven't been run, this could be dangerous.

Issue #1439, opened 2021-01-06
Get spatial coordinates on the details page


mccalluc:
Down the road, we might want to show how the sample is positioned in the organ... but for now, it would be fine just to put some numbers on the screen, to demo that the connection is there.

May require index work.

mccalluc:
from #1457
ngehlenborg:
> I would prefer if we could come up with a solution that opens the JSON in a separate view. I can't tell from your screenshot how long the expanded text is, but if our intention for now is to make this available to developers, I would prefer to not include it directly on the dataset/sample page. Could we use the approach that we implemented for entity-level view (open separate page via click on icon at top of page)?

mccalluc:
>  I'd like to avoid adding new code on the python side, which is where it needs ... but that can be done. Should the link be in the section here, or should it be another little icon button?

ngehlenborg:
>  I would add the link in the tissue section.

PR #1457, opened 2021-01-13
Show RUI JSON for Samples

mccalluc:
Might move us towards #1439... but I think the goal / design needs to be thought out more clearly. The `rui_location` is accessible... but is showing it (either as a blob of JSON, or as a set of numbers) useful to anyone? Should the real goal be to have the transparent human here?

(Whatever direction this takes, do we want analogous designs for Samples and Datasets? Only on Sample right now.)

(Branched from PR  #1458.)

### DIFF
--- a/CHANGELOG-show-rui-location.md
+++ b/CHANGELOG-show-rui-location.md
@@ -1,0 +1,1 @@
+- Show RUI location as a blob of JSON.
--- a/CHANGELOG-spatial-dev-search.md
+++ b/CHANGELOG-spatial-dev-search.md
@@ -1,0 +1,1 @@
+- Add a dev search facet to find spatially located datasets; Fix a bug in the facet for spatially located samples.
--- a/context/app/static/js/components/Detail/SampleTissue/SampleTissue.jsx
+++ b/context/app/static/js/components/Detail/SampleTissue/SampleTissue.jsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import { LightBlueLink } from 'js/shared-styles/Links';
 import SectionHeader from 'js/shared-styles/sections/SectionHeader';
 import SectionContainer from 'js/shared-styles/sections/SectionContainer';
-import { FlexPaper } from './style';
+import { FlexPaper, Pre } from './style';
 import SectionItem from '../SectionItem';
 
 function MetadataItem(props) {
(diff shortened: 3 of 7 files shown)


ngehlenborg (approved):
Based on my interpretation of the code this is looking good as a first step.

mccalluc:
@ngehlenborg : Sorry -- Should have included a screenshot: Hope you can understand why I'm hesitant.

Screen Shot 2021-01-13 at 8 35 27 AM

ngehlenborg (changes requested):
Ok, upon review of the screenshot I do not approve. I assumed that was somehow integrated into the dev-search. 

Could we expose the RUI JSON one a separate page, e.g., how we are providing the JSON for the entity pages? 

Or we could try to design a foldable view that maps the data to simple UI?

In the medium term we should aim to show the sample location in the body view (from the CCF EUI), assuming we can put together a component that would allow us to do that. CC @ilan-gold @john-conroy

mccalluc:
Sorry: I should have made sure there was a design first. Just filed #1462.

mccalluc:
Pivot: This is what we decided in the meeting:
- At a minimum, just show that it is registered, like the samples
- Nils would like to be able show body view on any sample/dataset page.
- Also show neighboring samples.
- Want to make an integrated search in year 4.
- Blocking on IU may slow us down unacceptably.

mccalluc:
@Nils:
- We're still ahead of Tiffany's design, but I think this is a stop-gap, and giving us a view on the JSON will make it easier to get to the next phase.
- This PR only covers sample details. I've filed #1478 for dataset details, and dev-search already has filters to find examples.
- I've put the JSON in a summary fold-down, which I think looks good enough, and is the right level of effort for something which should be temporary.
- With line-wrapping and a smaller font, it looks ok.
- I've also made the display of the section contingent on there actually being rui_location: Many samples do not have it.
Screen Shot 2021-01-21 at 5 46 41 PM

ngehlenborg:
@mccalluc I would prefer if we could come up with a solution that opens the JSON in a separate view. I can't tell from your screenshot how long the expanded text is, but if our intention for now is to make this available to developers, I would prefer to not include it directly on the dataset/sample page. Could we use the approach that we implemented for entity-level view (open separate page via click on icon at top of page)?

mccalluc:
@ngehlenborg : I'd like to avoid adding new code on the python side, which is where it needs ... but that can be done. Should the link be in the section here, or should it be another little icon button?

ngehlenborg:
I would add the link in the tissue section.

mccalluc:
Ok, the front-end work here isn't the right direction. Closing, and starting a new branch.

PR #1469, opened 2021-01-19
dev search view toggle

mccalluc:
- Provide two different view toggles, one for the main search, and one for dev.
- For the dev search, provide a stub where a CCF view could be filled in.

On dev-search, we now have this:
Screen Shot 2021-01-19 at 3 08 20 PM

@tsliaw : Only question for you right now is whether that is a good button for CCF view?

@bherr2 : I'm thinking that you will be responsible for filling in the React in `ResultsCCF.jsx`... Does that sound ok to you, or do we need to provide some intermediate wrapper that will handle an interface that you'll specify?

@john-conroy : Code review; As usual, there may be some questionable decisions here. There are now two `SearchViewSwitch` exports:
```
export { SearchViewSwitch as default, DevSearchViewSwitch };
```
Both of these components are created with a factory function, and separating them into two separate files would mean exposing that factory to the outside... but I'd prefer not to do that, just to keep the exports as components. The other consequence is that, now that there's not just a default export, that would need to be copied to the index.js as well... So instead I just got rid of the directory, and moved it `SearchViewSwitch.jsx` inside `SearchBarLayout/`. Thoughts?

### DIFF
--- a/CHANGELOG-dev-search-toggle.md
+++ b/CHANGELOG-dev-search-toggle.md
@@ -1,0 +1,2 @@
+- Provide two different view toggles, one for the main search, and one for dev.
+- For the dev search, provide a stub where a CCF view could be filled in.
--- a/context/app/static/js/components/Search/DevResults/DevResults.jsx
+++ b/context/app/static/js/components/Search/DevResults/DevResults.jsx
@@ -1,13 +1,41 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Hits } from 'searchkit'; // eslint-disable-line import/no-duplicates
+import { ViewSwitcherHits } from 'searchkit';
 
 import ResultsTable from '../ResultsTable';
+import ResultsCCF from '../ResultsCCF';
 
 function DevResults(props) {
   const { sortOptions, hitsPerPage, tableResultFields, detailsUrlPrefix, idField, resultFieldIds } = props;
 
   return (
+    <ViewSwitcherHits
+      hitsPerPage={hitsPerPage}
+      hitComponents={[
+        {
+          key: 'table',
+          title: 'Table',
+          listComponent: (
+            <ResultsTable
+              resultFields={tableResultFields}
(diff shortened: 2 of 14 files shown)


bherr2:
Yeah, this will probably be enough for us to work off of. One thing: could you provide documentation/typings for the props that are given to CCF? I'm not familiar enough with the libraries used here.

tsliaw (approved):
Button icon looks good

mccalluc:
@bherr2 : Just copied over the proptypes from the ResultsTable:
```
ResultsCCF.propTypes = {
  sortOptions: PropTypes.arrayOf(PropTypes.object).isRequired,
  resultFields: PropTypes.arrayOf(PropTypes.object).isRequired,
  detailsUrlPrefix: PropTypes.string.isRequired,
  idField: PropTypes.string.isRequired,
  hits: PropTypes.arrayOf(
    PropTypes.shape({
      sort: PropTypes.array,
      _id: PropTypes.string,
      _index: PropTypes.string,
      _score: PropTypes.number,
      _source: PropTypes.object,
      _type: PropTypes.string,
    }),
  ),
};
```
I think the most important is `hits`, with `_source` being the most important field in the entries of that array.

bherr2:
Great, thanks! I take it hits is straight from the response object from the search-api?

mccalluc:
@bherr2 : yes, this should be straight from ES. If this looks like an ok place to start, please give it a green check. Thanks!

john-conroy (commented):
I think the factory function is fine. More concerned that we are further intertwining search and dev-search when they seem to be diverging.

mccalluc:
@john-conroy : I think they diverged because dev-search didn't have the view modes... but now it does, so it's reasonable to try to reuse code?

@bherr2 : please complete review.

bherr2 (approved):

john-conroy:
I think it's fine to merge, I just wonder if there is a better way to reuse the code than conditionals inside the components themselves.

PR #1471, opened 2021-01-19
hierarchical facets

### LINKED ISSUE #1438: Make a demo of hierarchical facets in the UI
As we get more ontological information, we should let users select facets by hypernyms. This might be useful for anatomy, or assay types. Eventually we'd want to make sure there was data on the backend to support it... but for now, how does searchkit handle it? Feel free to do some grouping on the client-side that eventually we want in the index. 

I'd suggest seeing what the default style is first, and then loop back with Tiffany, rather than designing from a blank slate.

mccalluc:
Demo hierarchical facets by combining the existing status and access level facets. 
Screen Shot 2021-01-19 at 5 12 37 PM

Note that this UI does not include checkboxes, and that unlike the other facets, only one option can be selected at a time. There is a separate [HierarchicalRefinementFilter](http://searchkit.github.io/searchkit/stable/components/navigation/hierarchical-refinement-filter.html) that can handle multiple values... but it also requires a significantly different index structure.

questions, mostly for Tiffany:
- Do you think this will be sufficient? or should we prioritize the indexing work for `HierarchicalRefinementFilter`? Or maybe we use this for now, while not excluding a change to the index in the future...
- For demo, does it make sense to treat the status as a hierarchy? Or maybe something in dev-search?
- Are there other facets that really should be treated as hierarchical now? Down the road, perhaps anatomy, but right now the metadata doesn't say anything about anatomy below the organ level

Fix #1438

### DIFF
--- a/CHANGELOG-hierarchical-facets.md
+++ b/CHANGELOG-hierarchical-facets.md
@@ -1,0 +1,1 @@
+- Demo hierarchical facets with publication status.
--- a/context/app/static/js/components/Search/Search.scss
+++ b/context/app/static/js/components/Search/Search.scss
@@ -30,11 +30,11 @@ $secondary-gray: #636363;
 }
 
 .sk-layout__filters {
-  height: fit-content;
-  height:-webkit-fit-content;
-  height:-ms-fit-content;
-  height:-moz-fit-content;
-  background-color: white;
+    height: fit-content;
+    height:-webkit-fit-content;
+    height:-ms-fit-content;
+    height:-moz-fit-content;
+    background-color: white;
 }
 
 .sk-layout__filters {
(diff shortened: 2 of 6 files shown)


tsliaw:
Do you think you can indent the "Published" more instead of being in-line with "Status"? Similar to the screenshot below. Status does make sense as a hierarchy. 
Screen Shot 2021-01-20 at 9 28 07 AM

As for other facets to be treated as a hierarchy, nothing as of right now, but eventually organ or maybe specimen type would make sense to be treated as a hierarchy.

mccalluc:
padded to align with the labels above:
Screen Shot 2021-01-21 at 4 47 46 PM

tsliaw (approved):

john-conroy (approved):

PR #1483, opened 2021-01-25
Mccalluc/expose rui take 2

### LINKED ISSUE #1439: Get spatial coordinates on the details page

mccalluc:
@ngehlenborg : Looks like this.

Screen Shot 2021-01-25 at 5 30 43 PM

Any tweaks to the wording? Or you can delegate this to Tiffany.

*draft*: Need to handle  Datasets.

Fix #1439

### DIFF
--- a/CHANGELOG-ccf-json.md
+++ b/CHANGELOG-ccf-json.md
@@ -1,0 +1,1 @@
+- Add a link to the CCF JSON, and a new route to serve the JSON.
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -127,16 +127,18 @@ def details_rui_ext(type, uuid, ext):
         abort(404)
     client = _get_client()
     entity = client.get_entity(uuid)
+    # For Samples...
     if 'rui_location' in entity:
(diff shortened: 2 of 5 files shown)


ngehlenborg:
Could you just put the link on "spatial coordinates of this sample" (or "dataset")?

mccalluc:
@tsliaw : Nils has given direction on Samples. What do you think should be done for Datasets? maybe add a "Tissue" section to the page, highlighting the tissue the dataset is derived from? I'm thinking that down the road, if the same component is used in either context, it will set us up well for when we have a visualization. (As opposed to just adding the link to some other section that is already on the datasets page.)

ngehlenborg:
@mccalluc The question about datasets is also somewhat related to provenance (more directly to the question what information is passed along the provenance chain). Let's include that in the list of things to discuss.

ngehlenborg (approved):

tsliaw (approved):

Issue #260, opened 2020-05-13
We shouldn't assume prov-json is too stable.


mccalluc:
More like tech-debt than a bug, but we shouldn't assume too much about the stability of prov-json. It's a serialization of a much more flexible datastructure, and while I don't expect it to change, tying ourselves too tightly too it is like parsing xml with regexes. The easiest thing to point at are the colon-prefixes: These could be changed, without changing the semantics... but it would break us.

Some options:
- See if some other part of the response has the information we need. (best?)
- Propose additions to response which would provide what we need.
- Pull in a real prov-json library. (worst?)

cc @john-conroy

john-conroy:
I think we can close this?

PR #1487, opened 2021-01-26
Mccalluc/new prov structure

### LINKED ISSUE #1434: Provenance graph disappears when pressing on

### LINKED ISSUE #1464: Rework provenance widget to use new Entity API v2.0 /provenance/ call
@mccalluc This is a placeholder for a change that needs to be done in the Portal UI when the new Entity API is done.  The Entity API /entities/<id>/provenance WS call will return the attributes standardized to the same names that are used in search-api.  Also, on entity, the attributes layered under the metadata attribute will now be at the top level.  More details to come...

mccalluc:
Fix #1434. Fix #1464.

@yuanzhou : 
- If possible, I hope https://github.com/hubmapconsortium/entity-api/issues/106 can be addressed before this is merged. Not a blocker, per se, but I want to be sure we don't get ahead of ourselves.
- Would you ever anticipate that the prefix strings would change? A proper parser would resolve these to the URIs... but if they are never going to change, that flexibility would be gratuitous. (see #260)

### DIFF
--- a/context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
+++ b/context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
@@ -3,50 +3,51 @@ import PropTypes from 'prop-types';
 
 import SectionItem from 'js/components/Detail/SectionItem';
 import { renderProvVis } from '../ProvVis';
-import { StyledTypography, StyledLink, FlexPaper } from './style';
+import { StyledTypography, FlexPaper } from './style';
 import '@hms-dbmi-bgm/react-workflow-viz/dist/react-workflow-viz.min.css';
 
 function ProvGraph(props) {
   const { provData } = props;
   const isOld = 'ex' in provData.prefix;
   const idKey = isOld ? 'hubmap:displayDOI' : 'hubmap:hubmap_id';
+  const timeKey = isOld ? 'prov:generatedAtTime' : 'hubmap:created_timestamp';
+  const typeKey = isOld ? 'prov:type' : 'hubmap:entity_type';
   const runRenderProvVis = () =>
     renderProvVis('prov-vis-react', provData, {
       getNameForActivity: (id, prov) => {
         const activity = prov.activity[id];
-        return `${activity['prov:type']} - ${activity[idKey]}`;
+        return `${activity['hubmap:creation_action']} - ${activity[idKey]}`;
       },
       getNameForEntity: (id, prov) => {
         const entity = prov.entity[id];
         // NOTE: The initial entity node was not included in the sample data;
         // Fallback to ID, if needed. https://github.com/hubmapconsortium/prov-vis/issues/15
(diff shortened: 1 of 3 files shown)


yuanzhou:
@mccalluc I'll address the API difference issue next week. Regarding the prefix used in the provenance json, the current version is using `hubmap:` for HuBMAP data, and a few `prov:` since they are the standard (I'm not sure since I didn't write the code). They can be changed if we need to change the prefix, but less likely I think. @shirey what's your thought on this?

shirey:
@yuanzhou Not sure about the context of why you're asking about the attribute prefixes (what is the original question?), but I would think that using the same as previous should work.

yuanzhou:
@shirey because @mccalluc asked in the beginning:
> Would you ever anticipate that the prefix strings would change? A proper parser would resolve these to the URIs... but if they are never going to change, that flexibility would be gratuitous. (see #260)

I agree that the prefix would not need to be changed, it's just a prefix constant defined in the code.

shirey:
@yuanzhou I'd say leave these as is for now.  It seems to be compliant with the PROV-JSON standard which was the agreed upon standard.  If this is too cumbersome we can change it in the future.

mccalluc:
> And if by any chance it has to be changed down the road, @mccalluc maybe just put it in configuration?
If it does change, that will be the `prefix` section of the JSON, so I could pull it from there... but it sound like it should be stable for now, so I won't invest time there. Thanks for giving me the broader view.

yuanzhou (approved):

context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
```diff
@@ -3,48 +3,52 @@ import PropTypes from 'prop-types';
...
-import { StyledTypography, StyledLink, FlexPaper } from './style';
+import { StyledTypography, FlexPaper } from './style';
 import '@hms-dbmi-bgm/react-workflow-viz/dist/react-workflow-viz.min.css';
 
 function ProvGraph(props) {
   const { provData } = props;
+  const isOld = 'ex' in provData.prefix;
+  const idKey = isOld ? 'hubmap:displayDOI' : 'hubmap:hubmap_id';
+  const timeKey = isOld ? 'prov:generatedAtTime' : 'hubmap:created_timestamp';
+  const typeKey = isOld ? 'prov:type' : 'hubmap:entity_type';
```
john-conroy, line 14: Could make sense to store the details in an object in a config file or similar - otherwise looks good.

john-conroy (approved):

context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
mccalluc: Down the road, I just want to get rid of `*Key` ... They are really only needed during the transition. I'll file an issue for the cleanup (#1497) ... I don't think an outside config makes sense... it would be another thing to worry about switching between environments. (Though this attitude could be taken too far!)

PR #1496, opened 2021-02-01
John conroy/saved list page

john-conroy:
The foundation for a saved list 'detail' page which users can access by clicking on a saved list in the 'My Lists' page. Refactors the summary used in existing detail pages for use with saved lists and fixes date formats to match current designs. Also combines the stores into a single store.

Encoding/decoding the list title used in the url path fixed in #1500.


context/app/static/js/stores/useSavedEntitiesStore.js
```diff
@@ -16,6 +16,37 @@ const useSavedEntitiesStore = create(
...
           delete state.savedEntities[entityUuid];
         }),
+      savedLists: {},
+      createList: ({ title, description }) =>
+        set((state) => {
+          state.savedLists[title] = {
+            // the Donor, Sample, and Datasets are objects to avoid duplicates. Normally they would be sets, but objects work better with local storage
+            Donor: {},
+            Sample: {},
+            Dataset: {},
```
mccalluc, line 26: Prefer is just keep keys lowercase, if possible.

mccalluc (commented):
One small comment... but not reading this as closely as the first few, so ping me if something needs more attention.

context/app/static/js/stores/useSavedEntitiesStore.js
john-conroy: As we discussed, keeping this for now, but we may revisit later.

PR #1500, opened 2021-02-02
John conroy/delete saved list

john-conroy:
Allows the user to delete a saved list permanently from the saved list 'detail' page. Also fixes issues with encoding/decoding list titles used in the url path for navigation.

Ran into an issue deleting the list on the page itself so instead I added the list to a queue to be deleted once the user is navigated to the 'My Lists' page after selecting delete. Let me know if you have any alternative ideas.


mccalluc:
> Ran into an issue deleting the list on the page itself so instead I added the list to a queue to be deleted once the user is navigated to the 'My Lists' page after selecting delete. Let me know if you have any alternative ideas.

Perhaps when being on a missing list is more like a 404 and less like an code problem, this will be smoothed out? Happy to defer for now.

PR #1501, opened 2021-02-02
John conroy/edit saved list

john-conroy:
Enables the user to edit a saved list's description and title from the list detail page. Refactor the text fields to be reusable across components.

### DIFF
--- a/context/app/static/js/components/savedLists/EditListDialog/EditListDialog.jsx
+++ b/context/app/static/js/components/savedLists/EditListDialog/EditListDialog.jsx
@@ -1,0 +1,62 @@
+import React, { useState } from 'react';
+import Button from '@material-ui/core/Button';
+
+import useSavedEntitiesStore from 'js/stores/useSavedEntitiesStore';
+import DialogModal from 'js/shared-styles/DialogModal';
+import { StyledTitleTextField, StyledDescriptionTextField } from './style';
+
+const useSavedEntitiesStoreSelector = (state) => ({ createList: state.createList, deleteList: state.deleteList });
+
+function EditListDialog({ dialogIsOpen, setDialogIsOpen, listDescription, listTitle, setEditedListTitle }) {
+  const [title, setTitle] = useState(listTitle);
+  const [description, setDescription] = useState(listDescription);
+
+  const { createList, deleteList } = useSavedEntitiesStore(useSavedEntitiesStoreSelector);
+
+  function handleTitleChange(event) {
+    setTitle(event.target.value);
+  }
+
+  function handleDescriptionChange(event) {
+    setDescription(event.target.value);
+  }
+
+  const handleClose = () => {
+    setDialogIsOpen(false);
+    setTitle(listTitle);
+    setDescription(listDescription);
+  };
+
+  function handleSubmit() {
+    createList({ title, description });
+    setEditedListTitle(title);
+    deleteList(listTitle);
+    setDialogIsOpen(false);
+  }
+
+  return (
+    <DialogModal
+      title={`Edit ${listTitle}`}
+      content={
+        <>
+          <StyledTitleTextField handleChange={handleTitleChange} title={title} />
+          <StyledDescriptionTextField handleChange={handleDescriptionChange} description={description} />
+        </>
+      }
+      actions={
+        <>
+          <Button onClick={handleClose} color="primary">
+            Cancel
+          </Button>
+          <Button onClick={() => handleSubmit()} color="primary" disabled={title.length === 0}>
+            Save
+          </Button>
+        </>
+      }
+      isOpen={dialogIsOpen}
+      handleClose={handleClose}
+    />
+  );
+}
+
+export default EditListDialog;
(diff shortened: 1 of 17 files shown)


context/app/static/js/components/savedLists/listTextFields/index.js
```diff
@@ -0,0 +1,43 @@
...
+function TitleTextField({ handleChange, title }) {
+  return (
+    <PrimaryOutlinedTextField
+      autoFocus
+      margin="dense"
+      id="title"
+      label="Title"
+      fullWidth
+      variant="outlined"
+      placeholder='Like “Spleen-Related Data” or “ATAC-seq Visualizations"'
```
mccalluc, line 16:
```suggestion
      placeholder='Like “Spleen-Related Data” or “ATAC-seq Visualizations”'
```

context/app/static/js/pages/SavedList/SavedList.jsx
```diff
@@ -20,22 +21,36 @@ function getListAndItsEntities(savedLists, listTitle) {
 
 function SavedList({ listTitle }) {
   const decodedTitle = decodeURIComponent(listTitle);
+  const [editedListTitle, setEditedListTitle] = useState(false);
+
+  const currentTitle = editedListTitle || decodedTitle;
```
mccalluc, line 26: Bigger picture: I don't think we should use the title as a key, or put it in the URL (except, down the road, perhaps as a stub which isn't actually needed for URL resolution). So right here, maybe there's just one title, instead of having fallback logic?

context/app/static/js/pages/SavedList/SavedList.jsx
```diff
@@ -20,22 +21,36 @@ function getListAndItsEntities(savedLists, listTitle) {
...
-      <Typography variant="h2">{decodedTitle}</Typography>
+      <Typography variant="h2">{currentTitle}</Typography>
       <RightAlignedButtonRow
         leftText={
           <Typography variant="body1" color="primary">
             {Object.keys(listEntities).length}
           </Typography>
         }
-        buttons={<SavedListMenuButton listTitle={decodedTitle} />}
+        buttons={
```
mccalluc, line 44: This might also be more idiomatic as children?

context/app/static/js/stores/useSavedEntitiesStore.js
```diff
@@ -65,6 +65,11 @@ const useSavedEntitiesStore = create(
           state.listsToBeDeleted = [];
         });
       },
+      deleteList: (listTitle) => {
+        set((state) => {
+          delete state.savedLists[listTitle];
```
mccalluc, line 70: Am I understanding it correctly, that one of the big selling points of immer is that we can use idioms like this that mutate the object, instead of making work-arounds?

mccalluc (commented):
Maybe sort it out latter, but I think there are places where it might be more readable to use child components, instead of putting components in props. 

For all of these, you might want to merge it down to your feature branch, and then do some final cleanup there. Your call.

context/app/static/js/stores/useSavedEntitiesStore.js
john-conroy: Yep!

context/app/static/js/pages/SavedList/SavedList.jsx
john-conroy: As we discussed, in situations where there are multiple groups of components that need to be passed down, components as props can make sense but prefer children where possible.

context/app/static/js/components/savedLists/EditListDialog/EditListDialog.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <DialogModal
+      title={`Edit ${listTitle}`}
+      content={
+        <>
+          <StyledTitleTextField handleChange={handleTitleChange} title={title} />
+          <StyledDescriptionTextField handleChange={handleDescriptionChange} description={description} />
+        </>
+      }
```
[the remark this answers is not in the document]
john-conroy: Same comment about child comments as the one below.

context/app/static/js/components/savedLists/EditListDialog/EditListDialog.jsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  function handleDescriptionChange(event) {
+    setDescription(event.target.value);
+  }
+
+  const handleClose = () => {
+    setDialogIsOpen(false);
+    setTitle(listTitle);
+    setDescription(listDescription);
```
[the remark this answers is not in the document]
john-conroy: We're using set here to reset the form field values when the dialog is cancelled so the next time it is opened it will open with the current values.

context/app/static/js/components/savedLists/EditListDialog/EditListDialog.jsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  function handleSubmit() {
+    createList({ title, description });
+    setEditedListTitle(title);
+    deleteList(listTitle);
+    setDialogIsOpen(false);
```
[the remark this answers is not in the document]
john-conroy: I thought this made the most sense, I'll look into it again. Thanks.

context/app/static/js/pages/SavedList/SavedList.jsx
john-conroy: As we discussed, I'll update everything to use a generated ID instead of the title once we get it deployed for people to test.

Issue #1228, opened 2020-09-18
Mockup User-Customizable Collections


tsliaw:
Notes:
- Consider user with multiple devices
- Version 1 might have to be on local storage
- There should be "shopping cart" of data, which can be used to make a user-customizable collection

mccalluc:
Tiffany needs to rewind... identify subset to implement.

mccalluc:
Needs Nils feedback. Big!

tsliaw:
Approved Figma Design:
- [Detail Page "Save" Addition](https://www.figma.com/file/EQSyXT2wv6lh9cYfsggRql/My-Lists-Page?node-id=90%3A261)
- [My Lists Page](https://www.figma.com/file/EQSyXT2wv6lh9cYfsggRql/My-Lists-Page?node-id=2121%3A6311)
-[List Detail Page](https://www.figma.com/file/EQSyXT2wv6lh9cYfsggRql/My-Lists-Page?node-id=38%3A570)

PR #1504, opened 2021-02-02
John conroy/remove from lists edit saved status

john-conroy:
Checkboxes are now checked once you open the dialog based on whether the list already contains the entity. Clicking submit will add the entity to any checked lists and will remove the entity from unchecked lists. There is a lot of looping here, let me know if you can think of any alternatives to reduce it.

### DIFF
--- a/context/app/static/js/components/Detail/SaveEditEntityButton/SaveEditEntityButton.jsx
+++ b/context/app/static/js/components/Detail/SaveEditEntityButton/SaveEditEntityButton.jsx
@@ -9,7 +9,7 @@ function SaveEditEntityButton({ uuid, entity_type }) {
   const savedEntities = useSavedEntitiesStore(useSavedEntitiesSelector);
 
   return uuid in savedEntities ? (
-    <StyedEditSavedStatusButton uuid={uuid} />
+    <StyedEditSavedStatusButton uuid={uuid} entity_type={entity_type} />
   ) : (
     <StyledSaveEntityButton uuid={uuid} entity_type={entity_type} />
   );
--- a/context/app/static/js/components/savedLists/EditSavedStatusButton/EditSavedStatusButton.jsx
+++ b/context/app/static/js/components/savedLists/EditSavedStatusButton/EditSavedStatusButton.jsx
@@ -3,7 +3,7 @@ import Button from '@material-ui/core/Button';
 
 import EditSavedStatusDialog from 'js/components/savedLists/EditSavedStatusDialog';
 
-function EditSavedStatusButton({ uuid, ...rest }) {
+function EditSavedStatusButton({ uuid, entity_type, ...rest }) {
   const [dialogIsOpen, setDialogIsOpen] = useState(false);
 
   return (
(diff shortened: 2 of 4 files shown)


mccalluc (commented):
@john-conroy , checking this branch out, here are my usabilty thoughts. Tomorrow morning I'll start going through the code on all of these, including this one.

- From http://localhost:5001/my-lists, "Create New List", there doesn't seem to be anything to do with the new list I've created.
- on http://localhost:5001/my-lists/This%20is%20my%20list, what is the number, currently "0", below the title? Should there be more explanation?
- Presumably, "more options" will have something besides delete, at some point? Fine for now.
- Using names as keys in URLs may cause problems... if the name changes, links break.
Perhaps not such a big deal if it's all local, but it's not a good precedent.
- No warning about name collisions when creating lists: The dialog closes; no error.
- Behavior for http://localhost:5001/my-lists/no-such-list should be more like a 404, less like an code error.
- "My Saves Lists" -> "My Saved Lists"?
- Checkbox in header can not be unclicked, and unclicking items in list does not clean checkbox in header. Compare with github, as an example.
- When there are multiple saves, the incremental loading of "My Saves List" is disconcerting... Should be able to fire off a single request to get all UUIDs of interest in one response. 
- In the "My Saves List" section, after clicking "Add to", and selecting a target, there is no change in the counts on the page.
- Should "Add to" be disabled if no items are selected? And/or, in the dialog message, include a count of current items?
- Similarly with the trash can to delete.
- In search results, after initially having the whole rect as a click target, we backed off and limitted it to the ID... should we do something similar here and limit it to the title?
- I don't actually see anything of my saves on http://localhost:5001/my-lists/This%20is%20my%20list... This is still TODO?

john-conroy:
> From http://localhost:5001/my-lists, "Create New List", there doesn't seem to be anything to do with the new list I've created.

There will be an additional table in the list detail page which will allow you to remove items from the lists. The rows will be links to the entity detail pages.

> on http://localhost:5001/my-lists/This%20is%20my%20list, what is the number, currently "0", below the title? Should there be more explanation?

I need to add 'Items' next to the number.

> Using names as keys in URLs may cause problems... if the name changes, links break. Perhaps not such a big deal if it's all local, but it's not a good precedent.

I thought about this. Should I generate IDs and store a lookup table of sorts in local storage?

> No warning about name collisions when creating lists: The dialog closes; no error.

Work in progress, thanks!

> Behavior for http://localhost:5001/my-lists/no-such-list should be more like a 404, less like an code error.

Ok, I'll figure this out.

> Checkbox in header can not be unclicked, and unclicking items in list does not clean checkbox in header. Compare with github, as an example.

@tsliaw how should this behave? I think he's referring to the table in the my lists page.

> When there are multiple saves, the incremental loading of "My Saves List" is disconcerting... Should be able to fire off a single request to get all UUIDs of interest in one response.

I was trying to avoid storing more metadata for each entity in the local storage and as a result have to make requests to get that data. If we're not concerned with added size/complexity in storage, I can just add it there and avoid the request, but that could cause the data to not always be fresh. Is there a way to send a request for metadata for several uuids?

> In the "My Saves List" section, after clicking "Add to", and selecting a target, there is no change in the counts on the page.

When clicking 'Add to' you are adding the saved entities to the lists below and the counts in the Donors/Samples/Datasets does change. Maybe I am misunderstanding this or there's an issue I'm not experiencing.

> Should "Add to" be disabled if no items are selected? And/or, in the dialog message, include a count of current items?
Similarly with the trash can to delete.

Both of these will soon only displayed when at least one item is selected, thanks.

> In search results, after initially having the whole rect as a click target, we backed off and limitted it to the ID... should we do something similar here and limit it to the title?

If I'm remembering correctly we did that because we were having issues selecting the data inside the table cells. Do we think people will want to again select the list name or similar in these rows? @tsliaw what is your opinion?

> I don't actually see anything of my saves on http://localhost:5001/my-lists/This%20is%20my%20list... This is still TODO?

Yes I wanted to get your feedback on everything as is, thanks.

john-conroy:
@mccalluc designs can be found in #1228 if it would be useful to look at them.

PR #1517, opened 2021-02-05
John conroy/undo separation of list entities by type

john-conroy:
Converges the saved entities in each list into a single object instead of being separated by entity type to allow easier access at the cost of additional looping to determine count by type.

### DIFF
--- a/context/app/static/js/components/savedLists/EditSavedStatusDialog/EditSavedStatusDialog.jsx
+++ b/context/app/static/js/components/savedLists/EditSavedStatusDialog/EditSavedStatusDialog.jsx
@@ -15,7 +15,7 @@ const useSavedEntitiesSelector = (state) => ({
 
 function getSavedListsWhichContainEntity(savedLists, savedEntity) {
   return Object.entries(savedLists).reduce((acc, [title, obj]) => {
-    return savedEntity in obj ? [...acc, title] : acc;
+    return savedEntity in obj.savedEntities ? [...acc, title] : acc;
   }, []);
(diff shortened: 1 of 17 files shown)


context/app/static/js/components/savedLists/EditSavedStatusDialog/EditSavedStatusDialog.jsx
```diff
@@ -13,29 +13,26 @@ const useSavedEntitiesSelector = (state) => ({
...
 });
 
-function getSavedListsWhichContainEntity(savedLists, savedEntity, entity_type) {
+function getSavedListsWhichContainEntity(savedLists, savedEntity) {
   return Object.entries(savedLists).reduce((acc, [title, obj]) => {
-    if (savedEntity in obj[entity_type]) {
-      return [...acc, title];
-    }
-    return acc;
+    return savedEntity in obj.savedEntities ? [...acc, title] : acc;
```
mccalluc, line 18:
```suggestion
    return (savedEntity in obj.savedEntities) ? [...acc, title] : acc;
```
In case I forget order of operations.

mccalluc (approved):
Seems like a good simplification.

context/app/static/js/components/savedLists/EditSavedStatusDialog/EditSavedStatusDialog.jsx
john-conroy: Worth a prettier ignore?

PR #1518, opened 2021-02-08
John conroy/fix saved list links

john-conroy:
Adds links to table IDs and moves select row on click to the checkbox only. Moves link in saved list panels from entire panel to the title.

Do we want the panels used elsewhere (Collections page) to look and function the same way?

Screen Shot 2021-02-08 at 10 13 31 AM


mccalluc (approved):
Looks good.

PR #1519, opened 2021-02-08
John conroy/enforce unique list titles

john-conroy:
I wrote the text myself so please let me know what you would like me to replace it with. Also made changes to reset the create list form fields on exit.

Screen Shot 2021-02-08 at 11 29 39 AM

### DIFF
--- a/context/app/static/js/components/savedLists/CreateListDialog/CreateListDialog.jsx
+++ b/context/app/static/js/components/savedLists/CreateListDialog/CreateListDialog.jsx
@@ -28,6 +28,11 @@ function CreateListDialog({ dialogIsOpen, setDialogIsOpen }) {
     setDialogIsOpen(false);
   };
 
+  function handleExit() {
+    setTitle('');
+    setDescription('');
+  }
+
   function handleSubmit() {
     if (!(title in savedLists)) {
       createList({ title, description });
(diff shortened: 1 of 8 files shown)


context/app/static/js/components/savedLists/CreateListDialog/CreateListDialog.jsx
```diff
@@ -25,14 +28,24 @@ function CreateListDialog({ dialogIsOpen, setDialogIsOpen }) {
...
+      setDialogIsOpen(false);
+    } else {
+      setShouldDisplayWarning(true);
+    }
   }
 
   return (
     <DialogModal
       title="Create New List"
+      warning={shouldDisplayWarning && 'A list with that title already exists.'}
```
mccalluc, line 48:
```suggestion
      warning={shouldDisplayDuplicateTitleWarning && 'A list with that title already exists.'}
```
... In case there are other failure modes in the future? ... but it might be getting too long?

mccalluc (commented):
So I don't actually see this as essential: We're moving in a direction where the lists have IDs, so nothing will blow up if a name is reused.... and every bit of logic we add now is something that can break in the future. 

If Tiffany or Nils want this, listen to them... but I'd have a slight preference for keeping the code simpler.

context/app/static/js/components/savedLists/CreateListDialog/CreateListDialog.jsx
john-conroy: I'll modify in the future if we need another warning. I'll also add this to the list of feedback areas.

PR #1524, opened 2021-02-10
John conroy/fix saved list spacing

john-conroy:
I opted to use extra divs to add spacing for ease of use and to avoid having to deal with more complicated props interactions. Let me know if you think it makes more sense to add the margin to the components themselves.

Screen Shot 2021-02-10 at 11 20 59 AM
Screen Shot 2021-02-10 at 11 20 49 AM

### DIFF
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -192,12 +192,12 @@ def my_lists():
     )
 
 
-@blueprint.route('/my-lists/<saved_list_title>')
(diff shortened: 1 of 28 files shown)


mccalluc (approved):
Looks good. I don't have a problem with `SpacingDiv`. Wondering if `Description` should moved to its own file so we can stick with default exports, but your call.

tsliaw (approved):

PR #1542, opened 2021-02-12
John conroy/saved list route tracking

john-conroy:
Fix page view tracking to track visits to a saved list page together instead of separating by uuid. Visits to any saved list page should show as `/my-lists/saved-list` instead of `/my-lists/{uuid}`.

### DIFF
--- a/context/app/static/js/components/Routes/useSendPageView.js
+++ b/context/app/static/js/components/Routes/useSendPageView.js
@@ -8,6 +8,11 @@ function useSendPageView(path) {
       ReactGA.pageview(pathWithoutUUID);
       return;
     }
+    // send path without ID for specific saved list page
+    if (path.match(/\/my-lists\/[\w-]+/)) {
+      ReactGA.pageview('/my-lists/saved-list');
+      return;
+    }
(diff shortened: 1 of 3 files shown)


context/app/static/js/components/Routes/useSendPageView.js
```diff
@@ -8,6 +8,11 @@ function useSendPageView(path) {
       ReactGA.pageview(pathWithoutUUID);
       return;
     }
+    // send path without ID for specific saved list page
+    if (path.match(/\/my-lists\/[\w-]+/)) {
```
mccalluc, line 12:
```suggestion
    if (path.startswith('/my-lists/') { // Distinguished by final slash. 
```
Avoid regexes if you can.

mccalluc (commented):
Prefer to avoid regexes, unless we need them.

context/app/static/js/components/Routes/useSendPageView.js
john-conroy: @mccalluc Well, I'm specifically looking for /my-lists/{uuid} where uuid is a dash separated string since just /my-lists is the lists page.

PR #1544, opened 2021-02-12
John conroy/track web vitals

### LINKED ISSUE #1511: Track site performance, and record in google analytics

john-conroy:
Fix #1511.

Adds tracking to each route for the four big web vitals metrics. The event will be linked to a page view but it will be the full path not the simplified path we use (so uuids will be included), if you think it makes sense I could add the simplified page view to one of the fields we send over.

@tsliaw you may want to look at the information you get for the events eventually and see if there's anything else you'd like me to add.

### DIFF
--- a/context/app/static/js/components/Routes/useSendWebVitals.js
+++ b/context/app/static/js/components/Routes/useSendWebVitals.js
@@ -2,6 +2,7 @@ import { useEffect } from 'react';
 import ReactGA from 'react-ga';
 import { getCLS, getFID, getLCP, getTTFB } from 'web-vitals';
 
+// copy-and-paste from https://github.com/GoogleChrome/web-vitals#using-analyticsjs
 function sendVitalsToGA({ name, delta, id }) {
   ReactGA.event({
     category: 'Web Vitals',
--- a/context/app/static/js/components/Routes/useSendWebVitals.js
+++ b/context/app/static/js/components/Routes/useSendWebVitals.js
@@ -1,0 +1,31 @@
+import { useEffect } from 'react';
+import ReactGA from 'react-ga';
+import { getCLS, getFID, getLCP, getTTFB } from 'web-vitals';
+
+function sendVitalsToGA({ name, delta, id }) {
+  ReactGA.event({
+    category: 'Web Vitals',
+    action: name,
+    // From web-vitals: The `id` value will be unique to the current page load. When sending
+    // multiple values from the same page (e.g. for CLS), Google Analytics can
+    // compute a total by grouping on this ID (note: requires `eventLabel` to
+    // be a dimension in your report).
+    label: id,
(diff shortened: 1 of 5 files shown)


context/app/static/js/components/Routes/useSendWebVitals.js
```diff
@@ -0,0 +1,31 @@
+import { useEffect } from 'react';
+import ReactGA from 'react-ga';
+import { getCLS, getFID, getLCP, getTTFB } from 'web-vitals';
+
+function sendVitalsToGA({ name, delta, id }) {
+  ReactGA.event({
+    category: 'Web Vitals',
+    action: name,
```
mccalluc, line 8:
```suggestion
    action: (name === 'CLS' ? 'CLS * 1000' : name),
```
If we ended up using a different multiplier, we'd want to distinguish new and old data?

mccalluc: or maybe not: Like not to tweak copy-and-paste code without a good reason. (Also wondering is other metrics are worth multiplying.)

context/app/static/js/components/Routes/useSendWebVitals.js
```diff
@@ -0,0 +1,31 @@
+import { useEffect } from 'react';
+import ReactGA from 'react-ga';
+import { getCLS, getFID, getLCP, getTTFB } from 'web-vitals';
+
+function sendVitalsToGA({ name, delta, id }) {
```
mccalluc, line 5:
```suggestion
// copy-and-paste from https://github.com/GoogleChrome/web-vitals#using-analyticsjs
function sendVitalsToGA({ name, delta, id }) {
```
credit / blame where it's due.

context/app/static/js/components/Routes/useSendWebVitals.js
```diff
@@ -0,0 +1,31 @@
+import { useEffect } from 'react';
+import ReactGA from 'react-ga';
+import { getCLS, getFID, getLCP, getTTFB } from 'web-vitals';
```
mccalluc, line 3: In the docs, `TTFB` isn't highlighted the way the other three are... Do you have any sense why not?

mccalluc (commented):
Just cite your sources, and good to merge.

context/app/static/js/components/Routes/useSendWebVitals.js
john-conroy: If you look in the beginning of the readme they list those three as 'core'.

context/app/static/js/components/Routes/useSendWebVitals.js
john-conroy: Sure, I should have been more detailed with my 'from web-vitals' comments.

PR #1546, opened 2021-02-12
John conroy/track outbound links

john-conroy:
The start of tracking outbound links. I wanted to get your thoughts before I went about replacing more of these.


mccalluc (commented):
I think this is fine as-is... but a couple thoughts:
- Would we like to make more of the treatment of outbound links (like `target="_blank" rel="noopener noreferrer"` and the the box and arrow icon) a default that the component would supply?
- Maybe something other than`Clicked`, perhaps `Navigated` or just `Followed`, so it will be easier to segment on-page interactions from others? but you already have `Outbound Link`, and that's probably enough.
- Is the page where the even happens already recorded by default?

Again, fine as-as, but if anything here resonates with you, great.

context/app/static/js/components/Detail/SampleTissue/SampleTissue.jsx
```diff
@@ -36,9 +37,9 @@ function SampleTissue(props) {
                 spatial coordinates of this sample
               </LightBlueLink>{' '}
               have been registered and it can be found in the{' '}
-              <LightBlueLink href="/ccf-eui" target="_blank" rel="noopener noreferrer">
+              <OutboundLink href="/ccf-eui" target="_blank" rel="noopener noreferrer">
```
mccalluc, line 40:
```suggestion
              <OutboundLink href="/ccf-eui">
```

context/app/static/js/components/files/FilesConditionalLink/FilesConditionalLink.jsx
```diff
@@ -1,16 +1,16 @@
...
-import { LightBlueLink } from 'js/shared-styles/Links';
+import OutboundLink from 'js/shared-styles/Links/OutboundLink';
 import { AlignedLink } from './style';
 
 function FilesConditionalLink(props) {
   const { hasAgreedToDUA, openDUA, href, children, ...rest } = props;
   if (hasAgreedToDUA) {
     return (
-      <LightBlueLink target="_blank" rel="noopener noreferrer" underline="none" href={href} {...rest}>
+      <OutboundLink href={href} {...rest}>
```
mccalluc, line 11:
```suggestion
      <OutboundLink href={href} underline="none" {...rest}>
```

mccalluc (commented):
Two typos... but for all I know, there might have been mistakes there originally, and you're fixing them now? Whatever you think best.

context/app/static/js/components/files/FilesConditionalLink/FilesConditionalLink.jsx
john-conroy: Underline comes free with LightBlueLink this was just old.

PR #1554, opened 2021-02-17
John conroy/empty saved list table

### LINKED ISSUE #1550: Change empty table to alert when no items in a list
This is what it currently looks like when there are no items in a list.
Screen Shot 2021-02-16 at 10 19 04 AM

Can you change it to look like this screenshot instead when there are no items? [Figma](https://www.figma.com/file/EQSyXT2wv6lh9cYfsggRql/My-Lists-Page?node-id=2300%3A19064)
Screen Shot 2021-02-16 at 10 18 33 AM

john-conroy:
I altered the text slightly since we're calling it 'My Saved Items' instead of 'My Saves List'. Let me know if I should change the text. Fix #1550.

Screen Shot 2021-02-17 at 11 34 06 AM

### DIFF
--- a/CHANGELOG-empty-saved-list-table.md
+++ b/CHANGELOG-empty-saved-list-table.md
@@ -1,0 +1,1 @@
+- Display message instead of empty table on saved list page.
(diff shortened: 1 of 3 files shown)


tsliaw (approved):
Looks good

PR #1570, opened 2021-02-19
John conroy/fix prov vis react

### LINKED ISSUE #154: In react, use prov-vis directly
with Jennifer's PR, we're still using the helper method, but that shouldn't be needed: We can just use the react component.

### LINKED ISSUE #1486: in ProvGraph, using StyledLink causes "TypeError: props.theme.palette is undefined"
I've switched to plain `<a>`, which fixes #1434, but it's a stop gap. ... This whole component is using React in the wrong way, so the real fix may be to address #154

john-conroy:
Fix #154. Fix #1486. From my perspective everything looks and performs as it should, but you're much more familiar with the prov vis so it may be worth testing locally. The test in `ProvVis.spec.js` looks a little strange still, but probably not worth revisiting until we add more testing.

Screen Shot 2021-02-19 at 3 41 09 PM

### DIFF
--- a/context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
+++ b/context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
@@ -2,7 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import SectionItem from 'js/components/Detail/SectionItem';
-import { renderProvVis } from '../ProvVis';
+import { LightBlueLink } from 'js/shared-styles/Links';
+import ProvVis from '../ProvVis';
 import { StyledTypography, FlexPaper } from './style';
 import '@hms-dbmi-bgm/react-workflow-viz/dist/react-workflow-viz.min.css';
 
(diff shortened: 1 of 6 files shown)


context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
```diff
@@ -12,55 +13,63 @@ function ProvGraph(props) {
...
+
+  function getNameForActivity(id, prov) {
+    const activity = prov.activity[id];
+    return `${activity['hubmap:creation_action']} - ${activity[idKey]}`;
+  }
+
+  function getNameForEntity(id, prov) {
+    const entity = prov.entity[id];
+    // NOTE: The initial entity node was not included in the sample data;
+    // Fallback to ID, if needed. https://github.com/hubmapconsortium/prov-vis/issues/15
```
mccalluc, line 25: That repo is archived, so this has fallen off the radar. Could you ...
- Confirm that the referenced entity is still not included in the API response...
- If it's not, ping Bill on Slack and ask if this is something they might fix...
  - if so, make sure there's a tracking issue on their side,
  - if not, make an issue on our side to strip out the useless node before rendering.

mccalluc (commented):
I'm currently blocked by #1573... I don't think it's related to this PR, per se, but it's preventing me from trying it out locally... probably fix that in a separate PR, merge it to master, and then merge master into this?

Also: if you can follow-up with the one TODO that is copied over, and see if the landscape is any different than it was before.

Code all looks good.

context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
john-conroy: Filed #1574

PR #1593, opened 2021-03-01
John conroy/descendant prov

### LINKED ISSUE #1595: Show derived entities in the provenance graph
Users should be able to click on an entity and elect to show derived entities in the graph.

john-conroy:
Fix #1595.

Adds a 'Show Descendants' button to show descendants of entity in provenance graph detail pane. This has worked for both simple cases such as this donor `190aec1937701a97952b34cfac5529a3` where the resulting provenance matches its only descendant sample `14e44841842d639b28bfdf9188a95a86` and sample `d1fe368864b494e4cace64d68c7c60ed` where multiple descendants exist. Incrementally revealing descendants seems to work, but it is hard to verify (such as instances where paths diverge only to converge later, but I haven't been able to find an example of this if it could exist).

We probably won't want to merge this to master until I add some tests and we do some manual testing.

Likely follow-ups:
- Fetch immediate descendants when detail pane is opened and disable the button if no immediate descendants exist. Should we display something else instead of the disabled button.
- Disable button if the button has already been clicked and immediate descendants are shown.

@tsliaw do we want to change the button to 'Show Derived Entities' instead of 'Show Descendants'?

Screen Shot 2021-03-01 at 3 20 08 PM

Screen Shot 2021-03-01 at 3 20 18 PM

### DIFF
--- a/context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
+++ b/context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
@@ -1,28 +1,12 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import Button from '@material-ui/core/Button';
 
 import SectionItem from 'js/components/Detail/SectionItem';
 import { LightBlueLink } from 'js/shared-styles/Links';
-import { AppContext } from 'js/components/Providers';
-import useImmediateDescendantProv from 'js/hooks/useImmediateDescendantProv';
-import useProvenanceStore from 'js/stores/useProvenanceStore';
+import ShowDerivedEntitiesButton from 'js/components/Detail/provenance/ShowDerivedEntitiesButton';
 import ProvVis from '../ProvVis';
-import { StyledTypography, FlexPaper } from './style';
+import { FlexPaper } from './style';
 import '@hms-dbmi-bgm/react-workflow-viz/dist/react-workflow-viz.min.css';
-import ProvData from '../ProvVis/ProvData';
-
-function createStepNameSet(steps) {
-  return new Set(steps.map((step) => step.name));
-}
-
-function removeExistingSteps(steps, newSteps) {
-  const nameSet = createStepNameSet(steps);
-  const uniqueNewSteps = newSteps.filter((step) => !nameSet.has(step.name));
-  return uniqueNewSteps;
-}
-
-const useProvenanceStoreSelector = (state) => ({ steps: state.steps, addDescendantSteps: state.addDescendantSteps });
 
 function ProvGraph(props) {
   const { provData } = props;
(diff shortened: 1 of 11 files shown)


tsliaw:
> do we want to change the button to 'Show Derived Entities' instead of 'Show Descendants'?
Changing the button to "Show Derived Entities" will make more sense to users in my opinion.

> Fetch immediate descendants when detail pane is opened and disable the button if no immediate descendants exist. Should we display something else instead of the disabled button.
 I think the disabled button should be good enough for the cases when no immediate descendants exist

mccalluc:
(Waiting for go-ahead from John before reviewing.)

john-conroy:
Screen Shot 2021-03-02 at 4 16 26 PM

john-conroy:
@mccalluc ready for review. I'm going to begin working on the follow ups mentioned above.

john-conroy:
One possible issue is that a new 'step' has an input which doesn't exist in the provenance, but that input is later revealed by a different node and the link between the two wouldn't be shown. Not sure if this could be an issue down the road.

context/app/static/js/stores/useProvenanceStore.js
```diff
@@ -0,0 +1,44 @@
...
+      descendantSteps.forEach((descendantStep) => {
+        descendantStep.inputs.forEach((input) => {
+          get().steps.forEach((step, stepIndex) => {
+            const outputIndex = step.outputs.findIndex((output) => output.name === input.name);
+            // if input exists as an output of an existing step add new step as target
+            if (outputIndex >= 0) {
+              const output = { step: descendantStep.name, name: input.name };
+              set((state) => {
+                state.steps[stepIndex].outputs[outputIndex].target.push(output);
+              });
```
john-conroy, line 29: I'm thinking I could first loop through the steps outside of this and create a map to get indexes.

context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
```diff
@@ -1,11 +1,30 @@
...
 
 import SectionItem from 'js/components/Detail/SectionItem';
 import { LightBlueLink } from 'js/shared-styles/Links';
+import { AppContext } from 'js/components/Providers';
+import { getImmediateDescendantProv } from 'js/hooks/useImmediateDescendantProv';
+import useProvenanceStore from 'js/stores/useProvenanceStore';
 import ProvVis from '../ProvVis';
 import { StyledTypography, FlexPaper } from './style';
 import '@hms-dbmi-bgm/react-workflow-viz/dist/react-workflow-viz.min.css';
+import ProvData from '../ProvVis/ProvData';
```
mccalluc, line 13: Sorry, I'm fuzzy on the overall directory structure, but up and back down worries me a little... maybe there's a utility that should be factored to a higher directory level?

context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
```diff
@@ -1,11 +1,30 @@
...
+
+function createStepNameSet(steps) {
+  const nameSet = new Set([]);
+  steps.forEach((step) => nameSet.add(step.name));
+  return nameSet;
```
mccalluc, line 18:
```suggestion
  return new Set(steps.map((step)=> step.name));
```

context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
```diff
@@ -1,11 +1,30 @@
...
+}
+
+function removeExistingSteps(steps, newSteps) {
+  const nameSet = createStepNameSet(steps);
+  const uniqueNewSteps = [...newSteps].filter((step) => !nameSet.has(step.name));
```
mccalluc, line 23:
```suggestion
  const uniqueNewSteps = newSteps.filter((step) => !nameSet.has(step.name));
```
I might be missing something?

context/app/static/js/components/Detail/provenance/ProvGraph/ProvGraph.jsx
```diff
@@ -27,39 +46,72 @@ function ProvGraph(props) {
   }
 
   function renderDetailPane(prov) {
```
mccalluc, line 48: I think this is a non-react-ish thing I put in a long time ago: Maybe it gets cleaned up at some point, but not a priority right now? Or maybe it still makes sense... I'm not really seeing the big picture. I'm worried about how deep functions are getting nested.

context/app/static/js/components/Detail/provenance/ProvVis/ProvVis.jsx
```diff
@@ -1,30 +1,39 @@
...
+          showIndirectFiles: true,
+          showParameters: false,
+          showReferenceFiles: true,
+        }}
+        parentItem={{ name: 'Is this used?' }}
+        steps={steps}
+      >
+        <Graph rowSpacingType="compact" minimumHeight={300} renderDetailPane={renderDetailPaneWithNode} />
+      </GraphParser>
+    </>
```
mccalluc, line 37: Wrapper doesn't seem necessary?
```suggestion
      <GraphParser
        parsingOptions={{
          parseBasicIO: false,
          showIndirectFiles: true,
          showParameters: false,
          showReferenceFiles: true,
        }}
        parentItem={{ name: 'Is this used?' }}
        steps={steps}
      >
        <Graph rowSpacingType="compact" minimumHeight={300} renderDetailPane={renderDetailPaneWithNode} />
      </GraphParser>
```

mccalluc (commented):
Things look good: Minor idiom suggestions. 

There's no way for you to know before making the call whether there are descendants to expand, right? I thinking about when I'm on a terminal node, and I click, and nothing happens... How do we communicate that there's nothing to add to the graph, but that the button press was successful?

Unless you've hit situations where the `O(n^3)` has actually been noticeable, I wouldn't invest time there... but if you have, please do!, or at least file an issue so it doesn't get lost.

Running it locally, I get a lot of
```
Warning: Failed prop type: Invalid prop `nodes[8].ioType` of type `boolean` supplied to `Graph`, expected `string`.
    in Graph (created by ProvVis)
    in GraphParser (created by ProvVis)
    in ProvVis (created by ProvGraph)
    in ProvGraph (created by ProvTabs)
    ...
```
and a couple
```
Warning: Failed prop type: The prop `edges` is marked as required in `Graph`, but its value is `undefined`.
    in Graph (created by ProvVis)
    in ProvVis (created by ProvGraph)
    in ProvGraph (created by ProvTabs)
    ...
```
Can that be cleared up, and then merge? Thanks!

john-conroy:
> There's no way for you to know before making the call whether there are descendants to expand, right? I thinking about when I'm on a terminal node, and I click, and nothing happens... How do we communicate that there's nothing to add to the graph, but that the button press was successful?

I believe that should be resolved by #1598.

john-conroy:
> Can that be cleared up, and then merge? Thanks!

All of the prop-types warnings were existing issues and not related to this new work. I can take a look at them, but perhaps in a separate PR.

> Sorry, I'm fuzzy on the overall directory structure, but up and back down worries me a little... maybe there's a utility that should be factored to a higher directory level?

> I think this is a non-react-ish thing I put in a long time ago: Maybe it gets cleaned up at some point, but not a priority right now? Or maybe it still makes sense... I'm not really seeing the big picture. I'm worried about how deep functions are getting nested.

All of this was pre-existing, I can add a general refactoring to the PR to clean up the prop types warnings if that works?

PR #1597, opened 2021-03-02
Mccalluc/missing timestamp

mccalluc:
With the new backend, (https://portal.dev.hubmapconsortium.org/browse/dataset/ae217499cd4a8b73bda077ee2893b399 ... requires login) Bill and Zhou noticed that some pages had react errors. Root problem is that the front end code expects `last_modified_timestamp` to be present.

- @yuanzhou : Perhaps it used to be the case that if an entity hadn't actually be modified, the create timestamp was filled in for the modified? I have a slight preference for retaining the former behavior... but it's a slight preference.
- @john-conroy : In `DetailDescription.jsx` this is already handled, but it's handled differently:
```
  const formattedModificationDate = modifiedTimestamp ? format(modifiedTimestamp, dateFormat) : 'Undefined';
```
Would that be preferable?

### DIFF
--- a/CHANGELOG-last_modified_timestamp.md
+++ b/CHANGELOG-last_modified_timestamp.md
@@ -1,0 +1,1 @@
+- Handle missing last_modified_timestamp.
--- a/context/app/static/js/components/entity-tile/EntityTileBottom/EntityTileBottom.jsx
+++ b/context/app/static/js/components/entity-tile/EntityTileBottom/EntityTileBottom.jsx
@@ -18,7 +18,9 @@ function EntityTileBottom(props) {
           <StyledDivider flexItem orientation="vertical" $invertColors={invertColors} />
         </React.Fragment>
       ))}
-      <Typography variant="body2">Modified {format(entityData.last_modified_timestamp, 'yyyy-MM-dd')}</Typography>
+      {entityData.last_modified_timestamp && (
+        <Typography variant="body2">Modified {format(entityData.last_modified_timestamp, 'yyyy-MM-dd')}</Typography>
+      )}
     </FixedWidthFlex>
   );
 }


yuanzhou:
@mccalluc thanks for finding the root cause and the quick fix! You are right about the `last_modified_timestamp`. In the new entity-api, such properties (`last_modified_timestamp`, `last_modified_user_sub`, `last_modified_user_email`, and `last_modified_user_displayname`) are auto-generated on updating/editing an entity. We don't generate them on entity creation simply to avoid duplicated metadata. 

From the neo4j database migration perspective, all the existing entity nodes that have not never been updated would retain this `last_modified_timestamp` property even if it's the same value as `created_timestamp`. The entity-api can also generate  `last_modified_timestamp` property for new entity creation on the fly, but this would result additional function calls, not a performance issue, just unnecessary. It seems like adding an existence check in portal-ui should take care of this at the minimum cost. @shirey any insights?

Note: to avoid confusion, in search-api, the mapping (https://github.com/hubmapconsortium/search-api/blob/entities-schema/src/elasticsearch/neo4j-to-es-attributes.json#L18) converts `created_timestamp` (defined in entity-api) into `create_timestamp` to be used in ES.

shirey:
@yuanzhou I would say we need to set the last_modified timestamps on CREATE.  This is, after all, the last time the entity was updated.

These timestamps need to be set at CREATE and UPDATE time.  From the ws perspective any time a POST/CREATE happens both the created and modified timestamps are set with the same date-time.  On PUT/UPDATE the modified timestamps are set, but the created timestamps never change again.

Same for last_modified_<userinfo> attributes.

yuanzhou:
@shirey no problem, I'll make the change in entity-api and add back the `last_modified_timestamp` to those nodes that don't have it in neo4j.

@mccalluc I'll let you know once the changes are in place, the portal-ui probably won't need any tweaks from this PR.

mccalluc:
Ok, I'll close this, but if you change course, it wouldn't be hard to add it back.

PR #1599, opened 2021-03-05
Mccalluc/partonomy facets

### LINKED ISSUE #1572: parse partonomy in jsonld at index-time for hierarchical facets
https://cdn.jsdelivr.net/gh/hubmapconsortium/hubmap-ontology@1.0.0/ccf-partonomy.jsonld

This would need to be joined with the info in the rui-json

mccalluc:
There won't be anything to see till https://github.com/hubmapconsortium/search-api/pull/242 is merged and reindex has been completed. I'm going to say this fixes #1572 ... I think the coverage right now is poor enough that we won't actually want to use in in the public-facing search for a while.

Draft, until the index has been updated so this can be exercised.

### DIFF
--- a/CHANGELOG-partonomy-dev-search.md
+++ b/CHANGELOG-partonomy-dev-search.md
@@ -1,0 +1,1 @@
+- Add partonomy facet to dev-search.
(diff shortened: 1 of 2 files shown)


john-conroy (approved):

mccalluc:
Actually, this doesn't need to wait as a draft: When the index is updated, we'll see how it works.

PR #1600, opened 2021-03-05
John conroy/refactor provenance

john-conroy:
Still getting the below prop-types warnings when opening the provenance graph, but the other warnings and code style issues identified in #1593 have been addressed. The code in `ShowDerivedEntitiesButton` isn't new and was taken from `renderDetailPane` in `ProvGraph`. Filed #1602.

> Warning: Failed prop type: The prop `nodes` is marked as required in `Graph`, but its value is `undefined`.

> Warning: Failed prop type: The prop `edges` is marked as required in `Graph`, but its value is `undefined`.

### DIFF
--- a/context/app/static/js/components/Detail/provenance/ProvVis/ProvData.js
+++ b/context/app/static/js/components/Detail/provenance/ProvVis/ProvData.js
@@ -35,7 +35,7 @@ export function makeCwlInput(name, steps, extras, isReference) {
 }
 
 // export only to test.
-export function makeCwlOutput(name, steps, extras) {
+export function makeCwlOutput(name, steps, extras, isReference) {
   const id = name;
   return {
     name,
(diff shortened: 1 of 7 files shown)


context/app/static/js/components/Detail/provenance/ProvVis/ProvData.js
```diff
@@ -46,6 +46,7 @@ export function makeCwlOutput(name, steps, extras) {
     meta: {
       global: true,
       in_path: true,
+      type: isReference ? 'reference file' : 'data file',
```
john-conroy, line 49: Initially we opted to only provide a type for input files with `makeCwlInput`, but it's causing prop-types warnings. Is  it fine to add the same logic to output files or should this be some other value. We need something here.

mccalluc (approved):
Thanks! and I see you've filed #1602... If you can add a bit of the stack trace there, and if you know anything now about why it's hard to fix, that'd be useful.

john-conroy:
Updated #1602.

PR #1611, opened 2021-03-09
upgrade commons

mccalluc:
@ilan-gold - If you could do some manual QA, just to make sure the vitessce-config generation is working as expected, that would be good. Thanks!


ilan-gold (approved):
Works for me!

PR #1634, opened 2021-03-12
John conroy/prettier should fail ci

### LINKED ISSUE #1565: prettier errors should be ci errors
from  #1563:
> Also fixes a prettier error from a code review suggestion from an earlier PR. These keep popping up, should we make prettier fail our ci to catch these earlier?

Yes, I think that would be good: I'd like the pre-commit checks to be a subset of UI, and if I'm understanding correctly, that's not currently the case? Thanks for noting this, @john-conroy .

john-conroy:
Fix #1565. Add linting, which includes prettier, to test script. I think we previously got this from dev-start when it was in the test script, but lost it when it was removed.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -57,16 +57,17 @@ start pytest
 pytest context/app
 end pytest
 
-start lint
 cd context
+
+start lint
 npm run lint
 end lint
 
 start npm-test
 npm run test
-cd -
 end npm-test
 
+cd -
 
 start docker
 ./docker.sh 5001  # Needs to match port in cypress.json.
--- a/test.sh
+++ b/test.sh
@@ -57,9 +57,12 @@ start pytest
 pytest context/app
 end pytest
 
+start lint
+cd context
+npm run lint
+end lint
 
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -57,9 +57,12 @@ start pytest
 pytest context/app
 end pytest
 
+start lint
+cd context
+npm run lint
+end lint
```
mccalluc, line 63:
```suggestion
cd -
end lint
```

mccalluc (commented):
Small preference to have the same state at the end of the block as at the start... Here, it's obvious enough, but if these get bigger, it's easier to miss.

... or, alternatively, pull the `cd` outside of the `start`/`end` block... So you have all the tests that run at one directory level, and then the `cd` to another directory stands out.

Or whatever you think best.

test.sh
john-conroy: Moved the cds outside of the start/end blocks.

mccalluc (approved):

PR #1635, opened 2021-03-12
John conroy/dataset page collections

### LINKED ISSUE #1579: create "Collections" section for dataset pages
Each dataset page should have a section that lists all collections that the dataset is part of. This should include the title of the collection and the DOI.

We can resuse the collection list design from the Collections overview page:

image

john-conroy:
Fix #1579.

@tsliaw We had two slightly different panels, one seen in the collections page where the title is purple and the entire panel is a link and another on the saved lists page where the title is light blue and only the title is a link. We should probably use one style throughout, right? Let me know which you prefer, but for now this PR is using the light blue link version for the collections page, saved lists page and new collections section.

Old style
Screen Shot 2021-03-11 at 9 18 53 PM

Collections Section with new style.

Screen Shot 2021-03-11 at 9 14 50 PM


mccalluc (approved):
Looks good. At some point, I still think I'd like to move more of these API look-ups to index-time, but not soon.

tsliaw (approved):

tsliaw:
Use the light blue link version, I forgot to update the mockup in Figma.

PR #1637, opened 2021-03-12
Make Vitessce Conf Testing More robust

ilan-gold:
If Vitessce configuration fails for any reason at any point, we should be handling it gracefully.

For example: https://portal.test.hubmapconsortium.org/browse/dataset/a598609e8a6c8ba6d0ca52b3e567040d on TEST should work now with this PR

I fixed the issue in #1638 directly by raising an error instead.

More generally, I have wrapped everything in try-except statements to always, at the minimum return at an empty conf.  I have also added more "errro handling" and "error raising" in the configuration-generation so we log out what is going wrong where and why.  Furthermore I added "malformed" tests for each assay in response to #1639 - if an assay is added without one, the test will fail.

I may have gone overboard with the logging.  Let me know.


context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -30,12 +33,15 @@
...
-    SEQFISH_FILE_REGEX
+    SEQFISH_FILE_REGEX,
+    CODEX_TILE_DIR
 )
 from .type_client import CommonsTypeClient
 
 
 class SeqFISHViewConf(ImagingViewConf):
+
+    @return_empty_json_if_error
```
ilan-gold, line 44: This is what I was referring to.  This may be too strong, especially given that the entire process in `client.py` is wrapped in a try-catch.

context/app/api/client.py
```diff
@@ -139,6 +140,14 @@ def get_vitessce_conf(self, entity):
...
+                vc = get_view_config_class_for_data_types(
+                    entity=entity, nexus_token=self.nexus_token
+                )
+                conf = vc.build_vitessce_conf()
+                return conf
+            except Exception:
+                current_app.logger.info(
+                    f'Building vitessce conf threw error: {traceback.format_exc()}'
+                )
+                return {}
```
mccalluc, line 153: - You will need to make sure you're getting the error reports from IEC if we do this... If they can't promise to run a grep every day for two weeks, I don't think we should do this.
- I don't think this should be left in indefinitely: File a followup issue, and let's take it out in the next sprint.

context/app/api/vitessce_confs/assay_confs.py
mccalluc: I actually prefer this to the try-catch at a higher level.

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -46,6 +52,11 @@ def build_vitessce_conf(self):
             ]
         )
         found_images = get_matches(file_paths_found, full_seqfish_reqex)
+        if len(found_images) == 0:
+            message = f'seqFish assay with uuid {self._uuid} has no matching files'
+            if not self._is_mock:
+                current_app.logger.info(message)
+            raise FileNotFoundError(message)
```
mccalluc, line 59:
```suggestion
            raise FileNotFoundError(f'seqFish assay with uuid {self._uuid} has no matching files')
```
Let `return_empty_json_if_error` do the logging

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -83,20 +94,38 @@ def _get_pos_name(self, image_path):
...
 
+    @return_empty_json_if_error
     def build_vitessce_conf(self):
         file_paths_found = [file["rel_path"] for file in self._entity["files"]]
         found_tiles = get_matches(file_paths_found, TILE_REGEX)
+        if len(found_tiles) == 0:
+            message = f'Cytokit SPRM assay with uuid {self._uuid} has no matching tiles'
+            if not self._is_mock:
+                current_app.logger.info(message)
+            raise FileNotFoundError(message)
```
mccalluc, line 112: ditto

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -83,20 +94,38 @@ def _get_pos_name(self, image_path):
...
+                base_name=tile,
+                imaging_path=CODEX_TILE_DIR
             )
-            confs.append(vc.build_vitessce_conf())
+            conf = vc.build_vitessce_conf()
+            if conf == {}:
+                message = f'Cytokit SPRM assay with uuid {self._uuid} has empty view config'
+                if not self._is_mock:
+                    current_app.logger.info(message)
+                raise CytokitSPRMViewConfigError(message)
```
mccalluc, line 127: ditto

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -106,11 +116,18 @@ def __init__(self, entity, nexus_token, is_mock=False):
...
     def build_vitessce_conf(self):
         file_paths_found = [file["rel_path"] for file in self._entity["files"]]
         found_images = get_matches(
             file_paths_found, self.image_pyramid_regex + r".*\.ome\.tiff?$",
         )
+        if len(found_images) == 0:
+            message = f'Image pyramid assay with uuid {self._uuid} has no matching files'
+            if not self._is_mock:
+                current_app.logger.info(message)
+            raise FileNotFoundError(message)
```
mccalluc, line 129: ditto

context/app/api/vitessce_confs/test_confs.py
```diff
@@ -16,32 +16,66 @@
...
     "seqfish": SeqFISHViewConf,
+    "sprm": SPRMViewConf,
 }
 
 
 def test_assays():
     parent_dir = Path(__file__).parent
-    for entity_file in (parent_dir / FIXTURES_INPUT_DIR).glob("*_entity.json"):
+    for entity_file in (parent_dir / FIXTURES_INPUT_DIR).glob(
+        "[!malformed]*_entity.json"
```
mccalluc, line 35: No: This doesn't exclude files beginning with `malformed`, it excludes files beginning with `m` or `a` or `l` ...

mccalluc (changes requested):
I'd suggest separating this into 2 PRs:
- Small PR to fix the reported errors,
- then, separately, a bigger PR to return `{}` even if there is an error... but don't think that should be left in forever. If we had better error reporting, perhaps, but I don't think we can rely on IEC to grep the logs on an on-going basis for us, and having silent failures is worse than noisy failures. (The NIH has signed off on the idea that this is "research-quality" software, so I don't think we should sweep problems under the rug.)

john-conroy (approved):
I agree with Chuck's concern about sweeping errors under the rug if we don't regularly get the logs.

Do any of the existing issues result in a bug in the ui or are these issues contained to the flask app?

ilan-gold:
@john-conroy They should not.  I am going to merge this now, and will be opening a new PR that follows up.

PR #1646, opened 2021-03-16
John conroy/sample specific datasets

### LINKED ISSUE #1628: Implement sample-specific dataset v1.0
Design is in #1584. [Figma](https://www.figma.com/file/awtd3Klx2hka54jyekr9pV/Detail-Pages?node-id=1466%3A1) 

There are some other changes in this mockup to the sample entity page that might not be noticeable at first glance. This includes:

- Move attribution section to the bottom of the entity page
- Remove the tissue section
- The tissue location link is moved to next to the JSON button

john-conroy:
Fix #1628.

Creates a new hook `useSearchData` which will be reusable for different queries. I originally had trouble passing the query to the `useEffect` as an object as a dependency causing it to keep updating so I instead just opted to pass the query as a string. I also moved some reused styles to shared-styles. I think the shared-styles are pretty untidy, but not sure what the best organization is there.

@tsliaw do you want me to center all of the values in the table cells?
@mccalluc Adding you just to take a look at the es query.

Screen Shot 2021-03-17 at 11 10 19 AM

Screen Shot 2021-03-17 at 11 10 29 AM


context/app/static/js/hooks/useSearchData.js
```diff
@@ -0,0 +1,35 @@
...
+          'Content-Type': 'application/json',
+          ...authHeader,
+        },
+      });
+      if (!response.ok) {
+        console.error('Search API failed', response);
+        return;
+      }
+      const results = await response.json();
+      // eslint-disable-next-line no-underscore-dangle
```
mccalluc, line 24:
```suggestion
```

context/app/static/js/hooks/useSearchData.js
```diff
@@ -0,0 +1,35 @@
...
+      const response = await fetch(elasticsearchEndpoint, {
+        method: 'POST',
+        body: query,
+        headers: {
+          'Content-Type': 'application/json',
+          ...authHeader,
+        },
+      });
+      if (!response.ok) {
+        console.error('Search API failed', response);
```
mccalluc, line 20: Might include the request as well? Or that might be clutter... your call.

context/app/static/js/hooks/useSearchData.js
```diff
@@ -0,0 +1,35 @@
...
+        return;
+      }
+      const results = await response.json();
+      // eslint-disable-next-line no-underscore-dangle
+      const resultEntity = results.hits.hits;
+      setSearchData(resultEntity);
```
mccalluc, line 26:
```suggestion
      setSearchHits(results.hits.hits);
```
`data` is really vague, and I think the search hits will be multiple "entities", not just one?

context/app/static/js/pages/Sample/Sample.jsx
```diff
@@ -35,15 +37,48 @@ function SampleDetail(props) {
...
+    }),
+    elasticsearchEndpoint,
+    nexusToken,
+  );
 
   const shouldDisplaySection = {
     protocols: Boolean(protocol_url),
     tissue: true,
     metadata: 'metadata' in assayMetadata,
+    datasets: Boolean(sampleSpecificDatasets.length),
```
mccalluc, line 78:
```suggestion
    datasets: sampleSpecificDatasets.length > 0,
```
More direct?

context/app/static/js/pages/Sample/Sample.jsx
```diff
@@ -35,15 +37,48 @@ function SampleDetail(props) {
...
+      },
+      _source: [
+        'uuid',
+        'display_doi',
+        'mapped_data_types',
+        'descendant_counts',
+        'origin_sample.mapped_organ',
+        'status',
+        'last_modified_timestamp',
+      ],
```
mccalluc, line 68: This is just going to get the first 10 by default. I think we want to set a much larger page size.

mccalluc (commented):
- Some concerns about naming, but use your judgement. 
- ES default is just to return 10 results, I think: I think we decided to set a really high limit, rather than worrying about pagination of results?
- I didn't dig into the history, but I'm guessing you merged in master, and that's why the CHANGELOG files seem to be disappearing in this PR?

Add the `size` parameter (I think that's it), and reconsider a couple names, and it's good. Thanks!

tsliaw:
Everything looks good. Do you think you enable sorting by columns? Feel free to do this in a separate PR though. 

>  do you want me to center all of the values in the table cells?

What do you mean by centering the values?

tsliaw (approved):

john-conroy:
@tsliaw all of the values are left aligned and it's only really noticeable for the derived datasets count column. Switching to centered may make that column look better, but the others worse and would also be inconsistent with the rest of our tables. So I'm happy to keep it the way it is if you are.

john-conroy:
@tsliaw I should probably add sorting for all of our tables, right?

context/app/static/js/pages/Sample/Sample.jsx
john-conroy: Added a size argument to the query, thanks!

tsliaw:
Yes all tables should have column sorting.

Keep the values left-aligned, but either in this PR or in the future, number values according to [Material Design](https://material.io/components/data-tables#specs) should be right-aligned. But feel free to do this in a separate PR since all of our tables currently have numbers left-aligned and they should be uniform throughout the site.

context/app/static/js/pages/Sample/Sample.jsx
```diff
@@ -35,15 +37,49 @@ function SampleDetail(props) {
     metadata,
     rui_location,
   } = assayMetadata;
+  const { elasticsearchEndpoint, nexusToken } = useContext(AppContext);
+
+  const { searchData: sampleSpecificDatasets } = useSearchHits(
+    JSON.stringify({
```
ilan-gold, line 43: Just trying to contribute, is there an advantage to maybe memoizing this first instead of `stringify` on every call?  I think `stringify` can be a little price-y in terms of performance but maybe I am splitting hairs/projecting my own concerns

context/app/static/js/hooks/useSearchHits.js
```diff
@@ -0,0 +1,33 @@
+import { useState, useEffect } from 'react';
+import { getAuthHeader } from 'js/helpers/functions';
+
+function useSearchHits(query, elasticsearchEndpoint, nexusToken) {
+  const [searchData, setSearchData] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
```
ilan-gold, line 6: Is `isLoading` actually used? I think it can be removed.

context/app/static/js/hooks/useSearchHits.js
```diff
@@ -0,0 +1,33 @@
+import { useState, useEffect } from 'react';
+import { getAuthHeader } from 'js/helpers/functions';
+
+function useSearchHits(query, elasticsearchEndpoint, nexusToken) {
```
ilan-gold, line 4: Should we file a follow-up issue to use this elsewhere in the codebase?  I don't think this is the PR to do this in but this seems like something we might do often (could be wrong though)?

ilan-gold (approved):

context/app/static/js/pages/Sample/Sample.jsx
ilan-gold: Also memo-izing can also be expensive so maybe not...

context/app/static/js/hooks/useSearchHits.js
john-conroy: I'll likely use this when working on the layout shift for the table of contents sidebar.

context/app/static/js/hooks/useSearchHits.js
john-conroy: I think we have one place we could use this now, thanks I'll file an issue.

context/app/static/js/pages/Sample/Sample.jsx
john-conroy: I ended up just using `useMemo` in `Sample` and pushing `stringify` inside the hook.

PR #1652, opened 2021-03-18
John conroy/start prov with donor entity node

### LINKED ISSUE #1609: Don't display donor registration node
The first node in the graph should be `Donor`: Neither the fake root node, nor the donor registration activity are useful.
Screen Shot 2021-03-12 at 9 16 49 AM

john-conroy:
Fix #1609.

@tsliaw It seems that the first node will always be light purple. Is that fine or should we make it light green to match the other entity nodes?

Screen Shot 2021-03-18 at 1 43 51 PM

### DIFF
--- a/CHANGELOG-start-prov-with-donor-entity-node.md
+++ b/CHANGELOG-start-prov-with-donor-entity-node.md
@@ -1,0 +1,1 @@
+- Update provenance graph to not display donor registration node.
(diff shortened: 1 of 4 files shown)


tsliaw (approved):
Yea make the donor entity match the light-green of the other entity nodes. Otherwise looks good.

ilan-gold:
@john-conroy It seems that http://localhost:5001/browse/sample/ebd4eff27f9bfc8505836816689dc14b on TEST doesn't quite work as you have described above, but I might be missing something:

Screen Shot 2021-03-18 at 4 10 23 PM

Can you reproduce?

john-conroy:
Nice catch, the activity names vary across environments right now.

ilan-gold (approved):

john-conroy:
Per conversation with @tsliaw, we're going to keep node colors as is until we have a new design in place.

PR #1659, opened 2021-03-19
Change container name to prod-portal-ui

yuanzhou:
Issue: https://portal-prod.stage.hubmapconsortium.org/ uses the PROD search-api configuration, but sometimes this domain returns results from the actual STAGE search-api. 

Fix: in case the gateway nginx confuses `portal-ui` with `portal-ui-prod` hostnames due to name-based matching rule when forwarding the domain to the container, I'm changing the hostname to be totally unique to nginx.

Will make the corresponding changes in gateway nginx configuration.

### DIFF
--- a/compose/hubmap.stage.yml
+++ b/compose/hubmap.stage.yml
@@ -20,9 +20,9 @@ services:
     networks:
       - gateway_hubmap
 
-  portal-ui-prod:
-    hostname: portal-ui-prod
-    container_name: portal-ui-prod
+  prod-portal-ui:
+    hostname: prod-portal-ui
+    container_name: prod-portal-ui
     image: hubmap/portal-ui:latest
 
     # Mount the app config in container to keep it outside the image.
--- a/compose/hubmap.stage.yml
+++ b/compose/hubmap.stage.yml
@@ -20,6 +20,9 @@ services:
     networks:
       - gateway_hubmap
 
+  # Service names need to be kept in sync with gateway:
+  # https://github.com/hubmapconsortium/gateway/blob/f2c2d5d/nginx/conf.d-stage/prod-portal-ui.conf
+  # Some part of the routing seems to be be based on left substrings, rather than names, so make sure left substrings are unique.
   prod-portal-ui:
     hostname: prod-portal-ui
     container_name: prod-portal-ui
--- a/CHANGELOG-container-names.md
+++ b/CHANGELOG-container-names.md
@@ -1,0 +1,1 @@
+- Tweak container names to be in sync with gateway.


mccalluc (commented):
> name-based matching rule when forwarding the domain to the container

This seems strange: Do you mean that the rule is based on left substring matches? That is different from name based matching, and it would better to fix it at the root. If we tripped on it once, we might trip on it again.

If that's hard, could you add a comment that links to the code that does this mapping, for future reference?

yuanzhou:
The fix in gateway: https://github.com/hubmapconsortium/gateway/pull/75/files

what I suspected is when the domain name portal-prod.stage gets router to nginx, nginx picks up the “portal-ui.conf” as the server configuration file instead of “portal-ui-prod.conf”. I had something similar came up with another project a year ago. By renaming, nginx would match the right server configuration file. 

On the docker end, everything is correct, each container uses the correct configuration. This also led me to think the root cause is nginx forwarding rather than two containers based on the same docker image. 

The changes in portal-ui is to make the naming consistent.

compose/hubmap.stage.yml
```diff
@@ -20,9 +20,9 @@ services:
     networks:
       - gateway_hubmap
 
-  portal-ui-prod:
-    hostname: portal-ui-prod
-    container_name: portal-ui-prod
+  prod-portal-ui:
```
mccalluc, line 23:
```suggestion
  # Service names need to be kept in sync with gateway:
  # https://github.com/hubmapconsortium/gateway/blob/f2c2d5d/nginx/conf.d-stage/prod-portal-ui.conf
  # Some part of the routing seems to be be based on left substrings, rather than names, so make sure left substrings are unique.
  prod-portal-ui:
```

mccalluc:
Thank you for the explanation. I've added a note, and will wait for tests to pass, then merge.

I still think it is really strange for nginx to be splitting names on dashes and matching the resulting left substrings: I can't imagine this is the behavior out of the box. Any chance there's a regex in gateway that should be getting the whole hostname, and instead is just `\w+` or something like that? In your court.

yuanzhou:
To be honest with you, I feel this is strange too. And I remember we had this same issue before and you added more startup logging via PR https://github.com/hubmapconsortium/portal-ui/pull/1453 to verify each container uses the right conf file. 

At one point, I thought about if it's DNS related. Both `portal.stage.hubmapconsortium.org` and ` portal-prod.stage.hubmapconsortium.org` are CNAME of `ingest.stage.hubmapconsortium.org`. And `ingest.stage.hubmapconsortium.org` is and A record for the IP of STAGE VM. It's possbile when one of the portal domains get looked up, the host coming back might be `ingest.stage.hubmapconsortium.org`. But in our case, they seem to be just swapped. And if I remember correctly, a restart of the gateway may (not always) resolve this, but not a long term fix.

> Any chance there's a regex in gateway that should be getting the whole hostname, and instead is just \w+ or something like that?

I don't think so. Because the gateway listens on the same port for these two domains, actually several other APIs too. 

- https://github.com/hubmapconsortium/gateway/blob/test-release/nginx/conf.d-stage/portal-ui.conf
- https://github.com/hubmapconsortium/gateway/blob/test-release/nginx/conf.d-stage/portal-ui-prod.conf

But only the above two get swapped. An experienced server admin at PSC once told me the naming of the conf files does matter, they even use prefix with a number to ensure the lexical order so the conf files get looked at in the correct order. This is another reason I wanted to try with the renaming but I didn't prefix with numbers because all other conf files on the same VM served by the same nginx don't have such issue.

Let's give it a try, and I'll restart the two containers a few times to verify if these changes in portal-ui and gateway would be able to solve this issue.

mccalluc (changes requested):
No harm in trying -- good luck getting to the root of this!

mccalluc (approved):
(Sorry, meant to approve.)

PR #1664, opened 2021-03-23
"previous" should be spelled out

mccalluc:
I don't know how long this example on dev will stay around, but: https://portal.dev.hubmapconsortium.org/browse/dataset/8f528475e216c283851e6941841d7173.json
```
"previous_revision_uuid": "8f528475e216c283851e6941841d7173"
```

Can you confirm that this doubly-linked list structure works with your designs, and then put the implementation issues in the v0.15 milestone? Thanks!

(Related back-end issue: https://github.com/hubmapconsortium/search-api/issues/233)

### DIFF
--- a/CHANGELOG-prev.md
+++ b/CHANGELOG-prev.md
@@ -1,0 +1,1 @@
+- For versioning, "previous" will be spelled out in the API response. Fix dev search.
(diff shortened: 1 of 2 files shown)


tsliaw:
This should work for v1.0 of the versioning mockups.

tsliaw (approved):

PR #1667, opened 2021-03-23
John conroy/fix donor prov

john-conroy:
For now we are reverting changes to remove donor registration node in provenance graphs on donor entity nodes. We'll keep the changes in effect for other entity types. I'll create an issue and add an item to the UI agenda in regards to thinking about the path forward. I probably didn't need to pass the `entity_type` down through all of the components and could have checked if `Object.keys(this.prov.activity).length === 1` instead, but this seems more sure. 
 
Donor

Screen Shot 2021-03-23 at 1 45 44 PM

Sample

Screen Shot 2021-03-23 at 1 45 52 PM

Also, merges changes missed from an error with `push.sh`.

### DIFF
--- a/context/app/static/js/components/Detail/provenance/ProvVis/ProvData.js
+++ b/context/app/static/js/components/Detail/provenance/ProvVis/ProvData.js
@@ -125,10 +125,12 @@ export default class ProvData {
 
   toCwl() {
     return Object.entries(this.prov.activity).reduce((acc, [activityId, activity]) => {
-      // activity names currently vary across environments
-      // TODO reduce to single value once apis are consistent
       if (
+        // activity names currently vary across environments
+        // TODO reduce to single value once apis are consistent
         !['Register Donor Activity', 'Create Donor Activity'].includes(activity['hubmap:creation_action']) ||
+        // donors only have a single step which can't be removed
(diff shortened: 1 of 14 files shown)


tsliaw (approved):

context/app/static/js/components/Detail/provenance/ProvVis/ProvData.js
```diff
@@ -126,7 +127,10 @@ export default class ProvData {
     return Object.entries(this.prov.activity).reduce((acc, [activityId, activity]) => {
       // activity names currently vary across environments
       // TODO reduce to single value once apis are consistent
-      if (!['Register Donor Activity', 'Create Donor Activity'].includes(activity['hubmap:creation_action'])) {
+      if (
+        !['Register Donor Activity', 'Create Donor Activity'].includes(activity['hubmap:creation_action']) ||
+        this.entity_type === 'Donor'
```
mccalluc, line 132: Can you move the comment above inside the `if`, and then add a new comment about `this.entity_type === 'Donor'` ... Right now, it reads as if the comment applies to everything.

mccalluc (commented):
One small tweak to the comments, and good to merge! I think what we decided was a UI meeting agenda item to figure out the next steps? Or whatever you think best.

PR #1669, opened 2021-03-24
Upgrade Vitessce + New Sequencing Visualization

### LINKED ISSUE #1644: scRNA-seq (10x Genomics) [Salmon] should have vis?
Joel believes this should have a visualization. Requires login: https://portal.test.hubmapconsortium.org/browse/dataset/99d1d2779907fed926f32413cae814af 

(I think this is one that used to error before #1640.)

ilan-gold:
This PR does a few things
1. Upgrade `Vitessce` to 1.1.7 to pick up features like faster image loading and gene
2. Upgrade `vitessce-python` to use new arguments for `AnnDataWrapper` that are needed for this to work properly
3. Add new rna-seq conf class for zarr-backed AnnData (and corresponding test)

For testing on TEST:
http://localhost:5001/browse/dataset/3b066aa664e1a32eaea957927baf92e2

Fixes  #1644


mccalluc:
(I'll take a look at this right now.)

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -196,6 +243,11 @@ def get_view_config_class_for_data_types(entity, nexus_token):
         return ImagePyramidViewConf(
             entity=entity, nexus_token=nexus_token)
     if "rna" in hints:
+        dag_names = [dag['name']
+                     for dag in entity['metadata']['dag_provenance_list'] if 'name' in dag]
```
mccalluc, line 247: `'name' in dag` is a tight enough check? Other things couldn't accidentally include that string? Contrived examples: "enamel" or "ornament".

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -83,7 +83,7 @@ def _build_assets_url(self, rel_path):
             assets_endpoint = MOCK_URL
         base_url = urllib.parse.urljoin(assets_endpoint, f"{self._uuid}/{rel_path}")
         token_param = urllib.parse.urlencode({"token": self._nexus_token})
-        return base_url + "?" + token_param
+        return base_url + "?" + token_param if use_token else base_url
```
mccalluc, line 86:
```suggestion
        return f'{base_url}?{token_param}' if use_token else base_url
```

mccalluc (commented):
Might tighten up one string match, but that might be paranoid: you know better than I what the strings are like. Thanks!

context/app/api/vitessce_confs/assay_confs.py
ilan-gold: Not sure I follow here - this is just to prevent calling dag['name'] when the dag object doesn't have a 'name' attribute (it seems to be optional).

PR #1685, opened 2021-03-26
Prevent Unnecessary Auth for Published Zarr Stores

ilan-gold:
For published datasets, the `assets` API should not need to authenticate requests (and in fact when people are not signed in, cannot).  Authenticating requests via a header (needed for zarr stores) causes pre-flight `OPTIONS` requests to be sent out: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests (You can see this by opening the network tab on a zarr-backed AnnData RNA-seq dataset).

This PR takes away the header for published datasets as it is unnecessary, which should give us a nice performance bump.

This also fixes an issue where an empty auth header returns a 401 regardless of whether or not the dataset is published (which can happen for people who are not signed in but datasets which are "Published").  I don' t think it's a bug with `assets`, but it does cause this dataset to not work when published.  This PR also (inadvertently) resolves that.

Example:
https://portal.test.hubmapconsortium.org/browse/dataset/3b066aa664e1a32eaea957927baf92e2

### DIFF
--- a/CHANGELOG-prevent-cors-preflight.md
+++ b/CHANGELOG-prevent-cors-preflight.md
@@ -1,0 +1,1 @@
+- Unnecessary auth headers for published zarr stores cause unecessary CORS preflight requests.
--- a/context/app/api/vitessce_confs/assay_confs.py
+++ b/context/app/api/vitessce_confs/assay_confs.py
@@ -219,7 +219,6 @@ class NullConf():
 
 
 def get_view_config_class_for_data_types(entity, nexus_token):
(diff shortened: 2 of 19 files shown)


john-conroy (approved):
I should probably use similar logic for our js requests?

ilan-gold:
@john-conroy I just checked and yes, you would see a performance benefit as well - I can see the `OPTIONS` request for UI elements that require the IEC infra like the provenance, search etc.

PR #1687, opened 2021-03-29
John conroy/assay type bar chart

john-conroy:
Still need to add the legend, tooltip, and drop down to swap organ type for donor sex. Style and dimensions will be more exact once we have a home page to fit the chart into, so a focus on code would be great.

Screen Shot 2021-03-31 at 12 56 09 PM


context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
```diff
@@ -0,0 +1,152 @@
...
+import { useChartPalette } from './hooks';
+
+const organTypesQuery = {
+  size: 0,
+  aggs: {
+    organ_types: { terms: { field: 'origin_sample.mapped_organ.keyword', size: 10000 } },
+  },
+};
+
+const assayTypesQuery = {
```
mccalluc, line 20: My preference would be to combine these and make fewer, larger ES queries... and down the road, the same query might get the facets we need for pseudo-search.

But not necessary right now, and I might be optimizing for the wrong thing... I guess I worry a bit about making lots of simultaneous requests, but if this makes the code simpler, might be worth it.

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
```diff
@@ -0,0 +1,152 @@
...
+  const { searchData: assayTypesData } = useSearchData(assayTypesQuery, elasticsearchEndpoint, nexusToken);
+
+  const { formattedData, maxSumDocCount } = useMemo(() => {
+    if (Object.keys(assayTypesData).length > 0) {
+      const f = addSumProperty(formatAssayData(assayTypesData));
+      sortBySumAscending(f);
+      const m = Math.max(...f.map((d) => d.sum));
+      return { formattedData: f, maxSumDocCount: m };
+    }
+    return { formattedData: [], maxSumDocCount: 0 };
```
mccalluc, line 68: Prefer to avoid `if`... Should be able to say the sum of an empty set is 0?

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
```diff
@@ -0,0 +1,152 @@
...
+          y={getDataType}
+          xScale={docCountScale}
+          yScale={dataTypeScale}
+          color={colorScale}
+        >
+          {(barStacks) =>
+            barStacks.map((barStack) =>
+              barStack.bars.map(
+                (bar) =>
+                  bar.width > 0 && (
```
mccalluc, line 109: ... on the same theme, what goes wrong if there are zero-with bars?

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
```diff
@@ -0,0 +1,152 @@
...
+                  ),
+              ),
+            )
+          }
+        </BarStackHorizontal>
+        <AxisLeft
+          hideTicks
+          scale={dataTypeScale}
+          stroke="black"
+          numTicks={33}
```
mccalluc, line 127: Where does this number come from?

mccalluc (commented):
(I started this when I was thinking it might be good to rush it out the door... but that was not the right call... so take these as my first impressions, but don't worry about responding to them now.)

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
john-conroy: Nothing goes wrong with the bar chart itself, but we get a bunch of warnings.

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
john-conroy: Thanks, need to fix that to get the length of the data array.

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
```diff
@@ -0,0 +1,152 @@
...
+  const { searchData: organTypesData } = useSearchData(organTypesQuery, elasticsearchEndpoint, nexusToken);
+
+  const buckets = organTypesData?.aggregations?.organ_types?.buckets || [];
+  const organTypes = buckets.map((b) => b.key);
+
+  const { searchData: assayTypesData } = useSearchData(assayTypesQuery, elasticsearchEndpoint, nexusToken);
+
+  const { formattedData, maxSumDocCount } = useMemo(() => {
+    if (Object.keys(assayTypesData).length > 0) {
+      const f = addSumProperty(formatAssayData(assayTypesData));
```
ilan-gold, line 63: Strong preference for spelling out variables to at least some degree, at least in situations like this.

context/app/static/js/components/home-revision/AssayTypeBarChart/utils.js
```diff
@@ -0,0 +1,35 @@
+import produce from 'immer';
+/* eslint-disable no-param-reassign */
```
ilan-gold, line 2: Is this only for the sort call below?

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
```diff
@@ -0,0 +1,152 @@
...
+            },
+          },
+        ],
+        size: 10000,
+      },
+    },
+  },
+};
+
+const getDataType = (d) => d.mapped_data_type;
```
ilan-gold, line 47: Can we declare this within the function?  Just feels a little out of place here since it is only used once.

ilan-gold (commented):
How do I go about testing this?  When I open the homepage, nothing seems changed.

ilan-gold:
Nevermind, figured it out :) Looks good

context/app/static/js/components/home-revision/AssayTypeBarChart/hooks.js
```diff
@@ -0,0 +1,13 @@
+import { useTheme } from '@material-ui/core/styles';
+
+function useChartPalette() {
+  const {
+    palette: { success, primary, secondary, error, warning, info },
+  } = useTheme();
+
+  const colorObjects = [success, primary, secondary, error, warning, info];
+
+  return [...colorObjects.map((c) => c.light), ...colorObjects.map((c) => c.main), ...colorObjects.map((c) => c.dark)];
```
ilan-gold, line 10: Just to be clear, the order of this doesn't really matter?  This is just a way of getting some colors that look nice and contrast in an order that ensures contrasting ones are closer together?

john-conroy: The order does matter, Tiffany wants to prioritize light then main and finally dark.

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
john-conroy: Sounds good, thanks.

ilan-gold:
Is this supposed to be resizable?

context/app/static/js/components/home-revision/AssayTypeBarChart/utils.js
john-conroy: This is for the immer portion which includes all of the reassignments of the draft param.

context/app/static/js/components/home-revision/AssayTypeBarChart/AssayTypeBarChart.jsx
john-conroy: Sounds good, I couldn't reuse `formattedData` or `maxSumDocCount` so I just opted for something quick. I'll think of different variable names.

context/app/static/js/components/home-revision/AssayTypeBarChart/hooks.js
ilan-gold: Ok yes, that is what I meant.  perfect.

john-conroy: Maybe a comment makes sense here?

ilan-gold: That could work.  At first glance, it does seem a bit random

context/app/static/js/components/home-revision/AssayTypeBarChart/utils.js
ilan-gold: Ah, right, ok, didn't catch that one.

PR #1691, opened 2021-03-31
John conroy/tools menu in header

### LINKED ISSUE #1689: Implement new {Tools,Atlas,...} menu
Nils:
> ... Unless the response from Katy and Rahul is controversial, just go ahead ...

Rahul:
> my preference is for Tools (or 'Mapping Tools') instead of Atlas. Thanks!

Katy:
> The ASCT+B tables are not a tool but the begin of an Atlas CCF reference system.
> What about “Atlas & Tools”?

Screen Shot 2021-03-30 at 11 22 17 AM

Replaces #1630 

Azimuth is: https://azimuth.hubmapconsortium.org/

john-conroy:
Fix #1689. @mccalluc I made Azimuth an outbound link, if I remember correctly we don't want the ccf links to be treated as outbound links?

Screen Shot 2021-03-31 at 10 41 45 AM

### DIFF
--- a/CHANGELOG-tools-menu-in-header.md
+++ b/CHANGELOG-tools-menu-in-header.md
@@ -1,0 +1,1 @@
+- Replace ccf dropdown in header with 'Atlas & Tools' and add link to Azimuth.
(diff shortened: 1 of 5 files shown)


mccalluc (commented):
A little confused -- I would vote for not having a special outside-link treatment for Azimuth: It's not in the mockup Rahul looked at and approved, and I think we want to make things seem well integrated (even if they aren't) -- but happy to be overridden on this.

john-conroy:
Do we want these links tracked as outbound links for google analytics? Should i make a similar component for tracking, but doesn't include `target="target="_blank" rel="noopener noreferrer"?

tsliaw (approved):

PR #1692, opened 2021-03-31
Louder python errors

mccalluc:
- @ilan-gold : Can you confirm that all of these are in fact errors? Or if they aren't serious, or the false positive rate is too high, perhaps not needed?
- @shirey : Is there log monitoring in place? I think we would want to know if these were being hit... Can your log monitor aggregate errors and send them to us once a day, or something like that?

### DIFF
--- a/context/app/api/vitessce_confs/base_confs.py
+++ b/context/app/api/vitessce_confs/base_confs.py
@@ -26,7 +26,7 @@ def return_empty_json_if_error(func):
         except Exception:
             class_obj = args[0]
             if not class_obj._is_mock:
-                current_app.logger.info(
+                current_app.logger.error(
                     f'Building vitessce conf threw error: {traceback.format_exc()}'
                 )
             return {}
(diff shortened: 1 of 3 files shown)


context/app/api/vitessce_confs/base_confs.py
```diff
@@ -237,7 +237,7 @@ def build_vitessce_conf(self):
                 if path not in file_paths_found:
                     message = f'SPRM file {path} with uuid "{self._uuid}" not found as expected.'
                     if not self._is_mock:
-                        current_app.logger.info(message)
```
ilan-gold, line 240: Could get rid of this logger call all together since the error is thrown anyway and the decorator handles it

ilan-gold (approved):

shirey:
@mccalluc we can add log monitoring  for the portal via AWS Cloudwatch as we do with other services, but I think to do that we may need to explicitly write portal logs to a file. Right now I think the only way to get to the portal logs is via "docker log", which I don't think we can monitor.

Looking for an agent to monitor Docker logs and not finding anything.  Docker logs do produce a file, but the location/name changes I think.

mccalluc:
Filed #1707 to get a view on the logs. I think this PR is good to merge.

Issue #1331, opened 2020-10-29
Implement versioning UI


mccalluc:
I understand that test now how a small number of items with versions: Talking with Bill, I believe the extra attributes are `next`, `previous`, and `latest`. Two subtasks:
- [ ] In search results, hide items which are not the latest.
- [ ] On the details page, implement UI for navigating between versions. Design is #1222.

mccalluc:
Blocked. Removing from milestone.

mccalluc:
(There is now at least one example, use dev-search on test or dev.)

mccalluc:
Bill provided information about the draft API:

> The new method to retrieve all revisions is on Entity API TEST at `https://entity-api.test.hubmapconsortium.org/datasets/<id>/revisions`
>
> The docs won’t be published in SmartAPI until we release, you can look at the direct yaml spec for the method here. It returns an array of all datasets in a revision chain given any dataset in the chain.
>
> The method has an optional query parameter `…/revisions?include_dataset=true` that will return the full dataset info for each dataset when set to true--  defaults to false if not included.
> - [entity-api-spec.yaml](https://github.com/hubmapconsortium/entity-api/blob/d0a967513338018edc1cefc3d449985755cbfe4f/entity-api-spec.yaml)

PR #1698, opened 2021-04-01
Mccalluc/warning if not latest version

mccalluc:
Towards #1331. They are not providing a `latest` version, only `next`: I've put a question on the agenda: "Is it important to be able to link to the latest version? Or is next sufficient for now?"

- @tsliaw : Close enough to the mock up?
- @john-conroy
  - for the bottom margin, would it be better to have my own derived component just for this?
  - ... and maybe find a different approach to the flex, as well?
Screen Shot 2021-04-01 at 3 01 15 PM

### DIFF
--- a/context/app/static/js/pages/Dataset/Dataset.jsx
+++ b/context/app/static/js/pages/Dataset/Dataset.jsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useContext } from 'react';
 import Typography from '@material-ui/core/Typography';
 
+import { Alert } from 'js/shared-styles/alerts';
 import { LightBlueLink } from 'js/shared-styles/Links';
 import Files from 'js/components/files/Files';
 import ProvSection from 'js/components/Detail/provenance/ProvSection';
@@ -67,6 +68,7 @@ function DatasetDetail(props) {
     status,
     mapped_data_access_level,
   } = assayMetadata;
+  const isLatest = !('next_revision_uuid' in assayMetadata);
 
   const { elasticsearchEndpoint, nexusToken } = useContext(AppContext);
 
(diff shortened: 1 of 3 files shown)


tsliaw (approved):
This looks fine for now, given we are only given 'next_uuid'

context/app/static/js/pages/Dataset/Dataset.jsx
```diff
@@ -96,6 +98,18 @@ function DatasetDetail(props) {
   // TODO: When all environments are clean, data_types array fallbacks shouldn't be needed.
   return (
     <DetailContext.Provider value={{ display_doi, uuid, mapped_data_access_level }}>
+      {!isLatest && (
+        <Alert severity="warning">
+          <span>
```
john-conroy, line 103: Do you need the span?

context/app/static/js/shared-styles/alerts/index.js
```diff
@@ -23,6 +23,7 @@ const StyledAlert = styled(OutlinedAlert)`
   :not(svg) {
     color: ${(props) => props.theme.palette.text.primary};
   }
+  margin-bottom: 16px;
```
john-conroy, line 26: I don't really like default margins unless it's based on a prop.

john-conroy (commented):
> ... and maybe find a different approach to the flex, as well?
 
Which flex?

context/app/static/js/pages/Dataset/Dataset.jsx
mccalluc: Without the span, this gets broken into multiple lines. There is almost certainly a better way to do this!

context/app/static/js/shared-styles/alerts/index.js
mccalluc: ok, let me do that. Thanks for the feedback.

mccalluc:
What we have is fine for search / will use a new API when rendering the details page to show link to latest.
@shirey - update here when the API is available.

mccalluc:
@john-conroy : I'm going to interpret "comment" as "ok to merge with changes", but please file an issue if anything seems really wrong ... This will need to be followed up when the data we need is available.

john-conroy:
I don't think you re-requested review, but looks good.

PR #1699, opened 2021-04-01
John conroy/assay type bar chart legend

john-conroy:
Screen Shot 2021-04-01 at 3 36 17 PM


context/app/static/js/components/home-revision/AssayTypeBarChartContainer/style.js
```diff
@@ -0,0 +1,16 @@
+import styled from 'styled-components';
+
+const Flex = styled.div`
+  display: flex;
+`;
+
+const ChartWrapper = styled.div`
+  width: 1000px;
+  height: 800px;
+`;
```
ilan-gold, line 10: I guess this is not a comment about this PR specifically, but is there no resizing here?  It seems like the legend does react a bit to resizing.

john-conroy: The sizing here is just temporary until I fit it into the home page.

PR #1700, opened 2021-04-01
Mccalluc/exclude older versions from search

mccalluc:
Towards #1331

(This is a bit of machinery we could also use to hide pyramids after vis-lifting, #1690.)

### DIFF
--- a/CHANGELOG-exclude-older-datasets.md
+++ b/CHANGELOG-exclude-older-datasets.md
@@ -1,0 +1,1 @@
+- Exclude Datasets which have been superseded from search results.
--- a/context/app/static/js/components/Search/SearchWrapper.jsx
+++ b/context/app/static/js/components/Search/SearchWrapper.jsx
@@ -27,6 +27,7 @@ function SearchWrapper(props) {
     type,
     isLoggedIn,
     isDevSearch,
+    defaultQuery,
     resultsComponent: ResultsComponent,
   } = props;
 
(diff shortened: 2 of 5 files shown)


mccalluc:
I can replicate test failures locally... In cypress, it's not seeing the tutorial... Oh: I bet there are mock responses set for the cypress test, and since this changes the ES query, it's not matching....

mccalluc:
... hmm, no, not seeing any overly tight matches: It's just
```
        cy.intercept('POST', 'https://search*.hubmapconsortium.org/portal/search', {
          statusCode: 200,
          body: defaultSearch,
        });
```
and I can see in the Cypress console that the pattern match is satisfied.

In this branch, it sees the panel to begin the tutorial, and clicks, but then the tutorial never appears. It still works outside of Cypress, and if I switch to the master branch that also makes Cypress happy.

@john-conroy : Can you think of anything fragile about the way the tutorial is launched that would explain this? Right now, it's not clear to me why Cypress sees what it sees, nor is it clear why this code change would even bother Cypress.

john-conroy (approved):
Updated the cypress fixture data fixes it. As we discussed, maybe adding a default query modifies what `search-kit` expects in a response.

Do we need to do anything for dev-search or keep it as is?

mccalluc:
@john-conroy : 
> Do we need to do anything for dev-search or keep it as is?

I'd say keep it as is for now... If there's something weird going on, dev-search will get everything and we don't have to worry about any magic hiding some records. Might revisit in the future.

PR #1708, opened 2021-04-05
John conroy/bar chart color scale dropdown

john-conroy:
Adds the dropdown and addresses a few of the suggestions from the main pr. Filtering data to remove processed datasets in a future PR.

Screen Shot 2021-04-05 at 3 44 14 PM

Screen Shot 2021-04-05 at 3 44 06 PM


context/app/static/js/components/home-revision/AssayTypeBarChartContainer/hooks.js
```diff
@@ -0,0 +1,25 @@
...
+import { getAssayTypeBarChartData } from './utils';
+
+function useChartPalette() {
+  const {
+    palette: { success, primary, secondary, error, warning, info },
+  } = useTheme();
+
+  const colorObjects = [success, primary, secondary, error, warning, info];
+
+  return [...colorObjects.map((c) => c.light), ...colorObjects.map((c) => c.main), ...colorObjects.map((c) => c.dark)];
```
mccalluc, line 13: As I understand it, you need a palette for the bars, and since we've got a bunch of colors with different semantics for other parts of the site, they can be re-used here? I don't really care much myself... but my sense is that these uses are a little bit different: We don't actually want to draw attention to any particular color in the chart, but that is the point of some of these colors.

mccalluc (commented):
Some little comments: I think it could be merged as-is... but I also think my comments are pretty superficial... Let me know if there's anything particular to focus on?

context/app/static/js/components/home-revision/AssayTypeBarChartContainer/hooks.js
john-conroy: @tsliaw, thoughts?

john-conroy: As we discussed, we're not immediately concerned with everything pointed out by @mccalluc, but we could expect someone else to be concerned.

PR #1710, opened 2021-04-05
John conroy/filter out processed datasets

john-conroy:
Do we want to check on something more specific than the presence of a single bracket? Wanted to avoid regex, but maybe we should be more restrictive.

Screen Shot 2021-04-05 at 4 38 11 PM
Screen Shot 2021-04-05 at 4 38 04 PM

### DIFF
--- a/context/app/static/js/components/home-revision/AssayTypeBarChartContainer/utils.js
+++ b/context/app/static/js/components/home-revision/AssayTypeBarChartContainer/utils.js
@@ -4,7 +4,9 @@ import produce from 'immer';
 function formatAssayData(assayData, colorKey) {
   const formattedData = assayData.aggregations.mapped_data_types.buckets.reduce((acc, d) => {
     const snakeCaseDataType = d.key.mapped_data_type.replace(/ /g, '_');
-    // TODO get datasets to display from index
+    // TODO: Get datasets to display from index, instead of depending on patterns in names.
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/home-revision/AssayTypeBarChartContainer/utils.js
```diff
@@ -3,8 +3,12 @@ import produce from 'immer';
 
 function formatAssayData(assayData, colorKey) {
   const formattedData = assayData.aggregations.mapped_data_types.buckets.reduce((acc, d) => {
+    const snakeCaseDataType = d.key.mapped_data_type.replace(/ /g, '_');
+    // TODO get datasets to display from index
```
mccalluc, line 7:
```suggestion
    // TODO: Get datasets to display from index, instead of depending on patterns in names.
    // The first step is having a hierarchy for assay types:
    // https://github.com/hubmapconsortium/portal-ui/issues/1688
```

I like to reference the blocking issue in TODOs. It forces me to check that there is, in fact, an issue, and if someone is looking at this code in the future, it will give them the bigger picture so they can carry on the work.

mccalluc (commented):
Add a bit more context to the comment. For now, I'm fine with just looking for `[`... we know it's not what we want long term... but we also don't want to hold up the new home page on this.

PR #1711, opened 2021-04-05
Update client-side and server-side 401 messages

### LINKED ISSUE #1607: HuBMAP members with multiple Globus accounts might get confusing error pages when logging in
Users who have multiple Globus accounts (I ended up with a second one when testing something) who should have member access but are logged into their non-HuBMAP Globus account do get an error page with no indication about what is wrong. 

The error page should tell the user which globus account their are logged into an offer the option to log out of that account and log back in with the correct account for the HuBMAP portal. At a minimum, the error page should include a link to http://app.globus.org/.

This is also related to #1575.

image

mccalluc:
Fix #1607

I don't think the client-side error message is being shown right now... but understanding why not is a separate, low priority issue, and I think it makes sense to just keep the messages in sync for now.

### DIFF
--- a/CHANGELOG-friendlier-401.md
+++ b/CHANGELOG-friendlier-401.md
@@ -1,0 +1,1 @@
+- Help users who may be logged into the wrong globus account.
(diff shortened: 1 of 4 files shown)


john-conroy (changes requested):
I think this should produce warnings since a `<ul>` doesn't belong inside of a `<p>` (Typography component).

john-conroy (approved):

PR #1712, opened 2021-04-06
Remove references to "everything"

mccalluc:
@yuanzhou , can you post an update here when the index changes have made it to production? No hurry, I just want to make sure that hand-off happens. Keeping it as a draft until then.

@ilan-gold / @tsliaw / @john-conroy : While we can search against the new `all_text` field, we can't use it to highlight results: ES can't tell us what is in `all_text`, only that there was a match. For now, I'm just specifying the `description` field as the place to look when highlighting. If it's not there, there won't be highlighting. For example:

Screen Shot 2021-04-05 at 8 33 57 PM

(This is from an instance pointing at the DEV index, which has the new structure.)

Thinking about why there could be a hit, but no highlighting:
- `all_text` might be grabbing too much: Looking at the example here, I remember that we're excluding some parts of the document from `everything`, and that should be replicated with `all_text`. (I think ancestors and descendants should be excluded, but I'll need to check.)
- The match might be in a facet field, but just showing the value of the facet in bold doesn't really give us much: I think it would be fine to skip the highlighting in these cases.
- There might be other fields besides `description` where showing a match in context is useful.

After this is merged (probably with some tweaks) there will be a follow-up PR in `search-api` to get rid of `everything`

### DIFF
--- a/context/app/static/js/components/Search/DevResults/DevResults.jsx
+++ b/context/app/static/js/components/Search/DevResults/DevResults.jsx
@@ -29,7 +29,7 @@ function DevResults(props) {
       ]}
       sourceFilter={resultFieldIds}
       customHighlight={{
-        fields: { everything: { type: 'plain' } },
+        fields: { description: { type: 'plain' } },
       }}
     />
   );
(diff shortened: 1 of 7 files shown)


yuanzhou:
We are testing some other changes along with the mapping config on TEST. The full reindex went well today. Will let you know once they are released to PROD.

yuanzhou:
All the recently index changes in search-api (metadata.metadata, yaml config ...) have been released to PROD. I also did a full reindex and everything went well.

mccalluc:
replaced by #1809

PR #1726, opened 2021-04-09
Fix files header z-index

### LINKED ISSUE #1705: file browser has z-level bug
Row content on top of header instead of below header.

From: https://portal.hubmapconsortium.org/browse/dataset/f2647d9533956fdb12da6b1fc6254441

image

mccalluc:
Fix  #1705

### DIFF
--- a/CHANGELOG-file-z-index.md
+++ b/CHANGELOG-file-z-index.md
@@ -1,0 +1,1 @@
+- Fix the z-index of the Files header.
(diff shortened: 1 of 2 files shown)


john-conroy (changes requested):
I think we started to maintain z-index values in the theme, so probably best to put it there.

john-conroy (changes requested):
We would want the entity header to be above the files header when you scroll down.

mccalluc:
(Sorry for not getting it right no the first pass / Thanks for the feedback!)

mccalluc:
@john-conroy - Ping!

john-conroy (approved):

PR #1747, opened 2021-04-19
John conroy/home revision lower grid

john-conroy:
I remove the above the fold styles and modify spacing in #1766. So if you want to check out the changes locally, I'd do so from that branch.

The images here are small and don't necessitate the same treatment as the larger carousel images (or so I think).
 
Screen Shot 2021-04-22 at 2 29 19 PM

@tsliaw the hover for the external links reveals the white background for some of the images, do we want to use images without backgrounds?


context/app/static/js/components/home-revision/ExternalLinks/ExternalLinks.jsx
```diff
@@ -0,0 +1,49 @@
...
+        linkText="HuBMAP Publications"
+        description="Explore the list of related publications generated by members of the HuBMAP Consortium on their Google Scholar page."
+      />
+      <ExternalLink
+        img={<img src={ProtocolsIoLogo} height={56} width={100} alt="Protocols.io Logo" />}
+        linkHref="https://www.protocols.io/workspaces/human-biomolecular-atlas-program-hubmap-method-development/"
+        linkText="HuBMAP Protocols"
+        description="View the HuBMAP Consortium’s protocols in protocols.io."
+      />
+
```
mccalluc, line 38:
```suggestion
```

mccalluc (approved):
When/if I work with this code, I'd like understand what `GridArea` is for, and what `GridAreaContainer` is about, but it probably is more clear in context.

tsliaw (approved):

john-conroy:
GridArea is just a wrapper for a component which I use to set the `grid-area` used in conjunction with `grid-template-areas` with the grid itself. I opted to use wrappers and styling them there than the components themselves because it pertains more to the page's styles than the component's styles.

https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-areas

PR #1748, opened 2021-04-20
SPRM Stitched

ilan-gold:
Currently, nothing displays for cfa86388b3aacc0806a7f0524d3b942a because the image pyramid is not generated.  You can delete that part of the config though and something should display eventually - I have an update to the pipeline to decrease the number of sides on the polygon so that should help with load/render times:

```python
@return_empty_json_if_error
    def build_vitessce_conf(self):
        vc = VitessceConfig(name=self._base_name)
        dataset = vc.add_dataset(name="SPRM")
        zarr_path = f"anndata-zarr/{self._base_name}-anndata.zarr"
        adata_url = self._build_assets_url(zarr_path, use_token=False)
        anndata_wrapper = AnnDataWrapper(
            adata_url=adata_url,
            spatial_centroid_obsm="xy",
            spatial_polygon_obsm="poly",
            cell_set_obs=[
                "K-Means [Covariance] Expression",
                "K-Means [Mean-All-SubRegions] Expression",
                "K-Means [Mean] Expression",
                "K-Means [Shape-Vectors]",
                "K-Means [Texture]",
                "K-Means [Total] Expression",
            ],
            expression_matrix="X",
            factors_obs=[
                "K-Means [Covariance] Expression",
                "K-Means [Mean-All-SubRegions] Expression",
                "K-Means [Mean] Expression",
                "K-Means [Shape-Vectors]",
                "K-Means [Texture]",
                "K-Means [Total] Expression",
            ],
            request_init=self._get_request_init(),
        )
        dataset = dataset.add_object(anndata_wrapper)
        vc = self._setup_view_config_raster_cellsets_expression_segmentation(
            vc, dataset
        )
        return vc.to_dict()
```


context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -102,7 +105,7 @@ class CytokitSPRMConf(ViewConf):
     @return_empty_json_if_error
     def build_vitessce_conf(self):
         file_paths_found = [file["rel_path"] for file in self._entity["files"]]
-        found_tiles = get_matches(file_paths_found, TILE_REGEX)
+        found_tiles = get_matches(file_paths_found, f"{TILE_REGEX}|{STITCHED_REGEX}")
```
mccalluc, line 108:
```suggestion
        found_tiles = get_matches(file_paths_found, TILE_REGEX) or get_matches(file_paths_found, STITCHED_REGEX) 
```
My guess is that we don't expect to see match both `TILE_REGEX` and `STITCHED_REGEX`... but if we did, would it be better to just return one or the other? On the principle that we should try to be strict about what we return, to make the job easier downstream.

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -161,6 +164,35 @@ def __init__(self, entity, nexus_token, is_mock=False):
...
+class StitchedCytokitSPRMConf(ViewConf):
+    @return_empty_json_if_error
+    def build_vitessce_conf(self):
+        file_paths_found = [file["rel_path"] for file in self._entity["files"]]
+        found_regions = get_matches(file_paths_found, STITCHED_REGEX)
+        if len(found_regions) == 0:
+            message = (
+                f"Cytokit SPRM assay with uuid {self._uuid} has no matching regions"
+            )
+            raise FileNotFoundError(message)
```
mccalluc, line 176:
```suggestion
            raise FileNotFoundError(
                f"Cytokit SPRM assay with uuid {self._uuid} has no matching regions; "
                f"No file matches for '{STITCHED_REGEX}'."
            )
```

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -161,6 +164,35 @@ def __init__(self, entity, nexus_token, is_mock=False):
...
+                is_mock=self._is_mock,
+                base_name=region,
+                imaging_path=STITCHED_IMAGE_DIR,
+            )
+            conf = vc.build_vitessce_conf()
+            if conf == {}:
+                message = (
+                    f"Cytokit SPRM assay with uuid {self._uuid} has empty view config"
+                )
+                raise CytokitSPRMViewConfigError(message)
```
mccalluc, line 191:
```suggestion
                raise CytokitSPRMViewConfigError(
                    f"Cytokit SPRM assay with uuid {self._uuid} has empty view config for region '{region}'"
                )
```

mccalluc (changes requested):
I started suggesting changes where you define the message in the line before it's used... I don't see the advantage of this style, but I also don't think it's worrying about -- but still good to return some more information, particularly from the error inside a loop.

It might be safer not to combine the regexes with `|`... but it would only make a different in an error condition. Your call.

I'd like to avoid changes to package-lock.json, unless there are changes to package.json.

Could you add testing for the cases you're adding here? There's a lot of really detailed work, and I'd like us to have a very clear sense of what input we expect, and what output config we generated.

Particularly to give yourself time on the testing, I'd suggest not trying to get this in this afternoon, but I'll ask on slack to remind myself if anyone is waiting for this.

mccalluc (approved):
Thanks!

PR #1753, opened 2021-04-20
Mccalluc/more helpful 404

### LINKED ISSUE #522: More helpful 404 if UUID isn't 32 characters
mccalluc 
> I think there’s an issue to rephrase all the error messages I’ve written, so this wouldn’t be the last word, but perhaps: “404: Not Found. Check that URL is correct: UUIDs should be 32 characters.”

Nils
> Ok. And you would show the last part only if the UUID is incomplete?
(Or otherwise malformed)
I think that would be a good idea

mccalluc:
fix #522

(Resurfaced this because there was a link on slack this morning with a bad UUID.)

### DIFF
--- a/context/app/static/js/components/error/ErrorBody/ErrorBody.jsx
+++ b/context/app/static/js/components/error/ErrorBody/ErrorBody.jsx
@@ -55,7 +55,7 @@ function ErrorBody({ errorCode, urlPath, isAuthenticated, isGlobus401, isMainten
   }
 
   if (errorCode === 404) {
-    if (urlPath && urlPath.match(/^\/browse\//)) {
+    if (urlPath && urlPath.startsWith('/browse/')) {
       const uuid = urlPath.split('/').pop();
       if (uuid.length !== 32) {
         return (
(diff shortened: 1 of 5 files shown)


context/app/static/js/components/error/ErrorBody/ErrorBody.jsx
```diff
@@ -55,6 +55,17 @@ function ErrorBody({ errorCode, isAuthenticated, isGlobus401, isMaintenancePage
   }
 
   if (errorCode === 404) {
+    if (urlPath && urlPath.match(/^\/browse\//)) {
```
john-conroy, line 58: Could use `startsWith`, but maybe regex is better?

john-conroy (approved):

context/app/static/js/components/error/ErrorBody/ErrorBody.jsx
mccalluc: I'll go with `startsWith`: much easier to read, Thanks for the reminder!

PR #1763, opened 2021-04-22
Mccalluc/support support

### LINKED ISSUE #1576: vis-lifting: AF / PAS / IMS
Proposal:
- (1) When rendering the parents of AF / PAS / IMS pyramids, include a vitessce component, pointing at the pyramid files.
  - (1a) Also on these datasets, offer a second Files section which will allow download of pyramid files (either through the directory widget or globus?) ... without confusing the user about which are the real files, and which are the derived files. (@tsliaw -- I think this will need design.)
- (2) After vitessce is included, but not necessarily after the Files tweaks have been made, change the datatype of AF / PAS / IMS pyramids at index-time to something other than `Dataset` ... perhaps `Support`?
- (3) If someone does land on a `Support` entity, perhaps from an old URL, we can still show the page, but include a banner at the top suggesting the parent `Dataset` will be more useful. (@tsliaw - A little more design)

@ngehlenborg : This isn't quite what we discussed earlier, but I think introducing a new type will be cleaner / simpler than introducing a new facet, and then trying to make it a search default. After these first three types are covered, we can look at what remains, and tackle them the same way. This plan does not require any development work from IEC, just run-of-the-mill reindexing, and I think that's a feature, not a bug. One thing to confirm: will it be ok if (2) comes some time after (1)? It shouldn't be a big job... but if you want everything to happen simultaneously, that becomes harder to choreograph.

@tsliaw : If this seems good to Nils, feel free to file a separate issue for the design work, if it helps you manage everything.

### LINKED ISSUE #1762: Support "Support"
With https://github.com/hubmapconsortium/search-api/pull/267 and a reindex done on TEST, we now have `Support` entities that we can see in dev-search... but if you try to visit, it's a 404. Needs to be fixed.

mccalluc:
Make sure the new "Support" pages work. 
- Prov-grid is a little weird, with a new group for the "Support" entity. 
- <strike>Currently, no visualization, because `files` is missing in the JSON... this seems like an index problem? https://github.com/hubmapconsortium/search-api/issues/303 </strike> Bad test data: This one works: http://localhost:5001/browse/support/f9ae931b8b49252f150d7f8bf1d2d13f
- Also have proptype errors (#1764) as well as a failure from the Entities API (#1765) -- I don't think these are related to the details of this PR.

Fix #1762 (bug I just filed)
Fix #1576 (vis-lifting high-level)
Follow up to https://github.com/hubmapconsortium/search-api/pull/267

### DIFF
--- a/CHANGELOG-support-support.md
+++ b/CHANGELOG-support-support.md
@@ -1,0 +1,1 @@
+- UI support for new Support entity type.
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -102,7 +102,7 @@ class ApiClient():
         return entity
 
     def get_vitessce_conf(self, entity):
-        # First, try "vis-lifting": Display image pyramids on their parent entity pages, if possible.
+        # First, try "vis-lifting": Display image pyramids on their parent entity pages.
         if 'descendants' in entity \
                 and len(entity['descendants']) \
                 and 'data_types' in entity['descendants'][0] \
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -102,6 +102,7 @@ class ApiClient():
         return entity
 
     def get_vitessce_conf(self, entity):
+        # First, try "vis-lifting": Display image pyramids on their parent entity pages, if possible.
         if 'descendants' in entity \
                 and len(entity['descendants']) \
                 and 'data_types' in entity['descendants'][0] \
@@ -114,6 +115,8 @@ class ApiClient():
             # there will be a period of backward compatibility to allow us to migrate.
             derived_entity['files'] = derived_entity['metadata']['files']
             return self.get_vitessce_conf(derived_entity)
+
+        # Otherwise, just try to visualize the data for the entity itself:
         if 'files' not in entity or 'data_types' not in entity:
             return {}
         if self.is_mock:
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -10,7 +10,7 @@ import frontmatter
 from .api.client import ApiClient
 
 
-entity_types = ['donor', 'sample', 'dataset', 'collection']
+entity_types = ['donor', 'sample', 'dataset', 'support', 'collection']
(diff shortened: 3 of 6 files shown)


mccalluc:
I would lean towards getting this in... but pushing back is reasonable: Maybe we should understand the indexing problem https://github.com/hubmapconsortium/search-api/issues/303 first, and make sure we still see visualizations on Support pages.

ilan-gold:
>     * Currently, no visualization, because `files` is missing in the JSON... this seems like an index problem? [hubmapconsortium/search-api#303](https://github.com/hubmapconsortium/search-api/issues/303)

I am able to see a viz (on TEST): http://localhost:5001/browse/support/37b178deabc0512bd38170bb415b097d

Am I misunderstanding?

ilan-gold:
>     * Prov-grid is a little weird, with a new group for the "Support" entity.

What do you mean here? This makes sense to me since this is basically a dataset so there is another tab for it (marked as dataset).  Are you saying there should be a "Support" type for the graph?

mccalluc:
> > ```
> > * Prov-grid is a little weird, with a new group for the "Support" entity.
> > ```
> 
> What do you mean here? This makes sense to me since this is basically a dataset so there is another tab for it (marked as dataset). Are you saying there should be a "Support" type for the graph?

Sorry: A picture helps: Note the "Support" section in the lower left:

Screen Shot 2021-04-22 at 3 51 25 PM

I think this is fine, since people won't be going to these pages in the future, and it's not worth new dev effort.

ilan-gold (approved):

mccalluc:
Poor communication, but the corresponding ES changes were just released to production, so pages like https://portal.hubmapconsortium.org/browse/support/ca224d0c1a0144240858490e3df46f83 are 404. Reverting to an earlier index state is unfortunately not something IEC offers. I've just filed #1775 to do the audit John requested, but right now, this just needs to be merged and deployed. I'm sorry this was bungled.

PR #1767, opened 2021-04-22
Set random initial carousel image index

john-conroy:
Nils wants the initial carousel slide to be random.


context/app/static/js/components/home-revision/ImageCarouselContainer/ImageCarouselContainer.jsx
```diff
@@ -5,16 +5,19 @@ import ImageCarouselControlButtons from '../ImageCarouselControlButtons';
 import ImageCarouselCallToAction from '../ImageCarouselCallToAction';
 import { Flex, CallToActionWrapper } from './style';
 
+const numImages = 3;
```
mccalluc, line 8: Can this be connected to the logic that controls the wrap around when you go past the last slide? Want to avoid having the same value repeated in multiple locations in the code.

mccalluc (changes requested):

context/app/static/js/components/home-revision/ImageCarouselContainer/ImageCarouselContainer.jsx
john-conroy: I think this is the only place it is defined. Maybe I'm misunderstanding?

john-conroy: We pass this value to the control buttons as a prop.

mccalluc (approved):
Just merged my PR into this. Thanks for the feedback: All the slide info is in one array now.

PR #1768, opened 2021-04-22
Move image list up a level so the count does not need to be stored separately

mccalluc:


john-conroy (approved):
Do we want to move up the text in the call to action component as well?

PR #1769, opened 2021-04-22
Add Visualization Processing Docs

ilan-gold:
I was getting a few questions and after sitting in on a meeting, figured this would be a good thing to document, at least at a high level.  Feel free to move this somewhere else, make edits, or call for a separate repo.  It cuts across different parts so this may not be the right place.  This is the flow chart (it's probably somewhat idealized but this is how I would at least like to see the process):
portal-imaging-visualization-flowchart


mccalluc (changes requested):
Cool! I was going to ask if this was draw.io, and it looks like it is... I think they have an SVG format, which includes extra draw.io information, so you can make roundtrips with just one file, and not worry about exporting a PNG. 

(I like draw.io, but haven't used it recently because because dealing with files is just bit harder than just using Google Docs... but I think using draw.io is really the better answer in some ways: Really your call.)

We should make sure this is referenced from the main README, and clarify who the audience is -- What is for TMCs, what is for Devs?

The flowchart could be simplified with some rearrangement -- What if you did the "is spot data" check first, with just one diamond instead of three?

Formatting... for the longer blocks of text, left align... but maybe the details need to be parsed out?

ilan-gold:
I like the suggestion about the formatting/flow. Beyond that, I think the target audience is ourselves for answering the oft asked "what is going on with this dataset and how much work do we need to do to visualize it?" question that we get from other people at IEC like Peter, Joel, Matt etc.

ilan-gold:
I think the questions are meant for guiding/simplifying any discussion with external people (these are the main things that people ask and on some level are the only things that really matter to us), and the answers are for us to know what to do.

mccalluc:
Re-reading this, I think it could be flattened out... particularly with all the text, it might be good to try to keep it as text... maybe with collapsing details?

<details><summary>Has it validated?</summary>
... details about validation ...
</details>

<details><summary>Is there spot data?</summary>
... details ... THE END
</details>

<details><summary>Cytokit + SPRM?</summary>
... details ... THE END
</details>

<details><summary>Just SPRM?</summary>
... details ... THE END
</details>

<details><summary>For everything else...</summary>
... details ... THE END
</details>

ilan-gold:
@mccalluc That works for me.  I am a bit worried about future pipeline development complicating that structure (since once you add in spot data, each of the three final options in your dropdowns will likely have special paths for spot data).  I'll check in the current state of the svg (which is cleaned up) but will display the information as a dropdown in the README.md.  ThanksQ

ilan-gold:
I added the dropdowns and an updated process graph (for future safe-keeping).  I also added information about the sequencing pipelines as things stand (although those are somewhat simpler because there is really only RNA-seq visualization).

visualization-docs/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+<details><summary>Has the data been validated via ingest-validation-tools and confirmed to be viewable using Avivator (which loads data almost identically to what is in the portal)?</summary>
+To this end, we should ask the TMC to follow the instructions below for viewing their data in Avivator to make sure it looks right (should only need to be done for a single representative file): https://github.com/hms-dbmi/viv/tree/master/tutorial
```
mccalluc, line 10:
```suggestion
If so, we should ask the TMC to follow the instructions below for viewing their data in Avivator to make sure it looks right (should only need to be done for a single representative file): https://github.com/hms-dbmi/viv/tree/master/tutorial
```

visualization-docs/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+If there is a z or t stack to the data, ensure that each "stack" is uploaded as a single file.
+
+If it is valid in these three senses (viewable in Avivator locally, passes `ingest-validation-tools`, and "stacks" are uploaded as single files), then ingestion may be done and pipeline processing may proceed.
+
+</details>
+
+<details><summary>Is there spot data?</summary>
+If the answer is "yes," we should run the image pyramid pipeline + offsets on the appropriate imaging data.  We currently do not have a pipeline for visualizing spot data.  Create a new class that inherits from ViewConf to visualize the data (raw imaging + spot data) when such a pipeline is created.  If there is segmentation data coming from the TMC or elsewhere, then that will need to be both processed (via <a href="https://github.com/hubmapconsortium/portal-containers/tree/master/containers/sprm-to-anndata">sprm-to-anndata.cwl from portal-containers</a> or a different pipeline that ideally outputs zarr-backed AnnData) and visualized as well
```
mccalluc, line 21:
```suggestion
If the answer is "yes," we should run the image pyramid pipeline + offsets on the appropriate imaging data.  We currently do not have a pipeline for visualizing spot data.  Create a new class that inherits from ViewConf to visualize the data (raw imaging + spot data) when such a pipeline is created.  If there is segmentation data coming from the TMC or elsewhere, then that will need to be both processed (via <a href="https://github.com/hubmapconsortium/portal-containers/tree/master/containers/sprm-to-anndata">sprm-to-anndata.cwl from portal-containers</a> or a different pipeline that ideally outputs zarr-backed AnnData) and visualized as well.
```

visualization-docs/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+In the above instructions they should only need to a) run the bioformats2raw-raw2ometiff pipeline and then b) drag-and-drop or select the input file using the "CHOOSE A FILE" button on avivator.gehlenborglab.org. There is no need for a web server.
+
+If there is a z or t stack to the data, ensure that each "stack" is uploaded as a single file.
+
+If it is valid in these three senses (viewable in Avivator locally, passes `ingest-validation-tools`, and "stacks" are uploaded as single files), then ingestion may be done and pipeline processing may proceed.
+
+</details>
+
+<details><summary>Is there spot data?</summary>
```
mccalluc, line 20: Will "spot data" be clear to readers? Do you mean "point data", ie, properties are associated a particular coordinate?

mccalluc (approved):
Thank you again. I'm wondering if "spot data" would be better described as "point data"... but you might be describing something different? Your call.

mccalluc:
On formatting: You can use markdown inside `<details>`, but there needs to be one blank line between what it should parse as HTML, and what it should parse as MD.

visualization-docs/README.md
ilan-gold: I think "spot" data is a term used for FISH assay outputs (that I have heard).  I'll add some clarification like "(for example is this a FISH assay with resolved locations of probes that you want to show)"

PR #1771, opened 2021-04-23
Relabel Collections "contributors"

### LINKED ISSUE #1755: On collections, change `Creator` to `Contributor`
Higher level issue: #1723

> Jonathan believes it should be "Contributor", rather than "Creator" which the Collections pages currently have.

Not necessary to wait for API field-name change

mccalluc:
Fix #1755. Not sure whether there are mock-ups which should be changed to keep things in sync?


tsliaw (approved):
Looks good. Will change mockups to match.

PR #1784, opened 2021-04-27
Mccalluc/support tiles icons

### LINKED ISSUE #1778: "Support" tiles are missing icon
If you navigate to a Support entity, and look at the tiles, it is missing its icon:
Screen Shot 2021-04-27 at 10 48 04 AM

Suggest just using the same icon as for datasets.

mccalluc:
Fix #1778

Screen Shot 2021-04-27 at 3 21 28 PM

### DIFF
--- a/CHANGELOG-support-tiles-icons.md
+++ b/CHANGELOG-support-tiles-icons.md
@@ -1,0 +1,1 @@
+- Show an icon in the tile for Support entities.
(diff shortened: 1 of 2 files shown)


john-conroy (approved):
Works for me, but maybe we want a distinct icon support?

mccalluc:
I'm ok with using the same icon:
- We won't be publicizing Support page
- Support is really just a kind of Dataset.

but feedback still welcome.

PR #1787, opened 2021-04-27
John conroy/home revision search

### LINKED ISSUE #1642: Fake free text search for homepage
looks like a free text search, but it's actually just looking at facet values. So, if I type "male", it returns 
> **Donors:**
> _male_
> **Samples:**
> _male_
> **Datasets:**
> _male_

And then I click on the one of interest.

Challenge: Our ES does not have left substring indexing, which is normally a requirement for type-ahead search. options:
- Wait until IEC provides it. We'll probably be waiting a long time.
- Store left substrings, instead of just indexing: This will blow up the index, and is really not the right thing to do.
- Fake it: Home page search was a hardcoded list of type-ahead terms: it doesn't even hit the search API.

Question: Are we supposed to support the same behavior on the search pages?

### LINKED ISSUE #1666: Collect facet values for pseudo-search
I think the best way will be to issue an elastic search query from the homepage to collect facet values. The lists aren't so long, so it's probably fine to just scan the lists linearly for left matches, rather than constructing a left-substring in-memory index. Question: should the query be sent on page load? Click in search field? Keypress?

Other possibilities: 
- Bill thinks ES mappings are feasible... but some time in the summer would be the soonest. Research PR: #1665.
- Script that devs would run to generate code... Kludgier, and it couldn't show unpublished docs when logged in.

john-conroy:
Fix #1642. Fix #1666.

Opted for `includes` instead of `startsWith` since users will likely want to search for keywords in assay types as shown in the snare example below. I'd like to do more testing, but that can come later.

Still having issues with keyboard accessibility, but I think we want to get this in now.

Screen Shot 2021-05-05 at 11 52 52 AM
Screen Shot 2021-05-05 at 11 52 29 AM
Screen Shot 2021-05-06 at 11 01 30 AM


mccalluc:
@john-conroy -- A couple react errors which you might already have a fix for:
```
Warning: Each child in a list should have a unique "key" prop.
Check the render method of `EntityCounts`
```
```
Warning: Failed prop type: The prop `id` is marked as required in `DropdownListbox`, but its value is `undefined`
```
```
Warning: Each child in a list should have a unique "key" prop.
Check the render method of `FacetSearchMenu`.
```

mccalluc:
We're missing the "Get Started" buttons on the slideshow. Don't worry about them for this PR, just trying to list the things that still need to be done.

context/app/static/js/components/home-revision/FacetSearch/FacetSearch.jsx
```diff
@@ -0,0 +1,134 @@
...
+      // order of objects in array matters, we want dataset matches to appear first.
+      setMatches(
+        [
+          { entityType: 'Dataset', aggs: datasetAggsData },
+          { entityType: 'Sample', aggs: sampleAggsData },
+          { entityType: 'Donor', aggs: donorAggsData },
+        ].reduce((acc, entityData) => {
+          const matchingTerms = getMatchingTerms(entityData.aggs, searchTerm);
+          if (Object.keys(matchingTerms).length > 0) {
+            const tempAcc = acc;
```
mccalluc, line 85: This isn't cloning `acc`, so there will still be side-effects, I think.

context/app/static/js/components/home-revision/FacetSearch/style.js
```diff
@@ -0,0 +1,33 @@
...
+const inputPadding = 8;
+
+const StyledInput = styled.input`
+  flex-grow: 1;
+  appearance: none !important;
+  outline: none;
+  border-radius: 8px;
+  border: 0;
+  padding: ${inputPadding}px;
+  font-size: 16px;
```
mccalluc, line 29: I'm not clear on when we use `rem` and when we use `px`.

tsliaw (approved):

context/app/static/js/components/home-revision/FacetSearchEntityListItems/FacetSearchEntityListItems.jsx
```diff
@@ -0,0 +1,41 @@
...
+    <li>
+      {Object.entries(matches).map(([k, v]) => {
+        return (
+          <>
+            <FacetLabel color="primary" variant="subtitle2">{`${labels[k]} (${entityType}s)`}</FacetLabel>
+            <StyledList>
+              {v.map((m) => {
+                const facetQueryParam =
+                  k === 'mapped_data_access_level'
+                    ? `mapped_status-mapped_data_access_level[0][0]=Published&mapped_status-mapped_data_access_level[1][0]=${encodeURIComponent(
```
mccalluc, line 16: What's the motivation for this section? Is it to make sure we only show published entities when you're logged in? The dashed param names are there for the hierarchical facets on search... but I'd like to avoid baking in dependencies on those details.

context/app/static/js/components/home-revision/FacetSearchMenu/FacetSearchMenu.jsx
```diff
@@ -0,0 +1,48 @@
...
+          {!isLoading && matchesExist && (
+            <List>
+              {Object.entries(matches).map(([k, v]) => (
+                <FacetSearchEntityListItems entityType={k} matches={v} labels={labels} />
+              ))}
+            </List>
+          )}
+          {!isLoading && !matchesExist && (
+            <StyledAlert severity="warning">No results found. Check your spelling or try a different term.</StyledAlert>
+          )}
```
mccalluc, line 41: Consider just doing the `isloading` check once, and then a `matchesExist ? :` ternary.

mccalluc (commented):
Some comments, some questions. I'd suggest fixing the react warnings, and leaving the rest for the future. (Perhaps file them as individual tiny issues, if there's a risk of them getting lost.)

I'd prefer not to do the release late in the afternoon.

context/app/static/js/components/home-revision/FacetSearchEntityListItems/FacetSearchEntityListItems.jsx
john-conroy: I added this to ensure what we link to mirrors what is accessible from the facet UI on the search page.

context/app/static/js/components/home-revision/FacetSearch/FacetSearch.jsx
```diff
@@ -0,0 +1,134 @@
...
+    elasticsearchEndpoint,
+    nexusToken,
+  );
+  const { searchData: datasetAggsData, isLoading: datasetAggsDataIsLoading } = useSearchData(
+    datasetAggsQuery,
+    elasticsearchEndpoint,
+    nexusToken,
+  );
+
+  const isLoading = donorAggsDataIsLoading && sampleAggsDataIsLoading && datasetAggsDataIsLoading;
```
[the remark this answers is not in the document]
john-conroy: Yep, that's right. Sorry overlooked this.

context/app/static/js/components/home-revision/FacetSearchMenu/FacetSearchMenu.jsx
john-conroy: I avoided using a ternary because I thought you'd prefer to avoid them. I'll make the change.

context/app/static/js/components/home-revision/FacetSearch/style.js
john-conroy: This was an oversight on my part. `rem` should be used for `font-size`.

context/app/static/js/components/home-revision/FacetSearchEntityListItems/FacetSearchEntityListItems.jsx
john-conroy: This is so the corresponding checkboxes/inputs are selected and you can reverse the selection from the UI. Otherwise we be sending them to a page which can only be altered by the url query params.

mccalluc: Oh, I get it: I wasn't thinking off the publication status as something the homepage search would cover. Thanks for being thorough... but I'm going to file a follow-up to suggest we pull this out: It's not unlikely that that facet will change, and things that need to be modified in tandem across a codebase make me nervous.

PR #1809, opened 2021-05-04
Mccalluc/everything to all text

mccalluc:
This builds on [earlier work in `search-api`](https://github.com/hubmapconsortium/search-api/pull/283/files#diff-e855724d0ae950616ef6c6ea7a3923e655591f3dbf74e5b8bb5a2d8d5a9b9b4dR15) -- We now have field, `all_text`, which is indexed but not stored. We can use it as the target for search, but to get highlighting, he also need to specify one or more additional stored fields. 

This PR does that, and in `config.js` specifies that  we are only interested in highlighting matches from the `description` field. One consequence of this is some searches will not produce highlighted matches. There are a couple ways of looking at this:
- When the match is against a short field, just showing me the value isn't a big gain. ie, if I search for "liver" and the only match is the organ, it's not doing much for the user just to print out an additional line with "liver" in bold for each search result.
- If there are additional fields we would like to search, we can add them to `fieldsToHighlight`... but [we want to minimize the number of fields we search](https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-search-speed.html#search-as-few-fields-as-possible).
- Or, alternatively, if we really want to be able to highlight on every hit, we could make `all_text` a stored field. We'd lose the space savings in the index that were an original motivation, but it would still simplify the the index-time python, so it's not an unreasonable approach.

I'm happy with this PR as-is... but you might not be, and that would be good feedback.

### DIFF
--- a/CHANGELOG-use-all-text.md
+++ b/CHANGELOG-use-all-text.md
@@ -1,0 +1,1 @@
+- Stop using `everything`: Instead, use `all_text`, and highlight matches from `description`.
--- a/context/app/static/js/components/Search/DevResults/DevResults.jsx
+++ b/context/app/static/js/components/Search/DevResults/DevResults.jsx
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { ViewSwitcherHits } from 'searchkit';
 
+import { customHighlight } from 'js/components/Search/config';
+
 import ResultsTable from '../ResultsTable';
 import ResultsCCF from '../ResultsCCF';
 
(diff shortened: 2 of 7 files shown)


mccalluc:
(Duplicate of #1712 ... probably go with the more recent one.)

tsliaw (approved):

john-conroy (approved):
I think it adds room for question. Users will be able see matches, but won't be able to immediately understand why they are matches since there is no highlight. Also, it seems like the majority of matches will not have a highlight. That being said, I don't have a strong preference. If you think the space savings are worth the cost, this looks good to me.

PR #1817, opened 2021-05-06
More cells api demo

### LINKED ISSUE #1774: Fill in Colab examples for Cells API
Related to #1751 and #1750 -- Start filling in the Flask API with the [examples from the notebook](https://colab.research.google.com/drive/19dLSxd5rg_C7nOisIafaxr5A9PH5GjEm).

mccalluc:
Will fix #1774

@tsliaw , Can you confirm that the forms and and the flask endpoints have the right parameters for the questions you'll want to support?

cc @SFD5311

### DIFF
--- a/context/app/routes_cells.py
+++ b/context/app/routes_cells.py
@@ -64,8 +64,8 @@ def cell_counts_for_datasets():
     return response
 
 
-@blueprint.route('/cells/cell-expressions-in-dataset.json', methods=['POST'])
-def cell_expressions_in_dataset():
+@blueprint.route('/cells/cell-expression-in-dataset.json', methods=['POST'])
+def cell_expression_in_dataset():
     # For a single dataset we want to get the expression level of a given gene for all cells.
     # (In our discussion, we started by thinking about the set of matching cells, and then showing expression levels for the two groups, but that’s not needed.)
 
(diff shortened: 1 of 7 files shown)


mccalluc:
Add a sentence which explains the query

tsliaw (approved):

context/app/routes_cells.py
```diff
@@ -15,11 +15,76 @@ def cells_ui():
...
+def cell_counts_for_datasets():
+    # Select a set of datasets where cells express a given gene, and for each dataset,
+    # give the number of matching cells, and the total number of cells.
+
+    uuids = request.args.getlist('uuid')
+    gene_name = request.args.get('gene_name')
+    min_gene_expression = request.args.get('min_gene_expression')
+
     client = Client(current_app.config['CELLS_API_ENDPOINT'])
-    datasets = client.select_datasets().get_list()[0:10]
```
SFD5311, line 21: Once the client changes are made, we should be able to do something like:
datasets = client.select_datasets(
        where='gene',
        has=[f'{gene_name} > {min_gene_expression}'],
        genomic_modality='rna', min_cell_percentage = min_cell_percentage)
results = datasets.get_list(values_included=[f'{gene_name} > {min_gene_expression}'])
And that will give you, along with the UUIDs for each dataset, the percentage of cells that meet the criteria provided.

mccalluc: Not quite clear what the change is: Would we get rid of `has=[f'{gene_name} > {min_gene_expression}']` and replace it with `values_included=[f'{gene_name} > {min_gene_expression}']`? (Not really a question... will be more clear when it's delivered. Thanks!)

Issue #1740, opened 2021-04-19
Document image pipeline


john-conroy:
From conversation in #1727, we want to standardize how we introduce images to the portal. Documentation should be added to the [portal-images readme](https://github.com/hubmapconsortium/portal-images).

mccalluc:
@john-conroy : This would be good to wrap up before you forget what you used to process the images.

Issue #1608, opened 2021-03-09
Search facets based on visualization hints?


mccalluc:
We'd like to be able to direct users to datasets based on the visualizations available for that dataset.

Using the visualization hints at index time would be one way to do this, and would be better than implementing a separate, parallel mapping from assay type to visualization type.

mccalluc:
@ngehlenborg - This came up at one meeting... Is this something you want carried forward? Should you or Tiffany identify the actual facet values we want to see in the UI?

PR #1825, opened 2021-05-10
Add client methods to fetch all data types with visualizations

john-conroy:
Currently returns `['AF', 'codex_cytokit', 'IMC', 'MALDI-IMS-neg', 'MALDI-IMS-pos', 'PAS', 'salmon_rnaseq_10x', 'sc_atac_seq_sci', 'seqFish', 'sn_atac_seq', 'salmon_sn_rnaseq_10x']`

I've only attached the method to the `/home-revision` endpoint for debugging. Ideally we'd like to rework this as a script which can write the assay types to a file, but I'm having issues providing an app context for it to work similar to the pytest fixtures in `test_routes_main.py` and `test_routes_errors.py`.

### DIFF
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -208,13 +208,13 @@ class ApiClient():
     def get_assays_with_visualizations(self):
         tc = TypeClient(current_app.config["TYPE_SERVICE_ENDPOINT"])
         assays_with_visualizations = []
-        assays = [assay.name for assay in tc.iterAssays()]
+        assays = [assay for assay in tc.iterAssays()]
         for assay in assays:
-            entity = self.get_most_recent_dataset_for_assay(assay)
+            entity = self.get_most_recent_dataset_for_assay(assay.name)
             if entity:
                 vc = self.get_vitessce_conf(entity)
                 if any(vc):
-                    assays_with_visualizations.append(assay)
+                    assays_with_visualizations.append(assay.description)
         return assays_with_visualizations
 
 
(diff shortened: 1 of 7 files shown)


context/app/api/client.py
```diff
@@ -167,6 +168,55 @@ def _get_mock_vitessce_conf(self):
...
+        }
+
+        response_json = self._request(
+            current_app.config['ELASTICSEARCH_ENDPOINT']
+            + current_app.config['PORTAL_INDEX_PATH'],
+            body_json=query)
+        return response_json['hits']['hits'][0]["_source"] if len(response_json['hits']['hits']) > 0 else None
+
+    def get_assays_with_visualizations(self):
+        tc = TypeClient(current_app.config["TYPE_SERVICE_ENDPOINT"])
```
mccalluc, line 209: The goal is a script a developer can run to produce an up-to-date list of visualize-able dataset types? I would put it in a separate directory from the runtime code, and either hardcode the type service endpoint, or make it a command-line parameter.

(One possible shape for this: The output is checked in as a configuration file, and the test suite checks that there aren't new visualizable assay types.) (Having dependencies on live APIs during tests isn't great, but this would be small, limited scope... and it's really a consequence of using microservices for things that might have been better as shared configuration. But that's not worth fighting at this point.)

mccalluc (changes requested):

mccalluc:
@john-conroy / @ilan-gold:
- Ilan sets up a nice codex configuration
- John documents portal-images steps (#1740)
- Ilan is a guinea pig for the documentation, and updates the carousel, pointing to the new CODEX example, and using the new screenshot.
- At some point, record in the index whether a dataset has visualization: #1608 (This will likely be based on a pre-baked list of assay types, like in this PR.)

PR #1830, opened 2021-05-11
John conroy/document image pipeline

### LINKED ISSUE #1740: Document image pipeline

john-conroy:
Fix #1740. I kept it pretty minimal, let me know if more information (perhaps regarding using srcset) would be useful.

### DIFF
--- a/CHANGELOG-document-image-pipeline.md
+++ b/CHANGELOG-document-image-pipeline.md
@@ -1,0 +1,1 @@
+- Add image processing guide to readme.
--- a/CHANGELOG-document-image-pipeline.md
+++ b/CHANGELOG-document-image-pipeline.md
@@ -1,1 +1,1 @@
-- Add image processing guide to readme.
+- Fix image carousel srcset.
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ Images should displayed using the `source srcset` attribute. You should prepare 
 
 Finally after processing, the images should be checked into the [`portal-images`](https://github.com/hubmapconsortium/portal-images) submodule.
 
+Images provided for the homepage carousel should have a 16:9 ratio and abide by the above rules.
+
 ### React File Structure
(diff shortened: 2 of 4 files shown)


ilan-gold (commented):
I think this all makes sense conceptually but going from screenshot->end result of these instructions is not clear to me right now.  Are there some tools I can use to achieve these tasks?

john-conroy:
You can use Preview to crop the images, and something like ImgOptim or TinyPng to optimize the images.

john-conroy:
Ok, should be updated now. Let me know if it's helpful or if more information is needed.

john-conroy:
I'll add a comment in the carousel component.

mccalluc (changes requested):
Thank you for writing this up. I think this lays out the pieces, but I think it would be good to go in a different direction.
- This readme is getting long, so how about if we put image processing instructions in `portal-images`, and just link from here?
- In that repo, I would like there to be a script that does the image resizing and optimization.
- The discrepancy between the docs (`original size and at 75%, 50% and 25%`) and code (`src320w, src640w, src1280w, src1392w`) is odd now, and will only be more confusing in the future.
- Do you have guidance about getting a 16:9 aspect ratio? Like, if I'm trying to get a screenshot of a tool, and I don't want to end up cropping out important stuff at the edges, how should this be done? Maybe a bookmarklet to reopen the current url in a new, big, 16:9 window? https://developer.mozilla.org/en-US/docs/Web/API/Window/open#window_features

For now, I'd suggest moving this content to the other repo, and filing issues for the subtasks you don't get to.

john-conroy:
@mccalluc I deleted the readme changes. There is a small code fix here I'd like to get in.

PR #1834, opened 2021-05-11
Remove return_empty_json_if_error

### LINKED ISSUE #1645: Better Error Reporting to Remove Try-Catch in Vitessce Configuration Generation
Re #1637's aggressive try-catching to prevent 500 errors during Vitessce configuration-generation: 

> then, separately, a bigger PR to return {} even if there is an error...but don't think that should be left in forever. If we had better error reporting, perhaps, but I don't think we can rely on IEC to grep the logs on an on-going basis for us, and having silent failures is worse than noisy failures. (The NIH has signed off on the idea that this is "research-quality" software, so I don't think we should sweep problems under the rug.)

mccalluc:
Fix #1645

This is a judgement call, so push back if you don't think it's wise. Justifications:
- The same reasons that we don't want to use try-catch to silently ignore all errors.
- While you and John are watching the logs now, that's an after-the-fact notice, and I think you're only looking at PROD: Is there's a problem with a new unpublished dataset, we want the problem to be unignorable.
- There are lots of places where we do depend on particular structures in API responses, and we have had errors when our assumptions were wrong... and that has forced us to fix our code, or get fixes into the APIs.

### DIFF
--- a/CHANGELOG-vitessce-try-catch.md
+++ b/CHANGELOG-vitessce-try-catch.md
@@ -1,0 +1,1 @@
+- Remove the try-catch around Vitessce view-conf code.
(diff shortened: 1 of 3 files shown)


mccalluc:
Talked with Ilan: better approach is to return a short error message viewconf, but keep the try-catch.

Issue #1812, opened 2021-05-06
Add info button on CollectionCitation


mccalluc:
![Screen Shot 2021-05-05 at 10 50 27 AM](https://user-images.githubusercontent.com/16737715/117186479-b693cc00-ad8f-11eb-9023-76c581155275.png)

Followup from PR #1801.

PR #1839, opened 2021-05-11
Mccalluc/datacite collections

mccalluc:
This is in the mockup in #1812, but I don't see an issue.
- @tsliaw -- Review design.
- @john-conroy -- Code review. Not sure I've organized this well.


context/app/static/js/components/Detail/Citation/style.js
```diff
@@ -1,8 +1,25 @@
...
+
+const FlexRight = styled.div`
+  display: flex;
+  margin-left: auto;
+`;
+
+const StyledOpenInNewRoundedIcon = styled(OpenInNewRoundedIcon)`
+  font-size: 1.1rem;
+  vertical-align: text-bottom;
+  margin-left: 0.2rem;
```
john-conroy, line 22:
```suggestion
  margin-left: ${(props) => props.theme.spacing(0.5)}px;
```

context/app/static/js/components/Detail/Citation/Citation.jsx
```diff
@@ -15,12 +16,25 @@ function buildNLMCitation({ contributors, title, timestamp }) {
...
-      <SectionHeader>Citation</SectionHeader>
+      <Flex>
+        <SectionHeader>Citation</SectionHeader>
+        <FlexRight>
+          <Button
+            color="primary"
+            variant="contained"
+            component="a"
+            target="_blank"
+            href={`https://search.datacite.org/works/${doi}`}
```
john-conroy, line 32: We shouldn't need to encode the doi, right?

john-conroy (commented):
I would lean towards creating a reusable button component `OutboundLinkButton` or similar which leverages the existing `OutboundLink` component.

If we think this will only be used by collections I'd put it in the collections dir in components (where other components that are in detail should probably be).

tsliaw (approved):

context/app/static/js/components/Detail/Citation/Citation.jsx
mccalluc: I don't think so: If the service were structured differently, we might, but this works with a string like `10.35079/hbm876.xnrh.336` right in the path URL: We don't want to encode the `/` or anything like that.

mccalluc:
> I would lean towards creating a reusable button component `OutboundLinkButton` or similar which leverages the existing `OutboundLink` component.

Done, sort-of: Just putting `OutboundLink` inside the Button doesn't work because then you can only click exactly on the text, not anywhere else in the button. Instead, I've factored out the google analytics callback, so it can be called onClick, either form the original Link, or the new Button.

PR #1840, opened 2021-05-11
Kludge for Safari text color

### LINKED ISSUE #1790: button font color for data search page tour launch too dark
image

(observed in Safari, v0.17.2 - 2021-04-28)

mccalluc:
I feel like at the root, it's a safari bug: When I go to dev tools, it does list the color as `#fff`, and I think the rendering has fixed itself while I was inspecting, without making any actual changes. Fix #1790.

### DIFF
--- a/context/app/static/js/components/tutorials/DatasetSearchPrompt/style.js
+++ b/context/app/static/js/components/tutorials/DatasetSearchPrompt/style.js
@@ -32,7 +32,6 @@ const StyledInfoIcon = styled(InfoIcon)`
 `;
 
 const StyledButton = styled(Button)`
-  color: #fff !important; // Safari sees the color, but doesn't always respect it unless forced.
   padding: 6px 36px;
 `;
 
--- a/context/app/static/js/components/tutorials/DatasetSearchPrompt/style.js
+++ b/context/app/static/js/components/tutorials/DatasetSearchPrompt/style.js
@@ -32,7 +32,7 @@ const StyledInfoIcon = styled(InfoIcon)`
 `;
 
 const StyledButton = styled(Button)`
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/tutorials/DatasetSearchPrompt/style.js
```diff
@@ -32,6 +32,7 @@ const StyledInfoIcon = styled(InfoIcon)`
 `;
 
 const StyledButton = styled(Button)`
+  color: #fff !important; // Safari sees the color, but doesn't always respect it unless forced.
```
john-conroy, line 35:
```suggestion
```

john-conroy: We shouldn't need this at all. Setting the color prop to primary should also set the text color white. I can't reproduce the dark text in my safari browser, could you check if removing this fixes the issue?

john-conroy: My mistake for adding the css in the first place.

mccalluc:
@john-conroy -- Huh: so I've applied that, and the problem comes back. You are unable to reproduce this on Safari? My version: `Version 14.1 (15611.1.21.161.5, 15611)`

Screen Shot 2021-05-12 at 5 03 56 PM

john-conroy:
Yeah, it's a known Safari/Mui bug for any buttons that are disabled at some point in their lifecycle. There's a temporary fix to add a `key` prop equal to the same condition as the disabled prop to force a re-render.

mccalluc:
Closing in favor of #1848.

PR #1841, opened 2021-05-11
John conroy/home revision animations

john-conroy:


john-conroy:
Closing for now, will revisit based on feedback.

PR #1843, opened 2021-05-12
John conroy/replace homepage

### LINKED ISSUE #1670: Implement Homepage Redesign
Design is in #1513 

Collecting facet values for search is #1666

john-conroy:
Fix #1670.


mccalluc (approved):
Great! I assume you deleted all the old stuff before moving the new stuff into its place, so we can be pretty sure there aren't unreferenced old things hanging around? Thanks.

PR #1846, opened 2021-05-13
Fix Submodule Pulls for Deployment

ilan-gold:
`portal-images` uses `main` not `master`.  We'll be migrating all of our repos now but for now this provides a hot-fix.


john-conroy (approved):
Thanks!

PR #1848, opened 2021-05-13
John conroy/temp fix safari disabled buttons

### LINKED ISSUE #1740: Document image pipeline

### LINKED ISSUE #1790: button font color for data search page tour launch too dark

john-conroy:
Fix #1790. Replaces #1840. 

This is a known Safari/Mui bug for any buttons that are disabled at some point in their lifecycle. The bug affects all updated properties, not just `color` so probably best to make sure everything is updated and just force the re-render by setting a key prop equal to the same condition as the disabled prop. 

We'll want to replace this once there is a fix, should I put a 'TODO' somewhere?

### DIFF
--- a/context/app/static/js/shared-styles/buttons/DisabledButton/DisabledButton.jsx
+++ b/context/app/static/js/shared-styles/buttons/DisabledButton/DisabledButton.jsx
@@ -1,0 +1,8 @@
+import React from 'react';
+import Button from '@material-ui/core/Button';
+
+function DisabledButton({ disabled, ...props }) {
+  return <Button key={disabled} disabled={disabled} {...props} />;
+}
+
+export default DisabledButton;
--- a/context/app/static/js/shared-styles/buttons/DisabledButton/DisabledButton.jsx
+++ b/context/app/static/js/shared-styles/buttons/DisabledButton/DisabledButton.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
-import Button from '@material-ui/core/Button';
+
+import { StyledButton } from './style';
 
 function DisabledButton({ disabled, ...props }) {
-  return <Button key={disabled} disabled={disabled} {...props} />;
+  return <StyledButton disabled={disabled} {...props} />;
 }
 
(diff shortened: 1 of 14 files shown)


context/app/static/js/shared-styles/buttons/DisabledButton/DisabledButton.jsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import Button from '@material-ui/core/Button';
+
+function DisabledButton({ disabled, ...props }) {
```
mccalluc, line 4: - Worried that `DisabledButton` may be confusing, because if disabled is false, it's not actually "disabled"... Maybe just `Button`, and then the only thing that needs to change is the import?
- Can you link to the Safari/MUI issue here, or some discussion about this? Is the idea that when there's a fix upstream, we can just go back to using regular `Button`?
- Maybe file a followup issue, if that's the plan, link to that too, in the TODO? I like there to be a bidirectional connection between TODOs and issues.

mccalluc (commented):
Maybe just call it `Button`? and add a comment about what the next steps are?

context/app/static/js/shared-styles/buttons/DisabledButton/DisabledButton.jsx
john-conroy: I'd prefer to use a name that distinguishes this instead of `Button`. Some places say 'disableable' is a word, but a bit of a jumble. Yes the idea is to go back to using regular `Button` once it is fixed. I sent you a link to the issue. I'll add a TODO and an issue.

mccalluc:
@john-conroy -- Can you get tests passing and merge?

john-conroy:
Using a css fix instead of forcing the re-render with `key` prop.

Issue #1647, opened 2021-03-17
Address layout shift for dynamic content on detail pages


john-conroy:
For sections (sample specific datasets table, dataset collections table) we don't know whether they should be displayed until data is fetched, we need to address layout shift for when the section is displayed and suddenly pushes other content down.

We discussed in the meeting putting a new component in place of the intended content if there isn't data to be displayed. For now the action is to determine which sections we'd need this for to get an idea of the scope.

PR #1849, opened 2021-05-13
John conroy/donor specific changes

### LINKED ISSUE #1729: Implement v1.0 of Donor-Specific Datasets
Design is in #1728

john-conroy:
Fix #1729. Towards #1647. Pointing towards your PR for now @mccalluc.

- Adds tabs components to `shared-styles` for easier reuse. Also shortens the tabs height to be closer to the design.
- Sets a fix height and displays a circular loader while loading to avoid the layout shift. Since we know the number of derived entities in advance from the flask metadata, I could adapt the size of the paper to the number of entities to avoid having the large empty space seen below. There would likely still be some, but less. We could use a [skeleton](https://material-ui.com/components/skeleton/) to avoid this altogether, but we should hold off unless we get strong feedback.
- We only show the alert for the derived datasets and samples section. If derived datasets do not exist for a sample, the section will not be shown.

I'm going to add a PR off of this to add tests and move the components into a directory within `Detail`.

Screen Shot 2021-05-20 at 4 55 46 PM
Screen Shot 2021-05-20 at 4 55 55 PM
Screen Shot 2021-05-20 at 5 01 12 PM

Derived samples or dataset don't exist

Screen Shot 2021-05-20 at 4 56 57 PM

Derived datasets on samples page

Screen Shot 2021-05-20 at 4 54 39 PM
Screen Shot 2021-05-20 at 4 55 32 PM


mccalluc:
Is the warning for "Derived samples or dataset don't exist" what we want? It feels too much like an error to me... but I guess we do the same thing when there are no search results... but that feels different to me because the user initiated a request that can't be fulfilled.

(If this doesn't seems off-kilter to anyone else, please don't let me derail things! Looking at the code now...)

context/app/static/js/components/Detail/DerivedDatasetsSection/DerivedDatasetsSection.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import DerivedEntitiesSectionWrapper from 'js/components/Detail/DerivedEntitiesSectionWrapper';
+import DerivedEntitiesTable from 'js/components/Detail/DerivedEntitiesTable';
+import DerivedEntitiesSectionHeader from 'js/components/Detail/DerivedEntitiesSectionHeader';
```
mccalluc, line 5: In general, do we want to avoid paths where higher directories are capitalized? ie, a directory of reusable components shouldn't also be a component in it's own right? But... I'm not sure that should be a hard-and-fast rule ... and I think just renaming the directory to `details` would cause problems because the the Mac FS is case insensitive?

context/app/static/js/components/Detail/DerivedDatasetsSection/DerivedDatasetsSection.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+function DerivedDatasetsSection({ datasets, uuid, isLoading, sectionId }) {
+  const entityType = 'Dataset';
+  return (
+    <DerivedEntitiesSectionWrapper
+      isLoading={isLoading}
+      sectionId={sectionId}
+      header={
```
mccalluc, line 13:
```suggestion
      headerComponent={
```
Since we're less scrupulous about proptypes, I wonder if it would be good to make the prop names more descriptive. I think this one popped out for me because the `header` prop here is a different thing than the `header` inside, and they might be confused?

context/app/static/js/components/Detail/DerivedEntitiesSection/DerivedEntitiesSection.jsx
```diff
@@ -0,0 +1,45 @@
...
+    },
+  ];
+  return (
+    <DerivedEntitiesSectionWrapper
+      isLoading={isLoading}
+      sectionId={sectionId}
+      header={
+        <DerivedEntitiesSectionHeader
+          header="Derived Samples and Datasets"
+          entityCountsText={[`${samples.length} Samples`, `${datasets.length} Datasets`].join(' | ')}
```
mccalluc, line 29:
```suggestion
          entityCountsText={`${samples.length} Samples | ${datasets.length} Datasets`}
```
Simpler for now... if we need to generalize, revisit.

context/app/static/js/components/Detail/DerivedEntitiesTable/DerivedEntitiesTable.jsx
```diff
@@ -0,0 +1,60 @@
...
+              },
+            }) => (
+              <TableRow key={display_doi}>
+                <TableCell>
+                  <LightBlueLink href={`/browse/dataset/${entityUUID}`} variant="body2">
+                    {display_doi}
+                  </LightBlueLink>
+                </TableCell>
+                <TableCell>{mapped_data_types || origin_sample?.mapped_organ}</TableCell>
+                <TableCell>{status || mapped_specimen_type}</TableCell>
```
mccalluc, line 48: The structure here is linked to the logic in `getColumnNames`. In general, I worry about places where complementary changes would need to be made in widely separated pieces of code. In particular I worry about how a request to add a column for just one type would be handled.

Maybe too much has been factored out? I'm wondering if the parallelism between the two cases should be thought of as an accident, and it might be better to split it out, so that the columns for one could be modified without worrying about the other?

copy-app-conf.sh
```diff
@@ -9,4 +9,4 @@ else
   echo "Using existing $APP_CONF."
 fi
 
-grep 'TODO' "$APP_CONF" && die "Replace 'TODO' in $APP_CONF."
+grep 'TODO' "$APP_CONF" || echo "No TODOs in $APP_CONF"
```
mccalluc, line 12: What motivates this? This script is used by both CI and, conceivably, new developers, and I worry about taking away the error message in the later usecase.

mccalluc (commented):
I don't want to block this PR, but I feel uncomfortable about the way columns for different entity types are configured.

context/app/static/js/components/Detail/DerivedDatasetsSection/DerivedDatasetsSection.jsx
john-conroy: I think we'd like to change it to `detail` but doing so right now is more trouble than it is worth?

copy-app-conf.sh
john-conroy: This isn't a change to master, this is the already in place fix required to make `dev-start` work.

context/app/static/js/components/Detail/DerivedEntitiesTable/DerivedEntitiesTable.jsx
john-conroy: Separated into two components, thanks!

tsliaw (approved):

PR #1850, opened 2021-05-17
Fix SPRM Stitched

ilan-gold:
The offsets are wrong in the generated data but something should show now regardless

`06589a380ca45237eddd0d9ea7ad2c9f` on TEST if you'd like to test should work now whereas before it would have no visualization

### DIFF
--- a/CHANGELOG-fix-new-sprm-vitessce.md
+++ b/CHANGELOG-fix-new-sprm-vitessce.md
@@ -1,0 +1,1 @@
+- Fix new stitched `AnnData` SPRM viewconf.
--- a/context/app/api/vitessce_confs/base_confs.py
+++ b/context/app/api/vitessce_confs/base_confs.py
@@ -78,7 +78,7 @@ class ViewConf:
         # See: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests
         use_request_init = False if self._entity["status"] == "Published" else True
         return request_init if use_request_init else None
-    
+
     def _get_file_paths(self):
         return [file["rel_path"] for file in self._entity["files"]]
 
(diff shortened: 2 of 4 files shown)


mccalluc (commented):
Code looks good... but when I try it on FF on localhost:
```
8691082915633190000 exceeds MAX_SAFE_INTEGER. Precision may be lost. Please report if you get this message to https://github.com/geotiffjs/geotiff.js/issues
```
and the spinners never go away. (I've waited a couple minutes.) In the console:
```
cleared cell-sets; waiting on 2: ["cells","raster"]
cleared cell-sets; waiting on 1: ["cells"] 
cleared cells; waiting on 1: ["raster"] 
cleared cells; waiting on 0: [] 
cleared expression-matrix; waiting on 1: ["raster"] 
cleared expression-matrix; waiting on 0: []
```
From that, I would think all the data has loaded?

Having a visualization that doesn't work is one step ahead of not having a visualization. I think it might be ok to file an issue for the remaining work, and merge this... but it would be a little safer to get a working example before merging. Your call.

ilan-gold:
@mccalluc Yes the backing data is wrong.

This is the tradeoff of doing things this way and the reason I previously had not - they are almost surely going to break the first few (or more) times around.  I don't mind merging this since it's not our fault that the data is broken.  There's no new work for us here.

mccalluc (approved):
> I don't mind merging this since it's not our fault that the data is broken. There's no new work for us here.

Cool -- As long as you're pretty sure it's a data problem, merging now seems good. (Would it be worth trying to catch anything in a validation phase before visualization?) Thanks!

ilan-gold:
> > I don't mind merging this since it's not our fault that the data is broken. There's no new work for us here.
> 
> Cool -- As long as you're pretty sure it's a data problem, merging now seems good. (Would it be worth trying to catch anything in a validation phase before visualization?) Thanks!

I don't think there's much we could do here.  They ran the offsets pipeline on the wrong file so the offsets are wrong 🤷‍♂️

Issue #1804, opened 2021-05-03
on vis-lifted pages, add a note in the prov graph


mccalluc:
... unless there’s an easy way just to not show it at all

PR #1857, opened 2021-05-21
John conroy/remove support prov

john-conroy:
Towards #1804. Draft, ignore the code `Ancestor Note` and the `Search` page for now, I'll rework them both if we decide to go this route.

The original task was to remove the provenance section for support pages, but doing so left no way of easily navigating to the parent dataset(s) as prompted by the alert. This PR removes the provenance section for support entities and also adds a link to either the parent dataset detail page if only a single parent dataset exists or the search page if multiple parent datasets exist.

I also had to add a z-index fix for the alert now that it has a link. Without changing the z-index, the link would be un-clickable due to the offset padding/negative margin we needed in `SectionContainer` for the anchor tags in the table of contents to link to the correct location on the page for each section.

Questions

- If we only expect a single dataset parent for each support could we do away with the link to the search page and the new note?
- What language should we be using? I know we're trying to move away from vague terms like 'entity'. Does 'Parent dataset' fit? 'Ancestor' for something farther back?
- Do I describe the z-index/ anchor tags issue in the comments clearly enough?

Screen Shot 2021-05-25 at 12 27 38 PM

Screen Shot 2021-05-25 at 12 34 28 PM

Screen Shot 2021-05-25 at 12 34 49 PM

### DIFF
--- a/context/app/static/js/components/Detail/SupportAlert/SupportAlert.jsx
+++ b/context/app/static/js/components/Detail/SupportAlert/SupportAlert.jsx
@@ -3,16 +3,17 @@ import React from 'react';
 import { LightBlueLink } from 'js/shared-styles/Links';
 import { Typography } from '@material-ui/core';
 import { StyledAlert } from './style';
-import { getImmediateDatasetAncestors } from './utils';
 
-function SupportAlert({ immediate_ancestors, uuid }) {
+function SupportAlert({ uuid }) {
   // There should usually be only one parent, but this is more robust, and we want to keep it simple.
   return (
     <StyledAlert severity="warning">
       <Typography variant="body2">
         “Support” entities provide derived, low-level data for visualizations. Navigate to{' '}
-        <LightBlueLink href={`/search?descendant_ids[0]=${uuid}&entity_type[0]=Dataset`}>the parent dataset</LightBlueLink> for
-        a view of this information in context.
+        <LightBlueLink href={`/search?descendant_ids[0]=${uuid}&entity_type[0]=Dataset`}>
+          the parent dataset
+        </LightBlueLink>{' '}
+        for a view of this information in context.
       </Typography>
     </StyledAlert>
   );
--- a/context/app/static/js/components/Detail/SupportAlert/SupportAlert.jsx
+++ b/context/app/static/js/components/Detail/SupportAlert/SupportAlert.jsx
@@ -9,20 +9,17 @@ function SupportAlert({ immediate_ancestors, uuid }) {
   const immediateDatasetAncestors = getImmediateDatasetAncestors(immediate_ancestors);
 
   // We only expect a single immediate parent dataset for each support, but best to provide a fallback.
-  const singleDatasetAncestor = immediateDatasetAncestors.length === 1;
+  const hasSingleDatasetAncestor = immediateDatasetAncestors.length === 1;
+  const parentDatasetHref = hasSingleDatasetAncestor
+    ? `/browse/dataset/${immediateDatasetAncestors[0]}`
+    : `/search?descendant_ids[0]=${uuid}&entity_type[0]=Dataset`;
 
   return (
     <StyledAlert severity="warning">
       <Typography variant="body2">
         “Support” entities provide derived, low-level data for visualizations. Navigate to{' '}
-        {singleDatasetAncestor ? (
-          <LightBlueLink href={`/browse/dataset/${immediateDatasetAncestors[0]}`}>the parent dataset</LightBlueLink>
-        ) : (
-          <LightBlueLink href={`/search?descendant_ids[0]=${uuid}&entity_type[0]=Dataset`}>
-            the parent datasets
-          </LightBlueLink>
-        )}{' '}
-        for a view of this information in context.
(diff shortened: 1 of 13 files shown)


context/app/static/js/components/Detail/SupportAlert/SupportAlert.jsx
```diff
@@ -0,0 +1,31 @@
...
+    <StyledAlert severity="warning">
+      <Typography variant="body2">
+        “Support” entities provide derived, low-level data for visualizations. Navigate to{' '}
+        {singleDatasetAncestor ? (
+          <LightBlueLink href={`/browse/dataset/${immediateDatasetAncestors[0]}`}>the parent dataset</LightBlueLink>
+        ) : (
+          <LightBlueLink href={`/search?descendant_ids[0]=${uuid}&entity_type[0]=Dataset`}>
+            the parent datasets
+          </LightBlueLink>
+        )}{' '}
```
mccalluc, line 24:
```suggestion
          <LightBlueLink href={linkHref}>
            {linkText}
          </LightBlueLink>
```
Since the structure is the same in either case, I'd suggest pulling out the logic out so the react can be a simple template.

context/app/static/js/components/Search/AncestorNote.jsx
```diff
@@ -4,17 +4,17 @@ import Typography from '@material-ui/core/Typography';
 
 import { LightBlueLink } from 'js/shared-styles/Links';
 
-function AncestorNote(props) {
-  const { entity } = props;
-
+function AncestorNote({ entity, ancestorOrDescendantNote }) {
   let message = '...';
+
+  const text = ancestorOrDescendantNote === 'ancestor' ? 'Derived from' : 'Ancestor of';
```
mccalluc, line 10:
```suggestion
function AncestorNote({ entity, isDerivedFrom }) {
  let message = '...';

  const text = isDerivedFrom ? 'Derived from' : 'Ancestor of';
```
Or make the default the other way? Not sure which is best, but the string equality check makes me nervous.

mccalluc (commented):
I think this is fine, but I'd put it on hold: After the call yesterday, I think Nils is going to want more changes to providence, and if we're going to be making higher-level changes, that may change the picture here. Sorry.

context/app/static/js/components/Search/AncestorNote.jsx
john-conroy: As I mentioned above, this isn't finished and will be reworked.

john-conroy:
@mccalluc I requested re-review. Pending approval, I'll rename `AncestorNote` to `SearchNote` and add some tests.

context/app/static/js/components/Detail/SupportAlert/SupportAlert.jsx
```diff
@@ -0,0 +1,28 @@
...
+    : `/search?descendant_ids[0]=${uuid}&entity_type[0]=Dataset`;
+
+  return (
+    <StyledAlert severity="warning">
+      <Typography variant="body2">
+        “Support” entities provide derived, low-level data for visualizations. Navigate to{' '}
+        <LightBlueLink href={parentDatasetHref}>the parent dataset{!hasSingleDatasetAncestor && 's'}</LightBlueLink> for
+        a view of this information in context.
+      </Typography>
+    </StyledAlert>
```
mccalluc, line 24:
```suggestion
  // There should usually be only one parent, but this is more robust, and we want to keep it simple.
  return (
    <StyledAlert severity="warning">
      <Typography variant="body2">
        “Support” entities provide derived, low-level data for visualizations. Navigate to{' '}
        <LightBlueLink href={`/search?descendant_ids[0]=${uuid}&entity_type[0]=Dataset`}>the parent dataset</LightBlueLink> for
        a view of this information in context.
      </Typography>
    </StyledAlert>
```

context/app/static/js/components/Detail/SupportAlert/SupportAlert.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import { LightBlueLink } from 'js/shared-styles/Links';
+import { Typography } from '@material-ui/core';
+import { StyledAlert } from './style';
+import { getImmediateDatasetAncestors } from './utils';
+
+function SupportAlert({ immediate_ancestors, uuid }) {
+  const immediateDatasetAncestors = getImmediateDatasetAncestors(immediate_ancestors);
+
```
mccalluc, line 10:
```suggestion
```

context/app/static/js/components/Detail/SupportAlert/utils.js
```diff
@@ -0,0 +1,8 @@
+function getImmediateDatasetAncestors(immediate_ancestors) {
```
mccalluc, line 1: Propose not to use this.

mccalluc (commented):
I would lean towards keeping it simple, even if the end result is a little less polished. Every bit of logic we add is something someone may need to understand and tweak in the future... but since it's already done, I wouldn't block this, either.

(Let me know if this isn't definite enough, or I'm otherwise missing the point.)

PR #1860, opened 2021-05-24
John conroy/fix master changelog test

john-conroy:
Fixing `test.sh` issue introduced by a dependabot change causing master builds to fail without a changelog (should pass).


mccalluc (approved):
Thanks: That looks right, given that I see this on the failing master builds:
```
  GITHUB_REF: refs/heads/master
  GITHUB_HEAD_REF: 
  GITHUB_BASE_REF: 
```

Issue #1605, opened 2021-03-08
Jupyter Demo for Vitessce Backend Python


ilan-gold:
We recently committed in #1581 the integration of vitessce-python for the backend.  The next step would be to modularize this to some degree and allow users to launch Jupyter notebooks based on the current dataset they are working with, including the visualization configuration for the detail page they are on.

This could use Binder or Google Collab, I imagine, or we could host our own server.

mccalluc:
Lighter weight: Just serve a notebook with the view-conf embedded.
- Finding an environment to run it on is left as an exercise for the reader.
- Ideally, we would have some intermediate representation between the uuid and the whole view-conf, but the idea of serving a notebook doesn't need to block on that.

(Could imagine cramming the viewconf into a url parameter, but that wouldn't continue to work if someone saved the notebook, and we've have the usual problems with huge url parameters.)

mccalluc:
- [ ] Return named conf/notebook-cells tuples instead of just viewconfs.
- [ ] Only return notebook-cells when we've got something better than the json viewconf.
- [ ] ... and when we have notebook-cells, show a download link for the notebook.
- [ ] Try adding executable reprs to the wrapper classes in `vitessce-python`.
- [ ] ... and try constructing the notebooks from them.
- [ ] come up with a list of datasets to exercise every view conf notebook.
- [ ] testing!

PR #952, opened 2020-07-31
John conroy/collections description

john-conroy:
Fix #944. Also slightly increases spacing between sections on entity detail pages.

Screen Shot 2020-07-31 at 11 48 55 AM

PR #1873, opened 2021-05-27
Mccalluc/serve notebooks

mccalluc:
Towards #1605. @ilan-gold , considered as a proof of concept, is this much reasonable? If it's easy for you, do you want to fill out the TODO with the rendering command? (This is syntactically valid, but I have not actually started up a kernel and tried to run it.)

### DIFF
--- a/CHANGELOG-notebook.md
+++ b/CHANGELOG-notebook.md
@@ -1,0 +1,1 @@
+- Add an endpoint which serves a Jupyter notebook with the viewconf for a dataset.
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -140,9 +140,9 @@ def details_notebook(type, uuid):
 Visualization for [{entity['display_doi']}]({request.base_url.replace('.ipynb','')})
         """.strip()),
         new_code_cell("""
-'!pip install vitessce==0.1.0a9',
-'!jupyter nbextension install --py --sys-prefix vitessce',
-'!jupyter nbextension enable --py --sys-prefix vitessce'
+!pip install vitessce==0.1.0a9
+!jupyter nbextension install --py --sys-prefix vitessce
+!jupyter nbextension enable --py --sys-prefix vitessce
         """.strip()),
         new_code_cell('from vitessce import VitessceConfig'),
         new_markdown_cell('TODO: Generate the appropriate `conf` using `vitessce-python`'),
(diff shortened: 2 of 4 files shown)


ilan-gold:
We should also add some sort of validation for this to Vitessce, the schema is not strong enough

ilan-gold:
And in Vitessce: #952

mccalluc:
@ilan-gold : 
- Thanks for filling out the notebook! 
- Bigger picture, do you share the sense that this is something Nils wants, and it makes sense to complete this much, merge, and get feedback... and later we can expose it in the UI? (That is PR #1874.)
- Am I interpreting the situation correctly that those issues are blockers? 
- Style: maybe combine the python into one cell? Or do smaller chunks make sense?
- Content: We could inject more metadata, or more instructions... but maybe what's there is fine?

Do you want to see this through to completion? Or would you prefer that it stay with me?

ilan-gold:
>     * Bigger picture, do you share the sense that this is something Nils wants, and it makes sense to complete this much, merge, and get feedback... and later we can expose it in the UI? (That is PR #1874.)

Really do not know for sure, but I think that plan makes sense - do this PR then get feedback on the (functional) notebook output.

>     * Am I interpreting the situation correctly that those issues are blockers?

Indeed :(

>     * Style: maybe combine the python into one cell? Or do smaller chunks make sense?

No preference really - I think we have it separated for room for comments (check out some of the notebooks at https://vitessce.github.io/vitessce-python/ for reference)

>     * Content: We could inject more metadata, or more instructions... but maybe what's there is fine?

I was thinking about this too - I think we can wait for Nils' feedback.

> Do you want to see this through to completion? Or would you prefer that it stay with me?

I do not have a preference, but the 3D stuff + ongoing bitmask/Cytokit work takes precedence due to the upcoming data releases so this could languish for a while if given to me.  More likely though is that I just won't have the bandwidth to keep up with the back-and-forth feedback loop rather than the actual engineering.

mccalluc:
@ilan-gold : Following from the conversation at the meeting, I can make a branch that tries to add more explanation into the viewconf classes? I'm not at all sure that the experiment will be successful, but it's something I could do... and it would also let me get more familiarity with the view conf generation.

So: Nothing for you to do right now.

mccalluc:
@ilan-gold -- Could you approve this much? After this, I would do a bit in assay_confs.py, and then vitessce-python. Exposing it in the UI would come last. (#1874)

ilan-gold (approved):

ilan-gold:
Please do request my review once you are ready with the `assay_confs` so we can test it out with "real data"

PR #1875, opened 2021-05-28
Mccalluc/info on api endpoints

mccalluc:
Provide more information on the services page, and make it part of QA to confirm that we're looking at the production APIs.
Screen Shot 2021-05-28 at 11 23 58 AM

(Need to get IEC to provide the cells as part of the gateway status, and I'd like the gateway to come from and envvar instead of being hardcoded.)


context/app/static/js/components/ServiceStatusTable/ServiceStatusTable.jsx
```diff
@@ -1,42 +1,90 @@
...
+          apiName: 'entity-api',
+          endpointUrl: entityEndpoint,
+          response: gatewayStatus.entity_api,
+          noteFunction: (api) => `Neo4j: ${api.neo4j_connection}`,
+        }),
+        buildServiceStatus({
+          apiName: 'gateway',
+          endpointUrl: gatewayEndpoint,
+          response: gatewayStatus.gateway,
+          noteFunction: () => '',
```
ilan-gold, line 63: Why no `gateway` note?

ilan-gold (approved):

tsliaw (approved):

context/app/static/js/components/ServiceStatusTable/ServiceStatusTable.jsx
ilan-gold: Ah I see.  Ok nevermind

PR #1876, opened 2021-05-28
Gateway config

mccalluc:
@yuanzhou -- I think it makes sense for the cells and gateway endpoints to be provided as configurations... Is this something you can configure in each of the environments?

@john-conroy / @ilan-gold / @tsliaw : You'll need to add these to your configuration.

Extends #1875


yuanzhou:
@mccalluc the `CELLS_API_ENDPOINT` had already been added to all configuration files a while ago when we first talked about it. 

I've just added the `GATEWAY_ENDPOINT` in each configuration file. PORD for example:

````
CELLS_API_ENDPOINT = 'https://cells.api.hubmapconsortium.org/api/'

GATEWAY_ENDPOINT = 'https://gateway.api.hubmapconsortium.org'
````
Please make sure the portal-ui would be able to handle the trailing slash correctly, otherwise I can remove the trailing slash from cells API, not sure why it was there. I don't use trailing slash for all other API endpoints though.

**Comment by mccalluc** [comment]
> Please make sure the portal-ui would be able to handle the trailing slash correctly,

Yep, the trailing slash is what we're expecting right now. If you would like more consistency, feel free to add a new name (keeping the old one for now), and we could migrate to it, eventually:

```
XMODALITY_ENDPOINT = https://cells.test.hubmapconsortium.org
```

Is anything else needed on this PR, or can you approve?

john-conroy (approved):

yuanzhou:
I've added `XMODALITY_ENDPOINT = https://cells.test.hubmapconsortium.org` for TEST with corresponding changes to DEV/STAGE too.

For PROD, `XMODALITY_ENDPOINT = https://cells.api.hubmapconsortium.org`

yuanzhou (approved):

mccalluc:
> I've added `XMODALITY_ENDPOINT = https://cells.test.hubmapconsortium.org` for TEST with corresponding changes to DEV/STAGE too.
> 
> For PROD, `XMODALITY_ENDPOINT = https://cells.api.hubmapconsortium.org`

Thanks! just filed a new issue to use that: it will be good to have things consistent, but this PR doesn't need to block on that.

PR #1877, opened 2021-05-28
Mccalluc/serve notebooks plus introspection

mccalluc:
Proof of concept -- breaks a lot of things -- but it also produces a notebook for http://localhost:5001/browse/dataset/3d2c9295c0e05f4d81d23ee95e9dad23.ipynb that looks like:

```
vc = VitessceConfig(name="HuBMAP Data Portal")
vc.add_dataset(name="Visualization Files")
images = []
images.append(OmeTiffWrapper(img_url='https://assets.test.hubmapconsortium.org/ca224d0c1a0144240858490e3df46f83/ometiff-pyramids/processedMicroscopy/VAN0013-LK-206-2-PAS_FFPE.ome.tif?token=', offsets_url='https://assets.test.hubmapconsortium.org/ca224d0c1a0144240858490e3df46f83/output_offsets/processedMicroscopy/VAN0013-LK-206-2-PAS_FFPE.offsets.json?token=', name=Path('ometiff-pyramids/processedMicroscopy/VAN0013-LK-206-2-PAS_FFPE.ome.tif').name))
dataset = dataset.add_object(MultiImageWrapper(images))
vc = self._setup_view_config_raster(vc, dataset)
```
ie, the steps to build the viewconf are spelled out, instead of just giving the JSON that comes out the end of the pipe.

The idea is that on each `ViewConf` subclass, we would change `build_vitessce_conf` to `build_vitessce_conf_cells`: This would return a (named) tuple containing both a viewconf, and a list of notebook cells to construct the same conf. (I think keeping it as a single method, instead of splitting it in two, makes sense for development purposes, since the two flows need to be kept in sync.)

I think we have close to 20 of these classes, and this is one of the simpler ones: _This is a big job._ To make it more tractable, a couple pieces of preliminary work may help:
- For all the internal classes, make sure there is a `repr()` which can be executed to reproduce the object. That will allow us to drop completed datastructures into the notebook, instead of repeating the steps it took to build them.
- Pull out functions so that their definitions could be included into the notebook via introspection.

-----------------

## Alternatives (which I don't like any better)
- Dump the whole class definition into the notebook using inspection... we could dump only the the classes needed for a particular data set, but it would still be a lot of opaque code.
- Make the notebook the primary representation of some of this code, with code and text cells interleaved.
- Give up on getting the same viewconf into the notebook as on the website: If the notebook view was simpler and more generic, that might save work.
- Work on vitessce, so the view-conf object has mutator methods: The setup would still be somewhat opaque, but we could make it easier for them to tweak just the things they want to tweak.

### DIFF
--- a/context/app/api/vitessce_confs/base_confs.py
+++ b/context/app/api/vitessce_confs/base_confs.py
@@ -2,6 +2,7 @@ import urllib
 from pathlib import Path
 import re
 
+import nbformat
 from flask import current_app
 from vitessce import (
     VitessceConfig,
@@ -119,7 +120,10 @@ class ImagePyramidViewConf(ImagingViewConf):
         self.image_pyramid_regex = IMAGE_PYRAMID_DIR
         super().__init__(entity, nexus_token, is_mock)
 
-    def build_vitessce_conf(self):
+    def build_vitessce_conf_cells(self):
+        cells = [
+            nbformat.v4.new_code_cell('# TODO: vitessce_conf = ...')
+        ]
         file_paths_found = self._get_file_paths()
         found_images = get_matches(
             file_paths_found, self.image_pyramid_regex + r".*\.ome\.tiff?$",
@@ -147,7 +151,7 @@ class ImagePyramidViewConf(ImagingViewConf):
         conf = vc.to_dict()
         # Don't want to render all layers
         del conf["datasets"][0]["files"][0]["options"]["renderLayers"]
-        return conf
+        return (conf, cells)
 
 
 class ScatterplotViewConf(ViewConf):
--- a/context/app/api/vitessce_confs/base_confs.py
+++ b/context/app/api/vitessce_confs/base_confs.py
@@ -121,9 +121,7 @@ class ImagePyramidViewConf(ImagingViewConf):
         super().__init__(entity, nexus_token, is_mock)
 
     def build_vitessce_conf_cells(self):
-        cells = [
-            nbformat.v4.new_code_cell('# TODO: vitessce_conf = ...')
-        ]
+        cells = []
         file_paths_found = self._get_file_paths()
         found_images = get_matches(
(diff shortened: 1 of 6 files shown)


context/app/api/vitessce_confs/base_confs.py
```diff
@@ -131,8 +133,18 @@ def build_vitessce_conf(self):
             raise FileNotFoundError(message)
 
         vc = VitessceConfig(name="HuBMAP Data Portal")
+        cells.append(nbformat.v4.new_code_cell(
+            'vc = VitessceConfig(name="HuBMAP Data Portal")'
+        ))
+
         dataset = vc.add_dataset(name="Visualization Files")
+        cells.append(nbformat.v4.new_code_cell(
+            'vc.add_dataset(name="Visualization Files")'
```
mccalluc, line 142:
```suggestion
            'dataset = vc.add_dataset(name="Visualization Files")'
```

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -142,12 +154,29 @@ def build_vitessce_conf(self):
                     img_url=img_url, offsets_url=offsets_url, name=Path(img_path).name
                 )
             )
+            cells.append(nbformat.v4.new_code_cell(
+                f"images.append(OmeTiffWrapper(img_url='{img_url}', offsets_url='{offsets_url}', name=Path('{img_path}').name))"
+            ))
+        # NOTE: If OmeTiffWrapper had a repr which reproduced the object, we could just say:
+        #    f"images = {images}"
```
mccalluc, line 161: Or, better, do it at the very end, with `MultiImageWrapper`.

mccalluc:
Replaced in part by #1894, which uses namedtuples instead. Will be reimplementing the notebook generation in a tidier way.

PR #1879, opened 2021-06-01
Restart hubmap-auth for STAGE redeploy

yuanzhou:
To fix portal.stage and portal-prod.stage being switched.

Due to PR https://github.com/hubmapconsortium/portal-ui/pull/1659 doesn't really address the issue.

### DIFF
--- a/redeploy.sh
+++ b/redeploy.sh
@@ -33,8 +33,8 @@ docker rmi hubmap/portal-ui:latest
 echo 'starting...'
 docker-compose -f \$COMPOSE_CONFIG up -d
 
-# Need to restart the hubmap-auth container on STAGE to avoid
-# portal.stage.hubmapconsortium.org and portal-prod.stage.hubmapconsortium.org get swapped
+# We don't understand why the prod-stage instance is getting the wrong configuration.
+# We hope turning it off-and-on will help, but this is not a good situation.
 if [ "$TARGET" = "stage" ]; then
     echo 'Restart the hubmap-auth container... for STAGE only'
     docker restart hubmap-auth
--- a/redeploy.sh
+++ b/redeploy.sh
@@ -32,6 +32,16 @@ docker rmi hubmap/portal-ui:latest
 
 echo 'starting...'
 docker-compose -f \$COMPOSE_CONFIG up -d
+
+# Need to restart the hubmap-auth container on STAGE to avoid
+# portal.stage.hubmapconsortium.org and portal-prod.stage.hubmapconsortium.org get swapped
+if [ "$TARGET" = "stage" ]; then
+    echo 'Restart the hubmap-auth container... for STAGE only'
+    docker stop hubmap-auth
+    sleep 5
+    docker start hubmap-auth
+fi
+
 echo 'portal running?' \`docker ps | grep portal-ui\`
 
 docker exec -it portal-ui cat package.json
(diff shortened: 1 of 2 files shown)


redeploy.sh
```diff
@@ -32,6 +32,16 @@ docker rmi hubmap/portal-ui:latest
...
 echo 'starting...'
 docker-compose -f \$COMPOSE_CONFIG up -d
+
+# Need to restart the hubmap-auth container on STAGE to avoid
+# portal.stage.hubmapconsortium.org and portal-prod.stage.hubmapconsortium.org get swapped
+if [ "$TARGET" = "stage" ]; then
+    echo 'Restart the hubmap-auth container... for STAGE only'
+    docker stop hubmap-auth
+    sleep 5
+    docker start hubmap-auth
```
mccalluc, line 42: `sleep` always worries me: We're just asking for a race condition. Is there a reason not to use `docker restart`?

redeploy.sh
```diff
@@ -32,6 +32,16 @@ docker rmi hubmap/portal-ui:latest
 
 echo 'starting...'
 docker-compose -f \$COMPOSE_CONFIG up -d
+
+# Need to restart the hubmap-auth container on STAGE to avoid
+# portal.stage.hubmapconsortium.org and portal-prod.stage.hubmapconsortium.org get swapped
+if [ "$TARGET" = "stage" ]; then
+    echo 'Restart the hubmap-auth container... for STAGE only'
```
mccalluc, line 39: Is there any harm in doing this for every instance? I prefer that things be consistent.

mccalluc (changes requested):
> It worked when I tested several times. Apparently that fix didn't fully address the real issue because I noticed the switch again couple weeks ago when portal-ui was redeployed on STAGE. But I didn't follow up with that and didn't get a chance to try with another fix. Sorry about that.
>
> Based on my testing on STAGE, this PR should be able to fix the issue.

My understanding is that we don't have a reliable reproducer for the original bug, and we don't really understand it, and that this is still just a guess, ie, "Have you tried turning it off and on?"

That might be the best we can do for now, and if that's the case it should be explained in the script, so if someone in the future actually does understand the problem, they don't get confused by our kludges.

redeploy.sh
yuanzhou: STAGE is less frequently being used by other developers and this `portal-prod.stage` is only available on STAGE. We don't have this issue on DEV/TEST/PROD because we are only running one portal-ui instance. So this is a unique use case on STAGE only.

redeploy.sh
yuanzhou: I didn't even realize the existence of this `docker restart`. There's no reason not to use it. Feel free to use it.

mccalluc, line 42:
```suggestion
    docker restart hubmap-auth
```
> There's no reason not to use it. Feel free to use it.

redeploy.sh
```diff
@@ -32,6 +32,16 @@ docker rmi hubmap/portal-ui:latest
 
 echo 'starting...'
 docker-compose -f \$COMPOSE_CONFIG up -d
+
+# Need to restart the hubmap-auth container on STAGE to avoid
+# portal.stage.hubmapconsortium.org and portal-prod.stage.hubmapconsortium.org get swapped
```
mccalluc, line 37:
```suggestion
# We don't understand why the prod-stage instance is getting the wrong configuration.
# We hope turning it off-and-on will help, but this is not a good situation.
```

mccalluc (approved):

PR #1893, opened 2021-06-07
Mccalluc/endpoints

### LINKED ISSUE #1871: Gateway API URL should not be hard coded
The status page depends on a hard coded path in `context/app/static/js/components/ServiceStatusTable/hooks.js`:
```
const gatewayUrl = 'https://gateway.api.hubmapconsortium.org/status.json';
```

This should be provided as an envvar like the other APIs.

### LINKED ISSUE #1880: Use `XMODALITY_ENDPOINT`
> I've added `XMODALITY_ENDPOINT = https://cells.test.hubmapconsortium.org` for TEST with corresponding changes to DEV/STAGE too.
>
> For PROD, `XMODALITY_ENDPOINT = https://cells.api.hubmapconsortium.org`

_Originally posted by @yuanzhou in https://github.com/hubmapconsortium/portal-ui/issues/1876#issuecomment-853093236_

mccalluc:
Fix #1880
Fix #1871 (Mostly done already)

You will need to update your git-ignored `app.conf`.


example-app.conf
```diff
@@ -22,16 +22,18 @@ ELASTICSEARCH_ENDPOINT = 'https://search-api.test.hubmapconsortium.org'
 TYPE_SERVICE_ENDPOINT =  'https://search-api.test.hubmapconsortium.org'
 ENTITY_API_BASE =        'https://entity-api.test.hubmapconsortium.org'
 ASSETS_ENDPOINT =        'https://assets.test.hubmapconsortium.org'
-CELLS_API_ENDPOINT =     'https://cells.test.hubmapconsortium.org/api/'
+XMODALITY_ENDPOINT =     'https://cells.api.hubmapconsortium.org'
```
ilan-gold, line 25: Why are these all the same?  Shouldn't they be TEST, DEV etc.?

mccalluc: Eventually... but for now @SFD5311 is deploying straight to the "production" instance, and the other instances are too far behind to be useful for us.

ilan-gold: Ok!

john-conroy (approved):

example-app.conf
SFD5311: @mccalluc I would recommend recommend keeping the '/api/' at the end of the URL

mccalluc: @SFD5311 -- Can you expand on that? We will still be making the same API calls, but this makes the top-level configurations that Zhou sets more consistent. This is somewhat like the Elasticsearch... We always are hitting a particular path under the host, but that's a lower-level detail.

SFD5311: Oh, I see.  Yeah, if the intention is to add that '/api/' suffix to XMODALITY_ENDPOINT before passing it to the client, or to tweak the client to add the '/api/' suffix to the end of whatever is passed to it, then that should be fine.  I'd thought that the '/api/' just got lost in translation.  My mistake.

mccalluc:
(Notified Tiffany on slack.)

PR #1894, opened 2021-06-07
Mccalluc/conf cells

mccalluc:
- Method is returning a tuple now.
- Change the method name to reflect that.
- `get` is shorter than `build`, but I don't care strongly.


ilan-gold (approved):
Just to clarify, there are no cells here?  This is just adding the structure?  Looks good.

mccalluc:
> Just to clarify, there are no cells here? This is just adding the structure? Looks good.

Yep, no cells, just setting up the structure so they can be returned. Thanks!

PR #1895, opened 2021-06-07
Mccalluc/conf builder

mccalluc:
Stop me if I'm confused, but could we clarify that a viewconf is the data structure is these produce? `Builder` seems ok to me... `Factory` is a particular pattern where you use a function to create objects, rather than calling the constructor directorly. Does this seem ok/useful?


ilan-gold (commented):
Sounds good!

mccalluc:
(@ilan-gold ... assuming that counts as approval? Going to merge, but let me know if you have any reservations.)

PR #1902, opened 2021-06-08
John conroy/show derived samples for sample tiles

### LINKED ISSUE #1803: Show derived sample count on every sample tile.

john-conroy:
Fix #1803. Shows derived samples in sample entity tiles. Also refactored code to used provided descendant counts and consolidate utility functions for tile components across the repo.

Screen Shot 2021-06-08 at 4 39 58 PM


context/app/static/js/components/Detail/provenance/ProvTableTile/ProvTableTile.jsx
```diff
@@ -30,10 +34,10 @@ function ProvTableTile(props) {
           id={id}
           invertColors={isCurrentEntity}
           entityData={entityData}
-          descendantCounts={displayDescendantCounts}
+          descendantCounts={descendantCountsToDisplay}
         />
       )}
-      {isLastTile && entity_type !== 'Donor' && allDescendantCounts[entity_type] > 0 && (
+      {isLastTile && entity_type !== 'Donor' && descendantCounts?.[entity_type] > 0 && (
```
mccalluc, line 40: Again, I think this is fine in this case, but I'd be careful about the idiom... It might be safer to pull out the key check. What I worry about is someone refactoring this down the road, perhaps changing the direction of the comparison, and being surprised because any comparison with `undefined` is false.

mccalluc (approved):
Looks good. The comments are not _this-should-change_... more like _in-some-situations-this-idiom-might-cauase-confusion-in-the-future_.

PR #1904, opened 2021-06-08
VItessce 1.1.10 (Bitmask support)

ilan-gold:
This PR upgrades `vitessce-python` and `vitessce` to allow for bitmasks, specifically for stitched SPRM + Cytokit

To test this out you can change these lines in `SPRMAnnDataViewConfBuilder`
```python
img_url, offsets_url = self._get_img_and_offset_url(
            found_image_file, self._imaging_path,
        )
        image_wrapper = OmeTiffWrapper(
            img_url=img_url, offsets_url=offsets_url, name=self._image_name
        )
        bitmask_img_url, bitmask_offsets_url = self._get_img_and_offset_url(
            found_bitmask_file, self._mask_path,
        )
        bitmask_wrapper = OmeTiffWrapper(
            img_url=bitmask_img_url,
            offsets_url=bitmask_offsets_url,
            name=self._mask_name,
            is_bitmask=True
        )
```
to 
```python
img_url, offsets_url = self._get_img_and_offset_url(
            found_image_file, self._imaging_path,
        )
        image_wrapper = OmeTiffWrapper(
            img_url=img_url, name=self._image_name
        )
        bitmask_img_url, bitmask_offsets_url = self._get_img_and_offset_url(
            found_bitmask_file, self._mask_path,
        )
        bitmask_wrapper = OmeTiffWrapper(
            img_url=bitmask_img_url,
            name=self._mask_name,
            is_bitmask=True
        )
```

i.e remove the currently incorrect offsets.  With this change, everything should render though for bitmask datasets: 
http://localhost:5001/browse/dataset/1eeb205f9cebbf0198dae7c4a4643cdd (larger)
and
http://localhost:5001/browse/dataset/06589a380ca45237eddd0d9ea7ad2c9f (much smaller)


context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -182,6 +182,9 @@ def get_conf_cells(self):
                 is_mock=self._is_mock,
                 base_name=region,
                 imaging_path=STITCHED_IMAGE_DIR,
+                mask_path=STITCHED_IMAGE_DIR.replace('expressions', 'mask'),
```
mccalluc, line 185: This is probably paranoid, but would it be good to make this a tighter match? ... or maybe that would introduce fragility you don't want...

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -177,39 +177,33 @@ def _setup_scatterplot_view_config(self, vc, dataset):
 
 class SPRMViewConfBuilder(ImagePyramidViewConfBuilder):
 
-    def _get_sprm_image(self):
-        image_file_regex = f"{self._imaging_path}/{self._base_name}.ome.tiff?"
+    def _make_full_image_path(self):
```
mccalluc, line 180: Could this just be `_get`? One challenge for me as I get familiar with the code is understanding what the methods do, and whether they have side effects.

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -177,39 +177,33 @@ def _setup_scatterplot_view_config(self, vc, dataset):
...
-        found_image_files = get_matches(file_paths_found, image_file_regex)
+        found_image_files = get_matches(file_paths_found, path_regex)
         if len(found_image_files) != 1:
             message = f'Found {len(found_image_files)} image files for SPRM uuid "{self._uuid}".'
             raise FileNotFoundError(message)
         found_image_file = found_image_files[0]
         return found_image_file
 
-    def _add_sprm_image_to_view_conf(self, found_image_file, vc, dataset):
+    def _get_sprm_ometiff_wrapper(self, found_image_file):
```
mccalluc, line 192:
```suggestion
    def _get_ometiff_wrapper(self, found_image_file):
```

Is there anything SPRM specific about this? Or are there other OmeTiffWrapper that we need to distinguish this from?

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -177,39 +177,33 @@ def _setup_scatterplot_view_config(self, vc, dataset):
...
-        dataset = self._add_sprm_image_to_view_conf(found_image_file, vc, dataset)
-        return vc, dataset
 
 
 class SPRMJSONViewConfBuilder(SPRMViewConfBuilder):
     def __init__(self, entity, nexus_token, is_mock=False, **kwargs):
         # All "file" Vitessce objects that do not have wrappers.
         super().__init__(entity, nexus_token, is_mock)
         self._base_name = kwargs["base_name"]
+        self._image_name = kwargs["base_name"]
```
mccalluc, line 206: This is correct? Assigning the same value to two separate variables is a little odd... perhaps add a note to make it clear that this is right?

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -269,20 +272,27 @@ def __init__(self, entity, nexus_token, is_mock=False, **kwargs):
         # All "file" Vitessce objects that do not have wrappers.
         super().__init__(entity, nexus_token, is_mock)
         self._base_name = kwargs["base_name"]
+        self._mask_name = kwargs["mask_name"]
+        self._image_name = kwargs["image_name"]
         self._imaging_path = f"{self.image_pyramid_regex}/{kwargs['imaging_path']}"
+        self._mask_path = f"{self.image_pyramid_regex}/{kwargs['mask_path']}"
```
mccalluc, line 278: Is is generally true in this codebase that `*_path` variables are actually regexes? If that holds consistently, great... if not, maybe clean up some names, though it doesn't need to be this PR.

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -269,20 +272,27 @@ def __init__(self, entity, nexus_token, is_mock=False, **kwargs):
...
+
+    def _make_bitmask_image_path(self):
```
mccalluc, line 280:
```suggestion
    def _get_bitmask_image_path(self):
```

Or maybe save this for a future PR, if it's a lot of changes.

context/app/api/vitessce_confs/paths.py
```diff
@@ -7,6 +7,6 @@
 STITCHED_IMAGE_DIR = "stitched/expressions"
 SPRM_JSON_DIR = "output_json"
 TILE_REGEX = r"R\d+_X\d+_Y\d+"
-STITCHED_REGEX = r"reg\d+_stitched_expressions"
+STITCHED_REGEX = r"reg\d+"
```
mccalluc, line 10: Sorry not to have the whole picture, but `'_stitched_expressions'` just is moved downstream, right? I'm not sure if other paths also use `STITCHED_REGEX`, but that might be something to make sure of.

context/requirements.txt
```diff
@@ -5,6 +5,6 @@ requests==2.22.0
 pyyaml==5.4
 python-datauri==0.2.8
 python-frontmatter==0.5.0
-vitessce==0.1.0a9
+vitessce==0.1.0a11
```
mccalluc, line 8: What's the relationship between the vitessce python  and the vitessce js version numbers? Down the road, maybe it would make sense for the version numbers to match?

mccalluc (commented):
Lots of little comments, but I think they are all this-seems-a-little-odd-but-probably-ok, or lets-clean-this-up-at-some-point. I have not done a local check out... Let me get through some other things this morning, and I'll loop back to this, if that would be helpful.

context/app/api/vitessce_confs/base_confs.py
ilan-gold: I can try to clean this up - they have switched more than a few times (regexes turning into strings, strings into regex) and are used interchangeably (i.e strings as regex) so it's been hard to have consistent names for what every class uses them for.

context/app/api/vitessce_confs/paths.py
ilan-gold: Yes - this is now just like `TILE_REGEX` above it.  I can add a note about this.

context/requirements.txt
ilan-gold: I think this time is a coincidence that the numbers are even close, but we can try for that maybe.  There are fairly independent large development pushes in the respective libraries though.

ilan-gold:
> Lots of little comments, but I think they are all this-seems-a-little-odd-but-probably-ok, or lets-clean-this-up-at-some-point. I have not done a local check out... Let me get through some other things this morning, and I'll loop back to this, if that would be helpful.

Yes more testing super welcome!!! Thanks!

mccalluc:
To clarify, the change I should make in local environment is this?
```diff
--- a/context/app/api/vitessce_confs/base_confs.py
+++ b/context/app/api/vitessce_confs/base_confs.py
@@ -317,7 +317,7 @@ class SPRMAnnDataViewConfBuilder(SPRMViewConfBuilder):
             found_image_file, self._imaging_path,
         )
         image_wrapper = OmeTiffWrapper(
-            img_url=img_url, offsets_url=offsets_url, name=self._image_name
+            img_url=img_url, name=self._image_name
         )
         found_bitmask_file = self._check_sprm_image(self._make_bitmask_image_path())
         bitmask_img_url, bitmask_offsets_url = self._get_img_and_offset_url(
@@ -325,7 +325,6 @@ class SPRMAnnDataViewConfBuilder(SPRMViewConfBuilder):
         )
         bitmask_wrapper = OmeTiffWrapper(
             img_url=bitmask_img_url,
-            offsets_url=bitmask_offsets_url,
             name=self._mask_name,
             is_bitmask=True
         )
```
(Your before example is missing the assignment to `found_bitmask_file`.)

ilan-gold:
@jswelling @mruffalo The current backing data for these datasets is broken as we have discussed (SPRM issues, stitiching issues, no `offsets.json` or incorrect `offsets.json` etc.) but I believe that this PR should display correctly in the portal when those are fixed.  Right now, of course, everything is very broken.

mccalluc (approved):
Ilan talked me through this: With this PR, it will error out because the new data on TEST is not yet quite in the right shape... But I believe that everyone is on the same page, and when the data is the right shape, the visualization will work. 

Suggested that Matt be looped in to confirm that their plans are still the same.

context/app/api/vitessce_confs/base_confs.py
ilan-gold: Yes, unlike with the stitched version there is no `_stitched_expressions` to worry about

context/app/api/vitessce_confs/assay_confs.py
ilan-gold: I think this is OK for now. I see what you mean here but I think this OK especially since this may have to be reused for non Cytokit data.

PR #1905, opened 2021-06-08
John conroy/fix summary date alignment

### LINKED ISSUE #1869: creation and modification top aligned instead of spread out
Screen Shot 2021-05-27 at 1 27 42 PM

With long descriptions, the creation and modification date are spread out, at least on collection pages. @tsliaw -- Should these be top aligned instead? Or if the status-quo is what we want, please close. Thanks!

john-conroy:
Fix #1869. Is this what you both were thinking it should look like? I opted to just pass props to the div to allow us to use `styled`, but I could expand using margin specific props.

Screen Shot 2021-06-08 at 4 35 49 PM
Screen Shot 2021-06-08 at 4 35 55 PM

### DIFF
--- a/CHANGELOG-fix-summary-date-alignment.md
+++ b/CHANGELOG-fix-summary-date-alignment.md
@@ -1,0 +1,1 @@
+- Fix spacing for detail summary creation and modification dates.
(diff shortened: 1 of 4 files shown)


mccalluc (approved):
Looks great to me!

tsliaw (approved):

PR #1907, opened 2021-06-09
John conroy/fix bar chart responsiveness

### LINKED ISSUE #1883: homepage barchart key goes off-screen at narrow widths
Before it goes to the mobile layout, the legend is forced off-screen as the window gets narrower:

Screen Shot 2021-06-03 at 11 09 11 AM

Perhaps less margin at the right? Or move the legend below? cc: @john-conroy

john-conroy:
Fix #1883. Adding `width: 0px` seems to be the fix and forces another dimensions recalc. Please take a look locally and let me know what you think.

### DIFF
--- a/CHANGELOG-fix-bar-chart-responsiveness.md
+++ b/CHANGELOG-fix-bar-chart-responsiveness.md
@@ -1,0 +1,1 @@
+- Fix homepage bar chart responsiveness.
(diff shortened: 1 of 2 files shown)


mccalluc (approved):
When resizing, it takes a moment for the graph to redraw, and the key can overlap with it, but I think that's fine.

PR #1908, opened 2021-06-10
Mccalluc/readme and qa

mccalluc:
- @tsliaw -- I've added `cells` to the QA routine... and I think the features that are stable probably don't need to be checked as diligently now, but this is a judgement call: Do whatever you think will find the most bugs!
- @Roselkisma -- I've rearranged the README, putting several of the less important parts in fold-downs, and adding a bit of new information at the end. As a new person to the project, you're in a really good place to judge whether this documentation is clear.

### DIFF
--- a/CHANGELOG-readme-and-qa.md
+++ b/CHANGELOG-readme-and-qa.md
@@ -1,0 +1,1 @@
+- Update the README and manual QA instructions.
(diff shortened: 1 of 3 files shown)


tsliaw (approved):

Roselkisma:
The documentation looks clear to me.

PR #1937, opened 2021-06-22
Mccalluc/get all genes

### LINKED ISSUE #1888: Gene type-ahead search Flask API / simple demo
- Add a query to get all names to the demo page.
- Depending on how it performs, consider where it should be stored... Client? Server at startup? Baked into image?
- Add a left substring search flask endpoint.

@mruffalo -- I think you mentioned that the list could be provided in some other way besides hitting the API. How would we keep a static file up to date, if we went that route?

cc @SFD5311 , @tsliaw

mccalluc:
- @ilan-gold: 
  - code review, particularly whether caching the full list and just iterating over it feels ok.
- @tsliaw:
  - Does 10 matches seem ok?
  - Gene and protein name search need to be separate for the UI?
  - Right now it's a substring anywhere, not necessarily at the start of the term... Does that seem right?
 - cc @SFD5311 

Fix #1888

### DIFF
--- a/context/app/routes_cells.py
+++ b/context/app/routes_cells.py
@@ -39,16 +39,10 @@ def _get_protein_ids(app):
     return protein_ids
 
 
-# def _first_n_matches(strings, substring, n):
-#     first_n = islice((s for s in strings if substring in s), n)
-#     return [{'string': s} for s in first_n]
-
-
 def _first_n_matches(strings, substring, n):
     substring_lower = substring.lower()
     # TODO: Python 3.8 allows assignment within comprehensions.
     first_n = list(islice((s for s in strings if substring_lower in s.lower()), n))
-    # return [{'string': s} for s in first_n]
     offsets = [s.lower().find(substring_lower) for s in first_n]
     return [{
         'full': s,
--- a/context/app/routes_cells.py
+++ b/context/app/routes_cells.py
@@ -23,6 +23,8 @@ def _get_client(app):
 
 # TODO: In python 3.9, functools.cache would be better.
 gene_symbols = None
+
+
 def _get_gene_symbols(app):
     client = _get_client(app)
     global gene_symbols
@@ -30,7 +32,10 @@ def _get_gene_symbols(app):
         gene_symbols = [gene["gene_symbol"] for gene in client.select_genes().get_list()]
     return gene_symbols
 
+
 protein_ids = None
+
+
 def _get_protein_ids(app):
     client = _get_client(app)
     global protein_ids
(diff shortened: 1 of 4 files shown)


context/app/routes_cells.py
```diff
@@ -19,6 +21,55 @@ def _get_client(app):
...
+# def _first_n_matches(strings, substring, n):
+#     first_n = islice((s for s in strings if substring in s), n)
+#     return [{'string': s} for s in first_n]
+
+
+def _first_n_matches(strings, substring, n):
+    substring_lower = substring.lower()
+    # TODO: Python 3.8 allows assignment within comprehensions.
+    first_n = list(islice((s for s in strings if substring_lower in s.lower()), n))
+    # return [{'string': s} for s in first_n]
```
mccalluc, line 51:
```suggestion
def _first_n_matches(strings, substring, n):
    substring_lower = substring.lower()
    # TODO: Python 3.8 allows assignment within comprehensions.
    first_n = list(islice((s for s in strings if substring_lower in s.lower()), n))
```

context/app/static/js/pages/CellsAPIDemo/CellsAPIDemo.jsx
```diff
@@ -13,6 +14,24 @@ function CellsAPIDemo(props) {
     <>
       <Alert severity="warning">UI is still being designed, but we want to demonstrate that the API works.</Alert>
 
+      <SectionHeader>Search genes by substring</SectionHeader>
+      <p>
+        <b>Given</b> a subtring from a gene name.
```
ilan-gold, line 19:
```suggestion
        <b>Given</b> a substring from a gene name.
```

context/app/static/js/pages/CellsAPIDemo/CellsAPIDemo.jsx
```diff
@@ -13,6 +14,24 @@ function CellsAPIDemo(props) {
...
+      </p>
+      <p>
+        <b>Return</b> a list of the first few matches.
+      </p>
+      <SearchBySubstring targetEntity="genes" />
+
+      <SectionHeader>Search proteins by substring</SectionHeader>
+      <p>
+        <b>Given</b> a subtring from a protein name.
```
ilan-gold, line 28:
```suggestion
        <b>Given</b> a substring from a protein name.
```

ilan-gold (approved):

tsliaw:
> Does 10 matches seem ok?

I was assuming it was going to be a scrollable list of all the matches, but if that is too much, 10 should be fine.

> Gene and protein name search need to be separate for the UI?

Yes

> Right now it's a substring anywhere, not necessarily at the start of the term... Does that seem right?

For genes, I think it should be matching at the start of the term since they're gene symbols. As for proteins, I think substring anywhere fits better. If you would like to be uniform between query types, then let's do substring anywhere for both.

tsliaw (approved):

PR #1938, opened 2021-06-23
Mccalluc/get all genes autosuggest

### LINKED ISSUE #1924: Gene type-ahead search UI implementation
backend is #1888

mccalluc:
Extend #1937
Fix #1924

@tsliaw : Design review: Is the bolding of the the matching substring useful?
@john-conroy : Code review

Screen Shot 2021-06-23 at 2 34 01 PM

### DIFF
--- a/context/app/routes_cells.py
+++ b/context/app/routes_cells.py
@@ -39,16 +39,10 @@ def _get_protein_ids(app):
     return protein_ids
 
 
-# def _first_n_matches(strings, substring, n):
-#     first_n = islice((s for s in strings if substring in s), n)
-#     return [{'string': s} for s in first_n]
-
-
 def _first_n_matches(strings, substring, n):
     substring_lower = substring.lower()
     # TODO: Python 3.8 allows assignment within comprehensions.
     first_n = list(islice((s for s in strings if substring_lower in s.lower()), n))
-    # return [{'string': s} for s in first_n]
     offsets = [s.lower().find(substring_lower) for s in first_n]
     return [{
         'full': s,
--- a/context/app/routes_cells.py
+++ b/context/app/routes_cells.py
@@ -23,6 +23,8 @@ def _get_client(app):
 
 # TODO: In python 3.9, functools.cache would be better.
 gene_symbols = None
+
+
 def _get_gene_symbols(app):
     client = _get_client(app)
     global gene_symbols
@@ -30,7 +32,10 @@ def _get_gene_symbols(app):
         gene_symbols = [gene["gene_symbol"] for gene in client.select_genes().get_list()]
     return gene_symbols
 
+
 protein_ids = None
+
+
 def _get_protein_ids(app):
     client = _get_client(app)
     global protein_ids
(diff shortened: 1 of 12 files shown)


tsliaw:
Looks good

tsliaw (approved):

context/app/routes_cells.py
```diff
@@ -19,6 +21,54 @@ def _get_client(app):
...
+
+protein_ids = None
+
+
+def _get_protein_ids(app):
+    client = _get_client(app)
+    global protein_ids
+    if protein_ids is None:
+        protein_ids = [protein["protein_id"] for protein in client.select_proteins().get_list()]
+    return protein_ids
```
john-conroy, line 44: How often are we meant to retrieve `gene_symbols`/`protein_ids`?

john-conroy: Also, is it worth refactoring? It seems like they could share code.

context/app/routes_cells.py
```diff
@@ -19,6 +21,54 @@ def _get_client(app):
...
+    substring_lower = substring.lower()
+    # TODO: Python 3.8 allows assignment within comprehensions.
+    first_n = list(islice((s for s in strings if substring_lower in s.lower()), n))
+    offsets = [s.lower().find(substring_lower) for s in first_n]
+    return [{
+        'full': s,
+        'pre': s[:offset],
+        'match': s[offset:offset + len(substring)],
+        'post': s[offset + len(substring):]
+    } for s, offset in zip(first_n, offsets)]
```
john-conroy, line 57: I'd add some unit tests here.

context/app/static/js/pages/CellsAPIDemo/SearchBySubstring.jsx
```diff
@@ -0,0 +1,64 @@
...
+  async function handleChange(event) {
+    const { target } = event;
+    setSubstring(target.value);
+
+    if (target.value === '') {
+      setOptions([]);
+      return;
+    }
+
+    const urlParams = new URLSearchParams();
```
john-conroy, line 23: Is this the start of making queries accessible by URL? Is there a common pattern for managing this?

context/app/routes_cells.py
mccalluc: > How often are we meant to retrieve `gene_symbols`/`protein_ids`?

Once per deployment, and then they would be cached in memory. 
- Slow enough that I think it needs to be pre-fetched.
- Results don't change fast enough that I'd worry about caching on the timescale of a week.
- Down the road, might make warming the cache part of the deployment process.

> Also, is it worth refactoring? It seems like they could share code.

There are memoization decorators that we could use... but since it's not a built-in right now, decided not to add too much magic... could revisit this in the future.

context/app/static/js/pages/CellsAPIDemo/SearchBySubstring.jsx
mccalluc: I'm not sure if I understand your point, but one possibility would be, instead of having all these different URLs scattered in different components, we have a central service with methods that encapsulate the HTTP details. I've just filed #1945.... but if that doesn't get to your concern, let me know?

PR #1940, opened 2021-06-23
Cells Sliders

mccalluc:
@john-conroy : Code review
@tsliaw : Design review: Are sliders that look like this ok? (Would like to keep to the defaults where possible.) Width/positioning will change when the whole page is handled.
Screen Shot 2021-06-23 at 3 58 29 PM

### DIFF
--- a/context/app/static/js/pages/CellsAPIDemo/DatasetsSelectedByGene.jsx
+++ b/context/app/static/js/pages/CellsAPIDemo/DatasetsSelectedByGene.jsx
@@ -3,6 +3,7 @@ import React, { useState } from 'react';
 import Paper from '@material-ui/core/Paper';
 import TextField from '@material-ui/core/TextField';
 import Button from '@material-ui/core/Button';
+import Slider from '@material-ui/core/Slider';
 
 import ResultsTable from './ResultsTable';
 
@@ -38,8 +39,6 @@ function DatasetsSelectedByGene(props) {
     const { name } = target;
     const setFields = {
       geneName: setGeneName,
-      minGeneExpression: setMinGeneExpression,
-      minCellPercentage: setMinCellPercentage,
     };
     setFields[name](event.target.value);
   }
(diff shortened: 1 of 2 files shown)


context/app/static/js/pages/CellsAPIDemo/DatasetsSelectedByGene.jsx
```diff
@@ -38,29 +39,36 @@ function DatasetsSelectedByGene(props) {
...
     setFields[name](event.target.value);
   }
 
   return (
     <Paper>
       <TextField label="gene name" value={geneName} name="geneName" variant="outlined" onChange={handleChange} />
-      <TextField
-        label="min gene expression"
+
+      <p id="min-gene-expression-label">Minimun gene expression</p>
```
tsliaw, line 50:
```suggestion
      <p id="min-gene-expression-label">Minimum gene expression</p>
```

context/app/static/js/pages/CellsAPIDemo/DatasetsSelectedByGene.jsx
```diff
@@ -38,29 +39,36 @@ function DatasetsSelectedByGene(props) {
...
+        max={100}
+        onChange={(e, value) => {
+          setMinGeneExpression(value);
+        }}
+        aria-labelledby="min-gene-expression-label"
       />
-      <TextField
-        label="min cell percentage"
+
+      <p id="min-cell-percentage-label">Minimun cell percentage</p>
```
tsliaw, line 61:
```suggestion
      <p id="min-cell-percentage-label">Minimum cell percentage</p>
```

tsliaw (commented):
Are you planning to add number labels to these sliders in a different PR? The style of the slider looks fine since it's from Materials, but the sliders do need min/max number labels and when dragging/hovering, the current number should appear. 

Screen Shot 2021-06-23 at 1 28 11 PM

[Materials](https://material.io/components/sliders#discrete-sliders)

context/app/static/js/pages/CellsAPIDemo/DatasetsSelectedByGene.jsx
john-conroy, line 50: Use label instead of p?

john-conroy: I think MUI has a label component for inputs/forms.

context/app/static/js/pages/CellsAPIDemo/DatasetsSelectedByGene.jsx
```diff
@@ -38,29 +39,36 @@ function DatasetsSelectedByGene(props) {
...
     const setFields = {
       geneName: setGeneName,
-      minGeneExpression: setMinGeneExpression,
-      minCellPercentage: setMinCellPercentage,
     };
     setFields[name](event.target.value);
   }
 
   return (
     <Paper>
```
john-conroy, line 47: Should we use a form here? There are differing opinions on their necessity.

mccalluc: I'm hoping that's something you can sort out when you take these little pieces and make the real thing out of them!

context/app/static/js/pages/CellsAPIDemo/DatasetsSelectedByGene.jsx
mccalluc: Sure, I'll use that instead ... but John is going to have a lot of these to fix, in any event.

mccalluc:
Treating "comment" as "good-to-merge-with-tweaks", so I'll merge.

PR #1942, opened 2021-06-24
Move towards allowing more than gene/rna searches

### LINKED ISSUE #1891: Allow codex / atac to be searched in cells demo

mccalluc:
Filing as draft for now: If this is the right course, I will want to add UI fields for modality and target entity (if there's not a better name).
- @john-conroy : Please comment on API design... does adding more parameters seem like the best choice? Or maybe part of this could go in the URL path?
- @tsliaw : Am I correct in thinking that we want to be able to search by genes *or* proteins, rna *or* atac? Does the UI design support all these possibilities? Are there others I should be thinking of?

Will fix #1891.


tsliaw:
>  Am I correct in thinking that we want to be able to search by genes or proteins, rna or atac? Does the UI design support all these possibilities? Are there others I should be thinking of?

If I understood what you said correctly, then yes that is what we want. Right now the UI design allows for searching by genes or protein, and within genes, search by rna or atac. Don't believe there are other combinations at the moment for the first iteration.

john-conroy (approved):
I'm still working to stitch together the terms used in the design and the vars here, but `query type` should be in the URL path and everything else makes sense as a param. So I think that would mean moving `target_entity` to the path?

PR #1950, opened 2021-06-28
Fix Offsets Location for Cytokit AnnData

ilan-gold:
I think my assumption that the `offsets.json` file is in the wrong place from #1904 was wrong, and based on the fact that the offsets file is in that place from the old SPRM which led me to believe it was run on the non-image pyramid, which was to be somewhat expected given that we hadn't seen a completely clean run yet.  To recap, the only runs we had seen included the incorrect offsets file, but in the right place, which led me to believe it was in the wrong place given past behavior.  Hopefully the change in test case behavior and the two different new `_get_ometiff_image_wrapper` (one for the old way, one for the new) functions highlight how this current change makes (more) sense.

### DIFF
--- a/context/app/api/vitessce_confs/base_confs.py
+++ b/context/app/api/vitessce_confs/base_confs.py
@@ -189,14 +189,6 @@ class SPRMViewConfBuilder(ImagePyramidViewConfBuilder):
         found_image_file = found_image_files[0]
         return found_image_file
 
-    def _get_ometiff_image_wrapper(self, found_image_file):
-        img_url, offsets_url = self._get_img_and_offset_url(
-            found_image_file, self._imaging_path_regex,
-        )
-        return OmeTiffWrapper(
-            img_url=img_url, offsets_url=offsets_url, name=self._image_name
-        )
-
 
 class SPRMJSONViewConfBuilder(SPRMViewConfBuilder):
     def __init__(self, entity, nexus_token, is_mock=False, **kwargs):
@@ -224,6 +216,14 @@ class SPRMJSONViewConfBuilder(SPRMViewConfBuilder):
                 "data_type": dt.EXPRESSION_MATRIX,
             },
         ]
+    
+    def _get_ometiff_image_wrapper(self, found_image_file):
+        img_url, offsets_url = self._get_img_and_offset_url(
+            found_image_file, self._imaging_path_regex,
+        )
+        return OmeTiffWrapper(
+            img_url=img_url, offsets_url=offsets_url, name=self._image_name
+        )
 
     def get_conf_cells(self):
         found_image_file = self._check_sprm_image(self._get_full_image_path())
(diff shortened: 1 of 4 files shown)


ilan-gold:
I think a separate PR could include adding a check for the offsets file to the Vitessce Conf generation to make sure it is there without 404'ing, but that might be redundant since we do have nice error handling for it in Vitessce

ilan-gold:
Also, `http://localhost:5001/browse/dataset/d7cfb9e8beba2b6e65ce587b0b74b5bb` on PROD should just "work" now

mccalluc (changes requested):
What I see in this PR is a method being moved down from a class to its two subclasses. There don't seem to be any other references to `SPRMViewConfBuilder`, so I'm not clear on why this produces a change in behavior... Maybe there's a change to the method that I'm not seeing?

I also get the sense that some of these classes are there to support pipeline outputs formats that we're moving away from? Getting some of this context that's clear to you into documentation on the classes might help.

Zooming might be good?

ilan-gold:
> What I see in this PR is a method being moved down from a class to its two subclasses. There don't seem to be any other references to `SPRMViewConfBuilder`, so I'm not clear on why this produces a change in behavior... Maybe there's a change to the method that I'm not seeing?
> 
> I also get the sense that some of these classes are there to support pipeline outputs formats that we're moving away from? Getting some of this context that's clear to you into documentation on the classes might help.

I can make this clearer.
 
> > To recap, the only runs we had seen included the incorrect offsets file
> 
> It's "had seen", because the example this morning should be a _correct_ offsets file? Would it be worthwhile to get a pipeline person so look at the test fixtures in this PR? It feels like you've had to make inferences about the right structure.
> 
> Zooming might be good?

Right, this morning was the correct offsets file.  Previously they had run the `offsets` pipeline on the wrong file, but deposited it in the right place, which is a different place than the old SPRM pipeline, all of which led me to believe that the placement was also wrong.

ilan-gold:
@mccalluc I updated the changes so hopefully it's a bit clearer what's being changed.  I will open separate PR's for a few other small things, like docstrings.

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -189,9 +189,9 @@ def _check_sprm_image(self, path_regex):
         found_image_file = found_image_files[0]
         return found_image_file
 
-    def _get_ometiff_image_wrapper(self, found_image_file):
+    def _get_ometiff_image_wrapper(self, found_image_file, found_image_path):
         img_url, offsets_url = self._get_img_and_offset_url(
-            found_image_file, self._imaging_path_regex,
+            found_image_file, found_image_path,
```
mccalluc, line 194:
```suggestion
            found_image_file, re.escape(found_image_path),
```
If the second arg is treated as an RE, let's escape it?

mccalluc (commented):
Thanks! I think that is better than copying the whole method. You might want to use named arguments: they are of similar types, and the semantics don't make it clear which is first and which second... but it's a judgement call.

Suggest adding a `re.escape`... but if I'm confused, and it's not really treated as an regex, never mind.

context/app/api/vitessce_confs/base_confs.py
ilan-gold: Ok!

PR #1953, opened 2021-06-28
Vitessce 1.1.11

ilan-gold:
Closes 1673

To test:

- [Lightsheet](http://localhost:5001/search?mapped_data_types[0]=Lightsheet%20Microscopy&entity_type[0]=Dataset) should allow 3D and be functional
- [IMC](http://localhost:5001/search?mapped_data_types[0]=Imaging%20Mass%20Cytometry&entity_type[0]=Dataset) should be the same (both of the above, only the datasets in QA)

Other than that everything should look/work identical to what is on PROD right now with the exception of the small UI change of the z-slider placement (for example, for [seqFish](http://localhost:5001/browse/dataset/eaad5ed4b3aa2186bd87e05ff28b593e)) - importantly, there should be no visual cues for 3D outside of the above two data types even though the unstitched CODEX and seqFISH both do have z-stacks.

### DIFF
--- a/context/app/api/vitessce_confs/assay_confs.py
+++ b/context/app/api/vitessce_confs/assay_confs.py
@@ -66,7 +66,8 @@ class SeqFISHViewConfBuilder(ImagingViewConfBuilder):
             pos_name = self._get_pos_name(images[0])
             vc = VitessceConfig(name=pos_name)
             dataset = vc.add_dataset(name=pos_name)
-            for img_path in sorted(images, key=self._get_hybcycle):
+            sorted_images = sorted(images, key=self._get_hybcycle)
+            for img_path in sorted_images:
                 img_url, offsets_url = self._get_img_and_offset_url(
                     img_path, IMAGE_PYRAMID_DIR
                 )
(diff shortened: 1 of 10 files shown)


context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -78,7 +79,11 @@ def get_conf_cells(self):
                     )
                 )
             dataset = dataset.add_object(MultiImageWrapper(image_wrappers))
-            vc = self._setup_view_config_raster(vc, dataset)
+            vc = self._setup_view_config_raster(
+                vc,
+                dataset,
+                disable3d=[self._get_hybcycle(img_path) for img_path in sorted_images]
```
mccalluc, line 85:
```suggestion
                disable_3d=[self._get_hybcycle(img_path) for img_path in sorted_images]
```
but beyond that, it's really not clear to me what this parameter does... Thought it was a boolean at first. I'd like to use nouns for things / verbs for functions.

mccalluc (commented):
I think the variable name could be more clear. Beyond that looks reasonable, but I'm also not sure what's really going on.

context/app/api/vitessce_confs/assay_confs.py
ilan-gold: It's just a pass-through prop to the vitessce layer.  The camel-case thing can be fixed though in the Python.

ilan-gold: It allows us to "turn off" the option for 3D for seqfish and CODEX because we do not want to allow users to enter 3D mode for those assays.

john-conroy (approved):
I couldn't find any issues with the examples I looked at in 3D.

PR #1960, opened 2021-06-29
Add Docstrings/Comments to Vitessce Classes

### LINKED ISSUE #1951: Add Docstrings for all Vitessce Conf Classes
Currently on the base class has docstrings (and doctests) but all classes should at the minimum be documented.

ilan-gold:
Fixes #1951 and following up on [this comment](https://github.com/hubmapconsortium/portal-ui/pull/1950#pullrequestreview-693951261)

Maybe you could go through Chuck and tell me if this is clear/helpful?  And where to improve?

### DIFF
--- a/context/app/api/vitessce_confs/assay_confs.py
+++ b/context/app/api/vitessce_confs/assay_confs.py
@@ -165,7 +165,7 @@ class RNASeqViewConfBuilder(AbstractScatterplotViewConfBuilder):
 class ATACSeqViewConfBuilder(AbstractScatterplotViewConfBuilder):
     """Wrapper class for creating a JSON-backed scatterplot for "first generation" ATAC-seq data like
     https://portal.hubmapconsortium.org/browse/dataset/d4493657cde29702c5ed73932da5317c
-    from h5ad-to-arrow.cwl (August 2020 release).
+    from h5ad-to-arrow.cwl.
     """
     def __init__(self, entity, nexus_token, is_mock=False):
         super().__init__(entity, nexus_token, is_mock)
--- a/context/app/api/vitessce_confs/assay_confs.py
+++ b/context/app/api/vitessce_confs/assay_confs.py
@@ -44,7 +44,8 @@ from .paths import (
 
 
 class SeqFISHViewConfBuilder(ImagingViewConfBuilder):
-    """Wrapper class for generating Vitessce configurations, one per position, with the hybridization cycles
+    """Wrapper class for generating Vitessce configurations,
+    one per position, with the hybridization cycles
     grouped together per position in a single Vitessce configuration.
     """
     def get_conf_cells(self):
@@ -108,7 +109,8 @@ class CytokitSPRMViewConfigError(Exception):
 
 
 class TiledSPRMViewConfBuilder(ViewConfBuilder):
-    """Wrapper class for generating many "first generation" non-stitched JSON-backed SPRM Vitessce configurations,
+    """Wrapper class for generating many "first generation"
+    non-stitched JSON-backed SPRM Vitessce configurations,
     one per tile per region, via SPRMJSONViewConfBuilder.
     """
     def get_conf_cells(self):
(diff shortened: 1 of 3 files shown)


context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -105,6 +109,10 @@ class CytokitSPRMViewConfigError(Exception):
 
 
 class TiledSPRMViewConfBuilder(ViewConfBuilder):
+    """Wrapper class for generating many "first generation"
```
mccalluc, line 112: Do we have an approximate date for the last dataset of this form? Maybe comment on what replaces it?

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -133,6 +141,9 @@ def get_conf_cells(self):
 
 
 class RNASeqViewConfBuilder(ScatterplotViewConfBuilder):
+    """Wrapper class for creating a JSON-backed scatterplot for "first generation" RNA-seq data like
+    https://portal.hubmapconsortium.org/browse/dataset/c019a1cd35aab4d2b4a6ff221e92aaab
```
mccalluc, line 145: Ditto, but linking to an example is great!

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -42,6 +45,8 @@ def get_conf_cells(self):
 
     def _replace_url_in_file(self, file):
         """Replace url in incoming file object
+        :param dict file: File dict which will have its rel_path replaced by url
+        rtype: dict The file with rel_path replaced by url
```
mccalluc, line 49:
```suggestion
        :rtype: dict The file with rel_path replaced by url
```
I'm not a master of pydoc, but change syntax, here and below.

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -74,6 +82,18 @@ def _build_assets_url(self, rel_path, use_token=True):
...
+        This is needed for non-public zarr stores because the client forms URLs for zarr chunks,
+        not the above _build_assets_url function.
+
+        >>> entity = {"uuid": "uuid", "status": "QA"}
+        >>> vc = ViewConfBuilder(entity=entity, nexus_token='nexus_token', is_mock=True)
+        >>> vc._get_request_init()
+        {'headers': {'Authorization': 'Bearer nexus_token'}}
+        >>> entity = {"uuid": "uuid", "status": "Published"}
+        >>> vc = ViewConfBuilder(entity=entity, nexus_token='nexus_token', is_mock=True)
+        >>> vc._get_request_init() # Returns None because dataset is Published i.e public
```
mccalluc, line 95:
```suggestion
        >>> assert vc._get_request_init() is None # Returns None because dataset is Published i.e public
```

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -166,6 +207,7 @@ def get_conf_cells(self):
             raise FileNotFoundError(message)
         vc = VitessceConfig(name="HuBMAP Data Portal")
         dataset = vc.add_dataset(name="Visualization Files")
+        # The sublcass initializes _files in its __init__ method
```
mccalluc, line 210: This is helpful: From the top, one thing that's hard to get a handle on is which classes, like this one, are abstract. perhaps add `Abstract...` to the class name? There's also python's [`ABC`](https://docs.python.org/3/library/abc.html)... worth reading about, but I don't know that they'd give us much.

mccalluc (commented):
This is great -- Thank you!
- Check the pydoc syntax for `rtype`.
- Consider making abstract classes explicit.
- Is the "first generation" "second generation" distinction one that's being made outside? Wonder if approximate dates would help, or pointing from one format to what succeeded it?

context/app/api/vitessce_confs/assay_confs.py
ilan-gold: I can try to add some dates to these

ilan-gold:
The "generation" difference is apparent if you know where to look (besides the visualization just looking different)....it's the CWL/commit hash for the pipeline.

context/app/api/vitessce_confs/assay_confs.py
ilan-gold: Hmm, there's no date filter in the UI so it's hard.  I'll try to ballpark it but also link to the cwl.

PR #1967, opened 2021-06-30
Change menu

### LINKED ISSUE #1966: Update docs menu
The current menu is:
- Overview
- Software
- FAQ
- About

@computationdoc says:
> top level Documentation menu only would have three items in new set up: 
> - the new Documentation submenu, 
> - the current FAQ submenu,
> - the current About submenu

Questions:
- By "Documentation submenu", do you mean mean ["Technical"](https://portal.hubmapconsortium.org/docs/technical)?
- Should the current ["Overview"](https://portal.hubmapconsortium.org/docs) page be removed?

mccalluc:
Fix #1966.

Filing as draft. Problems need to be resolved before this can be merged.
- There is no longer anything at `/docs`: It is not good to make pages that previously worked into 404s. If the `technical.md` is really conceived as the entry point, perhaps it should be at `/docs`? Or maybe `/docs` should redirect to `/docs/technical`? We don't have redirection machinery in place, so that would be new development effort.
- There are a lot of documentation pages (`/docs/datasets`, `/docs/release-summer-2020`, `/docs/data-states`, `/docs/assays`) that are no longer referenced. Should the pages be removed (and replaced with redirects), or should links to these pages be added to the documentation?

The specs I have had up this this point have be late, vague, changing, and inconsistent. Suggest that there be reflection on the process to avoid repeating this.


computationdoc:
Agree that the specs were not properly requested due to misunderstanding on my part of how the documentation and ui menus are different bits of technology and not following that through with proper requests.

I have just reviewed everything with Peter - note that I believe some of the noted docs pages ARE referenced in the new page and some are deprecated so please review with Peter who will bring this to simple coherence now for you.

Then, yes, we need to reflect on the process failure. In that regard I note that there was a specific discussion in which you were known to be not available - one key thing to take forward is not going deep in an area of ownership by a technical person without them present in the discussion - and certainly when they are brought back in to then resolve to consistent, clear, stable, spec at earliest possible time.

JCS

pecan88:
Hi @mccalluc  - so, apologies in advance as I may be mixing content unintentionally across #1966 and #1967  - hopefully this summary of a conversation I had with @computationdoc just now will help clear some of the questions up.

For the menu, we'd like it to list the following for the portal navigation where "Documentation" is in the top bar nav ...

Documentation
-> Technical
-> FAQ
-> About


Re: "no longer anything at /docs": agree and for expediency, my thinking is we place the technical.md content at /docs/ to avoid the 404 and we submit a feature request / enhancement to add redirection in the next sprint, and then adjust as needed as part of regular sprint planning & work.  Happy to follow your suggestions / guidance here as well if you think there's a more correct way.   If you prefer, we can leave the ".md" name as "technical.md" or rename it to "documentation.md".  I will make content page specific updates via a future PR as I know this may create some inconsistencies between the URL text.md and the title text on the page. 


Re: "no longer referenced" doc pages:
"/docs/datasets" exists and is linked within the "technical.md" document content
"/docs/release-summer-2020" exists and we can archive / remove per your recommendation / desire
"/docs/data-states" exists and is linked within the "technical.md" document content
"/docs/assays" exists and is linked within the "technical.md" document content


I will confirm / add @computationdoc  as a reviewer to this issue to ensure we're following something more like GitHub change "protocol" - please correct me as needed.

computationdoc:
Thank you - Peter will ensure I press the right buttons to approve at the right time, but what he’s stated here is fully correct and approved by me - JCS

mccalluc:
@pecan88 - This is inconsistent.

> my thinking is we place the technical.md content at /docs/ to avoid the 404

`/docs` is served by `docs.md`, which is being deleted in this PR as it stands. Moving `technical.md` from `portal-docs` to `portal-ui` is doable, but then you request:

>  If you prefer, we can leave the ".md" name as "technical.md" or rename it to "documentation.md"

That suggests you still want `/docs/technical` or `/docs/documentation`. I just need a single, consistent spec, and bouncing it back to me does not help.

ngehlenborg:
@mccalluc We should implement this now and we will deal with the other issues through the regular process:

```
Documentation
   Technical -> https://portal.hubmapconsortium.org/docs/technical
   FAQ -> https://portal.hubmapconsortium.org/docs/faq
   About -> https://portal.hubmapconsortium.org/docs/about
```

computationdoc (approved):

pecan88 (approved):

ngehlenborg (approved):

PR #1982, opened 2021-07-01
Roselkisma/flaskpg

Roselkisma:
I used useSearchData to get some numbers from the Elasticsearch to create a table about the donor's race. #1930


end-to-end/artillery/queries/datasets.json
```diff
@@ -65,9 +65,9 @@
       "filter": { "term": { "entity_type.keyword": "Dataset" } },
       "aggs": {
         "donor.mapped_metadata.sex.keyword": {
-          "terms": { "field": "donor.mapped_metadata.sex.keyword", "size": 5 }
+          "terms": { "field": "donor..keyword", "size": 5 }
```
john-conroy, line 68: I imagine this got changed while you were looking for example queries. These are queries used in load testing that should be kept as is. I would revert these changes.

CHANGELOG-table.md
```diff
@@ -0,0 +1 @@
+- I created a table about the donor's race.
```
mccalluc, line 1:
```suggestion
- Created a table about the donor's race.
```
Style: In the context of the changelog, we usually just have lots of sentence fragments. The "I" doesn't matter.

CHANGELOG-hello.md
```diff
@@ -0,0 +1 @@
+- added Hello World to a page.
```
mccalluc, line 1: Get rid of this file: only need one CHANGELOG per PR.

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -0,0 +1,92 @@
...
+             "terms": {
+                 "field": "mapped_metadata.sex.keyword"
+             }
+         }
+     }
+ }
+
+ */
+
+const donorSexQuery = {
```
mccalluc, line 36: Either race or sex is a fine example to start with, but make sure the variable name matches the content.

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -0,0 +1,92 @@
...
+      terms: {
+        field: 'mapped_metadata.race.keyword',
+      },
+    },
+  },
+};
+
+// change to match the data
+const columns = [
+  { id: 'whie', label: 'White' },
```
mccalluc, line 58:
```suggestion
  { id: 'white', label: 'White' },
```

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -0,0 +1,92 @@
...
+  { id: 'black', label: 'Black or African American' },
+];
+
+function Vis() {
+  const { elasticsearchEndpoint, nexusToken } = useContext(AppContext);
+  // eslint-disable-next-line no-unused-vars
```
mccalluc, line 64:
```suggestion
```
Maybe searchData wasn't used in an earlier version? shouldn't need to disable the warning now.

context/app/static/js/pages/Vis/Vis.jsx
mccalluc: ... but see comments below.

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -0,0 +1,92 @@
...
+              <TableRow>
+                {columns.map((column) => (
+                  <HeaderCell key={column.id}>{column.label}</HeaderCell>
+                ))}
+              </TableRow>
+            </TableHead>
+            <TableBody>
+              <TableRow>
+                <TableCell> {searchData.aggregations['mapped_metadata.race'].buckets[0].doc_count}</TableCell>
+                <TableCell> {searchData.aggregations['mapped_metadata.race'].buckets[1].doc_count}</TableCell>
```
mccalluc, line 82: You assume that the order of the columns you've given above will match the order of the buckets... with more data, it might not. You need to pull the label for the bucket from the `searchData`, rather than hard coding.

And then in `HeaderCell`, it would be fine to use the `label` for the `key`... The `key` just needs to be unique, and when you have pulled the bucket names from `searchData`, they will be.

end-to-end/artillery/queries/datasets.json
```diff
@@ -65,9 +65,9 @@
...
         },
-        "donor.mapped_metadata.sex.keyword_count": {
+        "donor..keyword_count": {
```
mccalluc, line 70: I don't think you want to touch this file at all. Revert to original.

mccalluc (changes requested):
Lots of little changes, but looks like the right direction. CI is complaining about python formatting:

```
  ./context/app/routes_main.py:216:1: E302 expected 2 blank lines, found 1
  ./context/app/routes_main.py:230:1: E302 expected 2 blank lines, found 1
  Try: autopep8 --in-place --aggressive -r . --exclude node_modules,ingest-validation-tools
  Error
```

context/app/routes_main.py
```diff
@@ -209,6 +209,19 @@ def dev_search():
         flask_data=flask_data
     )
```
mccalluc, line 211:
```suggestion


```

context/app/routes_main.py
```diff
@@ -209,6 +209,19 @@ def dev_search():
...
+    flask_data = {
+        'endpoints': _get_endpoints(),
+        'title': title
+    }
+    return render_template(
+        'pages/base_react.html',
+        title=title,
+        types=entity_types,
+        flask_data=flask_data
+    )
```
mccalluc, line 225:
```suggestion


```

mccalluc (changes requested):
This is fix the pyflake problem.

end-to-end/artillery/queries/datasets.json
Roselkisma: I did not touch this file. This is how it looks on my computer. I do not know why it looks different after I push the changes to github, because I did not do anything to this file. 
 
Screen Shot 2021-07-07 at 1 06 19 PM

end-to-end/artillery/queries/datasets.json
Roselkisma: I did not touch this file. This is how it looks on my computer. I do not know why it looks different after I push the changes to github, because I did not do anything to this file. 
 
Screen Shot 2021-07-07 at 1 06 19 PM

end-to-end/artillery/queries/datasets.json
john-conroy: If that's the case, you can use `git pull` to pull down the current code and change it back and commit it afterwards.

john-conroy: There are other ways to do it, but that's probably the most straightforward.

Roselkisma (commented):
I made all the changes

end-to-end/artillery/queries/datasets.json
mccalluc: @Roselkisma -- We understand that this is what your local version has, the problem is that this is a change from what is in master. (Red is master / green is yours.) It was an accident, it just needs to be undone. Not a big deal.

mccalluc (changes requested):
I think there is some confusion. Let's find time to talk?

end-to-end/artillery/queries/datasets.json
mccalluc, line 68:
```suggestion
          "terms": { "field": "donor.mapped_metadata.sex.keyword", "size": 5 }
```

end-to-end/artillery/queries/datasets.json
mccalluc, line 70:
```suggestion
        "donor.mapped_metadata.sex.keyword_count": {
```

mccalluc (changes requested):
@Roselkisma -- Here are the last two changes that we discussed. If it's ok, I'd like to get this PR merged so we switch the default branch.

mccalluc (approved):

PR #1987, opened 2021-07-06
Mccalluc/docs redirects

### LINKED ISSUE #1969: Docs redirect mechanism for `/docs`
To support changes in #1967: When docs pages are moved or removed, we want to make sure we don't just break links. Suggest something like this: a file like `docs.redirect` would handle the `/docs` route: The contents would be read, and a redirect is served to user.

If there is an alternative that is less home-baked / weird, I'd be happy to do that instead.

mccalluc:
Fix #1969

### DIFF
--- a/CHANGELOG-redirects.md
+++ b/CHANGELOG-redirects.md
@@ -1,0 +1,1 @@
+- Going forward, if a docs page is no longer needed, instead of just deleting it, replace `whatever.md` with `whatever.redirect`, and change the content to the new path.
--- a/context/app/markdown/docs.redirect
+++ b/context/app/markdown/docs.redirect
@@ -1,0 +1,1 @@
+/docs/technical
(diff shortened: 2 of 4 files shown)


john-conroy (commented):
What's the motivation behind this?

mccalluc:
@john-conroy 
> What's the motivation behind this?

At the end of the kerfuffle last week, the new `/docs/technical` page was in place... and in Jonathan's view, it's now the main entry point to the documentation, and the old link to `/docs` was removed from the  banner menu.

I think a redirect would be good, in case someone follows the old link, perhaps from search results... but maybe I'm worrying unnecessarily?

john-conroy:
Fair enough, I was curious if there was some requirement to preserve docs.

john-conroy (approved):
Should we document this pattern somewhere outside of the changelog?

mccalluc:
@john-conroy 
> Should we document this pattern somewhere outside of the changelog?

Now that it's merged, I'll add a note to `portal-docs`, and slack the folks that typically edit it.

Issue #1979, opened 2021-07-01
Migrate to new ES names


mccalluc:
The table below shows all the normalized fields (left column) with duplicate data and their current non-normalized fields:

| Neo4j field (normalized) | Existing ES field (to be deprecated) |
| ----------- | ----------------- |
| hubmap_id |  display_doi |
| submission_id | hubmap_display_id |
| (Dataset) ingest_metadata | metadata |
| image_file_metadata | portal_uploaded_image_files |
| (Donor) label | lab_name |
| created_timestamp | create_timestamp |

_Originally posted by @yuanzhou in https://github.com/hubmapconsortium/search-api/issues/341#issuecomment-871516477_

yuanzhou:
Related to this migration: once it's completed in portal-ui, I'll need to bump the ES fields limit back down to 5000. It was increased to 6000 as a temporary fix to the https://github.com/hubmapconsortium/search-api/issues/338.

mccalluc:
Update:
- `display_doi` and `create_timestamp` are done. 
- `hubmap_display_id` is not present: `grep hubmap_display_id -r . --exclude-dir context/node_modules`
- `lab_name` is not present: `grep lab_name -r . --exclude-dir context/node_modules`
- `portal_uploaded_image_files` is not present: `grep portal_uploaded_image_files -r . --exclude-dir context/node_modules`

That leaves `metadata` -> `ingest_metadata`. Because of the different ways `metadata` can be used, I can't just search and replace, so this might be tricky, but I'll get started.

yuanzhou:
@mccalluc I did a quick check using to verify:
- `hubmap_display_id` presents in https://portal.test.hubmapconsortium.org/browse/dataset/b6eba6afe660a8a85c2648e368b0cf9f.json and it has the same value as `submission_id`
- `lab_name` presents in https://portal.test.hubmapconsortium.org/browse/donor/95a62662006235867084078a1e10bebf.json and it has the same value as `label`. `portal_uploaded_image_files` also presents in this same json.

Does this mean even though they present but they are not being used based on your `grep` results?

mccalluc:
@yuanzhou : I'm not sure what you're trying to demonstrate with the linked examples, except that the fields are in fact duplicated? If you can deploy the index without duplicated fields to dev or test or stage, so we can preview it, that would be appreciated. Closing this issue, since we decided that `metadata` doesn't need to be touched.

PR #1990, opened 2021-07-06
display_doi -> hubmap_id

mccalluc:
Just a search and replace: Seems to work locally, but have hardly been exhaustive.
Towards #1979


john-conroy (approved):
So both of these values exist in the metadata now?

PR #1996, opened 2021-07-07
Mccalluc/global warning

### LINKED ISSUE #1995: "Partial availability" message
If an envvar is set (TBD) show a banner with that message. Right now, because of a power outage at PSC, some things don't work, but we don't want to go down to maintenance mode.

mccalluc:
@john-conroy : Filing as draft.
- With this in addition to the endpoints, I should introduce a function that creates a starter dict for `flaskData`.
- Does this feel like the right place to render the message? As far as the logic goes, I'd prefer not to carry it down into the `Route`... but on the other hand, I think the presentation would be better if this were deeper in the DOM.

Fix #1995

### DIFF
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -40,9 +40,13 @@ def _get_endpoints():
     }
 
 
+def _get_global_alert():
+    return current_app.config.get('GLOBAL_ALERT')
+
+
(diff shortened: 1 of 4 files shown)


mccalluc:
Closing PR, and will reopen from `main`.

PR #1997, opened 2021-07-07
Clear localstorage during QA

### LINKED ISSUE #1964: Add clear-DUA to manual QA bookmarklet
We want to make it easy for the QA person to hit all the functionality of the site.

mccalluc:
Fix #1964, but more general: Just clear all of local storage when we QA... This will also clear lists, but that seems ok to me?


mccalluc:
@tsliaw -- Going to go ahead and merge this so we can switch the default branch.

PR #2003, opened 2021-07-08
Mccalluc/global warning

### LINKED ISSUE #1995: "Partial availability" message

mccalluc:
This would be useful if half the hybrid cloud goes down again. Filing as draft because the display of the warning isn't right yet.... wanted to check the direction with @john-conroy before going farther.

Fix #1995


john-conroy (commented):
What would be the entire process for adding a message? Updating the app conf file and docker?

mccalluc:
> What would be the entire process for adding a message? Updating the app conf file and docker?

Update app conf, and restart the container: Would not need to make a new image. 

(Might be possible to restart at a lower level than the container, [perhaps uWSGI](https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html#standard-default-boring-graceful-reload-aka-sighup), but I think we should leave that to IEC if they want something more complicated.)

I'll fix the build and the UI.

mccalluc:
Screen Shot 2021-07-13 at 11 58 36 AM Screen Shot 2021-07-13 at 11 58 55 AM

There seems to be a bit more vertical space than necessary on the homepage, but I don't think it's something I want to introduce special styling for, since other pages look good. 
@john-conroy - Code review.
@yuanzhou - Interface review: If there's another partial outage / partial downtime / whatever, IEC will be able to use this?

john-conroy:
Screen Shot 2021-07-15 at 2 55 48 PM

The alert goes outside the bounds we set for most content when I run it locally. Otherwise looks good.

yuanzhou:
@mccalluc I normally bring up the maintenance of portal-ui on PROD for:
- portal-ui redeployment or release since the portal-ui itself will be down for a short time
- backend API (search-api, entity-api, uuid-api) release
- full reindex of Elasticsearch

When would be good to use this global warning rather then the maintenance page? With this global warning message, the portal-ui will still function though.

mccalluc:
@yuanzhou 

> When would be good to use this global warning rather then the maintenance page?

No. This is not intended to replace the maintenance page, but it does give you more flexibility in communicating with users. Situations where it might be used:
- Part of the hybrid cloud is down. If the portal is in a state were it could still be of some use, we'd prefer not to totally disable it.
- We know there will be a scheduled outage, and we'd like to give advance notice.
- If, in the future, the whole portal has been supplanted by some other service, we could have a deprecation notice here.

yuanzhou (approved):
Looks good to me and I'll add this configuration for all environments:

````
#GLOBAL_ALERT_MD = 'Alert message here! [More info!](http://example.com/)'
````

mccalluc:
@john-conroy : Styling is fixed, and I think this is ready... but there's something weird on the preview pages: Unless I add an explicit `z-index`, it doesn't show up. If you have any ideas, happy to clean up, but I don't think it's worth sinking any time into.

john-conroy (approved):

PR #2005, opened 2021-07-12
Mccalluc/derived from sample

### LINKED ISSUE #2002: "Derived from sample" persists on search page, even when filter is removed.
Visit a page like https://portal.hubmapconsortium.org/search?ancestor_ids[0]=9c13d044a1529cb1f2596bc0b0325dbb&entity_type[0]=Dataset, and then remove the `Ancestor ID: 9c13d044a1529cb1f2596bc0b0325dbb` filter. It disappears, and the search is unconstrained, but the note "Derived from sample HBM998.KNJQ.234" in still above the search input box.

mccalluc:
Fix #2002. Not a high priority, but investigating the bug drifted into fixing the bug.

### DIFF
--- a/context/app/static/js/components/Search/SearchWrapper.jsx
+++ b/context/app/static/js/components/Search/SearchWrapper.jsx
@@ -64,15 +64,13 @@ function SearchWrapper(props) {
       setMessage(null);
       return query;
     }
-    const provMusts = musts.filter(
-      (must) => 'ancestor_ids.keyword' in must.term || 'descendant_ids.keyword' in must.term,
-    );
+    const labels = {
+      'ancestor_ids.keyword': 'Derived from',
+      'descendant_ids.keyword': 'Ancestor of',
+    };
+    const provMusts = musts.filter((must) => Object.keys(labels).some((labelKey) => labelKey in must.term));
     setMessage(
       provMusts.map((must) => {
-        const labels = {
-          'ancestor_ids.keyword': 'Derived from',
-          'descendant_ids.keyword': 'Ancestor of',
-        };
         return Object.entries(must.term).map(([k, v]) => (
           <LookupEntity uuid={v} elasticsearchEndpoint={elasticsearchEndpoint} nexusToken={nexusToken}>
             <SearchNote label={labels[k]} />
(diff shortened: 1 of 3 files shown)


context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -49,9 +53,40 @@ function SearchWrapper(props) {
     };
   }, [searchkit, setSearchHitsCount]);
 
+  const [message, setMessage] = useState(null);
+  const { elasticsearchEndpoint, nexusToken } = useContext(AppContext);
+  searchkit.setQueryProcessor((query) => {
```
john-conroy, line 58: It seems searchkit has a `QueryAccessor` is this more along the lines of what we should use?

context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -49,9 +53,40 @@ function SearchWrapper(props) {
...
+    if (!musts) {
+      setMessage(null);
+      return query;
+    }
+    const labels = {
+      'ancestor_ids.keyword': 'Derived from',
+      'descendant_ids.keyword': 'Ancestor of',
+    };
+    const provMusts = musts.filter(
+      (must) => 'term' in must && Object.keys(labels).some((labelKey) => labelKey in must.term),
```
john-conroy, line 72: Are there 'musts' we wouldn't want to display notes for?

context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -49,9 +53,40 @@ function SearchWrapper(props) {
...
+    );
+    setMessage(
+      provMusts.map((must) => {
+        return Object.entries(must.term).map(([k, v]) => (
+          <LookupEntity uuid={v} elasticsearchEndpoint={elasticsearchEndpoint} nexusToken={nexusToken}>
```
john-conroy, line 77: This is the only place we're using `LookupEntity` would one of our hooks fit instead?

context/app/static/js/components/Search/SearchWrapper.jsx
mccalluc: Will look into that. Thanks!

mccalluc:
Issue was closed won't-fix, and this code would need to be redone with the search replacement in any case. Closing.

PR #2008, opened 2021-07-13
`descendant_counts` fallback

mccalluc:
@yuanzhou, if you log in, you can see that https://portal.hubmapconsortium.org/browse/sample/6de04b223f7a976ef1129f10816cd5cb.json just has 
```
"descendant_counts": {}
```
This change gets the portal working again...
We don't have a definitive schema, so I can't say this JSON is "wrong", but it does seem like something has changed. If you think the upstream code is doing what you want, we can accommodate it... but if you're not sure why this changed, I would prefer not to sweep it under the rug.

Involve Bill if needed?

### DIFF
--- a/CHANGELOG-descendant_counts-fallback.md
+++ b/CHANGELOG-descendant_counts-fallback.md
@@ -1,0 +1,1 @@
+- Provide a fallback if `descendant_counts` is empty.
--- a/context/app/static/js/components/Detail/provenance/ProvTable/ProvTable.jsx
+++ b/context/app/static/js/components/Detail/provenance/ProvTable/ProvTable.jsx
@@ -20,7 +20,7 @@ function ProvTable(props) {
     typesToSplit.map(() => []),
   );
 
-  const descendantCounts = assayMetadata.descendant_counts.entity_type;
+  const descendantCounts = assayMetadata.descendant_counts.entity_type || {};
 
   return (
     <FlexContainer>


yuanzhou:
I am totally not aware of the existence of this `descendant_counts` and I can't find it in the search-api `indexer.py` at all like other runtime generated fields `descendant_ids`.

@shirey do you recall anything about this field?

john-conroy (commented):
I believe there are other components where we depend on `descendant_counts` they may already have a fallback or check that the key exists inside the metadata, but might be worth checking if you haven't already.

Also, could be worthwhile to move a default value higher up in the app?

shirey:
@mccalluc @yuanzhou @john-conroy Looking at the example above `6de04b223f7a976ef1129f10816cd5cb`- this is a Sample that has no descendants.  It seems like the translation code used to generate this field doesn't handle the case.  When there are no descendants our entity-api must just exclude the field vs sending an empty array of descendants?

mccalluc:
@shirey : 
>  When there are no descendants our entity-api must just exclude the field vs sending an empty array of descendants?

My preference is for an empty array, but obviously we can handle the status quo: That's what this PR does. I'm asking you to clarify whether the API is functioning as it should, or not, and we'll go forward from there.

mccalluc:
From standup, Bill thinks the fallback is the way to go. I will loop back to this and look for other dependencies as John suggests, but might file anything else I see as a new issue.

yuanzhou:
@mccalluc @shirey currently the entity-api returns an empty array when the target entity has no descendants.

For example: `https://entity.api.hubmapconsortium.org/descendants/6de04b223f7a976ef1129f10816cd5cb`

mccalluc:
> currently the entity-api returns an empty array when the target entity has no descendants

Ok, then either we're not talking about the same thing, or something was weird when I was looking at it. Will investigate.

mccalluc:
> currently the entity-api returns an empty array when the target entity has no descendants

@yuanzhou , to clarify, do you mean empty array or empty object? Because what I see is
`"descendant_counts": {}` rather than `"descendant_counts": []`.

Can you approve this PR if it looks good to you?

yuanzhou:
@mccalluc I think we are talking about different things. The call in search-api to entity-api to retrieve descendants of a given entity returns an empty array if no descendants found, example:  `https://entity.api.hubmapconsortium.org/descendants/6de04b223f7a976ef1129f10816cd5cb`

As I commented earlier: 
>I am totally not aware of the existence of this descendant_counts and I can't find it in the search-api indexer.py at all like other runtime generated fields descendant_ids.
I could see it's being used by the portal index transform code: https://github.com/hubmapconsortium/search-api/search?q=descendant_counts

So is this `"descendant_counts": {}` added by the portal index transform code for portal-ui use only?

yuanzhou (approved):

mccalluc:
> So is this "descendant_counts": {} added by the portal index transform code for portal-ui use only?

No: This PR is in `portal-ui`, not `search-api`. I'm going to go ahead and merge this as-is, since it has approval, even if there is still confusion.
- The problem it fixes it pretty serious.
- The intervention is small and of very limited scope.
- While I think having consistent data structures provided upstream would be great, I'm not hopeful about that, given the miscommunication we've seen.

PR #2009, opened 2021-07-14
list indexed datasets

### LINKED ISSUE #1975: Get a list of the UUIDs which are indexed
Because the indexing lags, we can't assume that all instances of a given data type will be indexed... but the API should be able to tell us. Just add this to the demo page... @tsliaw will figure out how to use this information.

mccalluc:
Fix #1975. @tsliaw -- I believe it's in your court to think about how this information would be used in the UI?

### DIFF
--- a/CHANGELOG-list-uuids.md
+++ b/CHANGELOG-list-uuids.md
@@ -1,0 +1,1 @@
+- List datasets indexed by Cells API.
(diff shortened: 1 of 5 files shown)


tsliaw (approved):

mccalluc:
Realizing that what we actually want is a list per-assay: Sean is working on the back-end for this: https://github.com/hubmapconsortium/cross_modality_query/issues/29

 Will reopen.

PR #2010, opened 2021-07-14
John conroy/track tutorial

### LINKED ISSUE #1976: Do stats exist for tutorial usage?
If yes, bring to the next UI meeting; If no, ask whether it's worth adding, and file a new issue.

john-conroy:
Fix #1976. It was easiest to put the `reactga` calls inside the store methods, does this make sense to you?


context/app/static/js/stores/useSearchDatasetTutorialStore.js
```diff
@@ -1,22 +1,46 @@
...
+    set({ runSearchDatasetTutorial: val });
+  },
   searchDatasetTutorialStep: 0,
-  incrementSearchDatasetTutorialStep: () =>
+  incrementSearchDatasetTutorialStep: () => {
+    ReactGA.event({
+      category: 'Dataset Search Tutorial',
+      action: 'Clicked next step',
+      value: get().searchDatasetTutorialStep,
+    });
```
mccalluc, line 21: I might add a helper function so you could use less boilerplate:
```suggestion
    recordEvent('Clicked next step', get().searchDatasetTutorialStep);
```

mccalluc: or, even more abbreviated:
```diff
- const useSearchDatasetTutorialStore = create((set, get) => ({
+ const useSearchDatasetTutorialStore = create((set, get) => {
+   function recordEvent(action) {
+      ReactGA.event({
+        category: 'Dataset Search Tutorial',
+        action,
+        value: get().searchDatasetTutorialStep,
+      });
+    }
+   return {
```

mccalluc (commented):
> calls inside the store methods, does this make sense to you?

Is your concern that it's a little weird to have side-effects like this inside the store? it seems ok to me. You might factor out some of the boiler plate, but your call. Looks good!

PR #2011, opened 2021-07-14
Mccalluc/query multiple genes

### LINKED ISSUE #1943: Query multiple genes API

### LINKED ISSUE #1944: Query multiple genes UI
Builds on #1943

mccalluc:
Towards #1943, Fix #1944. I don't actually remember whether the multiple values in `has` create the union or the intersection... @SFD5311, can you remind us which it is, and @tsliaw, can you confirm that that's what we want in the UI?

### DIFF
--- a/CHANGELOG-query-multiple-genes.md
+++ b/CHANGELOG-query-multiple-genes.md
@@ -1,0 +1,1 @@
+- In the Cells Demo, demonstrate how multiple genes could be queried.
--- a/context/app/routes_cells.py
+++ b/context/app/routes_cells.py
@@ -78,7 +78,7 @@ def proteins_by_substring():
 
 @blueprint.route('/cells/datasets-selected-by-<target_entity>.json', methods=['POST'])
 def datasets_selected_by_level(target_entity):
-    name = request.args.get('name')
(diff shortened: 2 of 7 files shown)


SFD5311:
@mccalluc The default value for logical operator across query types is "or," so by default we're taking the union.  The API supports overriding this default, but we didn't expose that parameter to the client.  Also worth mentioning here that the logical operator parameter in this case is actually applied to the cells in the datasets and not the datasets themselves.  So if we provide multiple genes as inputs, the results will be the datasets in which min_cell_percentage of cells express one gene or the other.  If what's desired is the datasets in which min_cell_percentage of cells express gene 1 or min_cell_percentage of cells express gene 2, the set union operator would be needed for that.

mccalluc:
@tsliaw -- can you review what Sean wrote above, and see if that's the behavior we want to offer in the UI, or not?

tsliaw:
@mccalluc  I'm not quite sure if that's the behavior we want, so I will put this on the UI agenda for Nils to decide

mccalluc:
Ok, I think the UI work here is good, but I'll unlink the API issue, and add a note there about our plan to get clarification.

Issue #1977, opened 2021-06-30
Make a list from a results set: UI


mccalluc:
We decided that the cells work can begin by adding the functionality to make a list from a results set. Chuck isn't perfectly clear on the details: clarification might be needed. Since this is not just the results on a single page, but _all_ of the results, backend work is a prerequisite: #1926.

mccalluc:
I think this comes from an earlier design where the cells flow would start from a results set. In any case:
- The design we've implemented for cells doesn't need this...
- but it was actually implemented for the TSV download.

Closing.

PR #2012, opened 2021-07-14
Proof of concept that we could fire off follow-up queries

mccalluc:
Towards #1975 / #1977. 
- @tsliaw : This demonstrates that we can generate an up-to-date list of all the UUIDs that would match the current set of facets, not just the first page of results. Would this be enough for #1975, or is anything else required?
- @john-conroy : It feels kind of hacky -- Do you have any other ideas?

### DIFF
--- a/CHANGELOG-uuids-from-facets.md
+++ b/CHANGELOG-uuids-from-facets.md
@@ -1,0 +1,1 @@
+- Demonstrate that we can get all matching UUIDs for a set of facets, not just the first page.
--- a/README-all-uuids.md
+++ b/README-all-uuids.md
@@ -1,0 +1,1 @@
+- Make a list of all UUIDs available on the search page.
--- a/README-all-uuids.md
+++ b/README-all-uuids.md
@@ -1,1 +1,0 @@
-- Make a list of all UUIDs available on the search page.
--- a/context/app/static/js/components/Routes/Routes.jsx
+++ b/context/app/static/js/components/Routes/Routes.jsx
@@ -80,7 +80,7 @@ function Routes(props) {
   if (urlPath.startsWith('/search')) {
     return (
       <Route>
-        <Search title={title} />
+        <Search title={title} getAllUUIDs />
       </Route>
     );
   }
--- a/context/app/static/js/components/Search/SearchWrapper.jsx
+++ b/context/app/static/js/components/Search/SearchWrapper.jsx
@@ -1,9 +1,10 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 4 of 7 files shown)


tsliaw (approved):
This should be what we want from a "Select All" button from the search page

mccalluc:
Huh: CI fails:
```
  FAIL app/static/js/components/Header/LoginButton/LoginButton.spec.js
    ● should be login button when not authenticated
  
      TypeError: Cannot read property 'querySelector' of null
  
        at HitsAccessor.scrollIfNeeded (node_modules/searchkit/src/core/accessors/HitsAccessor.ts:20:18)
        at HitsAccessor.setResults (node_modules/searchkit/src/core/accessors/HitsAccessor.ts:14:10)
        at node_modules/searchkit/src/core/AccessorManager.ts:129:35
        at arrayEach (node_modules/lodash/_arrayEach.js:15:9)
        at forEach (node_modules/lodash/forEach.js:38:10)
        at AccessorManager.setResults (node_modules/searchkit/src/core/AccessorManager.ts:129:5)
        at SearchkitManager.setError (node_modules/searchkit/src/core/SearchkitManager.ts:298:20)
        at SearchRequest.setError (node_modules/searchkit/src/core/SearchRequest.ts:33:22)

  Test Suites: 1 failed, 1 skipped, 45 passed, 46 of 47 total
  Tests:       1 failed, 3 skipped, 145 passed, 149 total
```
but locally:
```
Test Suites: 1 skipped, 46 passed, 46 of 47 total
Tests:       3 skipped, 146 passed, 149 total
```
and when I run just the failing test:
```
 PASS  app/static/js/components/Header/LoginButton/LoginButton.spec.js
  ✓ should be login button when not authenticated (107 ms)
  ✓ should be logout button when authenticated (149 ms)
  ✓ should display User when user_email is empty (17 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
```
Since this PR does add async, wonder if this is a race condition? Investigating...

mccalluc:
Ah: The previous test is `app/static/js/pages/search/Search.spec.js`, which passes, but with warnings. My guess is that I didn't wrap the earlier test, and so when the async returns, the effects spill into the next test.

mccalluc:
(Now failing on Cypress...)

mccalluc:
@john-conroy : I will take you up on your offer to look at this. My sense is that there's something about the react state change that confuses Cypress... but what I'm doing feels pretty kludgy regardless, so there might be a better approach that would incidentally also make Cypress happy.

mccalluc:
(This is not on the critical path: John's going to work on the actual cells page, and then we can revisit this... maybe make tweaks to get this working / maybe find a different approach.)

mccalluc:
This has been rendered unnecessary by #2244; Closing.

Issue #1932, opened 2021-06-22
Figure out how to answer an interesting question with data from elasticsearch


mccalluc:
Identifying the question is #1931

Roselkisma:
Based on question 1 on the survey "Which patient characteristics would you like to view when assessing the diversity of the population in a biomedical dataset?" I think that the best way to visualize this question is with a table. I attached a picture of how I think the table should look like. I was thinking that I will need to add the list of donors and then the user will choose the donor they are looking for. 

![donor's table](https://user-images.githubusercontent.com/85570580/125347600-1f1b9e00-e329-11eb-88c3-adcc611613c5.PNG)

mccalluc:
That's not really what we're looking for: For an individual donor, we can always go to that donor's page to get the details. Some suggestions/possibilities:
- How do the causes of death of Hubmap donors compare to the the general population, when stratified by age?
- Are blood-type / weight / height representative of the American population?
- When stratified by any one variable, does uneven sampling along other variables show up?

mccalluc:
@Roselkisma -- I think the last screenshot with ethnicity rows and gender columns does this. Can you make a PR, and link this issue to it?

PR #2013, opened 2021-07-15
This table combine the ethnicity and gender of the donors

Roselkisma:
I wrote the labels manually, I don't know if there is another way to do it. #1932


context/package.json
```diff
@@ -30,6 +31,7 @@
     "pretty-bytes": "^5.3.0",
     "prop-types": "^15.7.2",
     "react": "^16.12.0",
+    "react-chartjs-2": "^3.0.3",
```
mccalluc, line 34:
```suggestion
```

context/package.json
```diff
@@ -18,6 +18,7 @@
     "@visx/tooltip": "^1.7.2",
     "ajv": "^6.12.3",
     "bowser": "^2.11.0",
+    "chart.js": "^3.4.1",
```
mccalluc, line 21:
```suggestion
```

context/package-lock.json
```diff
@@ -30499,6 +30504,14 @@
...
       }
     },
+    "react-chartjs-2": {
+      "version": "3.0.3",
+      "resolved": "https://registry.npmjs.org/react-chartjs-2/-/react-chartjs-2-3.0.3.tgz",
+      "integrity": "sha512-jOFZKwZ8sMLkddewZ/tToxuu4pYimAvvY5I6uK+hCpSFT16Pvo2bdHhUoZ0X87zu9I+dx2I+JCqaLN6XhmrbDg==",
+      "requires": {
+        "lodash": "^4.17.19"
+      }
+    },
```
mccalluc, line 30514:
```suggestion
```

context/package-lock.json
```diff
@@ -15694,6 +15694,11 @@
       "integrity": "sha512-mT8iDcrh03qDGRRmoA2hmBJnxpllMR+0/0qlzjqZES6NdiWDcZkCNAk4rPFZ9Q85r27unkiNNg8ZOiwZXBHwcA==",
       "dev": true
     },
+    "chart.js": {
+      "version": "3.4.1",
+      "resolved": "https://registry.npmjs.org/chart.js/-/chart.js-3.4.1.tgz",
+      "integrity": "sha512-0R4mL7WiBcYoazIhrzSYnWcOw6RmrRn7Q4nKZNsBQZCBrlkZKodQbfeojCCo8eETPRCs1ZNTsAcZhIfyhyP61g=="
+    },
```
mccalluc, line 15701: Just to keep things tidy, I'd say revert this now, and we can bring it back when you need it. It's good to keep PRs on topic, and not let unrelated changes creep in.
```suggestion
```
... or, alternatively, add the charting to this PR... but I would lean towards keeping it small.

mccalluc (changes requested):

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -33,55 +33,76 @@ import useSearchData from 'js/hooks/useSearchData';
...
+          },
+          {
+            'mapped_metadata.race': {
+              terms: {
+                field: 'mapped_metadata.race.keyword',
+              },
+            },
+          },
+        ],
+        size: 10000,
```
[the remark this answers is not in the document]
john-conroy: Actually, for these composite queries the inner size param needs to be set to non 0.

context/package.json
john-conroy: Alternatively you can just run `npm uninstall chart.js` and it will uninstall the package.

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -33,55 +33,94 @@ import useSearchData from 'js/hooks/useSearchData';
...
-  const { buckets } = searchData.aggregations['mapped_metadata.race'];
+
+  function getCount(buckets, bloodType, race) {
+    const filtered = buckets.filter(
+      (b) => b.key['mapped_metadata.blood_type'] === bloodType && b.key['mapped_metadata.race'] === race,
+    );
+    return filtered.length ? filtered[0].doc_count : 0;
+  }
+  function getKeyValues(buckets, key) {
+    return [...new Set(buckets.map((b) => b.key[key]))];
```
mccalluc, line 78:
```suggestion
    return new Set(buckets.map((b) => b.key[key]));
```
This came up in the call: I think you can just return a set here, instead of converting it back to a list... but I could be wrong.

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -33,55 +33,94 @@ import useSearchData from 'js/hooks/useSearchData';
...
+    Object.keys(searchData).length && (
+      <Paper>
+        <StyledTableContainer>
+          <Table stickyHeader>
+            <TableHead>
+              <TableRow>
+                <HeaderCell> </HeaderCell>
+                {getKeyValues(buckets, 'mapped_metadata.blood_type').map((type) => (
+                  <HeaderCell> {type} </HeaderCell>
+                ))}
```
mccalluc, line 92: This is good, but you can't be certain the order coming out of this function will be the same as what you list below. I'd suggest taking this in stages:
- In this PR, replace this `map` with an explicit list.
- After it's merged, come back and add `const bloodTypes = getKeyValues(buckets, 'mapped_metadata.blood_type')` at the top, and use it here, and in the rows below.
- and finally wrap the rows on the table in a similar loop for race.

mccalluc (changes requested):
See comment: For now, replace `map` with an explicit list. You can clean it up later, but we want to be very sure we never introduce code that could just give us the wrong information.

context/app/static/js/pages/Vis/Vis.jsx
Roselkisma: You are right, we talked about it, I forgot to take it out.

mccalluc (approved):

PR #2021, opened 2021-07-15
Template updates for CCF EUI 2.0

bherr2:


context/app/templates/pages/ccf-eui.html
```diff
@@ -2,21 +2,9 @@
 <html lang="en">
 
 <head>
-  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-133341631-3"></script>
```
bherr2, line 5: FYI we removed the google analytics tags above as they are now handled internal to the EUI

mccalluc (approved):
Works for me. FWIW, 3D interactions are really slow/laggy for me... It seems like scroll events are getting queued up and then processed in order, and the rendering can't keep up.

bherr2:
Yeah, performance issues have been a persistent problem we're working through.

PR #2022, opened 2021-07-15
John conroy/add tutorial to storybook

john-conroy:
Just a bit of refactoring to make the transition to storybook for the tutorial components.


mccalluc (approved):
Looks good. Are there any cases where you would want there to be actions on the click, or is that not what you'd want to do with storybook?

john-conroy:
Cases where the actions are contained within the component will be shown. Cases where the action is determined by props, I think we can use our judgement unless there is a specific action/story that should be shown. Often you'll have parent component stories, but in this case the parent component is the Search page. Storybook also can be used to demonstrate pages, but that seems too heavyweight for now.

john-conroy:
I'm still getting a feel for this as I go so please share any thoughts on best practices.

PR #2024, opened 2021-07-19
Fix Codex carousel View Config.

### LINKED ISSUE #2023: Broken Carousel Link for Codex
image

- The bitmask controls should not be there (raster instead, as in the photo)
- 3d should be turned off via `disable3d` prop

ilan-gold:
FIxes #2023 i.e makes it work as it did before (and just like all the other CODEX datasets).


ilan-gold:
I am getting 500 errors on the actual Raster image @mccalluc do you see this as well?

ilan-gold:
Screen Shot 2021-07-19 at 1 35 26 PM

mccalluc:
@ilan-gold :
- Yes, I see the same 500s. Frustrating. You'll follow-up with IEC? Also feel free to ask them why their monitoring didn't alert anyone to the problem, before we discovered it for ourselves.
- I would really prefer that we get the backward compatibility in Vitessce, and see it working, before we merge this PR. I know that means a little bit of brokenness on the portal a little longer, but I think the discipline of backward compatibility is something we need to prioritize.

ilan-gold:
That's fine, I can close this then.  The PR to fix this then would just be updating Vitessce.

PR #2028, opened 2021-07-20
Upgrade Vitessce to 1.1.12

### LINKED ISSUE #2023: Broken Carousel Link for Codex

ilan-gold:
Upgrades Vitessce + python package to fix the bug with the carousel Codex view config without code changes, ensuring backwards compatibility

Fixed #2023

### DIFF
--- a/CHANGELOG-fix-vitessce-carousel.md
+++ b/CHANGELOG-fix-vitessce-carousel.md
@@ -1,0 +1,1 @@
+- Fix Vitessce carousel demo by upgrading Vitessce to 1.1.12 + python package to 0.1.0a11 to gain backwards compatibility.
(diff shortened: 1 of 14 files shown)


mccalluc (approved):
Thanks! Couple questions about the big picture, but nothing that blocks this PR:

- Are the `12`s in the version numbers likely to remain in sync, or is that an accident?
- Would it make sense to link the schema numbers to the version numbers? (I feel like I asked this before, and you explained that it doesn't, but I forget the reasoning. Sorry for the bother, again.)

ilan-gold:
Those numbers are an accident.  I am not sure it makes sense to do that, since the schema changes so rarely.  New features almost never necessitate a new scheme version, so I don't think tying the two together makes sense.  For example, adding things to the schema doesn't even necessitate a new schema version as those things aren't necessary.

Thanks!

PR #2029, opened 2021-07-20
upgrade caniuse

mccalluc:
This is really just procrastination. Something to think about would be how to approach housekeeping tasks... Maybe we'd     set a reminder to do this every _n_ sprints?


john-conroy:
Sure, I'm happy to do it every two sprints or whichever cadence you might prefer.

john-conroy (approved):

mccalluc:
`n` can be pretty large, but I think a schedule would be good. Thanks!

PR #2030, opened 2021-07-20
John conroy/add entity tile to storybook

john-conroy:
I figured I should add it while I'm here.


mccalluc (approved):
Approve, but... I'm not sure that we really want the story book to be this granular: `EntityTile` is the unit of reuse, but I'm not sure we need examples of `EnitityTileFooter` and `EntityTileBody`. I could be thinking about it wrong.

PR #2035, opened 2021-07-27
Mccalluc/unpublished hbm 500

### LINKED ISSUE #2033: 500 from non-public HBM url
I logged in, added an unpublished item to my lists, logged out, and tried to visit it, and got a 500:

https://portal.hubmapconsortium.org/browse/HBM655.RVNL.232

The HBM redirect fail with an access-denied, instead of 500, if the user can't see an item.

mccalluc:
Fix #2033

### DIFF
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -94,17 +94,7 @@ class ApiClient():
             body_json=query)
 
         hits = response_json['hits']['hits']
-
-        if len(hits) == 0:
-            if (uuid and len(uuid) == 32 or hbm_id) and not self.nexus_token:
-                # Assume that the UUID is not yet published:
-                # UI will suggest logging in.
-                abort(403)
-            abort(404)
-        if len(hits) > 1:
-            raise Exception(f'ID not unique; got {len(hits)} matches for {query}')
-        entity = hits[0]['_source']
-        return entity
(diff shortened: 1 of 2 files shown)


context/app/api/client.py
```diff
@@ -173,6 +163,46 @@ def _get_mock_vitessce_conf(self):
...
+
+    >>> _get_entity_from_hits([{'_source': 'fake-entity'}])
+    'fake-entity'
+
+    '''
+    if len(hits) == 0:
+        if (uuid and len(uuid) == 32 or hbm_id) and not has_token:
+            # Assume that the UUID is not yet published:
+            # UI will suggest logging in.
+            abort(403)
```
john-conroy, line 198: Didn't we want to use a `404` in these cases?

mccalluc: No: I think this comment comes from before we talked, but the idea is that most of the time, people were getting 404s because they just weren't logged in, and so (usually) returning a 403 and telling them to log in will point them in the right direction.

(We can't actually check whether a given ID exists or not in the unpublished state unless we're logged in. This is partly to prevent information leakage... and partially because it's simpler not to have to deal with we-know-this-exists-but-can't-say-anything-else.)

mccalluc:
(Interpreting "comment" as "good-to-go-perhaps-with-tweaks".)

Issue #1935, opened 2021-06-22
Connect Elasticsearch to chart.js


mccalluc:
Connect #1930 to #1934 .

PR #2036, opened 2021-07-28
Blood type and race bar chart

Roselkisma:
This PR has a bar chart connected to the Elasticsearch #1935


Roselkisma:
Screen Shot 2021-07-28 at 3 41 07 PM

context/app/static/js/pages/Vis/DonorChart.jsx
```diff
@@ -0,0 +1,58 @@
...
+  const { elasticsearchEndpoint, nexusToken } = useContext(AppContext);
+
+  const { searchData } = useSearchData(donorQuery, elasticsearchEndpoint, nexusToken);
+  if (!('aggregations' in searchData)) {
+    return null;
+  }
+
+  function getCount(buckets, key1, key2) {
+    const filtered = buckets.filter((b) => b.key[xKey] === key1 && b.key[yKey] === key2);
+    return filtered.length ? filtered[0].doc_count : 0;
```
mccalluc, line 17:
```suggestion
    if (filtered.length > 1) {
      console.warn(`Expected at most one match for ${key1} and ${key2}: Got ${filtered.length}.`);
    }
    return filtered.length ? filtered[0].doc_count : 0;
```

Not essential, but if something weird happened, would be good to know.

context/app/static/js/pages/Vis/DonorChart.jsx
```diff
@@ -0,0 +1,58 @@
...
+  }
+  function getKeyValues(buckets, key) {
+    return [...new Set(buckets.map((b) => b.key[key]))];
+  }
+  const { buckets } = searchData?.aggregations.composite_data;
+
+  const labels = getKeyValues(buckets, xKey);
+  /* const races = getKeyValues(buckets, 'mapped_metadata.race');
+   */
```
mccalluc, line 26:
```suggestion
  const labels = getKeyValues(buckets, xKey);
```
Shouldn't usually check in commented out code.

context/app/static/js/pages/Vis/DonorChart.jsx
```diff
@@ -0,0 +1,58 @@
...
+    <>
+      <div className="header">
+        <h1 className="title">{title}</h1>
+      </div>
+      <Bar data={graphdata} options={options} />
+    </>
+  );
+}
+
+export default DonorChart;
```
mccalluc, line 58: After parameterizing the keys, there really isn't anything left that's particular to Donors... consider giving it a more generic name.

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -33,55 +34,100 @@ import useSearchData from 'js/hooks/useSearchData';
...
-  const { buckets } = searchData.aggregations['mapped_metadata.race'];
+  function getCount(buckets, bloodType, race) {
+    const filtered = buckets.filter(
+      (b) => b.key['mapped_metadata.blood_type'] === bloodType && b.key['mapped_metadata.race'] === race,
+    );
+    return filtered.length ? filtered[0].doc_count : 0;
+  }
+  function getKeyValues(buckets, key) {
+    return [...new Set(buckets.map((b) => b.key[key]))];
+  }
```
mccalluc, line 79: This same logic has been generalized in DonorChart.jsx -- We shouldn't have the same logic in two places.

context/app/static/js/pages/Vis/Vis.jsx
```diff
@@ -33,55 +34,100 @@ import useSearchData from 'js/hooks/useSearchData';
...
+          donorQuery={donorRaceSexQuery}
+          xKey="mapped_metadata.blood_type"
+          yKey="mapped_metadata.race"
+          colorKeys={['White', 'Black or African American', 'Hispanic']}
+          colors={['rgb(255, 99, 132)', 'rgb(54, 162, 235)', 'rgb(75, 192, 192)']}
+          title="Blood Type and Race"
+        />
+        <StyledTableContainer>
+          <Table stickyHeader>
+            <TableHead>
```
mccalluc, line 97: I'd suggest factoring this out into a reusable component, like you did for the chart... Perhaps it's one component that could render both?

mccalluc (commented):
Looks good: Normally, I'd want to hold off on merging until all the comments are addressed, but in this case, I'd say merge it, but copy all of my comments to an issue, to make sure they aren't misplaced, and might be addressed some day.

context/app/static/js/pages/Vis/DonorChart.jsx
Roselkisma: Can the new name be VisChart - vis for visualization?

PR #2042, opened 2021-08-02
Mccalluc/log sliders

mccalluc:
- Adds a control for selecting values on a log scale. (The component actually returns the exponent, but we show the label as `10^n`, and on the other end, we calculate the exponent before making the request to the flask API.) (... down the road, maybe the Flask API should just take the exponent? Not sure.)
- Create `js/components/cells`, and put it in there.
- Add a story, so you can see how it renders. (The story uses state, which I think is a new thing for us, so this may or may not be the best way to do it.)


tsliaw:
For the label, can you also make the hover/drag label to the 10^n label instead of the exponent? 
Screen Shot 2021-08-02 at 12 14 54 PM

mccalluc:
@tsliaw -- Yes, I should do that, and I should have read farther down the in the documentation, because they have [an example of a log-scale slider](https://material-ui.com/components/slider/#non-linear-scale).

context/app/static/js/components/cells/LogSliderWrapper/LogSliderWrapper.stories.js
```diff
@@ -0,0 +1,18 @@
...
+
+import LogSliderWrapper from './LogSliderWrapper';
+
+export default {
+  title: 'Controls/LogSliderWrapper',
+  component: LogSliderWrapper,
+};
+
+export const LogSliderWrapperStory = (args) => {
+  const [logValue, setLogValue] = useState(1);
```
john-conroy, line 11: I think this works for a single story, for components with more stories we'd likely want to use a decorator.  Looking at some of the state addons, I don't think an addon is something we need, but maybe I'm wrong.

john-conroy (approved):
Do we want to add some padding to the paper surrounding the sliders or defer until I start working on this?

mccalluc:
> Do we want to add some padding to the paper surrounding the sliders or defer until I start working on this?

@john-conroy - I think it would go on the slider, rather than styling the paper... but I could be confused. Either way, I'm fine leaving this until you start pulling the pieces together, so I'll just go ahead and merge this. Thanks!

PR #2045, opened 2021-08-03
Mccalluc/globus link if vis lifted

### LINKED ISSUE #1972: Show globus links for derived datasets
Links to globus for derived, vis-lifted dataset should be made available. Initial proposal (but wait for final wording from @ngehlenborg )
```diff
Files are available through the Globus Research Data Management System. 
Access dataset HBM423.JZTB.864 on Globus.
+ Access derived dataset HBM123.ABCD.456 on Globus
```
Maybe "Support"? That's how they are described it in the UI.

Question: Would it be ok if this was done for _all_ derived data sets, rather than depending on the vis lifting logic? 

_Originally posted by @mccalluc in https://github.com/hubmapconsortium/portal-ui/issues/1947#issuecomment-869841428_

mccalluc:
Fix #1972: @ngehlenborg , Your request was:
```diff
Files are available through the Globus Research Data Management System. 
Access dataset HBM423.JZTB.864 on Globus.
+ Data generated for visualization of this dataset are in support dataset HBM123.ABCD.456 on Globus.
```
Right now, this PR looks like:
Screen Shot 2021-08-03 at 12 56 13 PM
- The green checkmark beside "Bulk Data Transfer" is actually the result of an API call, to make sure the Globus directory is accessible. If we thought it was necessary for the base dataset, guessing it's necessary for the lifted one as well?
- We don't have the HuBMAP display ID... We could get it, but it's another API hit, and more complexity, for relatively little gain.

Will put this on the UI Meeting agenda.


ngehlenborg:
This looks fine apart from the spacing. There needs to be more space above "Bulk Data Transfer (Support)".

ngehlenborg (changes requested):
This looks fine apart from the spacing. There needs to be more space above "Bulk Data Transfer (Support)".

mccalluc:
@john-conroy : Sorry that this has gotten complicated:
- Where there's vis-lifting, pass more information in from the python side.
- Move Paper wrapper out from GlobusLink, since we'll want both GlobusLink instances to be on the same Paper.
- Tweak proptypes/default values to avoid some errors.
- Since the paper now depending on two asyncs, the please wait would have gotten more complicated. Since this is well below the fold, I think making it simpler is worth it, even if it could mean some display flicker.
- ... that in turn requires changing the tests.

john-conroy:
I'm getting the following error when checking out the example dataset.

Screen Shot 2021-08-16 at 4 08 25 PM

mccalluc:
@john-conroy -- Thanks for catching that: Logging in causes errors. Investigating.

context/app/static/js/components/files/Files/Files.jsx
```diff
@@ -40,7 +42,12 @@ function Files(props) {
       <SectionContainer id="files">
         <SectionHeader>Files</SectionHeader>
         {files && <FileBrowser files={files} />}
-        <GlobusLink uuid={uuid} hubmap_id={hubmap_id} />
+        <MarginTopDiv>
+          <DetailSectionPaper>
+            <GlobusLink uuid={uuid} hubmap_id={hubmap_id} />
+            {visLiftedUUID && <GlobusLink uuid={visLiftedUUID} isSupport />}
+          </DetailSectionPaper>
```
john-conroy, line 49: Slight preference towards pulling all of this into a component.

context/app/static/js/components/files/GlobusLink/style.js
```diff
@@ -6,18 +6,9 @@ const StyledTypography = styled(Typography)`
...
-  margin-top: ${(props) => props.theme.spacing(2)}px;
-`;
-
-const MarginTopDiv = styled.div`
-  margin-top: ${(props) => props.theme.spacing(2)}px;
-`;
-
 const Flex = styled.div`
   display: flex;
+  margin-top: ${(props) => props.theme.spacing(1)}px;
```
john-conroy, line 11: There should only be extra margin for the support link.

context/app/static/js/components/files/GlobusLink/GlobusLink.jsx
```diff
@@ -44,30 +39,32 @@ function GlobusLink(props) {
...
+    statusCode && (
+      <>
         <Flex>
           {statusCode === 200 ? (
             <StyledSuccessIcon data-testid="success-icon" />
           ) : (
             <StyledErrorIcon data-testid="error-icon" />
           )}
-          <StyledTypography variant="h6">Bulk Data Transfer</StyledTypography>
+          <StyledTypography variant="h6">Bulk Data Transfer {isSupport && '(Support)'}</StyledTypography>
```
john-conroy, line 51:
```suggestion
          <StyledTypography variant="h6">Bulk Data Transfer{isSupport && ' (Support)'}</StyledTypography>
```

john-conroy (changes requested):
Sorry for falling behind on this - a few style changes. Maybe make an issue for the layout shift so we can keep track of it in the event we clean it up one day.

context/app/static/js/components/files/Files/Files.jsx
```diff
@@ -40,7 +42,12 @@ function Files(props) {
       <SectionContainer id="files">
         <SectionHeader>Files</SectionHeader>
         {files && <FileBrowser files={files} />}
-        <GlobusLink uuid={uuid} hubmap_id={hubmap_id} />
+        <MarginTopDiv>
```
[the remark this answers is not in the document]
mccalluc: I'll wrap `FileBrowser` in `MarginBottomDiv` instead.

context/app/static/js/components/files/Files/Files.jsx
mccalluc: Done.

context/app/static/js/components/files/GlobusLink/style.js
mccalluc: I'll remove margin-top, add conditional `<br>` instead.

john-conroy: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

> Note: Do not use <br> to create margins between paragraphs; wrap them in <p> elements and use the CSS margin property to control their size.

john-conroy: Line breaks are specifically meant to be used inside of a `<p>` or similar.

mccalluc:
> Sorry for falling behind on this - a few style changes. Maybe make an issue for the layout shift so we can keep track of it in the event we clean it up one day.

Filed #2093.

With the latest commits it looks like:
Screen Shot 2021-08-23 at 11 43 51 AM

- http://localhost:5001/browse/dataset/e0ba24d0c4dab528e46aab3f67a1aae8#files
- http://localhost:5001/browse/dataset/1e9b9d5ac57ac8bd4093f2503675a2cd#files
- http://localhost:5001/browse/dataset/53cc1f9c21907d8f7880eed6828ba8a1#files

PR #2046, opened 2021-08-03
Added more graphs and style to the diversity page

Roselkisma:


.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-    "eslint.workingDirectories": ["./context"]
+    "eslint.workingDirectories": [
+        "./context"
+    ],
+    "python.pythonPath": "/usr/bin/python"
```
mccalluc, line 5:
```suggestion
    "eslint.workingDirectories": ["./context"]
```

Having IDE settings checked in is a grey-area: Sometimes it makes sense to share settings across a team... other times different people may have different preferences. In this case, different devs will be using python binaries at different locations, so this should be reverted.

CHANGELOG-chart.md
```diff
@@ -0,0 +1 @@
+- Chart and table.
```
mccalluc, line 1: These go into the change log: We want something that explains the new feature without assuming context.

mccalluc (changes requested):
@Roselkisma , there are changes in here that I think are already deployed. I'm not sure exactly what's going on, but you might try:
```
git checkout main
git pull
git checkout roselkisma/more-graphs-and-style
git merge main
# There may or may not be conflicts at this point.
# Merge and push when problems are resolved.
```
Then re-request review.

ngehlenborg:
@Roselkisma @mccalluc Has the route for this page already been changed to `.../diversity`?

PR #2047, opened 2021-08-03
Mccalluc/publications

mccalluc:
First draft:
Screen Shot 2021-08-03 at 3 43 33 PM
- Is other information required?
- Does article metadata need to be in a more structured form, or can it just be markdown?
  - If it should be more structured, rather than putting the data entry burden on the authors, maybe write a script to parse [bioarxive XML](https://www.biorxiv.org/content/early/2021/07/29/2021.07.28.454201.source.xml) and output our yaml?
- Do we need a section of boilerplate text for publications like we have for previews?

Still todo:
- Landing page which lists all publications. (Maybe not needed until there are a few?)
- Adding a link to a menu. (Unclear where it should go.)


context/app/publication/kidney-cell-states-atlas.md
```diff
@@ -0,0 +1,43 @@
...
+            'y': 0,
+            'w': 12,
+            'h': 4,
+            'props': { 'description': 'View #2 goes here' }
+           },
+        ]
+    }
+  ]
+---
+[Preprint on Biorxive](https://www.biorxiv.org/content/10.1101/2021.07.28.454201v1)
```
ngehlenborg, line 37:
```suggestion
[Preprint on BioRxiv](https://www.biorxiv.org/content/10.1101/2021.07.28.454201)
```

Unfortunately, bioRxiv adds the version number of the preprint (e.g. "v1") at the end of the URL once you visit the DOI link but redirects to the latest if it is not part of the URL. I removed it here.

ngehlenborg (changes requested):
It would be good to move the author list below the abstract in this case and format the preprint info and a link to the contact author (new) a bit more prominently. See mockup below. 

Can we insert the author names section between the "Manuscript" and "Contact" section?

image


We do not need the notification message at the top for these pages.

context/app/publication/kidney-cell-states-atlas.md
ngehlenborg: Obviously a moot point if we make the other requested changes ...

mccalluc:
@ngehlenborg -- I'm understanding the sections in order to be:
- Abstract
- Manuscript
- Authors
- Contact

Also: If you care strongly about the presentation, and we're hoping it will not be our responsibility to populate this in the future, then we probably want this information in structured yaml that'd we'd parse and then fill in a UI template, to ensure that each page is consistent. Is that what you want, or is the markdown fine? Either is ok, I just want to make sure the question is considered.

ngehlenborg:
The order of sections is correct.

Let's stick to the Markdown approach for the moment given the flexibility and since we don't have time to do a comprehensive  requirements analysis for a more structured approach. I see this current approach as version 0 and once we have seen a few publication examples, we can structure this further. The four sections included here will be available for all publications.

context/app/static/js/pages/Publication/Publication.jsx
```diff
@@ -0,0 +1,30 @@
...
+  // const { ... } = publicationMetadata;
+
+  return (
+    <>
+      <SectionContainer id="summary">
+        <Typography variant="subtitle1">Publication</Typography>
+        <SectionHeader variant="h1" component="h1">
+          {title}
+        </SectionHeader>
+        {/* Add description of Publications in general? */}
```
mccalluc, line 22:
```suggestion
```

context/app/static/js/pages/Publication/Publication.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+import Markdown from 'js/components/Markdown';
+import VisualizationWrapper from 'js/components/Detail/visualization/VisualizationWrapper';
+import SectionHeader from 'js/shared-styles/sections/SectionHeader';
+import SectionContainer from 'js/shared-styles/sections/SectionContainer';
+// import { StyledDescription } from './style';
```
mccalluc, line 8:
```suggestion
```

context/app/static/js/pages/Publication/Publication.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+function Publication(props) {
+  const { vitData, title, /* publicationMetadata, */ markdown } = props;
+
+  // const { ... } = publicationMetadata;
+
+  return (
```
mccalluc, line 15:
```suggestion
  const { vitData, title, markdown } = props;

  return (
```

mccalluc:
Now looks like:
Screen Shot 2021-08-04 at 9 31 00 AM
I'm going to branch from this and start work on the index page, but that doesn't need to block this.

ngehlenborg (changes requested):
Looking great apart from that `mailto:`

mccalluc:
With comments from Nils addressed, I'm going to assume it would have his approval. This way we can get it out in this release, and show it to the authors to get their buy-in, and get a start on authoring the vitessce config.

PR #2050, opened 2021-08-04
Mccalluc/publications index

mccalluc:
Complements
- #2047 

Screen Shot 2021-08-04 at 11 12 05 AM

@tsliaw -- Sorry not to have done design first, but do you have thoughts here, particularly if it's not a long list to start?
- Do we want some text at the top just to explain what the list is?
- Probably don't need a subhead?
- How should the list be styled?
- Should anything more than the title be in the list? 

Also: If you have thoughts about where the link to this page should be...
- Just in footer?
- Change "Preview" menu to something than can encompass both?
- De prioritize "Preview"? ... Maybe just add them to the home page carousel, since it's not core functionality?  

Process: Perhaps suggest direction here, and then we'll get confirmation from Nils at the next UI meeting before I'll do any more code.


mccalluc:
@tsliaw : Thanks!
- The pre-preprints won't have publication date, I don't think: That's why they're "pre-".
- Publication landing page: "Publications" in the title and subtitle doesn't make sense... Get rid of sub title?
- The "et al." probably means we should represent the authors as an actual list?
- Get Nils to fill in the Lorem ipsum?

(Nothing needs to be replied to... probably sort this out at the next UI meeting.)

mccalluc:
@ngehlenborg -- Notes from the portal call indicate some confusion about why we're doing this. Please make sure everyone is on board.

> - Publication Tracker [Bill/Chuck]
>   - Spreadsheet is active and being used
>   - Chuck wants to be able to present it to on his pages with an API
>       - Use the info from Google Scholar because it is public
>       - what is the use case for this?

mccalluc:
@tsliaw : Is this acceptable for now?
Screen Shot 2021-08-18 at 3 50 20 PM
differences/issues:
- Title is link, design just has it in bold.
- Concerned about duplication of author info.
- No publication date
- No sub title at the top.
- I don't have the content for the info box.

tsliaw (approved):

PR #2061, opened 2021-08-10
Configure EUI via web-component properties

bherr2:


mccalluc (approved):
Ran it locally and it seems to work. I'll go ahead and do the merge... Usually we leave it to PR authors, but this seems low risk, less hassle. Should be deployed Thursday afternoon.

Issue #2069, opened 2021-08-17
TSNE Features for SPRM


ilan-gold:
- [ ] Add factors/clustering to spatial plots
 - [ ] Fix naming [(`tnse` -> `tsne`)](https://github.com/hubmapconsortium/portal-containers/blob/master/containers/sprm-to-anndata/context/main.py#L64) in pipeline
 - [ ] Fix incorrect file usage for scatterplot i.e TSNE not clustering: https://github.com/hubmapconsortium/portal-containers/blob/master/containers/sprm-to-anndata/context/main.py#L56
 - [ ] Decide on new view configuration for adding scatterplot to visualization

ngehlenborg:
I believe this has been rolled out.

PR #2070, opened 2021-08-17
Add TSNE Features to SPRM Viz

ilan-gold:
For the first point here: #2069.  `cf5cf2de8fb34dd35734756c0e7cf0a1` on TEST is the test data.


context/app/api/vitessce_confs/base_confs.py
```diff
@@ -395,6 +396,7 @@ def get_conf_cells(self):
                 "K-Means [Shape-Vectors]",
                 "K-Means [Texture]",
                 "K-Means [Total] Expression",
+                "K-Means [tSNE_All_Features]"
```
mccalluc, line 399: I can't suggest an edit that spans all the rows, but can you define this list once, and then use if for both?

mccalluc (commented):
Suggest factoring out the duplicated list, unless I'm missing something. Let me know if you'd like me to load it locally and poke around.

context/app/api/vitessce_confs/base_confs.py
ilan-gold: Good call.

PR #2080, opened 2021-08-18
John conroy/vertical stacked bar chart

john-conroy:
I could possibly move the scales inside of the component to make it simpler to use at the cost of flexibility, but I think this is fine for now.


context/app/static/js/shared-styles/charts/VerticalStackedBarChart/VerticalStackedBarChart.stories.js
```diff
@@ -0,0 +1,74 @@
...
+
+VerticalStackedBarChart.args = {
+  parentWidth: 500,
+  parentHeight: 500,
+  visxData: [
+    {
+      cluster: 1,
+      matched: 3,
+      unmatched: 5,
+      total: 8,
```
mccalluc, line 35: Could you drop one of these? I'd worry about what it's supposed to do when they don't add up.

context/app/static/js/shared-styles/charts/VerticalStackedBarChart/VerticalStackedBarChart.stories.js
```diff
@@ -0,0 +1,74 @@
...
+  domain: ['matched', 'unmatched'],
+  range: ['#DA348A', '#6C8938'],
+});
+
+VerticalStackedBarChart.args = {
+  parentWidth: 500,
+  parentHeight: 500,
+  visxData: [
+    {
+      cluster: 1,
```
mccalluc, line 32:
```suggestion
      cluster: 'cluster A',
```
These are just the labels on the axis, right? I'd suggest strings in the demo.

mccalluc (commented):
Tweaks in the storybook. Also, the padding on the first column seems to be different. Looks good overall!
Screen Shot 2021-08-19 at 4 03 07 PM

PR #2081, opened 2021-08-18
Mccalluc/diversity cleanup

### LINKED ISSUE #2043: This same logic has been generalized in DonorChart.jsx -- We shouldn't have the same logic in two places.
This same logic has been generalized in DonorChart.jsx -- We shouldn't have the same logic in two places.

_Originally posted by @mccalluc in https://github.com/hubmapconsortium/portal-ui/pull/2036#discussion_r679455348_

mccalluc:
Going to say this fixes #2043, but there's still work to do, and Nils has filed new issues.

### DIFF
--- a/context/app/static/js/pages/Diversity/Diversity.jsx
+++ b/context/app/static/js/pages/Diversity/Diversity.jsx
@@ -14,7 +14,7 @@ import ProjectAttribution from './ProjectAttribution';
 import { PageTitleWrapper, PageTitle, ChartPaper, ChartTitle, DescriptionPaper } from './style';
 import { getKeyValues, getAgeLabels } from './utils';
 
-function termSource(field) {
+function makeTermSource(field) {
   const esField = `mapped_metadata.${field}`;
   const source = {};
   source[esField] = {
@@ -25,7 +25,7 @@ function termSource(field) {
   return source;
 }
 
-function histogramSource(field) {
+function makeHistogramSource(field) {
   const esField = `mapped_metadata.${field}`;
   const source = {};
   source[esField] = {
@@ -37,7 +37,7 @@ function histogramSource(field) {
   return source;
 }
 
-function compositeQuery(source1, source2) {
+function makeCompositeQuery(source1, source2) {
   return {
     size: 0,
     aggs: {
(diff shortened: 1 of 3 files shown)


context/app/static/js/pages/Diversity/Diversity.jsx
```diff
@@ -14,164 +14,44 @@ import ProjectAttribution from './ProjectAttribution';
...
-            'mapped_metadata.blood_type': {
-              terms: {
-                field: 'mapped_metadata.blood_type.keyword',
-              },
-            },
-          },
-        ],
-        size: 10000,
-      },
+function termSource(field) {
```
john-conroy, line 17: Should these function names be verbs?

context/app/static/js/pages/Diversity/Diversity.jsx
```diff
@@ -14,164 +14,44 @@ import ProjectAttribution from './ProjectAttribution';
...
-              },
-            },
-          },
-        ],
-        size: 10000,
-      },
-    },
-  },
-};
+const donorAgeRace = compositeQuery(histogramSource('age'), termSource('race'));
```
john-conroy, line 54: We can move this outside of the function to avoid it be called on each render since it doesn't depend on any props.

context/app/static/js/pages/Diversity/Diversity.jsx
```diff
@@ -237,38 +117,26 @@ function Diversity() {
...
-                  </TableRow>
-
-                  <TableRow>
-                    <HeaderCell> Hispanic </HeaderCell>
-                    {age.map((type) => (
-                      <TableCell> {getCount(buckets, type, 'Hispanic')} </TableCell>
-                    ))}
-                  </TableRow>
+                  {['White', 'Black or African American', 'Hispanic'].map((value) => (
+                    <TableRow>
```
john-conroy, line 126: You'll probably get a key warning without a key here.

context/app/static/js/pages/Diversity/Diversity.jsx
```diff
@@ -237,38 +117,26 @@ function Diversity() {
...
+                    </TableRow>
+                  ))}
                 </TableBody>
               </Table>
             </StyledTableContainer>
           </ChartPaper>
 
           <DonorChart
-            donorQuery={donorRaceSexQuery}
+            donorQuery={compositeQuery(termSource('race'), termSource('blood_type'))}
```
john-conroy, line 139: Same as above, all of these function calls can be pulled outside the component.

context/app/static/js/pages/Diversity/Diversity.jsx
```diff
@@ -14,164 +14,44 @@ import ProjectAttribution from './ProjectAttribution';
...
-          {
-            'mapped_metadata.sex': {
-              terms: {
-                field: 'mapped_metadata.sex.keyword',
-              },
-            },
-          },
-        ],
-        size: 10000,
+function compositeQuery(source1, source2) {
```
john-conroy, line 40: Some of these could be useful outside of this component, consider adding them to shared utils somewhere.

john-conroy (changes requested):

context/app/static/js/pages/Diversity/Diversity.jsx
mccalluc: Thanks for pointing this out... I was thinking it was simpler, but I wasn't thinking about how it changes the functionality.

mccalluc:
(@john-conroy : I've generally been using "request changes" as "loop me in again once the changes are made"... but I don't think that's what you intended here? Anyway, I'm planning to merge with changes when CI is good... but if you really want to re-review, I won't rush things in the future.)

PR #2084, opened 2021-08-18
Add flag

### LINKED ISSUE #2072: Publications page w/ `is_published` boolean in frontmatter
- Should include a link to the full list of publications.
- Should only list those with `is_published`: There will be others that may still be drafts.

mccalluc:
Fix #2072

### DIFF
--- a/context/app/static/js/pages/Publications/Publications.jsx
+++ b/context/app/static/js/pages/Publications/Publications.jsx
@@ -23,16 +23,19 @@ function Publications(props) {
         <Paper>
           <Table>
             <TableBody>
-              {Object.entries(publications || []).map(([path, publication]) => (
-                <TableRow key={path}>
-                  <TableCell>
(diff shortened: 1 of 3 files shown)


context/app/static/js/pages/Publications/Publications.jsx
```diff
@@ -23,16 +23,19 @@ function Publications(props) {
...
-                      <LightBlueLink href={`/publication/${path}`}>{publication.title}</LightBlueLink>
-                    </Typography>
-                    {publication.authors}
-                  </TableCell>
-                </TableRow>
-              ))}
+              {Object.entries(publications || [])
+                // eslint-disable-next-line no-unused-vars
+                .filter(([path, publication]) => publication?.is_public)
+                .map(([path, publication]) => (
```
john-conroy, line 29: You could combine these into a single loop, but I don't think it's necessary.

john-conroy (approved):

PR #2086, opened 2021-08-18
Fix for IMC Data Type Deprecation.

### LINKED ISSUE #2085: IMC 2D Broken Viz
`iterAssays` in `TypeClient` does not produce deprecated assays but the entity response, which is fixed and does not change, still contains `IMC`.  Therefore https://github.com/hubmapconsortium/portal-ui/blob/07958d154206cd74ca0f354a79c9400dd1691637/context/app/api/vitessce_confs/assay_confs.py#L301-L313 breaks due to the `_assays[data_type]` call for `data_type` set to `IMC`.

ilan-gold:
Fixes #2085

### DIFF
--- a/context/app/api/vitessce_confs/assay_confs.py
+++ b/context/app/api/vitessce_confs/assay_confs.py
@@ -296,24 +296,20 @@ class NullViewConfBuilder():
 
 
 _assays = None
-_type_client = None
 
 
 def _get_assay(data_type):
     "Return the assay class for the given data type"
     global _assays
-    global _type_client
-    if _type_client is None:
-        _type_client = TypeClient(current_app.config["TYPE_SERVICE_ENDPOINT"])
-    
-    global _assays
+
+    type_client = TypeClient(current_app.config["TYPE_SERVICE_ENDPOINT"])
     if _assays is None:
         # iterAssays does not include deprecated assay names...
-        _assays = {assay.name: assay for assay in _type_client.iterAssays()}
+        _assays = {assay.name: assay for assay in type_client.iterAssays()}
         
     if data_type not in _assays:
         # ... but getAssayType does handle deprecated names:
-        _assays[data_type] = tc.getAssayType(data_type)
+        _assays[data_type] = type_client.getAssayType(data_type)
     return _assays[data_type]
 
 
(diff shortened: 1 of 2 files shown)


context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -296,14 +296,24 @@ def get_conf_cells(self):
 
 
 _assays = None
+tc = None
```
mccalluc, line 299:
```suggestion
_type_client = None
```

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -296,14 +296,24 @@ def get_conf_cells(self):
...
         tc = TypeClient(current_app.config["TYPE_SERVICE_ENDPOINT"])
         _assays = {assay.name: assay for assay in tc.iterAssays()}
+    # iterAssays only returns current, and not deprecated, assays,
+    # which may be present in the entity.
+    # getAssayType can check for deprecated assays properly.
+    if data_type not in _assays:
+        assay = tc.getAssayType(data_type)
+        # The assay.name in this case would be the new name for the
+        # data_type so we cache the result using the data_type from the entity.
+        _assays[data_type] = assay
```
mccalluc, line 316:
```suggestion
    global _type_client
    if _type_client is None:
        _type_client = TypeClient(current_app.config["TYPE_SERVICE_ENDPOINT"])
    
    global _assays
    if _assays is None:
        # iterAssays does not include deprecated assay names...
        _assays = {assay.name: assay for assay in _type_client.iterAssays()}
        
    if data_type not in _assays:
        # ... but getAssayType does handle deprecated names:
        _assays[data_type] = tc.getAssayType(data_type)
```
This might be easier to read, but I don't think I've changed the logic.
- Why the typeclient global? Preference to hide details. I might be missing something.
- Split up the `_assays is None and tc is None`, since there are really two separate things happening inside.

mccalluc (commented):
I might reorder it a little bit, but the logic seems good. Thanks!

context/app/api/vitessce_confs/assay_confs.py
ilan-gold: If `type_client` weren't global (as things stood previously), we couldn't have accessed it after the first `if` block.  I suppose we could instantiate it every time.  I agree with the last point.

PR #2087, opened 2021-08-19
John conroy/refactor expandable row

john-conroy:
Upgrades `Zustand` to access their context provider and `Storybook`. Configures main webpack and storybook webpack to handle `.mjs` files.


mccalluc (approved):
- Would it be easy to add a little animation on the fold-up / fold-down?
- Storybook question: For something like a table row, would it make sense to frame it with a table? So much about how it works is hard to judge right now without context.

john-conroy:
That's useful feedback, I'll add a story which includes a table. I think there is still value in having the isolated story.

PR #2089, opened 2021-08-19
softer warning from entities API

### LINKED ISSUE #2088: Explaing Entities 403 in log
If the user doesn't have privs, there's an expected 403... but then we have an extra `console.error`:
```
console.error('Entities API failed', response);
```

If it's a 403, we shouldn't error: Instead, it'd be better to log a message that explains the 403 that preceded it is not unexpected.

mccalluc:
Fix #2088

### DIFF
--- a/context/app/static/js/components/files/GlobusLink/GlobusLink.jsx
+++ b/context/app/static/js/components/files/GlobusLink/GlobusLink.jsx
@@ -24,7 +24,12 @@ function GlobusLink(props) {
           headers: requestHeaders,
         });
         if (!response.ok) {
-          console.error('Entities API failed', response);
+          if (response.status === 403) {
+            // eslint-disable-next-line no-console
+            console.info('No error: 403 is an expected API response');
+          } else {
+            console.error('Entities API failed', response);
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/files/GlobusLink/GlobusLink.jsx
```diff
@@ -24,7 +24,12 @@ function GlobusLink(props) {
           headers: requestHeaders,
         });
         if (!response.ok) {
-          console.error('Entities API failed', response);
+          if (response.status === 403) {
```
john-conroy, line 27: Should we check for existence of a nexus token?

john-conroy (approved):

context/app/static/js/components/files/GlobusLink/GlobusLink.jsx
mccalluc: I don't know that that would change anything: You could be logged in, but still not have sufficient privs, and a 403 would still be an expected response. 

If I'm misunderstanding, let me know?

PR #2095, opened 2021-08-23
Simplify home page anchor link

john-conroy:
Removes the `useEffect` and handles everything in the callback.

### DIFF
--- a/context/app/static/js/pages/Home/Home.jsx
+++ b/context/app/static/js/pages/Home/Home.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useCallback, useState } from 'react';
+import React, { useCallback } from 'react';
 import { useTheme } from '@material-ui/core/styles';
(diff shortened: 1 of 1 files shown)


mccalluc (commented):
I don't want to hold things up if people are waiting on this, but do you think there are npm packages that could do this for us? Detail pages and others in the pipeline are going to have ToCs in the sidebar, and right now, those anchored links aren't going to work if people bookmark them.

Just googling... looks like other people are bothered by this too:
- https://www.npmjs.com/package/react-anchor-link-smooth-scroll
- https://gajus.medium.com/making-the-anchor-links-work-in-spa-applications-618ba2c6954a

john-conroy:
I'll merge for now and open an issue. I believe `react-router` and other frameworks with their own routing handle this, but I'm not sure if there are any easy to integrate, standalone packages we could introduce.

PR #2108, opened 2021-08-30
Mccalluc/tsv query

### LINKED ISSUE #2102: Limit dataset TSV by assay type
simple query parameter

mccalluc:
Fix #2102

### DIFF
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -155,13 +155,13 @@ def _make_query(constraints):
     >>> constraints = {'color': ['red', 'green'], 'number': ['42']}
     >>> query = _make_query(constraints)
     >>> from pprint import pp
-    >>> pp(query)
-    {'bool': {'must': [{'bool': {'should': [{'term': {'metadata.metadata.color.keyword': 'red'}},
-                                            {'term': {'mapped_metadata.color.keyword': 'red'}},
-                                            {'term': {'metadata.metadata.color.keyword': 'green'}},
-                                            {'term': {'mapped_metadata.color.keyword': 'green'}}]}},
-                       {'bool': {'should': [{'term': {'metadata.metadata.number.keyword': '42'}},
-                                            {'term': {'mapped_metadata.number.keyword': '42'}}]}}]}}
+    >>> pp(query['bool'])
+    {'must': [{'bool': {'should': [{'term': {'metadata.metadata.color.keyword': 'red'}},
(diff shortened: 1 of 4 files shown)


context/app/api/client.py
```diff
@@ -152,6 +145,40 @@ def get_vitessce_conf_cells_and_lifted_uuid(self, entity):
...
+                                   {'term': {'metadata.metadata.color.keyword': 'green'}},
+                                   {'term': {'mapped_metadata.color.keyword': 'green'}}]}},
+              {'bool': {'should': [{'term': {'metadata.metadata.number.keyword': '42'}},
+                                   {'term': {'mapped_metadata.number.keyword': '42'}}]}}]}
+    '''
+    shoulds = [
+        [
+            {"term": {f'{root}.{k}.keyword': v}}
+            for v in v_list
+            for root in ['metadata.metadata', 'mapped_metadata']
```
john-conroy, line 170: Do we expect the values for the properties in `metadata.metadata` and `mapped_metadata` to be the same?

john-conroy (approved):

context/app/api/client.py
john-conroy: Or is that the purpose of the `should`?

context/app/api/client.py
```diff
@@ -69,21 +69,14 @@ def get_all_dataset_uuids(self):
...
-        return self._get_all_entities_of_type(non_metadata_fields, 'Sample')
-
-    def get_all_datasets(self, non_metadata_fields):
-        return self._get_all_entities_of_type(non_metadata_fields, 'Dataset')
-
-    def _get_all_entities_of_type(self, non_metadata_fields, entity_type):
+    def get_entities(self, plural_lc_entity_type, non_metadata_fields, constraints):
+        entity_type = plural_lc_entity_type[:-1].capitalize()
         query = {
             "size": 10000,  # Default ES limit,
```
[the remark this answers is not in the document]
mccalluc: We're still in the hundreds, so when/if we get there, we can revisit... 

Alternatively: We could file an issue now, and every time this comes up, include a TODO referencing the issue, so we can search the codebase when we do want to tackle it.

What do you think?

context/app/api/client.py
mccalluc: yeah: The JSON document structure is a mess, and the "metadata" is in different locations for different entity types. The `should` (ie, `OR`) handles this: check all the possible paths, and see if it's under any of them.

PR #2109, opened 2021-08-30
Mccalluc/stub organ index

mccalluc:
- Split off a new routes file, for the routes based on sets of files on-disk.
- Rough draft of index page.

### DIFF
--- a/context/app/static/js/components/Routes/Routes.jsx
+++ b/context/app/static/js/components/Routes/Routes.jsx
@@ -171,14 +171,6 @@ function Routes(props) {
     );
   }
 
-  if (urlPath.startsWith('/publication')) {
-    return (
-      <Route>
-        <Organ organ={organ} />
-      </Route>
-    );
-  }
-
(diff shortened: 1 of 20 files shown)


context/app/static/js/components/Routes/Routes.jsx
```diff
@@ -143,14 +147,38 @@ function Routes(props) {
...
+  }
+
+  if (urlPath.startsWith('/publication')) {
+    return (
+      <Route>
+        <Organ organ={organ} />
+      </Route>
+    );
+  }
+
```
john-conroy, line 181: Could these be combined or will they separate down the road?

john-conroy (approved):

context/app/static/js/components/Routes/Routes.jsx
mccalluc: Bad copy and paste on my part: Thanks for catching... and I should add some tests.

PR #2110, opened 2021-08-30
Mccalluc/stub organ detail

mccalluc:
Before more gets added, does the over-all file organization seem right?

### DIFF
--- a/context/app/static/js/pages/Organ/Organ.jsx
+++ b/context/app/static/js/pages/Organ/Organ.jsx
@@ -1,9 +1,11 @@
 import React from 'react';
 
-import Paper from '@material-ui/core/Paper';
-
 import SectionHeader from 'js/shared-styles/sections/SectionHeader';
-import SectionContainer from 'js/shared-styles/sections/SectionContainer';
+import Asctb from 'js/components/organ/Asctb';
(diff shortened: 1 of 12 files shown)


context/app/static/js/pages/Organ/Organ.jsx
```diff
@@ -1,9 +1,35 @@
...
+  return (
+    <>
+      <SectionHeader variant="h1" component="h1">
+        {organ.title}
+      </SectionHeader>
+      {shouldDisplaySection.description && <Description description={organ.description} />}
+      {shouldDisplaySection.asctb && <Asctb asctb={organ.asctb} />}
+      {shouldDisplaySection.azimuth && <Azimuth description={organ.azimuth} />}
+      {shouldDisplaySection.assays && <Assays title={organ.title} />}
+      {shouldDisplaySection.samples && <Samples title={organ.title} />}
```
john-conroy, line 30: I think this is fine to merge for now, but as with the detail pages, we should agree on a best solution.

john-conroy (approved):

context/app/static/js/pages/Organ/Organ.jsx
mccalluc: @john-conroy : Let's talk more, after standup, maybe.

PR #2115, opened 2021-09-01
John conroy/cluster chart data

john-conroy:
- Adds a flask endpoint to fetch data from the cells API and perform some data wrangling. I'd appreciate a harder look at the python. Maybe here's probably a lighter weight library similar to pandas which would provide a `group_by` method?
- Adds a bar chart to visualize cells which match the query in dataset clusters. Right now it only works with a single gene, but will be expanded down the road. The cluster method is hard coded to the first type, but changed in #2118. The dimensions will also not be hardcoded in the future.
- Adds some basic diagnostic info on api response time.


context/app/routes_cells.py
```diff
@@ -66,6 +67,39 @@ def _first_n_matches(strings, substring, n):
     } for s, offset in zip(first_n, offsets)]
 
 
+def get_cluster_name_and_number(cluster_str):
```
mccalluc, line 70: Usually underscore functions that are purely internal to the module.

Can you use (preferably) [a dataclass](https://docs.python.org/3/library/dataclasses.html) or (less preferably) a named tuple? Inside a function, where the dependency is really near by, plain tuples are fine, but when the code is more spread out, structure is good.

For simple functions like this, putting in a doctest is easy:

```suggestion
def _get_cluster_name_and_number(cluster_str):
    '''
    >>> n_n = _get_cluster_name_and_number('cluster-name-number')
    >>> n_n.name
    'cluster-name'
    >>> n_n.number
    'number'
    '''
```

Also: I really don't like getting into splitting IDs / assuming semantics: It's too easy for the API maintainer to forget that part of the contract. Can you make sure there's an issue in the API, and reference it in a TODO here?

context/app/routes_cells.py
```diff
@@ -66,6 +67,39 @@ def _first_n_matches(strings, substring, n):
...
+    cluster_name_arr = cluster_str.split('-')
+    cluster_number = cluster_name_arr.pop()
+    cluster_name = '-'.join(cluster_name_arr)
+    return (cluster_name, cluster_number)
+
+
+def get_cluster_cells(cells, gene, min_gene_expression):
```
mccalluc, line 77: add doctest

context/app/routes_cells.py
```diff
@@ -66,6 +67,39 @@ def _first_n_matches(strings, substring, n):
...
+        for cluster in cell['clusters']:
+            cluster_name, cluster_number = get_cluster_name_and_number(cluster)
+            cluster_cells.append({'modality': cell['modality'], 'cluster_name': cluster_name,
+                                  'cluster_number': cluster_number,
+                                  'meets_minimum_gene_expression':
+                                  cell['values'][gene] >= min_gene_expression})
+    return cluster_cells
+
+
+def get_matched_cell_counts_per_cluster(cells):
```
mccalluc, line 89: add doctest

context/app/routes_cells.py
```diff
@@ -66,6 +67,39 @@ def _first_n_matches(strings, substring, n):
...
+    grouper = itemgetter("cluster_name", "cluster_number", 'modality')
+    clusters = {}
+    for key, grp in groupby(sorted(cells, key=grouper), grouper):
+        grp_list = list(grp)
+        cluster = dict(zip(["cluster_name", "cluster_number", 'modality'], key))
```
mccalluc, line 94:
```suggestion
    group_keys = ["cluster_name", "cluster_number", 'modality']
    grouper = itemgetter(*group_keys)
    clusters = {}
    for key, grp in groupby(sorted(cells, key=grouper), grouper):
        grp_list = list(grp)
        cluster = dict(zip(group_keys, key))
```

context/app/routes_cells.py
```diff
@@ -66,6 +67,39 @@ def _first_n_matches(strings, substring, n):
...
+        if not cluster['cluster_name'] in clusters:
+            clusters[cluster['cluster_name']] = []
```
mccalluc, line 96: I think a defaultdict would be better here.

context/app/routes_cells.py
```diff
@@ -66,6 +67,39 @@ def _first_n_matches(strings, substring, n):
...
+        cluster['matched'] = sum(item["meets_minimum_gene_expression"] for item in grp_list)
+        cluster['unmatched'] = len(grp_list) - cluster['matched']
```
mccalluc, line 98:
```suggestion
        cluster.update({
            'matched': sum(item["meets_minimum_gene_expression"] for item in grp_list)
            'unmatched': len(grp_list) - cluster['matched']
        })
```
Where possible, I like to have bigger data structures / comprehensions, instead of individual assignments. Easier to get a sense of what the big picture is.

context/app/routes_cells.py
```diff
@@ -157,3 +191,27 @@ def all_indexed_uuids():
...
+    uuid = request.args.get('uuid')
+    gene_name = request.args.get('gene_name')
+    min_gene_expression = request.args.get('min_gene_expression')
+    client = _get_client(current_app)
+
+    try:
+        cells = client.select_cells(where='dataset', has=[uuid])
+        cells_list = list(cells.get_list(values_included=gene_name))
+
+        # list() will call iterator behind the scenes.
```
mccalluc, line 211:
```suggestion
        cells_list = cells.get_list(values_included=gene_name)

```
`for` loop over `cells_list` should work as-is: Doesn't need to be an _actual_ `list`.

context/app/static/js/pages/CellsAPIDemo/DatasetClusterChart/DatasetClusterChart.jsx
```diff
@@ -0,0 +1,85 @@
...
+      });
+
+      const xScale = scaleBand({
+        domain: results[selectedClusterType].map((result) => result.cluster_number).sort((a, b) => a - b),
+        padding: 0.2,
+      });
+
+      const colorScale = scaleOrdinal({
+        domain: ['matched', 'unmatched'],
+        range: ['#DA348A', '#6C8938'],
```
mccalluc, line 28: theme?

mccalluc (changes requested):
I think the `groupby` is fine... but I still have trouble understanding what's going on in that function. I've made some suggestions, but maybe part of it just needs to be factored out into a little helper function? I wouldn't stop you from using pandas, but it's not clear to me that it's needed, either.

Can you make some tweaks, particularly adding some simple doc tests, and then ping me again? Thanks.

john-conroy:
We talked to resolve requested changes - it's approved to merge.

PR #2117, opened 2021-09-01
Add publications description

### LINKED ISSUE #2113: Publications note
> The following publications are a partial list of published HuBMAP research that uses data available through the HuBMAP Data Portal. The full list of HuBMAP-funded publications is available on [Google Scholar](https://scholar.google.com/citations?user=CtGSN80AAAAJ&hl=en).

mccalluc:
Fix #2113

### DIFF
--- a/CHANGELOG-publications-note.md
+++ b/CHANGELOG-publications-note.md
@@ -1,0 +1,1 @@
+- Add note explaining the limitted scope of the publications page.
--- a/context/app/static/js/pages/Publications/Publications.jsx
+++ b/context/app/static/js/pages/Publications/Publications.jsx
@@ -9,6 +9,7 @@ import TableRow from '@material-ui/core/TableRow';
 
 import { LightBlueLink } from 'js/shared-styles/Links';
 import SectionHeader from 'js/shared-styles/sections/SectionHeader';
+import SectionContainer from 'js/shared-styles/sections/SectionContainer';
 import Description from 'js/shared-styles/sections/Description';
 
 function Publications(props) {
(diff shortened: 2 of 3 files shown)


john-conroy (commented):
The `SectionContainer` is really only meant to be used on the detail pages to offset for the headers for anchor links. Is it mainly used here for spacing? Maybe we could comment or rename `SectionContainer` to make its intended use more clear?

mccalluc:
@john-conroy :
> The `SectionContainer` is really only meant to be used on the detail pages to offset for the headers for anchor links. Is it mainly used here for spacing? Maybe we could comment or rename `SectionContainer` to make its intended use more clear?

I guess I'm not clear about the right way to assemble the components for a typical page. Maybe there could be a prototypical example in storybook?

john-conroy:
I don't think there is much structure for pages outside of the detail pages (mostly just trying to visually match the mockups), but that is something we could move towards. I'll create an issue to create two page stories in storybook, one for a detail page and another for a 'typical' page?

mccalluc:
@john-conroy -- Without `SectionContainer` the description and the list have no separation. If not `SectionContainer`, what is the right idiom here?
Screen Shot 2021-09-07 at 11 02 30 AM

john-conroy:
I'd suggest refactoring to something like this, but it would require updating other pages which consume `SectionContainer`. So maybe this or something similar that would require less change?

```
const SectionContainer = styled.div`
  margin-bottom: ${(props) => props.theme.spacing(5)}px;
`;

// Offsets for anchor links
const DetailSectionContainer = styled(SectionContainer)`
  padding-top: ${headerHeight + entityHeaderHeight + 10}px;
  margin-top: -${headerHeight + entityHeaderHeight + 10}px;
`;
```

mccalluc:
(Sorting out the containers was done separately. Merge when tests pass.)

PR #2118, opened 2021-09-01
John conroy/cluster dropdown

john-conroy:
Adds a dropdown to allow users to select a cluster method. Still a little rough.

### DIFF
--- a/context/app/static/js/shared-styles/charts/VerticalStackedBarChart/VerticalStackedBarChart.jsx
+++ b/context/app/static/js/shared-styles/charts/VerticalStackedBarChart/VerticalStackedBarChart.jsx
@@ -75,6 +75,7 @@ function VerticalStackedBarChart({
                         x={bar.x}
                         y={bar.y}
                         width={bar.width}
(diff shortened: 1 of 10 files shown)


context/app/static/js/shared-styles/charts/VerticalStackedBarChart/VerticalStackedBarChart.jsx
```diff
@@ -75,7 +75,8 @@ function VerticalStackedBarChart({
                         x={bar.x}
                         y={bar.y}
                         width={bar.width}
-                        height={bar.height - strokeWidth}
+                        // TODO: Fix stroke overlap without shrinking bars
```
mccalluc, line 78: Just to say that I think this `TODO` is appropriate. There is a view that everything that could be fixed should be in the issue tracker... but I think we're working on so many fronts, and lots of things are going to be good-enough. It's really good to have the reminder here if we touch this code again... but adding another issue is overhead without much benefit.

mccalluc (approved):

PR #2121, opened 2021-09-02
Mccalluc/previews to resources

### LINKED ISSUE #2120: Change "previews" to "resources"
- Change "previews" to "resources"
- Contents of menu:
  - Publications
  - <hr>
  - Preview: Lorem...
  - Preview: Ipsum...
- but for now the publications section will be commented out.

_Originally posted by @mccalluc in https://github.com/hubmapconsortium/portal-ui/issues/2074#issuecomment-909618826_

mccalluc:
Fix #2120. The effect on the screen is negligible (change the menu title; prefix options with "preview") but it turned into a bigger refactor: Lots of renaming, and then it turned out that the Azimuth link wasn't beeing included in the narrow view, so move it inside the component, and then since the same pattern repeats 3 times, factor out just the parts that change between menus.


context/app/static/js/components/Header/Menu/Menu.jsx
```diff
@@ -32,28 +32,19 @@ function Menu(props) {
...
+              [toggleDocumentation, 'Documentation', openDocumentation, DocumentationLinks],
+            ].map(([onClick, label, isOpen, Component]) => (
+              <>
+                <DropdownMenuItem onClick={onClick}>
+                  {label}
+                  {isOpen ? <ArrowDropUpIcon /> : <ArrowDropDownIcon />}
+                </DropdownMenuItem>
+                {isOpen && <Component isIndented />}
+              </>
+            ))}
```
john-conroy, line 47: We could probably create a new component here like `Submenu` and move the open state into the component and only pass the string and links components as props.

mccalluc:
@john-conroy -- It's not clear to me why Cypress was having trouble right here: The general idea is that an element is on the screen, so cypress starts working with it, but then its visibility changes, and chained methods fail. This kludge is suggested [towards the end of a long thread](https://github.com/cypress-io/cypress/issues/695#issuecomment-558544974). I did try upgrading Cypress, but that didn't fix anything. (Upgrading might still be good, but I didn't want to muddy the waters in this PR.) Re-review?

end-to-end/cypress/integration/file-based-routes.spec.js
```diff
@@ -24,7 +24,8 @@ describe('file-based-routes', () => {
       // TODO: When we link to it from the menu, follow the link instead.
       cy.visit('/publication');
       cy.contains('Blue B. Lake');
-      cy.contains('An atlas of healthy and injured cell states').click();
+      // be.visible is a hack: https://github.com/cypress-io/cypress/issues/695
+      cy.contains('An atlas of healthy and injured cell states').should('be.visible').click();
```
john-conroy, line 28: Was this possibly considered 'not visible' before the hack because the test string provided isn't an exact match of what's on the page?

mccalluc: You mean it's just a substring? I don't think that's a problem, but let me check...

mccalluc: No, even if it's the full string, the failure is the same:
```diff
-      cy.contains('An atlas of healthy and injured cell states').should('be.visible').click();
+      cy.contains('An atlas of healthy and injured cell states and niches in the human kidney').click();
```
Good to confirm, though.

PR #2128, opened 2021-09-07
Upgrade Vitessce to 1.1.14

ilan-gold:
~~Feedback needed for new warning message.~~  Other than that, just new features.

### DIFF
--- a/context/package.json
+++ b/context/package.json
@@ -48,7 +48,7 @@
     "uuid": "^8.3.2",
     "vega": "^5.13.0",
     "vega-lite": "^4.13.1",
-    "vitessce": "^1.1.13",
+    "vitessce": "1.1.14",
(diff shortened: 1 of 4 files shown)


context/package.json
```diff
@@ -48,7 +48,7 @@
     "uuid": "^8.3.2",
     "vega": "^5.13.0",
     "vega-lite": "^4.13.1",
-    "vitessce": "^1.1.13",
+    "vitessce": "1.1.14",
```
mccalluc, line 51:
```suggestion
    "vitessce": "^1.1.14",
```
I'd keep the `^`: It's the normal practice with NPM. Without it I'm wondering if there's something wrong with a higher version.

mccalluc (commented):
keep the `^`?

Issue #2130, opened 2021-09-08
`components/files/` to `components/Detail/files/`?


mccalluc:
@john-conroy : Since this is only used on the detail pages, would it good to move it so it's under `Detail/`?

john-conroy:
Sure, also changing `Detail` to `detail` or something more meaningful would be good as well.

PR #2132, opened 2021-09-08
Mccalluc/section container refactor

### LINKED ISSUE #2131: refactor `SectionContainer`
I'd suggest refactoring to something like this, but it would require updating other pages which consume `SectionContainer`. So maybe this or something similar that would require less change?

```
const SectionContainer = styled.div`
  margin-bottom: ${(props) => props.theme.spacing(5)}px;
`;

// Offsets for anchor links
const DetailSectionContainer = styled(SectionContainer)`
  padding-top: ${headerHeight + entityHeaderHeight + 10}px;
  margin-top: -${headerHeight + entityHeaderHeight + 10}px;
`;
```

_Originally posted by @john-conroy in https://github.com/hubmapconsortium/portal-ui/issues/2117#issuecomment-914396627_

mccalluc:
- Fix #2131. 
- Towards #2117.
- Prompted #2130.

### DIFF
--- a/CHANGELOG-section-container.md
+++ b/CHANGELOG-section-container.md
@@ -1,0 +1,1 @@
+- Move `SectionContainer` to `PaddedSectionContainer`.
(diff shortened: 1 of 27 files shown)


john-conroy (approved):
'Padded' might be a bit misleading since there is now visible padding on the screen, but this works. Thanks!

mccalluc:
> 'Padded' might be a bit misleading since there is now visible padding on the screen, but this works. Thanks!

Thanks: Sorry, I'm not sure, either: Started with `Detail`, but it didn't seem right, because it's more about the extra space that's needed at the top, though the Detail pages are the only ones that have it... Anyway, it's an easy rename if it doesn't feel right.

PR #2135, opened 2021-09-09
John conroy/cells colors theme

john-conroy:
Retrieves colors from theme for cells ui charts. There was an issue where the changelog for one of the chart PRs wasn't compiled because it was missing the `.md` file extension.

### DIFF
--- a/CHANGELOG-cells-colors-theme.md
+++ b/CHANGELOG-cells-colors-theme.md
@@ -1,0 +1,1 @@
+- Get cells ui chart colors from theme.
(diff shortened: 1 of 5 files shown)


john-conroy:
I'll fix the missing hook deps, sorry I missed that.

mccalluc (commented):
Huh: I hadn't seen inline warnings from github actions like that before. I've tried a bit of google but haven't found a feature announcement... I guess my question is how does it know to flag these lines... Is it an interface other linters could use? And, would we want this to be a failure instead of a warning?

The PR looks fine.

PR #2137, opened 2021-09-09
John conroy/cleanup cells UI

john-conroy:
Remove demo components, move components into components dir, fix imports, and rename cells page. No material changes.


mccalluc (approved):
You could probably get rid of the `About`, too? But whenever; Looks good.

PR #2139, opened 2021-09-10
Update organ-util/descriptions.yaml

bherr2:


bherr2:
The webcomponent can be embedded by

1. Add to html head
```html
  <link href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500&amp;display=swap" rel="stylesheet">
  <link href="https://cdn.jsdelivr.net/gh/hubmapconsortium/ccf-ui@2/organ-info/styles.css" rel="stylesheet">
  <script src="https://cdn.jsdelivr.net/gh/hubmapconsortium/ccf-ui@2/organ-info/wc.js" async></script>
```

2. In your page, you can include the organ-info pane like this (right kidney for instance):
```html
<ccf-organ-info organ-iri="http://purl.obolibrary.org/obo/UBERON_0004538"></ccf-organ-info>
```

Full example below:
```html
<!doctype html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <title>HuBMAP CCF Organ Info</title>
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <link rel="icon" type="image/x-icon" href="favicon.ico">

  <link href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500&amp;display=swap" rel="stylesheet">
  <link href="https://cdn.jsdelivr.net/gh/hubmapconsortium/ccf-ui@2/organ-info/styles.css" rel="stylesheet">
  <script src="https://cdn.jsdelivr.net/gh/hubmapconsortium/ccf-ui@2/organ-info/wc.js" async></script>
</head>
<body class="mat-typography">
  <ccf-organ-info organ-iri="http://purl.obolibrary.org/obo/UBERON_0004538"></ccf-organ-info>
</body>
</html>
```

organ-utils/descriptions.yaml
```diff
@@ -4,11 +4,55 @@
...
+  name: Heart
+  description: A myogenic muscular circulatory organ found in the vertebrate
+    cardiovascular system composed of chambers of cardiac muscle. It is the
+    primary circulatory organ.
+http://purl.obolibrary.org/obo/UBERON_0004538:
+  name: Kidney (Left)
+  description: A kidney that is part of a left side of organism.
+http://purl.obolibrary.org/obo/UBERON_0004539:
+  name: Kidney (Right)
+  description: A kidney that is part of a right side of organism.
```
mccalluc, line 24:
```suggestion
http://purl.obolibrary.org/obo/UBERON_0002113:
  name: Kidney
  description: A paired organ of the urinary tract which has the production of urine as its primary function.
```

- @ngehlenborg -- My understanding of the latest plan is the we just want a single "Kidney" page. Is that correct? Later in this file, there is a single entry for "Lungs".
- @bherr2 -- Your webcomponent does not support generic "Kidney", correct? So the mapping from top-level to subordinate will need to be in a separate ccf.yaml (or whatever you want to call it) that you'll supply: I'll probably reference a raw url at github.

organ-utils/descriptions.yaml
```diff
@@ -4,11 +4,55 @@
...
+http://purl.obolibrary.org/obo/UBERON_0002106:
+  name: Spleen
+  description: the organ that functions to filter blood and to store red
+    corpuscles and platelets
```
mccalluc, line 28:
```suggestion
  description: The organ that functions to filter blood and to store red
    corpuscles and platelets.
```

mccalluc (changes requested):

organ-utils/descriptions.yaml
bherr2: We can special case it in our code. Probably something we should handle anyway.

bherr2:
We'll push out a new version of organ-info today that will let you use the Kidney IRI and it'll handle it properly. That organ is the only special case right now, but our code has been updated to handle other paired organs if/when they come up.

Issue #2129, opened 2021-09-07
Implement metadata table download on search pages


tsliaw:
See #2079 for design. [Figma Link](https://www.figma.com/file/pkIZbpJOAQnylo1zjyhaU6/Search-Page?node-id=1362%3A20055). 

The mockup is called v2.0. It is an update to how the search page currently looks on the portal and does not have the molecular & cellular query updates in case this feature gets implemented before that. The 'Add to List' feature/checkboxes in the search results is also in this mockup but does not have to be implemented for this. 

Datasets Dropdown Menu

- The dropdown menu should be a multi-selectable list of data types
- First option should be a option called "All Data Types" and should be selected automatically unless unselected by the user
- Preferably menu should be fixed height and scrollable

mccalluc:
@ngehlenborg -- With a more public exposure of the URL, I would really like the question in
- #2099 
to be resolved before the work here is deployed. Status quo is fine with me, but I'd like that confirmed.

mccalluc:
Declaring this complete. Rather than implement a type pull-down, let's try to get the the code worked out so we can get the current found set: Filed #2166.

Issue #2099, opened 2021-08-25
`/data` instead of `/api`?


mccalluc:
If the TSV URLs don't claim to be "APIs", will that be better? @ngehlenborg will check with Bill.

mccalluc:
`metadata`

PR #2140, opened 2021-09-10
Mccalluc/tsv download button

mccalluc:
Towards #2129: Pull down for Dataset assay types not yet implemented, but I think this is a fine intermediate state.

- @john-conroy -- Basing this off of `TileSortDropDown`... I think they are different enough that trying to factor out a common Button not worth it, but could be wrong.
- @tsliaw : as we add more buttons here, I'm not sure that we want to continue reserving space for the tiles sort control -- Would it work for the search box just to use what space there is?
- @ngehlenborg : This is blocked on the resolution of #2099 -- We should get that sorted out before linking to the TSVs more publicly.

### DIFF
--- a/context/app/static/js/components/Search/DownloadButton/DownloadButton.jsx
+++ b/context/app/static/js/components/Search/DownloadButton/DownloadButton.jsx
@@ -1,15 +1,20 @@
 import React from 'react';
+import ReactGA from 'react-ga';
 
 import { SecondaryBackgroundTooltip } from 'js/shared-styles/tooltips';
 
 import { StyledButton } from './style';
+
+function sendTsvEvent(event) {
+  ReactGA.pageview(event.currentTarget.href);
+}
 
 function DownloadButton(props) {
   const { type } = props;
   const lcPluralType = `${type.toLowerCase()}s`;
   return (
     <SecondaryBackgroundTooltip title="Download a TSV of the table metadata. This will not take into account any filtering done on the search page.">
-      <StyledButton href={`/api/v0/${lcPluralType}.tsv`} target="_blank" component="a">
+      <StyledButton href={`/api/v0/${lcPluralType}.tsv`} target="_blank" component="a" onClick={sendTsvEvent}>
         Download Metadata
       </StyledButton>
     </SecondaryBackgroundTooltip>
--- a/context/app/static/js/components/Search/DownloadButton/DownloadButton.jsx
+++ b/context/app/static/js/components/Search/DownloadButton/DownloadButton.jsx
@@ -2,8 +2,14 @@ import React from 'react';
 
 import { StyledButton } from './style';
 
-function DownloadButton() {
-  return <StyledButton>Download Metadata</StyledButton>;
+function DownloadButton(props) {
+  const { type } = props;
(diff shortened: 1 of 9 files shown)


context/app/static/js/components/Search/DownloadButton/DownloadButton.jsx
```diff
@@ -0,0 +1,19 @@
...
+import { SecondaryBackgroundTooltip } from 'js/shared-styles/tooltips';
+
+import { StyledButton } from './style';
+
+function DownloadButton(props) {
+  const { type } = props;
+  const lcPluralType = `${type.toLowerCase()}s`;
+  return (
+    <SecondaryBackgroundTooltip title="Download a TSV of the table metadata. This will not take into account any filtering done on the search page.">
+      <StyledButton href={`/api/v0/${lcPluralType}.tsv`} target="_blank" component="a">
```
john-conroy, line 12: Should we add a google analytics event for this?

context/app/static/js/components/Search/DownloadButton/style.js
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+import Button from '@material-ui/core/Button';
+
+const StyledButton = styled(Button)`
+  margin-left: ${(props) => props.theme.spacing(1)}px;
+  background-color: white;
+  height: 40px;
+  margin-right: ${(props) => props.theme.spacing(1)}px;
+  border-radius: 3px;
+  border: 1px solid ${(props) => props.theme.palette.secondary.main};
```
john-conroy, line 10: Looking at the design, I don't think there is a background color. So I think you can get most of this passing the `variant="outlined" color="primary"` props to the button component.

john-conroy: Also, I wouldn't add a fixed height for the button. I would add padding if you want to increase the size incase anyone increases the default font size in their browser.

context/app/static/js/components/Search/DownloadButton/DownloadButton.jsx
mccalluc: Will do: Doesn't hurt anything.

context/app/static/js/components/Search/DownloadButton/style.js
mccalluc: Thanks: I've simplified the CSS to this:
```
  margin: 0 ${(props) => props.theme.spacing(1)}px;
  height: 40px;
``` 
which gives this:
Screen Shot 2021-09-13 at 9 54 11 AM

but if I remove the height, it's too short. `TilesSortDropdown` also has `height: 40px`... Does it make sense for them to work the same way?

john-conroy: Ah ok, keep it how it is and hopefully sometime I'll take a look.

tsliaw:
@mccalluc I think we should still reserve space for the tile sort button since there's no other way to sort the tile view, but I'll look into turning that button into an icon button to reduce the amount of space it takes up at a later time.

mccalluc:
At the meeting, Nils said we should change this from `api` to `metadata`. Will do.

PR #2149, opened 2021-09-14
Mccalluc/new organs architecture

mccalluc:


mccalluc:
Ilan has confirmed that the next version of Vitessce will fix the errors. He'll release the next Vitessce version, and add a commit here later today or tomorrow. With that in place, Chuck will just do a few more tweaks here so we can something out, just to show the gaps.

Vitessce is going to continue to be on a fairly rapid release cadence, and with Deck.gl's architecture, pollution of the global namespace is unavoidable. For the long-run, blocking Vitessce features across the site until CCF can be upgraded doesn't work. If Bruce volunteers to package their component so we can avoid these conflicts, great, but an iframe is a pragmatic solution here: it's heavy-weight, but it does give us separate JS contexts.

mccalluc:
We talked a little more yesterday: The release cadences for Vitessce and the IU component just are different, Ilan's talked with the web.gl folks and global namespace pollution is not going away, so our best solution is the iframe. It is heavy-weight, and it limits some aspects of layout, but it's workable.

mccalluc:
@john-conroy -- I'm going to continue on this branch, but I'd like to get this much in so we can demonstrate to the other teams that we're integrating their data; The TODOs make it explicit what we still plan to do.

@ngehlenborg -- Depending on the importance of delivering before the end of the quarter, we could strip back some of the features, and implement them later... or we can keep the TODOs in there until everything is done.

@ilan-gold -- I think I'm going to refactor `VisualizationWrapper` so I can get a spinner, but without the "Visualization" title. Ok?

@tsliaw -- Do we have text for the info popovers? and Can you confirm that we don't need the old IU component at the bottom?

@tsliaw or @ngehlenborg -- Which of these do we want to show in the Azimuth section? The design calls for links to reference datasets, but I'm not sure we have that.

- [ ]  [Lake et al, bioRxiv 2021](https://www.biorxiv.org/content/10.1101/2021.07.28.454201v1)
- [ ]  [Stewart et al., Science 2019](https://science.sciencemag.org/content/365/6460/1461.long) [[Seurat Object](https://seurat.nygenome.org/azimuth/demo_datasets/kidney_demo_stewart.rds)]
- [ ]  This reference consists of 64,693 kidney nuclei (10x Genomics snRNA-seq) from healthy reference states generated by investigators in the Human Biomolecular Atlas Program (HuBMAP) and the Kidney Precision Medicine Project (KPMP). The dataset represents 21 samples across 13 patient donors. Samples were derived from nephrectomy, biopsy, or deceased donor tissue, spanning three anatomical regions: medulla, cortex and papilla. The annotations and the markers are provisional (especially levels 2 and 3) and will likely get curated as more data are added to the atlas. More details regarding cell-type specific positive and negative markers can be found in the supplementary tables 4, 5 and 8 [here](https://doi.org/10.1101/2021.07.28.454201). 

Screen Shot 2021-09-23 at 3 19 20 PM

ilan-gold:
@mccalluc that sounds reasonable to me.

ngehlenborg:
We can leave the TODOs in for now, I don't think it is critical that we roll this out before the end of the quarter.

Can you clarify your question about "Which of these do we want to show in the Azimuth section?"

ngehlenborg:
We should add the Uberon ID that we are using for this organ on the page. @tsliaw we should add a header for that kind of information (there might be other identifiers in the future, too). Can you please add that to the design?

mccalluc:
@ngehlenborg 
> Can you clarify your question about "Which of these do we want to show in the Azimuth section?"

Of the three check-box items, which, if any, do we want to show on our page? If none of those, what is it that should be rendered? I'm just not sure what we want to fill in the placeholder on the design.

context/app/static/js/components/organ/Azimuth/Azimuth.jsx
```diff
@@ -1,17 +1,43 @@
...
 
   return (
     <SectionContainer>
-      <SectionHeader>Description</SectionHeader>
-      <Paper>TODO: {JSON.stringify(azimuth)}</Paper>
+      TODO: Add info popover.
+      <SpacedSectionButtonRow
+        leftText={<SectionHeader>Reference-Based Analysis</SectionHeader>}
+        buttons={
+          <Button color="primary" variant="contained" component="a" target="_blank" href={config.applink}>
```
john-conroy, line 25: Could use the `OutboundLinkButton` component to track the event.

iframe-demo.html
```diff
@@ -31,5 +31,10 @@ <h2><code>/iframe/assay-barchart</code></h2>
       <iframe src="http://localhost:5001/iframe/assay-barchart" height="450" width="800" scrolling="no"></iframe>
       <!-- The element has more whitespace at the bottom than you probably need.-->
     </div>
+
+    <h2><code>/iframe/organ</code></h2>
+    <div>
+      <iframe src="http://localhost:5001/iframe/organ?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_0002113" height="604" width="916" scrolling="no"></iframe>
```
john-conroy, line 37: The intent is for this to just work locally?

context/app/static/js/components/organ/Azimuth/Azimuth.jsx
```diff
@@ -1,17 +1,43 @@
...
+  const demodataHtml = marked(config.demodata);
+  const detailsHtml = marked(config.details);
 
   return (
     <SectionContainer>
-      <SectionHeader>Description</SectionHeader>
-      <Paper>TODO: {JSON.stringify(azimuth)}</Paper>
+      TODO: Add info popover.
+      <SpacedSectionButtonRow
+        leftText={<SectionHeader>Reference-Based Analysis</SectionHeader>}
```
john-conroy, line 23: `SectionHeader` comes with some margin baked in so that might misalign things, but no concern if it looks ok.

context/app/static/js/components/organ/Description/style.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+import Paper from '@material-ui/core/Paper';
+
+const StyledPaper = styled(Paper)`
+  padding: 20px 40px 20px 40px;
```
john-conroy, line 5:
```suggestion
  padding: 20px 40px;
```

john-conroy (approved):
Mostly questions. We might want something to validate the yaml later on I imagine?

ngehlenborg:
> @ngehlenborg
> 
> > Can you clarify your question about "Which of these do we want to show in the Azimuth section?"
> 
> Of the three check-box items, which, if any, do we want to show on our page? If none of those, what is it that should be rendered? I'm just not sure what we want to fill in the placeholder on the design.

I see. Just use the one(s) listed under "Reference dataset(s)" on the Azimuth site: 

image

I would not duplicate the description from the Azimuth website for now and we don't have to link to the demo datasets.

iframe-demo.html
mccalluc: Nope! Just sloppy.

context/app/static/js/components/organ/Azimuth/Azimuth.jsx
mccalluc: Looks ok to me... but I don't have the best eye for this. There will definitely need to be a second (third, fourth, ...) pass.

context/app/static/js/components/organ/Description/style.js
mccalluc: I've also filed
- #2172

iframe-demo.html
mccalluc: oh, sorry: Yes, this is correct as-is. Maybe there's a better approach, but with all these these iframes, I just want to be able to sanity check them. For the other ones, it's also intended as a resource for Bill, so he can see which dimensions work well.

PR #2150, opened 2021-09-15
John conroy/protein queries

### LINKED ISSUE #2142: Add accordions for steps in cells UI
Three accordions should be added according to the design - 'Query Type', 'Parameters', and 'Results'.

### LINKED ISSUE #2143: Enable protein queries in cells ui
Currently we only handle gene expression queries.

john-conroy:
Fix #2142. Fix #2143. Adds protein queries to the cells ui. The requests needed for the charts for proteins don't seem to work, but this seems like an API issue.


context/app/routes_cells.py
```diff
@@ -94,7 +94,7 @@ def _get_cluster_name_and_number(cluster_str):
     return Cluster(name=cluster_name, number=cluster_number)
 
 
-def _get_cluster_cells(cells, gene, min_gene_expression):
+def _get_cluster_cells(cells, query_type, name, min_expression):
```
mccalluc, line 97: Not a hard cut-off, but somewhere between 3-5 args we should start using some kwargs.

context/app/routes_cells.py
```diff
@@ -124,81 +124,84 @@ def _get_cluster_cells(cells, gene, min_gene_expression):
...
-            cluster_cells.append({'modality': cell['modality'], 'cluster_name': cluster_name,
-                                  'cluster_number': cluster_number,
-                                  'meets_minimum_gene_expression':
-                                  cell['values'][gene] >= min_gene_expression})
+            cluster_cell = {'cluster_name': cluster_name,
+                            'cluster_number': cluster_number,
+                            'meets_minimum_expression':
+                            cell['values'][name] >= min_expression}
+            if query_type == 'gene':
+                cluster_cell['modality'] = cell['modality']
```
mccalluc, line 166:
```suggestion
                            cell['values'][name] >= min_expression,
                            'modality': cell.get('modality')}
```
I like to avoid flow-control if something else can work.

context/app/routes_cells.py
```diff
@@ -217,13 +220,15 @@ def _get_matched_cell_counts_per_cluster(cells):
                            'modality': 'Z',
                            'unmatched': 0}]}
     '''
-    group_keys = ["cluster_name", "cluster_number", 'modality']
+    group_keys = ["cluster_name", "cluster_number"]
+    if query_type == 'gene':
+        group_keys.append('modality')
```
mccalluc, line 225:
```suggestion
    group_keys = ["cluster_name", "cluster_number", 'modality']
```
With the suggestion above, `modality` will be present, though it may be `None`.

context/app/static/js/components/cells/CellExpressionHistogram/CellExpressionHistogram.jsx
```diff
@@ -5,8 +5,8 @@ import { useTheme } from '@material-ui/core/styles';
 import Histogram from 'js/shared-styles/charts/Histogram';
 import CellsService from 'js/components/cells/CellsService';
 
-function CellExpressionHistogram({ uuid, geneName }) {
-  const [geneExpressionData, setGeneExpressionData] = useState([]);
+function CellExpressionHistogram({ uuid, cellVariableName }) {
```
mccalluc, line 8: `cellVariableName` is a good name... Maybe apply it on the python side, too? Your call.

context/app/static/js/components/cells/DatasetClusterChart/DatasetClusterChart.jsx
```diff
@@ -45,13 +45,14 @@ function DatasetClusterChart({ uuid, geneName, minGeneExpression }) {
     async function fetchCellClusterMatches() {
       const response = await new CellsService().getClusterCellMatchesInDataset({
         uuid,
-        geneName,
-        minGeneExpression,
+        name: cellVariableName,
```
mccalluc, line 48: yeah, The same name everywhere might be good.

context/app/static/js/components/cells/QuerySelect/style.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+import TextField from '@material-ui/core/TextField';
+
+const StyledTextField = styled(TextField)`
+  margin: 24px 0px;
```
mccalluc, line 5: In general, should we be trying to pull units from the theme?

context/app/static/js/pages/Cells/Cells.jsx
```diff
@@ -1,66 +1,53 @@
...
-                setMinExpressionLog={setMinExpressionLog}
-                minCellPercentage={minCellPercentage}
-                setMinCellPercentage={setMinCellPercentage}
-                geneNames={geneNames}
-                setGeneNames={setGeneNames}
+    <AccordionSteps
+      steps={[
+        {
+          heading: '1. Query Type',
+          content: <QuerySelect setQueryType={setQueryType} setCellVariableNames={setCellVariableNames} />,
```
mccalluc, line 20: Made a similar comment elsewhere, but maybe a simpler datastructure for this. I guess I start to get nervous with nested context changes in JSX, but not worth a big refactor.

mccalluc (commented):
Looks good. Some style comments. 

I would lean towards keeping the shape of objects consistent, even if sometimes the value for a key will be `None`. If you do need to read from a dict in python where you're not sure a key is present, use `.get()` rather than an `if`.

Maybe use a longer name for `name` everywhere? I'm on the fence.

PR #2158, opened 2021-09-18
Add Slide-Seq Visualization

ilan-gold:
This PR adds an `_is_spatial` option to the `RNASeqAnnDataZarrViewConfBuilder` class (and its descendants) for enabling visualization of spatially resolved data, including but not limited SlideSeq.

### DIFF
--- a/context/app/api/vitessce_confs/assay_confs.py
+++ b/context/app/api/vitessce_confs/assay_confs.py
@@ -233,6 +233,11 @@ class RNASeqAnnDataZarrViewConfBuilder(ViewConfBuilder):
     https://portal.hubmapconsortium.org/browse/dataset/e65175561b4b17da5352e3837aa0e497
     """
 
+    def __init__(self, entity, nexus_token, is_mock=False):
+        super().__init__(entity, nexus_token, is_mock)
+        # SlideSeq is a spatially resolved RNA-seq assay that requires some special handling.
+        self._is_spatial = "salmon_rnaseq_slideseq" in self._entity["data_types"]
+
     def get_conf_cells(self):
         zarr_path = 'hubmap_ui/anndata-zarr/secondary_analysis.zarr'
         file_paths_found = [file["rel_path"] for file in self._entity["files"]]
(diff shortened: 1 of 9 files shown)


context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -340,8 +379,15 @@ def get_view_config_class_for_data_types(entity, nexus_token):
...
-            return RNASeqAnnDataZarrViewConfBuilder(entity=entity, nexus_token=nexus_token)
+        if "anndata-to-ui.cwl" in dag_names:
+            if "salmon_rnaseq_slideseq" in data_types:
+                return SpatialRNASeqAnnDataZarrViewConfBuilder(
+                    entity=entity,
+                    nexus_token=nexus_token
+                )
+            return RNASeqAnnDataZarrViewConfBuilder(
+                entity=entity, nexus_token=nexus_token
+            )
```
mccalluc, line 390: Could you refactor this whole function so that it just returns the _class_, and leaves it to the caller to construct the _instance_? I think all the constructors except for `NullViewConfBuilder` have the same signature, so it shouldn't be a a huge pain. So on just these lines the change looks like:

```suggestion
        if "anndata-to-ui.cwl" in dag_names:
            if "salmon_rnaseq_slideseq" in data_types:
                return SpatialRNASeqAnnDataZarrViewConfBuilder
            return RNASeqAnnDataZarrViewConfBuilder
```
The if-then tree is getting unwieldy, so cutting the number of lines in half is a win for readability.

At the caller, the change would be something like be:

```diff
-vc = get_view_config_class_for_data_types(
-    entity=entity, nexus_token=self.nexus_token
-)
+Builder = get_view_config_builder(entity)
+vc = Builder(entity, self.nexus_token)
```

This is the sort of thing [Python 3.10's pattern matching](https://www.python.org/dev/peps/pep-0636/) might be applied to, but I don't think it's worth getting into that here.

mccalluc (commented):
Looks good, but could you try the suggested refactor? Thanks.

context/app/api/vitessce_confs/assay_confs.py
ilan-gold: Yes!

PR #2164, opened 2021-09-22
Mccalluc/finish publications

### LINKED ISSUE #2071: Publication page: structured data above / markdown below vitessce
Probably in the yaml frontmatter... but it could be done by having a placeholder for vitessce in the body.

mccalluc:
Fix #2071

The rendering is clunky right now, but before worrying about that I'd like to make sure we're in the right place about how the data should be represented... and I'm not sure we are. This structure captures what we've decided we needed... but it's adhoc, redundant, and requires data entry by hand. I wonder if it would be better to render from some standard format for citations, instead of inventing our own. [zotero](https://www.biorxiv.org/highwire/citation/2080174/zotero) would almost work, but it doesn't give us corresponding author information. Maybe another format would... but before going into that, I'd like to know where you think this should go. @ngehlenborg ?

### DIFF
--- a/CHANGELOG-publications-frontmatter.md
+++ b/CHANGELOG-publications-frontmatter.md
@@ -1,0 +1,1 @@
+- Put publication details in frontmatter.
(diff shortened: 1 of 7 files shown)


ngehlenborg:
This is fine. I don't want to make it more complicated than necessary. The driving motivation to move some required fields into the front matter was to be able to have a clear separation between required and optional content, as well as the ability to display required content in a consistent manner.

ngehlenborg (approved):

PR #2179, opened 2021-09-27
Mccalluc/add uberon to organ list

mccalluc:
If the formatting is a bit awkward (and I fully expect it is) can you file an issue for the changes you'd like to see, and approve this if it's good-enough?

Screen Shot 2021-09-27 at 1 23 15 PM
Screen Shot 2021-09-27 at 1 23 01 PM


tsliaw (approved):
This is good enough for now. I would imagine there should be a tooltip in the future to explain what UBERON is, but not necessary for right now.

PR #2180, opened 2021-09-27
Mccalluc/rename entities table

mccalluc:
Just a reorganization to bring the EntitiesTable to the top, so that the organ page can reuse it without digging under `Details/`.


john-conroy (commented):
Seems like this is abstract enough it could be pulled into `shared-styles/tables`.

mccalluc:
> Seems like this is abstract enough it could be pulled into `shared-styles/tables`.

@john-conroy : `shared-styles/tables/` doesn't exist, but `shared-styles/Table/` does... but since it doesn't actually export a `Table` component, I'll rename, and this under that.

mccalluc:
@john-conroy -- Feel free to comment, but if tests pass, I'll go ahead and merge.

john-conroy (approved):

PR #2184, opened 2021-09-29
John conroy/fix loaders

john-conroy:
- Hides/reveals not conditionally renders `ExpandableRow` content.
- Adds spring to `ExpandableRow` expand.
- Pulls expand transition/spring into reusable hooks.

Nils wanted the charts in the table to be preserved when closed/re-opened. To do that I decided to hide the expanded content instead of not rendering it similar to what MUI does with its accordions. I also added some logic to the chart fetch hooks to make sure the hook is only run on the initial expand. Using a ref instead of state ensures the component is not updated when the value is changed. I attempted to move that logic into a reusable hook, but couldn't manage it quickly and cleanly.

### DIFF
--- a/context/app/static/js/components/cells/CellExpressionHistogram/CellExpressionHistogram.jsx
+++ b/context/app/static/js/components/cells/CellExpressionHistogram/CellExpressionHistogram.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import Typography from '@material-ui/core/Typography';
 import { useTheme } from '@material-ui/core/styles';
 
@@ -6,10 +6,11 @@ import Histogram from 'js/shared-styles/charts/Histogram';
 import CellsService from 'js/components/cells/CellsService';
 import { StyledSkeleton } from 'js/components/cells/CellsCharts/style';
 
-function CellExpressionHistogram({ uuid, cellVariableName, isLoading, finishLoading, loadingKey }) {
+function CellExpressionHistogram({ uuid, cellVariableName, isLoading, finishLoading, loadingKey, isExpanded }) {
   const [expressionData, setExpressionData] = useState([]);
   const [diagnosticInfo, setDiagnosticInfo] = useState({});
   const theme = useTheme();
+  const loadedOnce = useRef(false);
 
   useEffect(() => {
     async function fetchCellExpression() {
@@ -26,8 +27,15 @@ function CellExpressionHistogram({ uuid, cellVariableName, isLoading, finishLoad
       setExpressionData(response.map((d) => d.values[cellVariableName]));
       finishLoading(loadingKey);
     }
-    fetchCellExpression();
-  }, [uuid, cellVariableName, finishLoading, loadingKey]);
+    if (loadedOnce.current) {
+      return;
+    }
+
+    if (isExpanded) {
+      fetchCellExpression();
+      loadedOnce.current = true;
+    }
+  }, [uuid, cellVariableName, finishLoading, loadingKey, isExpanded]);
 
   return Object.values(isLoading).every((val) => !val) ? (
     <>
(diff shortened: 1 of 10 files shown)


mccalluc:
I'll start looking at the code, but I wonder if it would be easier / feasible to just do a CSS change, either inline, or with a class toggle? I know that's not the react idiom, but getting into the refs takes us away from the usual idiom in a way that might be more complicated?

context/app/static/js/components/cells/CellExpressionHistogram/CellExpressionHistogram.jsx
```diff
@@ -26,8 +27,15 @@ function CellExpressionHistogram({ uuid, cellVariableName, isLoading, finishLoad
...
       finishLoading(loadingKey);
     }
-    fetchCellExpression();
-  }, [uuid, cellVariableName, finishLoading, loadingKey]);
+    if (loadedOnce.current) {
+      return;
+    }
+
+    if (isExpanded) {
+      fetchCellExpression();
```
mccalluc, line 35: Sorry if I'm not getting it, but it feels like the goal is to avoid repeating an expensive process. Could `useMemo` be used here, instead of us keeping track for ourselves of what's already been done?

context/app/static/js/components/cells/DatasetClusterChart/DatasetClusterChart.jsx
```diff
@@ -61,8 +63,14 @@ function DatasetClusterChart({
       });
       setResults(response);
     }
-    fetchCellClusterMatches();
-  }, [cellVariableName, minExpression, queryType, uuid]);
+    if (loadedOnce.current) {
+      return;
+    }
+    if (isExpanded) {
+      fetchCellClusterMatches();
```
mccalluc, line 70: ditto

context/app/static/js/hooks/useExpand.js
```diff
@@ -0,0 +1,24 @@
...
+  });
+}
+
+function useExpandSpring(ref, initialElementHeight, toggle, optionalConfig) {
+  const { height = 0 } = useResizeObserver({ ref });
+  return useSpring({
+    overflowY: 'hidden',
+    height: toggle ? height : initialElementHeight,
+    opacity: toggle ? 1 : 0,
+    config: optionalConfig || config.default,
```
mccalluc, line 21: Not sure what's in the `config.default`, but would it make any sense to add new keys from optionalConfig, but keep others from the default?

context/app/static/js/hooks/useExpand.js
```diff
@@ -0,0 +1,24 @@
...
+  });
+}
+export { useExpandTransition, useExpandSpring };
```
mccalluc, line 24: I'm not really following this, or how it would be used. Is it UI-ish enough that a storybook would make sense? Or maybe some comments.

context/app/static/js/shared-styles/Table/ExpandableRow/style.js
```diff
@@ -0,0 +1,18 @@
...
+  ${(props) =>
+    !props.$isExpanded &&
+    css`
+      visibility: hidden;
+    `};
+`;
+
+const ExpandedCell = styled(TableCell)`
+  padding: ${(props) => (props.$isExpanded ? `${props.theme.spacing(2)}px` : 0)};
+  border-bottom: ${(props) => (props.$isExpanded ? ' 1px solid rgba(224, 224, 224, 1)' : 'none')};
```
mccalluc, line 15: theme for the color?

mccalluc (commented):
I think this looks good, though I don't understand why this isn't a job for `useMemo`. I do worry about the complexity of the front-end code, mostly with a concern for maintainability. I think you're doing everything well, so it's not really about problems now, as problems when someone else needs to understand it and tweak it.

context/app/static/js/shared-styles/Table/ExpandableRow/style.js
john-conroy: Added a comment as we discussed.

john-conroy:
We discussed the changes and @mccalluc is feeling more comfortable with the direction.

PR #2188, opened 2021-10-01
Mccalluc/donor sample metadata

### LINKED ISSUE #2163: See if sample and donor metadata is available on dataset documents
... if it is available, make a design issue to expose it.

mccalluc:
Fix #2163: 
- @tsliaw , This jumps the gun a little bit: Rather than waiting for a design, I suggest we just put it all in the sample table, and prefix with `donor.` or `sample.`. (The Donor and Sample metadata will be at the bottom of the list: Objects preserve order of keys, nowadays.). Acceptable?
- @john-conroy , Note the multiline comment. At one point, I thought about doing document validation at index time, but that didn't go anywhere, because of the number of failures. Does it make sense to try to make it robust against more edge cases, or take the time now to get a better handle on our document structure?

### DIFF
--- a/CHANGELOG-metadata.md
+++ b/CHANGELOG-metadata.md
@@ -1,0 +1,1 @@
+- Add donor and sample metadata.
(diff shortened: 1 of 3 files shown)


mccalluc:
Bill writes: 
> There is no guarantee which sample in the chain it will be on, and of course if there is no published data associated with a sample you may find chains without any sample metadata.

So to code to fetch sample metadata needs to be more flexible.

PR #2203, opened 2021-10-08
John conroy/dataset version traversal

john-conroy:
Towards #1331. Adds a dropdown to dataset page to enable users to traverse between dataset versions. The design includes all detail pages, but the API only provides routes for datasets and I believe we only expect dataset versions in the near future.

I also decided to show a disabled, blank button instead of a loader until the version data is returned by the API.

If you want to check it out locally, the `next_revision_uuid` is only available in test environments right now.

Screen Shot 2021-10-15 at 2 42 13 PM

### DIFF
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -391,7 +391,7 @@ def _get_image_pyramid_descendants(entity):
 
 def _get_latest_uuid(revisions):
     '''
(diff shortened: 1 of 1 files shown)


tsliaw:
Can you remove the grey background of the selected version in the menu? It looks like a hover effect when it isn't being hovered over. 
Can you also make the hover effect stretch the entire width? There's a slight white space at the end of the hover effect.
Otherwise, everything else looks good. 
Screen Shot 2021-10-18 at 9 36 53 AM

Issue #2170, opened 2021-09-23
cross-reference the organ pages from other places where organs are mentioned


mccalluc:
We will also need to cross-reference the organ pages from other places where organs are mentioned. That might be the trickiest part.

_Originally posted by @ngehlenborg in https://github.com/hubmapconsortium/portal-ui/issues/2090#issuecomment-904852710_

mccalluc:
Fixed by #2436

PR #2206, opened 2021-10-13
Mccalluc/link to organ page

### LINKED ISSUE #2252: Uneven line weights on details header (Firefox)
On FF, the weight of the vertical bars / pipes varies depending on layout / page width.

Screen Shot 2021-11-22 at 4 17 21 PM

@john-conroy : I don't think this is worth fixing on its own... but if you're in the neighborhood, worth thinking about.

mccalluc:
Towards #2170. (To close that issue, I'd want to link to a particular organ page, and not just the list.) The motivation for all of this is `Dataset.jsx`: It makes sense to treat the link to the assay and the link to the organ the same way... that entails pulling out the trailing vertical divider and making it explicit everywhere it is needed.

- Is this a good way of approaching the problem?
- Assuming I should set up a containing directory for `VerticalDivider`... unless it can be grouped with something else?
- Right now, this is introducing a vertical alignment problem on the right side of the summary:
Screen Shot 2021-10-13 at 12 21 14 PM

Filing as draft for now.

---

Updated, moved out of draft. Fixes #2252.


mccalluc:
@john-conroy -- Replaced the VerticalDivider with Typography around `|`... Treating it the same as the text that it it punctuates seems like the easiest way to sort out the base-line alignment issues.

context/app/static/js/shared-styles/VerticalDivider.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+function VerticalDivider() {
+  return (
+    <Typography variant="h6" component="p">
+      &nbsp;|&nbsp;
```
john-conroy, line 7: What prompted the change to a pipe?

john-conroy: I'd also prefer to rely on css for spacing.

context/app/static/js/components/detailPage/SummaryItem.jsx
```diff
@@ -1,15 +1,6 @@
...
 import Typography from '@material-ui/core/Typography';
-import Divider from '@material-ui/core/Divider';
-
-const VerticalDivider = styled(Divider)`
-  margin-left: ${(props) => props.theme.spacing(1)}px;
-  margin-right: ${(props) => props.theme.spacing(1)}px;
-  height: 15px;
-  background-color: ${(props) => props.theme.palette.text.primary};
-  align-self: center;
-`;
```
john-conroy, line 12: I think with the removal of the `VerticalDivider` this component has lost its use. The flex is only used to move the divider and icon in-line, so this is only really needed in situations where the `statusIcon` is passed.

context/app/static/js/pages/Dataset/Dataset.jsx
```diff
@@ -22,21 +21,24 @@ import { getAllCollectionsQuery } from 'js/helpers/queries';
...
       </SummaryItem>
-      <Typography variant="h6" component="p">
-        {origin_sample.mapped_organ}
-      </Typography>
+      <VerticalDivider />
+      <SummaryItem>
+        <LightBlueLink variant="h6" href="/organ">
+          {origin_sample.mapped_organ}
+        </LightBlueLink>
+      </SummaryItem>
```
john-conroy, line 41: As mentioned above, as is, `SummaryItem` not needed here.

context/app/static/js/pages/Sample/Sample.jsx
```diff
@@ -78,10 +78,13 @@ function SampleDetail(props) {
           description={description}
           group_name={group_name}
         >
-          <SummaryItem>{mapped_organ}</SummaryItem>
-          <Typography variant="h6" component="p">
-            {mapped_specimen_type}
-          </Typography>
+          <SummaryItem>
```
john-conroy, line 81: `SummaryItem` not needed.

context/app/static/js/components/detailPage/SummaryData/SummaryData.jsx
```diff
@@ -59,7 +60,9 @@ function SummaryData(props) {
             {['Dataset', 'Support'].includes(entity_type) && (
               <>
                 <SummaryItem statusIcon={<StatusIcon status={status} />}>{status}</SummaryItem>
+                <VerticalDivider />
                 <SummaryItem>{`${mapped_data_access_level} Access`}</SummaryItem>
```
john-conroy, line 64: No longer need `SummaryItem` here.

john-conroy (commented):
I think this is horizontal progress. The previous situation was flawed, but this has its own drawbacks. Not sure what the best direction is as I tried before and didn't make much progress.

context/app/static/js/components/detailPage/SummaryItem.jsx
john-conroy: For most cases you can just use `Typography` now.

mccalluc:
@john-conroy :
- I'll get rid of the `SummaryItem` component entirely, and just use `Typography`
- Use CSS rather than `&nbsp;`
- Difference of opinion about `Divider` vs `|`... For me, it's actually a good thing that `|` is harder to customize, because the default behavior all I want

mccalluc:
Screenshots of the current state:

wide:
Screen Shot 2022-01-18 at 1 25 01 PM

slightly narrower:
Screen Shot 2022-01-18 at 1 25 24 PM

Assuming the tests pass on CI, the one thing that still pops out for me is the vertical alignment of the colored circle.

mccalluc:
Using CSS margin to nudge the icon around:
Screen Shot 2022-01-18 at 1 49 05 PM

context/app/static/js/components/detailPage/StatusIcon/style.js
```diff
@@ -5,7 +5,7 @@ const ColoredStatusIcon = styled(LensIcon)`
   color: ${(props) => props.theme.palette[props.$iconColor].main};
   font-size: 16px;
   margin-right: 3px;
-  align-self: center;
+  margin-bottom: -2.5px;
```
john-conroy, line 8: Can we create a component for typography with an icon which relies on flex to center the items instead of specific margin values?

context/app/static/js/components/detailPage/entityHeader/EntityHeaderContent/EntityHeaderContent.jsx
```diff
@@ -28,10 +27,10 @@ function EntityHeaderContent({ hubmap_id, entity_type, data, shouldDisplayHeader
...
         <AnimatedFlexContainer style={props} key={key} maxWidth={vizIsFullscreen ? false : 'lg'}>
           {iconMap[entity_type]}
-          <EntityHeaderItem text={hubmap_id} />
-          {Object.entries(data).map(([k, v]) => (
-            <EntityHeaderItem text={v.value || `undefined ${v.label}`} key={k} />
-          ))}
+          {hubmap_id}
+          {Object.values(data)
+            .map((v) => v.value || `undefined ${v.label}`)
+            .join(' | ')}
```
john-conroy, line 33: I understand moving from the divider to the pipe for the detail page `Summary` as it doesn't affect the page's appearance, but `EntityHeader` items are meant to be separated by lines which are the full height of its container. Also something has happened and the font size of the text/icons in the `EntityHeader` is different than before.

john-conroy: Also missing a divider/pipe between the hubmap id and organ type.

john-conroy (changes requested):
Prod

Screen Shot 2022-01-18 at 5 16 27 PM

This branch

Screen Shot 2022-01-18 at 5 16 12 PM

Could I improve with better component naming, file organization or maybe adding some stories to make it more clear where a component is used?

john-conroy:
I don't think the 'slightly narrower' layout is great, but since it would require an even larger effort than this PR and we've been given the directive to focus on laptops/desktops it probably shouldn't block.

mccalluc:
Oh: This hadn't even been on my radar:

Screen Shot 2022-01-18 at 5 16 12 PM

That makes me think the premise of simplifying by getting rid of `SummaryItem` sent this in a bad direction. Do you think there's anything worth salvaging here? We can talk at standup.

mccalluc:
Closing and starting over. Trying to replace dividers with pipe characters doesn't work when it's pinned to the top of the window.

Issue #2217, opened 2021-10-22
Implement google analytics tracking for search pages


john-conroy:
@tsliaw this seems like a good starting place for the new wave of tracking.

PR #2228, opened 2021-11-01
John conroy/track search filters

john-conroy:
Towards #2217. Reimplements search filters with minor UI changes (suggested by Tiffany) to enable tracking. Only list, hierarchical, and checkbox filters are reimplemented. Ranger filters (with histogram and slider) weren't necessary to track and would require a higher level of effort.

Each filter action will send the following event to GA with title being the title of the facet group.
```
      ReactGA.event({
        category: 'Search Page Interactions',
        action: 'Facet',
        label: `${title}: ${label}`,
      });
```

Screen Shot 2021-11-02 at 12 34 30 PM


tsliaw:
Everything looks good. I have a question about is there a way to differentiate which facets are coming from the dataset search page vs the sample vs the donors? Or if the donor/sample/dataset search pages share a facet, will they appear as the same event?

john-conroy:
The page on which an event occurs is automatically tracked by Google Analytics.

tsliaw (approved):

john-conroy:
Updated to include page title.

Screen Shot 2021-11-03 at 12 33 29 PM

PR #2244, opened 2021-11-18
John conroy/download current set tsv

### LINKED ISSUE #2166: TSV should just be the current found set
If #2012 or something equivalent is merged, we'd like TSV download to just be the current found set. With this, the explanatory popover could probably be removed.

john-conroy:
Fix #2166.


mccalluc (changes requested):

context/app/static/js/components/Search/SearchWrapper.jsx
```diff
@@ -56,11 +60,24 @@ function SearchWrapper(props) {
...
       setSearchHitsCount(results.hits.total.value);
+      async function getAndSetAllUUIDs() {
+        const {
+          query: { query, post_filter },
+        } = searchkit.query;
+        const allResults = await fetchSearchData(
+          { query, post_filter, _source: false, size: 10000 },
+          elasticsearchEndpoint,
+          groupsToken,
+        );
```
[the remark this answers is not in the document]
john-conroy: I think I'd like to handle this in `fetchSearchData` in a separate PR.

context/app/routes_api.py
```diff
@@ -9,15 +9,21 @@
...
+@blueprint.route('/metadata/v0/<entity_type>.tsv', methods=['GET', 'POST'])
 def entities_tsv(entity_type):
-    entities = _get_entities(entity_type)
+    if request.method == 'GET':
+        all_args = request.args.to_dict(flat=False)
+        constraints = {k: all_args[k] for k in all_args.keys() - {'uuids'}}
+        entities = _get_entities(entity_type, constraints, request.args.getlist('uuids'))
+    else:
+        body = request.get_json()
+        entities = _get_entities(entity_type, {}, body.get('uuids'))
```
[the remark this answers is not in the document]
john-conroy: I've added a few errors responses to the POST method.

context/app/api/client.py
```diff
@@ -69,14 +69,14 @@ def get_all_dataset_uuids(self):
             raise Exception('At least 10k datasets: need to make multiple requests')
         return uuids
 
-    def get_entities(self, plural_lc_entity_type, non_metadata_fields, constraints):
+    def get_entities(self, plural_lc_entity_type, non_metadata_fields, constraints, uuids):
```
mccalluc, line 72: Getting long enough that kwargs would make sense.

context/app/routes_api.py
```diff
@@ -1,23 +1,45 @@
...
 
 from .utils import make_blueprint, get_client, get_default_flask_data
 
 
 blueprint = make_blueprint(__name__)
 
 
-@blueprint.route('/metadata/v0/<entity_type>.tsv')
+def _get_dict_subset(d, keys_to_remove):
+    return {k: d[k] for k in d.keys() - dict.fromkeys(keys_to_remove)}
```
mccalluc, line 13: `d.keys() - dict.fromkeys(keys_to_remove)` seems like it's a `set` - `dict`? I may be confused.

```suggestion
    return {k: v for k, v in d.items() if k not in keys_to_remove}
```
or 
```suggestion
    return {k: d[k] for k in d.keys() - keys_to_remove}
```

context/app/routes_api.py
```diff
@@ -1,23 +1,45 @@
...
+
+
+def _raise_api_error(status, message):
```
mccalluc, line 16: Consider renaming... would expect a method with this name to actually `raise`.

context/app/routes_api.py
```diff
@@ -1,23 +1,45 @@
...
+from flask import Response, abort, request, render_template, jsonify
 
 from .utils import make_blueprint, get_client, get_default_flask_data
 
 
 blueprint = make_blueprint(__name__)
 
 
-@blueprint.route('/metadata/v0/<entity_type>.tsv')
+def _get_dict_subset(d, keys_to_remove):
```
mccalluc, line 12: Consider renaming: Right now, just from the function name, it's not clear whether the given keys are the ones that stay or go. I might do:
```suggestion
def _drop_keys(d, keys_to_remove):
```

mccalluc (commented):
Probably rename a couple functions, but overall looks good. Tweak and merge.

PR #2248, opened 2021-11-22
Mccalluc/move files to detail

### LINKED ISSUE #2130: `components/files/` to `components/Detail/files/`?

mccalluc:
Fix #2130


john-conroy (changes requested):
Think you missed one. Otherwise looks good, thanks!

```
  /home/runner/work/portal-ui/portal-ui/context/app/static/js/components/ServiceStatusTable/ServiceStatusTable.jsx
  Error:   11:40  error  Unable to resolve path to module 'js/components/files/GlobusLinkMessage/style'  import/no-unresolved
```

PR #2249, opened 2021-11-22
Mccalluc/and check all caps dirs

mccalluc:
Follow-up to #2248, some adhoc bash to make sure any capitalized directory contains an index. This is kind of gratuitous, but we weren't too far from the target, so decided to just do it.


john-conroy (approved):
Would it be easy to extend the script so it could exclude a list or file of directory names? Not that we need to address this case now, but I think it'd be easier to address now than later.

mccalluc:
> Would it be easy to extend the script so it could exclude a list or file of directory names?

Sure: I'll add that, and then merge.

PR #2255, opened 2021-11-23
Mccalluc/donor sample field descriptions

### LINKED ISSUE #2251: Add UI field definitions for `donor.*` and `sample.*`
These keys aren't in the `field-descriptions.yaml`: Rather than adding them to the list, make the lookup smarter.

mccalluc:
Fix #2251.

@tsliaw , does the wording ("For the original donor: ...", "For the original sample: ...") seem good to you?


tsliaw:
Maybe "From the original donor/sample"? Either way seems fine to me though.

tsliaw (approved):

PR #2258, opened 2021-11-24
organ dataset chart

### LINKED ISSUE #2257: Datasets chart on Organ page
@john-conroy : Do you think it makes any sense to try to reuse the homepage `AssayTypeBarChart` here? There is some similarity, except: 
- Just want the chart: don't need the legend or the selector,
- ... because there's only one organ at a time,
- ... and probably don't need the hover behavior.

mccalluc:
Not sure if trying to reuse AssayTypeBarChart makes any sense.

Eventually, fix #2257.


mccalluc:
John is taking over on this. Closing.

PR #2259, opened 2021-11-29
Pad around publication subheadings

### LINKED ISSUE #2165: Fix the styling on the publication page
The [one example](https://portal.hubmapconsortium.org/publication/kidney-cell-states-atlas) publication isn't linked yet from the home page. (The TMC needs to configure it.)

With the #2164, Nils confirmed that the structure is right, but the rendering isn't pretty. Might just be a matter of getting the right styling on section titles, but loop in Tiffany if needed.

mccalluc:
Fix #2165?  @tsliaw , does this look ok?
Screen Shot 2021-11-29 at 9 40 30 AM


context/app/static/js/pages/Publication/style.js
```diff
@@ -4,6 +4,10 @@ import Paper from '@material-ui/core/Paper';
 const StyledPaper = styled(Paper)`
   padding: 20px 40px 20px 40px;
   margin-bottom: 16px;
+
+  h2 {
+    padding: 20px 0 20px 0;
+  }
```
john-conroy, line 10: Sorry if jumping in here is wrong, but I prefer to style the component itself unless it's somehow inaccessible.

mccalluc: The h2 here is raw html that we generate from the markdown and inject into react... so I think this is a case where it is inaccessible... I could add a comment to that effect?

(Jumping in is welcome! I might be misunderstanding something.)

john-conroy: You could use shorthand `padding: 20px 0px;`, but not a big deal.

john-conroy (approved):

context/app/static/js/pages/Publication/style.js
```diff
@@ -4,6 +4,10 @@ import Paper from '@material-ui/core/Paper';
 const StyledPaper = styled(Paper)`
   padding: 20px 40px 20px 40px;
   margin-bottom: 16px;
+
+  h2 {
+    padding: 20px 0 20px 0;
```
mccalluc, line 9:
```suggestion
    padding: 20px 0px;
```

PR #2264, opened 2021-11-30
John conroy/organ page chart

john-conroy:
There's still quite a bit left to refactor, but I think this is good for now. All of the files related to `AssayTypesBarChart` existed previously, but were moved to `shared-styles` so some of the changes look inflated.

We opted to remove data types with brackets in their name from the home page chart, but these are shown on organ pages. The chart should mirror the table above it.

As we discussed due to some of the issues with svg stroke and bar placement/dimensions I've moved to using a brightness filter instead of the outline when a user hovers over a bar.

Screen Shot 2021-12-03 at 3 36 47 PM
Screen Shot 2021-12-03 at 3 36 16 PM
Screen Shot 2021-12-03 at 3 35 56 PM


context/app/static/js/shared-styles/charts/AssayTypeBarChart/hooks.js
```diff
@@ -14,13 +14,15 @@ function useChartPalette() {
...
   return useMemo(() => {
     if (Object.keys(rawData).length > 0) {
-      return getAssayTypeBarChartData(rawData, colorKey);
+      const buckets = rawData.aggregations.mapped_data_types.buckets.filter(filterBucketsFunc || keepAllBuckets);
+      return getAssayTypeBarChartData(buckets, colorKey);
     }
     return { formattedData: [], maxSumDocCount: undefined };
-  }, [rawData, colorKey]);
+  }, [rawData, filterBucketsFunc, colorKey]);
 }
```
john-conroy, line 26: `filterBucketsFunc` here seems a little clumsy, but I wanted to get the optional filtering inside the hook.

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.jsx
```diff
@@ -1,13 +1,18 @@
...
+import React from 'react';
 
+import ChartWrapper from 'js/shared-styles/charts/ChartWrapper';
 import useSearchData from 'js/hooks/useSearchData';
-import { useChartPalette, useAssayTypeBarChartData } from './hooks';
-import { getAssayTypesCompositeAggsQuery } from './utils';
-import AssayTypeBarChart from '../AssayTypeBarChart/AssayTypeBarChart';
-import { Flex, ChartWrapper, LegendWrapper } from './style';
-import AssayTypeBarChartDropdown from '../AssayTypeBarChartDropdown';
+import { useSelectedDropdownIndex } from 'js/shared-styles/dropdowns/DropdownListbox';
```
mccalluc, line 5: Not a big deal, but generally the non-default exports from capitalized components are kind of private, I think?

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.jsx
```diff
@@ -1,13 +1,18 @@
...
+import AssayTypeBarChart from 'js/shared-styles/charts/AssayTypeBarChart/AssayTypeBarChart';
```
mccalluc, line 6: Repeated capitalized names in the path aren't typical.

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.jsx
```diff
@@ -19,14 +24,11 @@ const organTypesQuery = {
 const assayOrganTypesQuery = getAssayTypesCompositeAggsQuery('origin_sample.mapped_organ.keyword', 'organ_type');
 const assayDonorSexQuery = getAssayTypesCompositeAggsQuery('donor.mapped_metadata.sex.keyword', 'donor_sex');
 
-function AssayTypeBarChartContainer() {
-  const [selectedColorDataIndex, setSelectedColorDataIndex] = useState(0);
+const filterDataTypesWithBracket = (bucket) => !bucket.key.mapped_data_type.includes('[');
```
mccalluc, line 27:
```suggestion
const filterOutDataTypesWithBracket = (bucket) => !bucket.key.mapped_data_type.includes('[');
```
maybe?

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.jsx
```diff
@@ -65,47 +66,42 @@ function AssayTypeBarChartContainer() {
...
 
   return (
-    <Flex>
-      <ChartWrapper>
+    <ChartArea>
+      <ChartWrapper
+        margin={margin}
+        colorScale={colorScale}
+        dropdown={
+          <div>
```
mccalluc, line 83: Add a comment explaining why the `<div>` is needed?

context/app/static/js/components/organ/OrganDatasetsChart/OrganDatasetsChart.jsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <ChartArea>
+      {search.length > 1 ? (
+        <ChartWrapper margin={margin} colorScale={colorScale}>
+          <AssayTypeBarChart {...sharedProps} />
+        </ChartWrapper>
+      ) : (
+        <AssayTypeBarChart {...sharedProps} showTooltipAndHover={false} />
+      )}
+    </ChartArea>
```
mccalluc, line 67: What's the behavior for organs that aren't in the index at all, like "Brain"... I'm expecting there to be a check that skips the chart entirely if there is no data.

context/app/static/js/shared-styles/charts/StackedBar/StackedBar.jsx
```diff
@@ -0,0 +1,60 @@
...
+  const { length, thickness, discreteAxis, categoricalAxis } = propMap[direction];
+
+  const dimensionsAndPlacementProps = {
+    [length]: bar[length],
+    [discreteAxis]: bar[discreteAxis],
+    [thickness]: Math.min(bar[thickness], maxBarThickness),
+    [categoricalAxis]:
+      bar[thickness] > maxBarThickness
+        ? bar[categoricalAxis] + (bar[thickness] - maxBarThickness) / 2
+        : bar[categoricalAxis],
```
mccalluc, line 34:
```suggestion
      bar[categoricalAxis] + Math.max(0, bar[thickness] - maxBarThickness) / 2,
```
Maybe?

context/app/static/js/shared-styles/charts/StackedBar/style.js
```diff
@@ -0,0 +1,11 @@
+import styled, { css } from 'styled-components';
+
+const StyledRect = styled.rect`
+  ${(props) =>
+    props.$isHovered &&
```
mccalluc, line 5: Could you just use CSS `:hover` instead of relying on react?

mccalluc (commented):
Looks good: Can probably be merged as is, but a couple comments which may or may not be useful.

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.jsx
john-conroy: The component is nested within the same conditional used for the `Assays` component on the Organ page which I believe has the same conditions you would want for this component.

context/app/static/js/components/organ/OrganDatasetsChart/OrganDatasetsChart.jsx
john-conroy: The component is nested within the same conditional used for the Assays component on the Organ page which I believe has the same conditions you would want for this component.

Issue #2270, opened 2021-12-02
Make Lineup public


mccalluc:
- [x] #2134
- [ ] @ngehlenborg will review notes from Ajay and see if there are additional feature requests
- [x] #2271
- [x] #2278

mccalluc:
Completed with #2281

PR #2273, opened 2021-12-02
Mccalluc/lineup columns

mccalluc:
Thought about only applying categorical to fields that are actually constrained by an enum, but for pretty much everything but the id columns, this histogram at the top doesn't seem unuseful... like for protocol DOIs, it's nice to see at a glance if they all use the same protocol, or different protocols?

Filed a couple issues in other repos to clean up the source data:
- https://github.com/hubmapconsortium/search-api/issues/415
- https://github.com/hubmapconsortium/ingest-validation-tools/issues/1005

And filed one more for the final steps needed to wrap this up and make it public:
- #2270

@ngehlenborg, chime in if there is any feedback from Ajay that is relevant here.


context/app/static/js/pages/LineUpPage/LineUpPage.jsx
```diff
@@ -1,20 +1,52 @@
...
+    // Donors:
+    'hubmap_id',
+    'medical_history',
+    // Samples:
+    'donor.hubmap_id',
+    'sample_id',
+    // Datasets:
+    'description',
+    'library_adapter_sequence',
+    'library_id',
```
john-conroy, line 28: Is there a way to get this from `ingest-validation-tools` or is hard coding this the best route for now?

john-conroy (approved):

context/app/static/js/pages/LineUpPage/LineUpPage.jsx
mccalluc: I started off thinking I would want to distinguish the string fields from the enum fields based on their definitions... but it turned out that for most string fields, even if they aren't formally enums, only a small number of values are seen, and having a histogram at the top of the column doesn't seem like a bad thing.

This may be revisited: It wouldn't be that hard to have `field-types.yaml` figure out which fields are enums.

PR #2279, opened 2021-12-06
Refactor + Support Cell DIVE (And Future Segmentation + SPRM Datasets Hopefully)

ilan-gold:
This PR does two things:

1. Adds support for Cell Dive by
2. Refactoring the current code base to have a single `MultiImageSPRMAnndataViewConfBuilder` which should be used going forward for all segmentation + SPRM datasets.  The Cell DIVE assay uses this directly.  The `StitchedCytokitSPRMViewConfBuilder` now inherits from this class and overrides a few naming conventions.

Hopefully this has some staying power for future assays....please dig into this as much as you can Chuck, I definitely got lost in the sauce doing this as evidenced by my panicked slack messages.


mccalluc:
> overrides a few naming conventions.

In the slack thread, it had sounded like there were differences in the structures that didn't seem necessary... but this is fuzzy. If the only argument against uniformity is "it will take _n_ days to rerun", I would ask Nils to push back... but maybe there are other reasons?

I'll give it a scan now...

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -428,3 +427,62 @@ def _setup_view_config_raster_cellsets_expression_segmentation(self, vc, dataset
...
+        )
+
+    def _find_ids(self):
+        """Search the image pyramid directory for all of the names of OME-TIFF files
+        to use as unique identifiers.
+        """
+        file_paths_found = [file["rel_path"] for file in self._entity["files"]]
+        full_pyramid_path = IMAGE_PYRAMID_DIR + "/" + self._image_pyramid_subdir_regex
+        pyramid_files = [file for file in file_paths_found if full_pyramid_path in file]
+        found_ids = [re.sub(r".ome.tif(f?)", "", Path(image_path).name).replace(
```
mccalluc, line 459: either:
- `\.ome\.tiff?`... and probably anchor the end?
- or might be simpler just not to bother with regexes?:
```
Path(image_path).name.replace('.ome.tiff', '').replace('.ome.tif', '').replace('_' + self._expression_id, '')

File: context/app/api/vitessce_confs/base_confs.py [hunk 2]
```diff
@@ -428,3 +427,62 @@ def _setup_view_config_raster_cellsets_expression_segmentation(self, vc, dataset
...
+                mask_path=self._mask_pyramid_subdir_regex,
+                image_name=f"{id}_{self._expression_id}",
+                mask_name=f"{id}_{self._mask_id}"
+            )
+            conf = vc.get_conf_cells().conf
+            if conf == {}:
+                raise MultiImageSPRMAnndataViewConfigError(
+                    f"Cytokit SPRM assay with uuid {self._uuid} has empty view\
+                        config for id '{id}'"
+                )
```
mccalluc, line 486:
```suggestion
```
Unless this is somehow an expected error, probably doesn't belong.

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -428,3 +427,62 @@ def _setup_view_config_raster_cellsets_expression_segmentation(self, vc, dataset
...
+            confs.append(conf)
+        return ConfCells(confs if len(confs) > 1 else confs[0], None)
```
mccalluc, line 488: Could you simplify to:
```suggestion
        return ConfCells(confs, None)
```
Consistent typing of responses makes the code a lot simpler... and even if the downstream code already handles it, I think there's a good argument for moving in a direction where interfaces a simpler.

context/app/api/vitessce_confs/fixtures/output_expected/multi_image_sprm_anndata_conf.json
```diff
@@ -0,0 +1,406 @@
+[
```
mccalluc, line 1: If I understand it correctly, the files in `input_entity` and `output_expected` are parallel? Not essential, but it might be nice to rearrange the directory structure so that input/output pairs are next to each other, each directory containing just two files. Depending the name matching is a little fragile, it's possible for the lists to fall out of sync.

mccalluc (commented):
Main concern is the degree to which the directory structures should be the same across assays... I would default to requiring parallel structures, and only handle variation (even if it's easy) after it has acceptance from Nils.

ilan-gold:
@mccalluc The idea is that what I've implemented in `MultiImageSPRMAnndataViewConfBuilder` would be the the default going forward (there is no reference to assay in it and it does not require any regexes to look for certain region names or the like).  So the deviation would be old CODEX + Cytokit runs that have this weird `stitched` naming convention.  I will ping Nils and Joel about it.  Going forward, once we see https://github.com/hubmapconsortium/search-api/pull/419/files#diff-2bee40c75016c9731a187e68922eace757f9cc0ff2f5e551c963c2fbeba5e793R43 in the hints, we will automatically use this configuration.

mccalluc:
Is this the same thing that Vasyl and Matt is talking about on Slack?

Vasyl:
> Could someone please give me a concrete example of how the desired output should look like in case of CODEX and in case of CellDive.

Matt:
> Rename the stitched base directory to pipeline_output
> Rename the expressions directory to expr
> Drop stitched from any additional file names in which it's added

If the output is going to change, should this be put on hold?... maybe make it a draft?

ilan-gold:
The future outputs of Cytokit are going to change to fulfill what is expected by `MultiImageSPRMAnndataViewConfBuilder`.  I suppose it's worth waiting for that but the idea here is that nothing going forward will be different than what is expected by `MultiImageSPRMAnndataViewConfBuilder`.  Only old runs of Cytokit will be.

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -190,41 +190,24 @@ def __init__(self, entity, groups_token, is_mock=False):
...
+    def __init__(self, entity, groups_token, is_mock=False):
+        super().__init__(entity, groups_token, is_mock)
+        self._image_pyramid_subdir_regex = STITCHED_IMAGE_DIR
+        # The ids don't match exactly with the replacement because all image files have
+        # stitched_expressions appended while the subdirectory only has /stitched/
+        self._expression_id = 'stitched_expressions'
+        self._mask_pyramid_subdir_regex = STITCHED_IMAGE_DIR.replace(
+            'expressions', 'mask'
+        )
+        self._mask_id = 'stitched_mask'
```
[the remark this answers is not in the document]
ilan-gold: We'll need this forever to handle the old versions.  They won't be deleted.

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -428,3 +427,62 @@ def _setup_view_config_raster_cellsets_expression_segmentation(self, vc, dataset
...
+        self._mask_pyramid_subdir_regex = SPRM_PYRAMID_SUBDIR.replace(
+            self._expression_id, self._mask_id
+        )
+
+    def _find_ids(self):
+        """Search the image pyramid directory for all of the names of OME-TIFF files
+        to use as unique identifiers.
+        """
+        file_paths_found = [file["rel_path"] for file in self._entity["files"]]
+        full_pyramid_path = IMAGE_PYRAMID_DIR + "/" + self._image_pyramid_subdir_regex
```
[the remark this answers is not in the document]
ilan-gold: Totally right.  No need for this to be called regex.

context/app/api/vitessce_confs/base_confs.py
ilan-gold: This is a separate issue, I think.  We have been doing this for a while and the portal relies on the list/not-list to know whether to render a drop down.  Maybe a separate issue + PR?

context/app/api/vitessce_confs/base_confs.py
```diff
@@ -237,7 +237,7 @@ class SPRMViewConfBuilder(ImagePyramidViewConfBuilder):
     """
 
     def _get_full_image_path(self):
-        return f"{self._imaging_path_regex}/{self._image_name}.ome.tiff?"
+        return f"{self._imaging_path_regex}/{self._image_name}.ome.tif(f?)"
```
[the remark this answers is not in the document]
ilan-gold: Sounds good.

context/app/api/vitessce_confs/fixtures/output_expected/multi_image_sprm_anndata_conf.json
ilan-gold: It's sorted not parallel so they can come in any order but will always output in this order
```python
for id in sorted(found_ids):
```

context/app/api/vitessce_confs/assay_confs.py
```diff
@@ -362,6 +345,8 @@ def get_view_config_builder(entity):
     dag_names = [dag['name']
                  for dag in entity['metadata']['dag_provenance_list'] if 'name' in dag]
     if "is_image" in hints:
+        if "celldive_deepcell" in data_types:
+            return MultiImageSPRMAnndataViewConfBuilder
```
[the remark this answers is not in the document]
ilan-gold: Waiting on the new hints to finalize this.  Using `assay_names` in the meantime

context/app/api/vitessce_confs/base_confs.py
ilan-gold: This is expected if one of the underlying view configs (i.e a single in the multi) errors out.  If we want to change this behavior, we could, but this is just copy-and-pasted from the previous.  Errors result in empty view configs to prevent bad UI in the frontend.

mccalluc:
Update from Ilan:
> We were going to wait on a single CODEX + Cytokit run to make sure the directory structure for future CODEX + Cytokit runs matched that of the current Cell DIVE datasets.  This involved removing things like stitched from the names and folders

mccalluc:
Propose moving view conf generation to index-time: See https://github.com/hubmapconsortium/portal-visualization/pull/1

mccalluc:
- Replaced by https://github.com/hubmapconsortium/portal-visualization/pull/3.

Closing.

Issue #2167, opened 2021-09-23
UI support for retracted datasets


mccalluc:
@shirey will indicate what the flag we should look for is, and provide an example.

@ngehlenborg : I'm guessing we do not want to see retracted datasets in search results? The work that's being done in #2162 would need to be generalized to exclude these as well... or maybe at index time we add a flag that covers all the cases where a dataset should not be in search results?

ngehlenborg:
It should be possible to find retracted datasets through the search interface but by default they should be excluded. Could this be handled as a facet with value `is_retracted` that would be set to `false` for all queries unless explicitly requested?

@shirey could probably comment on how this is handled in the data model. I guess a dataset is either retracted or not retracted. Sometimes people might want to see all that were ever available.

ngehlenborg:
Was the agenda item discussed in any of our portal calls?

shirey:
This was discussed extensively during a portal call, though I'm not sure we got into the UI much.  From my end, I'm adding two fields:
  `sub_status`  which will be set to "Retracted" in this case (and potentially has other  uses later)
  `retraction_reason` a reason for retraction and is required on retraction.
  https://github.com/hubmapconsortium/entity-api/issues/220

mccalluc:
https://portal.test.hubmapconsortium.org/browse/dataset/04e7385339167e541ad42a2636e18398

```
status | "Published"
sub_status | "Retracted"
```

mccalluc:
Details page handing is confirmed: Need to exclude from public search. Currently, [DEV has one public retracted dataset](https://portal.dev.hubmapconsortium.org/dev-search?has_substatus=true).

PR #2287, opened 2021-12-07
handle retracted on details

mccalluc:
Towards #2167 (Still need to exclude from search results.)

Screen Shot 2021-12-07 at 1 01 27 PM

### DIFF
--- a/context/app/static/js/pages/Dataset/Dataset.jsx
+++ b/context/app/static/js/pages/Dataset/Dataset.jsx
@@ -65,9 +65,12 @@ function DatasetDetail(props) {
     last_modified_timestamp,
     description,
     status,
+    sub_status,
     mapped_data_access_level,
   } = assayMetadata;
   const isLatest = !('next_revision_uuid' in assayMetadata);
+
+  const combinedStatus = sub_status || status;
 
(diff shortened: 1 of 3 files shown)


context/app/static/js/pages/Dataset/Dataset.jsx
```diff
@@ -65,10 +65,13 @@ function DatasetDetail(props) {
     last_modified_timestamp,
     description,
     status,
+    sub_status,
     mapped_data_access_level,
   } = assayMetadata;
   const isLatest = !('next_revision_uuid' in assayMetadata);
 
+  const combinedStatus = sub_status || status;
```
john-conroy, line 73: `sub_status` always takes precedence over status? Could there be a situation where the `status` is more severe and worth attention?

john-conroy (approved):

context/app/static/js/pages/Dataset/Dataset.jsx
mccalluc: Maybe? Eventually? but right now the only usage is retracted.

tsliaw (approved):

PR #2296, opened 2021-12-09
Mccalluc/na in tsv

### LINKED ISSUE #2285: Metadata download: fill in “n/a” if the field not in row
So fields which belong to a type, but which were left blank, can be distinguished from fields that would never be present.

mccalluc:
- Extends #2295
- Fix #2285

@ngehlenborg -- Please confirm that "n/a" is a good string, or suggest an alternative.


ngehlenborg (changes requested):
Let's use "N/A".

ngehlenborg (approved):

john-conroy (approved):

PR #2303, opened 2021-12-15
Mccalluc/hide retracted

### LINKED ISSUE #2167: UI support for retracted datasets

mccalluc:
- fix #2167, I think

@tsliaw -- The linked issue doesn't refer to a design... Will just this in the header be enough on the detail page, or do we need more messaging? 

Screen Shot 2021-12-15 at 1 28 34 PM

@john-conroy -- code review

@ngehlenborg -- Filing as draft for now: I see that you made the comment below... I'll add a check-box facet for this on the public search, like we already have on dev-search... not sure that's the best call, but I don't feel strongly.

> It should be possible to find retracted datasets through the search interface but by default they should be excluded. Could this be handled as a facet with value is_retracted that would be set to false for all queries unless explicitly requested?

### DIFF
--- a/CHANGELOG-hide-retracted.md
+++ b/CHANGELOG-hide-retracted.md
@@ -1,0 +1,1 @@
+- Add retracted datasets to the list of things that are hidden from the default search.
--- a/context/app/static/js/helpers/__tests__/addRestrictionsToQuery.spec.js
+++ b/context/app/static/js/helpers/__tests__/addRestrictionsToQuery.spec.js
@@ -21,11 +21,18 @@ test('it should return the correct query', () => {
           },
           {
             bool: {
-              must_not: {
-                exists: {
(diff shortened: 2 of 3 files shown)


ngehlenborg (approved):

mccalluc:
Looking into it a little more, it's not obvious how a facet could override a default query: Those documents are already removed, and there's not a mechanism that would "add them back" to results.

So I think this PR in its current state should be considered enough.

john-conroy (approved):

tsliaw (approved):

PR #2307, opened 2021-12-16
exclude "separate/"

### LINKED ISSUE #2305: MALDI-IMS confused by too many pyramids
On [this support dataset for MALDI-IMS](https://portal.hubmapconsortium.org/browse/support/83add9e02e419755afda8fab18260f52#files) I see
```
ometiff-pyramids/
  ometiffs/
    VAN0031-LK....ome.tif
    separate/
	  # Lots of TIFFs!
```
What we want is for vitessce to only worry about the top TIFF, and not worry about the files under `separate`. (A different approach would be to change the pipeline so that only the one is produced, but that would be difficult: Right now it just recurses through the directory looking for things to tile.)

For now, Chuck is going to try to understand the existing portal code, and see if a fix is feasible there.

mccalluc:
- Fix #2305

Possible ways this could go wrong:
- MALDI-IMS starts putting TIFFs in a directory other than `separate/`.
- Some other assay starts using `*separate/`, and we shouldn't ignore it.


mccalluc:
NPM error:
```
npm ERR! 503 Service Unavailable - GET https://registry.npmjs.org/vega-crossfilter/-/vega-crossfilter-4.0.2.tgz
```

mccalluc:
Huh: Looking through the code, found this... Feels like it overlaps?
```
class IMSViewConfBuilder(ImagePyramidViewConfBuilder):
    """Wrapper class for generating a Vitessce configurations
    for IMS data that excludes the image pyramids
    of all the channels separated out.
    """

    def __init__(self, entity, groups_token, is_mock=False):
        super().__init__(entity, groups_token, is_mock)
        # Do not show the separated mass-spec images.
        self.image_pyramid_regex = (
            re.escape(IMAGE_PYRAMID_DIR) + r"(?!/ometiffs/separate/)"
        )
```

ilan-gold:
> Huh: Looking through the code, found this... Feels like it overlaps?
> 
> ```
> class IMSViewConfBuilder(ImagePyramidViewConfBuilder):
>     """Wrapper class for generating a Vitessce configurations
>     for IMS data that excludes the image pyramids
>     of all the channels separated out.
>     """
> 
>     def __init__(self, entity, groups_token, is_mock=False):
>         super().__init__(entity, groups_token, is_mock)
>         # Do not show the separated mass-spec images.
>         self.image_pyramid_regex = (
>             re.escape(IMAGE_PYRAMID_DIR) + r"(?!/ometiffs/separate/)"
>         )
> ```

Yes - what you are doing here would replace that so I think you could just remove this class

mccalluc:
> aren't the directory structures checked on ingest

Only the raw datasets coming from the labs; the pipeline outputs are specified in a similar manner, but it's a different mechanism.

I'll fix merge conflicts and remove `IMSViewConfBuilder`.

mccalluc:
@ilan-gold / @jswelling -- If you can review this, if it's still relevant. Thanks!

ilan-gold (approved):

jswelling (approved):
This should work.  Worth noting that the 'separate' subdirectory is present in the raw MALDI data, not just the image pyramid.  It is explicitly permitted by the associated schema in ingest-validation-tools: https://github.com/hubmapconsortium/ingest-validation-tools/blob/main/src/ingest_validation_tools/directory-schemas/ims.yaml

PR #2314, opened 2021-12-17
Mccalluc/new docs

### LINKED ISSUE #2312: Use `https://software.docs.hubmapconsortium.org/` instead of `portal-docs`
on 12/17/2021, @shirey said that they would be happy to host all documentation on the new gh-pages based site, and @ngehlenborg is ok with this too. Then we can get rid of the submodule and move the repo to the graveyard.

mccalluc:
- Fix #2312.

Following the decision made this morning, the documentation will be hosted separately: This PR changes the links to point to the docs in their new home.
- I believe the documentation team has responsibility for moving the docs.
- Please do not approve until the documentation has been moved: You might want to check each URL.
- There is no reason this process should be drawn out. If there is a blocker, please identify it.
- If the old urls should redirect to the new ones, `*.md` can be replaced with `*.redirect`. I'll leave this to your judgement.

### DIFF
--- a/CHANGELOG-move-docs.md
+++ b/CHANGELOG-move-docs.md
@@ -1,0 +1,2 @@
+- Get rid of `docs/` links and replace with `https://software.docs.hubmapconsortium.org/about`.
+  (In a follow-up PR we'll remove the submodule, but holding off to avoid merge conflicts.)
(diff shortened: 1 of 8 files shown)


mccalluc:
Waiting for https://github.com/hubmapconsortium/software-docs/pull/25

mccalluc:
Replaced by #2428.

PR #2315, opened 2021-12-17
Mccalluc/use get view config builder in tests

### LINKED ISSUE #2308: Use `get_view_config_builder` in `test_confs.py`?
@ilan-gold , to improve our test coverage, and avoid depending on test-only code, I'm wondering how you would feel about 
```diff
- AssayViewConfClass = AssayConfClasses[assay]
+ AssayViewConfClass = get_view_config_builder(entity)
```
in `test_confs.py`?

If I make this tweak (for the real thing, would figure out the mocking):
```diff
- type_client = TypeClient(current_app.config["TYPE_SERVICE_ENDPOINT"])
+ type_client = TypeClient('https://search.api.hubmapconsortium.org')
```
... then it fails with:
```
>                    for dag in entity['metadata']['dag_provenance_list'] if 'name' in dag]
E       KeyError: 'metadata'
```

Is there reason I couldn't just add the this information to the JSON fixtures? It feels like that that would a good part of the logic to have tested, too.

mccalluc:
- Fix #2308 

@ilan-gold -- Out of the box, the tests aren't passing, so I'm incrementally adding fields to the JSON as they seem to be needed... Guessing that the old way took a short-cut around some of the logic, and it's good to fill this out? ... but I'm really flying blind.
- Is the best thing to continue reading the error messages and adding the missing pieces to the JSON?
- ... or would it be a lot easier for you to fill in the gaps?
- ... or would it make sense to assume the current behavior is correct and just get new examples from production, and just throw out all of the old ones?


mccalluc:
I've just checked locally, and this does fix the problem with `Bulk ATAC-seq [BWA + MACS2]`, and the signatures are now consistent, which simplifies other parts of the code. Linked two more issues that this PR will resolve.

mccalluc (commented):
Make the util, and then John says it's good.

ilan-gold (approved):

PR #2317, opened 2021-12-26
new approach to vitessce fixtures

mccalluc:
One possible way to extend #2315:

- Move the builders into a subdirectory, just to keep things tidy.
- Get rid of all the old fixtures.
- Replace with paired `-entity.json` and `-conf.json` files.
- Organize in directories whose names should match the builder class name.
- Remove the mock flag: Provider Flask context in tests.

TODO:
- Strip down the `entity.json` to just the parts that matter.
- Add fixtures for all the other Builder classes.

Does this seem like a good direction?


mccalluc:
Talked with John -- He's happy enough with this that I'll merge it into the parent branch, and continue work there.

PR #2318, opened 2022-01-03
tweak docker.sh

mccalluc:
- `--follow` is always what I want.
- `--timestamp` is useful, too.

### DIFF
--- a/CHANGELOG-timestamp-local-docker.md
+++ b/CHANGELOG-timestamp-local-docker.md
@@ -1,0 +1,1 @@
+- When running docker locally, include a timestamp in the log, and make logging to STDOUT the default.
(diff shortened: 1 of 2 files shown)


john-conroy (approved):

mccalluc:
With `--follow`, the command never exits, and so CI just times out. Could background the process... but I'm having second thoughts about this whole thing... might just be better to give the follow as an example command...

CHANGELOG-timestamp-local-docker.md
```diff
@@ -0,0 +1 @@
+- When running docker locally, include a timestamp in the log, and make logging to STDOUT the default.
```
mccalluc, line 1:
```suggestion
- Simplify `docker.sh`: Run `docker logs` if that's what you need.
```

Issue #2320, opened 2022-01-04
Only let one cells result row be expanded at a time


mccalluc:
To avoid multiple requests competing for resources. There should be a textual explanation of what's going on.

tsliaw:
> Only one row can be expanded at a time due to complex computation of generating the results.

ngehlenborg:
How about this? "No additional results can be expanded while detailed data are being retrieved."

PR #2321, opened 2022-01-04
increase timeout

mccalluc:
Running the docker container locally, this prevented timeout errors. [Notes.](https://github.com/hubmapconsortium/portal-ui/blob/mccalluc/debug-cells/cells-lab-notebook.md)

John will do some work in the UI to prevent a single user from expanding multiple cells rows at the same time. 

- Issue: #2320

This is not a zero-risk change, but Nils is on board. Keeping as draft until Bill or Zhao can comment on whether we may also hit timeout errors from the gateway.

### DIFF
--- a/CHANGELOG-nginx-timeout.md
+++ b/CHANGELOG-nginx-timeout.md
@@ -1,0 +1,1 @@
+- In the NGINX conf, bump the timeout up to 600s, so that requests to the cells API won't time out.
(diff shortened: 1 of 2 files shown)


mccalluc:
Zhao has filed 
- https://github.com/hubmapconsortium/gateway/issues/158

and Bill had no objections during standup, so that will come along in time, but this can be merged now.

PR #2332, opened 2022-01-10
add knee

### LINKED ISSUE #2331: Add "Knee" to organs
Bill says IU has created "Knee"... Presumably it should be added to the organ pages.

The `descriptions.yaml` needs to be updated with a purl. For example:
```
http://purl.obolibrary.org/obo/UBERON_0001264:
  name: Pancreas
  description: "An endoderm derived structure that produces precursors of
    digestive enzymes and blood glucose regulating enzymes[GO]."
  has_iu_component: True
```

Reached out to @bherr2 on Slack.

mccalluc:
Fix #2331. Filing as draft until v3 is available from IU.


mccalluc:
Replaced by #2394. Closing.

PR #2355, opened 2022-01-12
John conroy/fix link to lineup ref

john-conroy:
Filed #2356 and #2357 followup issues. I know in the past we've opted to not use stories in tests, but I'm starting to think that in cases like these, where we are testing components which depend upon one another together and don't have many cases, it saves time and avoids copy/paste confusion.

- Fixes ref usage. 
- Adds reusable dropdown menu components.


john-conroy:
...not sure why Cypress is failing. I'll take a look.

mccalluc (approved):

Issue #2350, opened 2022-01-11
Redesign organ list page


mccalluc:
- The list of organs is growing, and the wide list doesn't use space efficiently.
- IU has icons... perhaps use them?
- Different organs have different resources underneath (3d model / search results / Azimuth)... Should the top-level list suggest what is available inside?
- Perhaps think about launching a notebook for an organ?

tsliaw:
Approved Design
[Figma Link](https://www.figma.com/file/OHx2hzPxER89yPFfWFWUCi/Organ%2FTissue-Pages?node-id=801%3A600)
<img width="981" alt="Screen Shot 2022-05-19 at 1 38 14 PM" src="https://user-images.githubusercontent.com/16737715/169400380-1049afd5-726d-4957-8f0d-03c06641a79b.png">

PR #2370, opened 2022-01-18
add organ icons to config

mccalluc:
Towards #2350.

@bherr2 -- Is there anything we could hit on your side that we could pull this mapping from, instead of duplicating it here? If not, fine, but wanted to check.


bherr2:
No, I think that's probably your best bet just to duplicate the mapping.

john-conroy (approved):

PR #2374, opened 2022-01-19
Create codeql-analysis.yml

mccalluc:
I've commented out the cron, but otherwise using their template verbatim. Curious what the output will look like... May or may not be useful.


john-conroy (approved):
Changelog, thanks!

mccalluc:
Besides the changelog, it's catching some worthwhile issues:
- Uncontrolled data used in path expression
- Information exposure through an exception
- URL redirection from remote source

I'll look into these...

mccalluc:
There are still a couple warnings, but I'll merge this much, and file a follow-up PR.

PR #2382, opened 2022-01-24
John conroy/images to cdn

john-conroy:
I haven't tried deploying the stack yet. TTL here refers to how long before CloudFront sends a new request to the s3 origin. To enforce browser caching we'd set the `Cache-Control: max-age=31536000` header when uploading the images to s3.

There is some followup work to automate the resizing and submission to s3 for images, but getting the images to s3/cloudfront is the priority.

### DIFF
--- a/CHANGELOG-images-to-cdn.md
+++ b/CHANGELOG-images-to-cdn.md
@@ -1,0 +1,1 @@
+- Move move images to a cloudfront cdn with s3 origin.
--- a/cloudformation-templates/portal-images.yml
+++ b/cloudformation-templates/portal-images.yml
@@ -1,0 +1,75 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: 'Cloudfront/S3 Origin for Portal Images'
+
+Resources:
+  S3Bucket:
+    DeletionPolicy: 'Retain' # retain bucket if stack is deleted
+    Properties:
+      AccessControl: 'Private'
+      BucketName: !Sub '${AWS::StackName}-s3-origin'
+    Type: 'AWS::S3::Bucket'
+
+  S3BucketPolicy:
+    Properties:
+      Bucket: !Ref S3Bucket
+      PolicyDocument:
+        Statement:
+          - Action:
+              - 's3:GetObject'
+            Effect: 'Allow'
+            Principal:
+              CanonicalUser: !GetAtt CfOriginAccessIdentity.S3CanonicalUserId
+            Resource:
+              - !Sub 'arn:aws:s3:::${S3Bucket}/*'
+    Type: 'AWS::S3::BucketPolicy'
+
+  CloudFrontDistribution:
+    Properties:
+      DistributionConfig:
+        DefaultCacheBehavior:
+          AllowedMethods:
+            - 'HEAD'
+            - 'GET'
+          CachedMethods:
+            - 'HEAD'
+            - 'GET'
+          Compress: false
+          ForwardedValues:
+            Cookies:
+              Forward: 'none'
+            Headers:
+              - 'Origin'
+            QueryString: false
+          MinTTL: 86400
+          MaxTTL: 31536000
+          DefaultTTL: 86400
+          TargetOriginId: !Sub 's3-origin-${S3Bucket}'
+          ViewerProtocolPolicy: 'redirect-to-https'
+        Enabled: true
+        HttpVersion: 'http2'
+        IPV6Enabled: false
+        Origins:
+          - DomainName: !GetAtt S3Bucket.DomainName
+            Id: !Sub 's3-origin-${S3Bucket}'
+            OriginPath: ''
+            S3OriginConfig:
+              OriginAccessIdentity: !Sub 'origin-access-identity/cloudfront/${CloudFrontOriginAccessIdentity}'
+        PriceClass: 'PriceClass_All'
+    Type: 'AWS::CloudFront::Distribution'
+
+  CloudFrontOriginAccessIdentity:
+    Properties:
+      CloudFrontOriginAccessIdentityConfig:
+        Comment: 'Access S3 bucket content only through CloudFront'
+    Type: 'AWS::CloudFront::CloudFrontOriginAccessIdentity'
+
+Outputs:
+  S3BucketName:
+    Description: 'S3 Bucket name'
+    Value: !Ref S3Bucket
+  CloudFrontDistributionID:
+    Description: 'CloudFront distribution ID'
+    Value: !Ref CloudFrontDistribution
+  CloudFrontDistributionDomainName:
+    Description: 'CloudFront distribution domain name'
+    Value: !GetAtt CloudFrontDistribution.DomainName


mccalluc (commented):
Just reading this through, without consulting any docs, it seems reasonable, but the bigger picture isn't clear to me...
- Do we still need portal-images? The images need to live somewhere, but I don't think that needs to be git... It could just be S3... but then the S3 bucket would already exist: I don't think we'd need to create it with terraform?
- Depending on the answer to the first question, how is this used? I'm not thinking it would run with every redeploy?
- What kind of speedup do we get from using the CDN instead of just S3? It is a bit more complexity.
- Are you thinking of putting JS assets here, too, or is that not even in the picture?

If you are basing this on any particular examples, please link to them... but if you're just reading the terraform/aws docs and implementing based on them, fine.

john-conroy:
> Do we still need portal-images? The images need to live somewhere, but I don't think that needs to be git... It could just be S3... but then the S3 bucket would already exist: I don't think we'd need to create it with terraform?

The images would be uploaded to the s3 bucket created by this cloudformation stack. Are you suggesting we'd create another bucket as a backup of sorts? Also, I generally prefer to use an infrastructure as code approach.

> Depending on the answer to the first question, how is this used? I'm not thinking it would run with every redeploy?

I would only create the stack once and we add images to the s3 bucket as needed. Normally we'd have this be a part of a ci/cd pipeline, but as we've discussed before we're not set up for that.

> What kind of speedup do we get from using the CDN instead of just S3? It is a bit more complexity.
It should be faster with most benefit to our non-U.S. users, but for our purposes, the speed benefit isn't critical. Some documents suggest that it could also be cheaper.

> Are you thinking of putting JS assets here, too, or is that not even in the picture?
JS assets would be a separate, but similar, stack.

> If you are basing this on any particular examples, please link to them... but if you're just reading the terraform/aws docs and implementing based on them, fine.

https://aws.amazon.com/blogs/networking-and-content-delivery/amazon-s3-amazon-cloudfront-a-match-made-in-the-cloud/

john-conroy:
It is worth noting that there are other services/cdns better featured and cheaper for delivering images. We want to keep our infrastructure to AWS, right?

john-conroy:
@mccalluc there are a few changes since you've lasted reviewed. 

- Replaced the images' src throughout the application with cdn urls.
- Added a default `cache-control` header for the CDN. This feature was added to cloudfront in Nov 2021 and detailed [here](https://aws.amazon.com/blogs/networking-and-content-delivery/amazon-cloudfront-introduces-response-headers-policies/). The policy itself was not written from an example, but made based on the cloudformation docs.
- Removed the `portal-images` submodule. 
- Updated the README.

README.md
```diff
@@ -67,7 +67,7 @@ These are concatenated by `push.sh`.
...
 
-Finally after processing, the images should be checked into the [`portal-images`](https://github.com/hubmapconsortium/portal-images) submodule.
+Finally after processing, the images should be added to the S3 bucket, `portal-ui-images-s3-origin`, to be delivered by the cloudfront CDN. SVG files larger than 5KB should also be stored in S3 and delivered by the CDN. SVG files smaller than 5KB can be included in the repository in `context/app/static/assets/svg/`. The CDN responds with a `cache-control: max-age=1555200` header for all items, but can be overridden on a per image basis by setting the `cache-control` header for the object in S3.
```
mccalluc, line 70:
```suggestion
Finally after processing, the images should be added to the S3 bucket, `portal-ui-images-s3-origin`,
to be delivered by the cloudfront CDN.
SVG files larger than 5KB should also be stored in S3 and delivered by the CDN.
SVG files smaller than 5KB can be included in the repository in `context/app/static/assets/svg/`.
The CDN responds with a `cache-control: max-age=1555200` header for all items,
but can be overridden on a per image basis by setting the `cache-control` header for the object in S3.
```
Slight preference for putting sentences/clauses on separate lines: Can make diffs easier to read.

context/.eslintrc.yml
```diff
@@ -10,6 +10,8 @@ extends:
   - plugin:import/errors
   - plugin:import/warnings
 parser: "babel-eslint"
+globals:
+  IMAGES_CDN_ORIGIN: readonly
```
mccalluc, line 14:
```suggestion
  CDN_URL: readonly
```
- May hold assets besides images someday?
- For the CDN, "origin" means where the assets originally come from: In our case, that's S3, and that's not what you mean here.

context/app/static/js/components/home/ImageCarouselContainer/utils.js
```diff
@@ -0,0 +1,7 @@
+function getCarouselImageSrcSet(key, path) {
+  return ['320', '640', '1280', '1392'].reduce((acc, width) => {
+    return { ...acc, [`src${width}w`]: `${path}/${key}-slide-${width}w.png` };
+  }, {});
+}
```
mccalluc, line 5:
```suggestion
function getCarouselImageSrcSet(key, url) {
  return ['320', '640', '1280', '1392'].reduce((acc, width) => {
    return { ...acc, [`src${width}w`]: `${url}/${key}-slide-${width}w.png` };
  }, {});
}
```

context/app/static/js/components/home/ImageCarouselContainer/utils.spec.js
```diff
@@ -0,0 +1,10 @@
+import { getCarouselImageSrcSet } from './utils';
+
+test('should return correct srcset', () => {
+  expect(getCarouselImageSrcSet('cat', 'images/animals')).toEqual({
+    src320w: 'images/animals/cat-slide-320w.png',
+    src640w: 'images/animals/cat-slide-640w.png',
+    src1280w: 'images/animals/cat-slide-1280w.png',
+    src1392w: 'images/animals/cat-slide-1392w.png',
+  });
```
mccalluc, line 9:
```suggestion
  expect(getCarouselImageSrcSet('cat', 'http://example.com/animals')).toEqual({
    src320w: 'http://example.com/animals/cat-slide-320w.png',
    src640w: 'http://example.com/animals/cat-slide-640w.png',
    src1280w: 'http://example.com/animals/cat-slide-1280w.png',
    src1392w: 'http://example.com/animals/cat-slide-1392w.png',
  });
```

context/build-utils/webpack.common.js
```diff
@@ -77,7 +77,13 @@ const config = {
       },
     ],
   },
-  plugins: [new CleanWebpackPlugin()],
+  plugins: [
+    new CleanWebpackPlugin(),
+    new webpack.DefinePlugin({
+      // update globals in eslintrc to fix undefined errors
+      IMAGES_CDN_ORIGIN: JSON.stringify('https://d3evp8qu4tjncp.cloudfront.net'),
```
mccalluc, line 84: I had been thinking you would pass this in with the other environment variables... I guess I have a slight preference for minimizing the range of techniques we use, so if someone is familiar with the other configurations in our codebase, they'll have the right intuition about how to work with this one.

(If you do decide to change it, I don't think it needs to be exactly the same, because for now while if would come from Flask, it wouldn't need to originate with an envvar Zhao would need to add.)

Also:
```suggestion
      CDN_URL: JSON.stringify('https://d3evp8qu4tjncp.cloudfront.net'),
```

context/app/static/js/components/home/ExternalLinks/ExternalLinks.jsx
```diff
@@ -1,42 +1,43 @@
...
-
 import ExternalLink from 'js/components/home/ExternalLink';
 
 function ExternalLinks() {
   return (
     <Paper>
       <ExternalLink
-        img={<img src={HubmapLogo} height={56} width={100} alt="HuBMAP Logo" />}
+        img={
+          <img src={`${IMAGES_CDN_ORIGIN}/hubmap-person-and-text-logo.png`} height={56} width={100} alt="HuBMAP Logo" />
```
mccalluc, line 11:
```suggestion
          <img src={`${CDN_URL}/hubmap-person-and-text-logo.png`} height={56} width={100} alt="HuBMAP Logo" />
```

context/app/static/js/components/home/ExternalLinks/ExternalLinks.jsx
```diff
@@ -1,42 +1,43 @@
...
+        }
         linkHref="https://hubmapconsortium.org/"
         linkText="HuBMAP Consortium"
         description="Navigate to the HuBMAP Consortium where members are developing the tools to create an open, global atlas of the human body at the cellular level."
       />
       <ExternalLink
-        img={<img src={NIHLogo} height={56} width={56} alt="NIH Logo" />}
+        img={<img src={`${IMAGES_CDN_ORIGIN}/nih-logo.png`} height={56} width={56} alt="NIH Logo" />}
```
mccalluc, line 18:
```suggestion
        img={<img src={`${CDN_URL}/nih-logo.png`} height={56} width={56} alt="NIH Logo" />}
```

context/app/static/js/components/home/ExternalLinks/ExternalLinks.jsx
```diff
@@ -1,42 +1,43 @@
...
         linkHref="https://commonfund.nih.gov/hubmap/"
         linkText="NIH Common Fund"
         description="Learn more about the NIH Common Fund, a component of the NIH budget which is managed by the Office of Stategic Coordination/Division of Program Coordination, Planning and Strategic Coordination."
       />
       <ExternalLink
-        img={<img src={GoogleScholarLogo} height={56} width={100} alt="Google Scholar Logo" />}
+        img={
+          <img src={`${IMAGES_CDN_ORIGIN}/google-scholar-logo.png`} height={56} width={100} alt="Google Scholar Logo" />
```
mccalluc, line 25:
```suggestion
          <img src={`${CDN_URL}/google-scholar-logo.png`} height={56} width={100} alt="Google Scholar Logo" />
```

context/app/static/js/components/home/ExternalLinks/ExternalLinks.jsx
```diff
@@ -1,42 +1,43 @@
...
+        }
         linkHref="https://scholar.google.com/citations?user=CtGSN80AAAAJ"
         linkText="HuBMAP Publications"
         description="Explore the list of related publications generated by members of the HuBMAP Consortium on their Google Scholar page."
       />
       <ExternalLink
-        img={<img src={ProtocolsIoLogo} height={56} width={100} alt="Protocols.io Logo" />}
+        img={<img src={`${IMAGES_CDN_ORIGIN}/protocols-io-logo.png`} height={56} width={100} alt="Protocols.io Logo" />}
```
mccalluc, line 32:
```suggestion
        img={<img src={`${CDN_URL}/protocols-io-logo.png`} height={56} width={100} alt="Protocols.io Logo" />}
```

context/app/static/js/components/home/ExternalLinks/ExternalLinks.jsx
```diff
@@ -1,42 +1,43 @@
...
         linkHref="https://www.protocols.io/workspaces/human-biomolecular-atlas-program-hubmap-method-development/"
         linkText="HuBMAP Protocols"
         description="View the HuBMAP Consortium’s protocols in protocols.io."
       />
       <ExternalLink
-        img={<img src={HubmapLogo} height={56} width={100} alt="HuBMAP Logo" />}
+        img={
+          <img src={`${IMAGES_CDN_ORIGIN}/hubmap-person-and-text-logo.png`} height={56} width={100} alt="HuBMAP Logo" />
```
mccalluc, line 39:
```suggestion
          <img src={`${CDN_URL}/hubmap-person-and-text-logo.png`} height={56} width={100} alt="HuBMAP Logo" />
```

context/app/static/js/components/home/ImageCarouselContainer/ImageCarouselContainer.jsx
```diff
@@ -2,74 +2,36 @@ import React, { useState } from 'react';
...
-import CCFSlide1280w from 'portal-images/ccf-slide-1280w.png';
-import CCFSlide1392w from 'portal-images/ccf-slide-1392w.png';
-
 import ImageCarousel from '../ImageCarousel';
 import ImageCarouselControlButtons from '../ImageCarouselControlButtons';
 import ImageCarouselCallToAction from '../ImageCarouselCallToAction';
 import { Flex, CallToActionWrapper } from './style';
+import { getCarouselImageSrcSet } from './utils';
+
+const getCDNCarouselImageSrcSet = (key) => getCarouselImageSrcSet(key, IMAGES_CDN_ORIGIN);
```
mccalluc, line 11:
```suggestion
const getCDNCarouselImageSrcSet = (key) => getCarouselImageSrcSet(key, CDN_URL);
```

mccalluc (commented):
A few comments: The one I'm most confident about is changing the name to something without `ORIGIN`... (Not sure if it's helpful to comment on all the lines this comes up, or if it's easier just to leave it to the owner to search and replace...)

For the rest, maybe it resonates with you, or not: Your call. Checked out the branch, and it looks good locally.

john-conroy:
I ultimately decided to use webpack to define the variable because I didn't want to have to deal with getting it at runtime from flask throughout the application.

PR #2386, opened 2022-01-26
Mccalluc/new vis submodule

mccalluc:
Keeping as draft until I sort out the tests, but does this seem like a horrible idea?

- Pull code into submodule: Preparing to move it into search-api.
- Reference just the subdirectory src with a symlink.

submodule version is pinned to the hash of [this PR](https://github.com/hubmapconsortium/portal-visualization/pull/1), pending review.


mccalluc:
This is blocked pending approval of https://github.com/hubmapconsortium/portal-visualization/pull/1, so I want to keep it as draft to make sure it doesn't get merged by mistake, but it can be reviewed in the current state.

ilan-gold (approved):

mccalluc:
Unblocked. @john-conroy , please review when you have a chance.

john-conroy (approved):
I'm not a huge fan of submodules, but most of the pain is upfront which we've already dealt with.

mccalluc:
@john-conroy : Thanks for the feedback. Down the road, one possibility is redoing this as a python package that could be pip-installed... but I didn't want to front-load that while we're still seeing if this architecture will work.

PR #2390, opened 2022-01-27
Updates to support CCF EUI v3.0

bherr2:


mccalluc:
@bherr2 -- Does this mean the new organs should be ready to go, or is that something separate? I'll check this branch out locally and make sure it works.

bherr2:
Yes, all new organs are integrated

mccalluc (commented):
The component on the organ pages hasn't changed that I can see: That's the right behavior?
I see that it no longer has the Elasticsearch index as a parameter... Does this mean that it is hard-coded to point at the production instance now? I think that's fine, but I'd like to make sure my understanding is correct.

bherr2:
Yes, we now use the ccf-api endpoint for querying. This update was mainly for the EUI, I haven't looked closely at the organ-info other than to make sure the works. Probably do need to update organ page info.

bherr2:
Is [this](https://github.com/hubmapconsortium/portal-ui/blob/eui-3.0/organ-utils/descriptions.yaml) what needs updated to add more organs to the organ-info page?

bherr2:
I think that piece may need some discussion to get right.

mccalluc:
> Is [this](https://github.com/hubmapconsortium/portal-ui/blob/eui-3.0/organ-utils/descriptions.yaml) what needs updated to add more organs to the organ-info page?

yep, but I've already made PRs for each new organ: [Liver](https://github.com/hubmapconsortium/portal-ui/pull/2335/files), for example.

I think this PR is fine to merge (we've been letting the original authors do the merge to avoid confusion), and then I'll take all those organ PRs, check that they look good, and merge... unless you want to?

mccalluc (approved):

bherr2:
Ok sounds good. You can do the first pass on the new organs and let me know if there are issues. Glad to help.

PR #2391, opened 2022-01-27
Update README diagram

mccalluc:
Could add more information about what kind of dependency each is... but that could get complicated, and even this partial picture may be useful.

### DIFF
--- a/CHANGELOG-diagram.md
+++ b/CHANGELOG-diagram.md
@@ -1,0 +1,1 @@
+- Update README diagram.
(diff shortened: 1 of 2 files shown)


john-conroy (approved):
Should the image S3/Cloudfront resources be added or does that `portal-ui` node encompass them?

mccalluc:
@john-conroy : I was thinking of limiting the scope to just repos... For the audience, I'm imaging a new developer that's been tasked with making a fix, and I'm trying to communicate that there is just a _huge_ range of repos to consider, beyond just `portal-ui`.

I think there could be more diagrams... I'd be hesitant to cram more into this one, though: It's already pretty messy.

Issue #2397, opened 2022-02-01
Provide maintenance time window information on maintenance page


ngehlenborg:
The maintenance page should display information about the planned window allocated for maintenance (the same information that is displayed as a warning in the time leading up to a planned downtime):

<img width="1465" alt="image" src="https://user-images.githubusercontent.com/1216518/151979632-26daab7a-5dfc-409b-9568-60f59b97b351.png">

We should also provide the option to display an anticipated downtime window for unexpected outages.

mccalluc:
Right now the text comes from `context/app/static/js/pages/Error/utils.js`:
```
  return { title: 'Portal Maintenance', subtitle: 'Portal unavailable for scheduled maintenance.' };
```
and `context/app/static/js/components/error/ErrorBody/ErrorBody.jsx`:
```
      <>
        While the portal is under maintenance, visit the{' '}
        <OutboundLink href="https://hubmapconsortium.org/">HuBMAP Consortium</OutboundLink> website.
      </>
```
I think the proposal here is the replace the latter with the value of `GLOBAL_ALERT_MD`: @yuanzhou -- From your end, does that make sense? Will you still be able to provide this configuration even when in maintenance mode?

yuanzhou:
The portal maintenance page is actually served from these static resources: https://github.com/hubmapconsortium/gateway/tree/test-release/nginx/html/portal-ui-maintenance. @mccalluc maybe you can add a simple json config file with a configurable alert message that overwrites the default message to be displayed on the index.html. I don't mind updating the information in two places.

mccalluc:
@yuanzhou: Would it be difficult to set `GLOBAL_ALERT_MD` on your side when in maintenance mode? It might be: You'd know better than me.

If it's possible, that would be my preference: Even if it's not hard for you to set the message via two different mechanisms, it is one more special case to document and to remember.

yuanzhou:
@mccalluc currently the gateway nginx serves whatever in that https://github.com/hubmapconsortium/gateway/tree/test-release/nginx/html/portal-ui-maintenance directory for requests made to `portal.hubmapconsortium.org` under maintenance mode. Consider this as a different static website placed in a totally different directory on the same VM, so it knows nothing about the real portal-ui and its configuration running from another directory. That's why `GLOBAL_ALERT_MD` won't be picked up by the maintenance page.

[...]

PR #2414, opened 2022-02-08
fix maintenance / add cypress test

mccalluc:
I think the maintenance build has been broken since 50b147ea4ac263f4d0ea9cac43ee7b93bb0c28f0.
Towards #2397

### DIFF
--- a/context/build-utils/webpack.maintenance.js
+++ b/context/build-utils/webpack.maintenance.js
@@ -52,7 +52,7 @@ const config = {
       },
       {
         test: /\.svg$/,
-        use: { loader: '@svgr/webpack' },
+        use: ['@svgr/webpack', 'url-loader'],
       },
       {
         test: /\.(png|jpg|gif)$/,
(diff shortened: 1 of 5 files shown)


context/build-utils/webpack.maintenance.js
```diff
@@ -52,7 +52,7 @@ const config = {
       },
       {
         test: /\.svg$/,
-        use: { loader: '@svgr/webpack' },
+        use: ['@svgr/webpack', 'url-loader'],
```
john-conroy, line 55: As we discussed in standup, we should just use  the `resolve` and `module` from `webpack.common.js`.

john-conroy (changes requested):

context/build-utils/webpack.maintenance.js
mccalluc: I'd like to tackle that in a separate PR #2416. Probably easy... but my first attempt isn't working, and I'd prefer not to block everything on this.

john-conroy (approved):

PR #2415, opened 2022-02-08
Mccalluc/fix maintenance load message

mccalluc:
- Towards #2397 ... wait to close till the bundle has been delivered.

Use `import()` to split the code, so there's a standalone file with just the message:
```
$ cat app/static/js/maintenance/public/message.bundle.js
(window.webpackJsonp=window.webpackJsonp||[]).push([[1],{62:function(e){e.exports=JSON.parse('"Portal unavailable for scheduled maintenance. While the portal is under maintenance, visit the [HuBMAP Consortium website](https://hubmapconsortium.org/)."')}}]);
```
@yuanzhou , will you be able to work with this file?

Changed the display a little bit, moving the subtitle into the message.

There might be a better way to work around webpack, but this is what I came up with.

### DIFF
--- a/CHANGELOG-maintenance.md
+++ b/CHANGELOG-maintenance.md
@@ -1,0 +1,1 @@
+- Get the maintenance page build working again, and add a cypress test.
--- a/context/app/static/js/components/error/ErrorBody/ErrorBody.jsx
+++ b/context/app/static/js/components/error/ErrorBody/ErrorBody.jsx
@@ -14,7 +14,7 @@ function ErrorBody({ errorCode, urlPath, isAuthenticated, isGlobus401, isMainten
 
   useEffect(() => {
     async function fetchMessage() {
-      setMessage((await import('./message.json')).default);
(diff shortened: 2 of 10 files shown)


john-conroy (commented):
I think modifying the bundle files is more hack-y than I'd prefer. Can we not just set up an s3 bucket or similar and put a message there?

yuanzhou:
The Gateway uses the static resource: https://github.com/hubmapconsortium/gateway/blob/test-release/nginx/html/portal-ui-maintenance/maintenance.bundle.js

We can replace that message string, but we don't want to do it in a minified js file. And it's extra work if I'll have to reset it after the maintenance mode is over.

Still I'd prefer a separate config file with overwriting message to be parsed by this js file for maintenance only.

mccalluc:
Ok, doesn't seem like this a direction that makes anyone happy. on the parent issue I've asked @yuanzhou where the configuration should live.

PR #2417, opened 2022-02-09
end-user docs for publication pages

mccalluc:
Coming out of the UI meeting, the proposal was a google doc... but I think it would be more helpful to keep this documentation near the files. The authors will have a set of examples at hand, and if the structure needs to change, we're more like to keep everything in sync.

Is this everything we should provide?


tsliaw (approved):

ngehlenborg (approved):
Thank you, that is looking very good!

PR #2418, opened 2022-02-09
Mccalluc/header redesign

mccalluc:
- Fix #2237 

Will file as draft until it's known where "my lists" should go. Let me know if I've missed anything? Currently looks like:
Screen Shot 2022-02-09 at 2 22 42 PM
Screen Shot 2022-02-09 at 2 15 42 PM

PR #2424, opened 2022-02-10
Mccalluc/collections search

mccalluc:
While working on #2418, realized that we didn't need the special index page any more: the collections are in ES. I don't see an issue for this, or a design, but there was a TODO. 

Still needs work:
- metadata download is 404... but there isn't really any metadata per se, so I'm not sure what to do there.
- Assays facet is showing raw strings... the mappings in search-api aren't being applied.
- What columns and facets would be appropriate? There isn't that much data available, but could present more.

Filing as draft to get feedback. Thanks!
Screen Shot 2022-02-10 at 2 55 33 PM
Screen Shot 2022-02-10 at 2 55 19 PM


john-conroy (commented):
I think this is helpful as a visual reference for designing a search page for collections, but probably shouldn't develop past here given the impending changes to the search pages.

mccalluc:
- @john-conroy : Makes sense -- Don't want to introduce merge conflicts.
- but ... @tsliaw, can you comment on the content here? Would this set of columns and facets be sufficient, or will we need more? I'd like to get it sorted out now.

tsliaw:
Facets (In Order):
* Affiliation (includes: Contacts, Group) ----- This is to match the other search pages' "Affiliation" facets

Columns (In Order):
* HuBMAP ID
* Title
* DOI (with link to DOI attached)---- Not sure about this one
* Last Modified

I think 'Assay Types' would also be nice as both facets and columns, but I'm not sure about this one given that there's multi-values for it.

PR #2436, opened 2022-02-17
link from organ names to organ list

mccalluc:
@tsliaw -- I can only think of three places where we have an organ name to link... Can you think of others? (Right now just linking to the list: Will follow-up with deeper links, which will require a bit more work.)


tsliaw (approved):
I can't think of other places that contain the organ names at the moment

PR #2460, opened 2022-02-24
Mccalluc/show doi

### LINKED ISSUE #2454: Implement DOI on Dataset pages
Depends on 
- #2453

Bill says `.registered_doi` and `.doi_url` should be available.

mccalluc:
Fix #2454
Screen Shot 2022-02-24 at 5 24 18 PM

(The icon seems a little low... but it seems a little low on Collections, too, and if it should change, I'd prefer that to be an issue separate from this: I don't feel great about the copy and paste of style.js.)

### DIFF
--- a/CHANGELOG-dataset-doi.md
+++ b/CHANGELOG-dataset-doi.md
@@ -1,0 +1,1 @@
+- Add DOI link on dataset pages.
--- a/context/app/static/js/pages/Dataset/Dataset.jsx
+++ b/context/app/static/js/pages/Dataset/Dataset.jsx
@@ -25,9 +25,16 @@ import { getSectionOrder } from 'js/components/detailPage/utils';
 import { combineMetadata, getCollectionsWhichContainDataset } from 'js/pages/utils/entity-utils';
 
 function SummaryDataChildren(props) {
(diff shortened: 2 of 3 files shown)


john-conroy (changes requested):
Remove the `vertical-align:text-bottom` and apply `display: flex` to the `OutboundLink` and the text/icon should be aligned. The dividers in the screenshot seem to be different sizes.

ngehlenborg:
@john-conroy and @mccalluc If we can get this out before next Monday it will be greatly appreciated. Thanks!

mccalluc:
@john-conroy , trying to apply your advice, it looked like this:
Screen Shot 2022-02-28 at 4 49 40 PM
In the last commit I add `align-items: center;` and now it looks like
Screen Shot 2022-02-28 at 4 49 58 PM

(This is also missing the preceding whitespace from the original screenshot.)

I've seen what look like variations in the weight of the vertical bar, too, but on the right side: I don't this PR has anything to do with that problem.

And, if the strings get long, we don't handle it well... but I don't think this is the place to sort that out.
Screen Shot 2022-02-28 at 4 54 34 PM

mccalluc:
- Replaced by #2472

PR #2462, opened 2022-02-25
Mccalluc/to notebook

mccalluc:
Filing as draft: I want to run `docker.sh` to confirm it works in that context. I've tested out [one notebook](localhost:5001/browse/dataset/44830922b634a9ecd4812b4f27ad4923.ipynb) on colab, and it worked!


mccalluc:
I think I have a local docker problem: on both this branch and on main, `docker.sh` fails with `failed to compute cache key: "/work/app/static/public" not found: not found`

PR #2469, opened 2022-02-28
Add email icons

### LINKED ISSUE #2016: Add mail icon on `mailto:` links
Parallel with
- #2015

mccalluc:
- Fix #2016

... but I'm not really happy with all the style.js, but that seems to be the way things are done? Is there a reason not to update `js/shared-styles/icons` so the export is usable out of the box, without additional styling?

(Most of these I've confirmed the rendering, but I haven't for the globus error messages.)

### DIFF
--- a/CHANGELOG-email-icons.md
+++ b/CHANGELOG-email-icons.md
@@ -1,0 +1,1 @@
+- Add email icons on mailto links.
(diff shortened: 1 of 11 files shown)


john-conroy (changes requested):
Make an `EmailLink` component to reduce the copy and paste. Use flex to align the icons/text.

john-conroy:
Icons have a default size from MUI, but that often doesn't match what we want to display. Icons with the same `font-size` can also be very different dimensions. I believe we've added the `$fontSize` prop to the icons in shared-styles. I believe early on in this project you mentioned you'd prefer explicit css and to not rely on props for styling, but that could be one solution.

mccalluc:
- Replaced by #2473

PR #2470, opened 2022-02-28
List contributors on dataset

### LINKED ISSUE #1756: Display contributors on datasets
High-level issue: #1723

@tsliaw : I believe the conclusion was to use the same table as on Collections? So the design will just be the same contributor and contact tables as on collections, but on a dataset page. Provide a mockup if that's easy, but just knowing where it goes in the ordering of the page is sufficient.

@shirey : Update here when there a dataset contributors is the Entity JSON.

mccalluc:
Fix #1756. Change the naming since the component is for more than Collections now. I've changed the id so it works with the dataset details ToC... Collections doesn't have a sidebar ToC, and the old id `datasets-table` is actually used twice on that page, so should be changed. I'm on the fence about adding `id` as an explicit parameter.


john-conroy (approved):
I think ultimately we should move towards how the organ pages are ordered where the section wrappers are moved out of the components and into the page.

Issue #2471, opened 2022-02-28
 Conform the icon sizes for OutboundIconLink


john-conroy:
We should conform the icon sizes for `OutboundIconLink` so that they're visually consistent and so you don't need to specify font-size every time. Maybe something like a mapping of the variant used for link typography to icon font size.

PR #2472, opened 2022-02-28
John conroy/doi link

### LINKED ISSUE #2454: Implement DOI on Dataset pages

john-conroy:
Fix #2454.

We should conform the icon sizes for `OutboundIconLink` so that they're visually consistent and so you don't need to specify font-size every time. Maybe something like a mapping of the variant used for link typography to icon font size. Filed #2471.

Icon alignment is not 100% pixel perfect, but seems good enough?

Screen Shot 2022-02-28 at 4 22 24 PM


context/app/static/js/shared-styles/Links/iconLinks/IconLink/IconLink.jsx
```diff
@@ -0,0 +1,29 @@
...
+      {iconPosition === 'start' && icon}
+      <StyledSpan $iconPosition={iconPosition}>{children}</StyledSpan>
+      {iconPosition === 'end' && icon}
+    </LinkComponent>
+  );
+}
+
+IconLink.propTypes = {
+  icon: PropTypes.element.isRequired,
+  iconPosition: PropTypes.oneOf(['start', 'end']),
```
mccalluc, line 20: I would vote for making this a boolean, that would only need to be specified when the non-default option is chosen... probably `iconOnLeft` or `iconAtStart`, and the default would be the icon at the end.

context/app/static/js/components/detailPage/SummaryItem.jsx
```diff
@@ -1,31 +1,41 @@
...
-  const { children, statusIcon } = props;
+function SummaryItem({ children, statusIcon, showDivider }) {
   return (
     <StyledDiv>
       {statusIcon}
       <Typography variant="h6" component="p">
         {children}
       </Typography>
-      <VerticalDivider orientation="vertical" flexItem />
+      {showDivider && <VerticalDivider orientation="vertical" flexItem />}
```
mccalluc, line 26: I think it would be simpler just to pull this out, and we'll explicitly give `<VerticalDivider orientation="vertical" flexItem />`. We're not saving many lines of code with this.

context/app/static/js/shared-styles/Links/iconLinks/IconLink/style.js
```diff
@@ -0,0 +1,28 @@
...
+
+const StyledSpan = styled.span`
+  ${(props) =>
+    props.$iconPosition === 'start'
+      ? css`
+          margin-left: ${props.theme.spacing(0.5)}px;
+        `
+      : css`
+          margin-right: ${props.theme.spacing(0.5)}px;
+        `};
```
mccalluc, line 25: Could you instead use left/right for iconMargin and then: 
```suggestion
  ${(props) =>
    css`
      margin-${props.$iconMargin}: ${props.theme.spacing(0.5)}px;
    `};
```

mccalluc (commented):
Cool -- Thanks for doing this. There are some places where I think it might be more maintainable to pull logic out of the lower levels and make things more explicit at the top... but it's fuzzy: Do what you think best.

PR #2473, opened 2022-03-01
John conroy/email link icons

### LINKED ISSUE #2016: Add mail icon on `mailto:` links

john-conroy:
Fix #2016.

I opted to add `mailto:` inside the `EmailLinkIcon` component. Should I use a proper name other than `href` to make this more clear?

Screen Shot 2022-03-01 at 11 38 31 AM

### DIFF
--- a/CHANGELOG-email-link-icons.md
+++ b/CHANGELOG-email-link-icons.md
@@ -1,0 +1,1 @@
+- Add icons to email links.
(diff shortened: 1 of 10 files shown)


mccalluc (commented):
Yeah, I'd suggest changing the parameter name to "email" to avoid confusion. Would you like to make `1rem` the default? Off-hand, I'm not sure how I would know to use a different size in a different context.

But looks good. Thanks!

PR #2476, opened 2022-03-02
Make datasets robust against missing contributors

mccalluc:
https://portal.test.hubmapconsortium.org/browse/dataset/550c5a19866fb79faf17ecb75ea6987f  is missing contributors. With this tweak, the page renders, but Vitessce still fails: I'm branch from this branch and make another PR for that.

### DIFF
--- a/CHANGELOG-missing-contributors.md
+++ b/CHANGELOG-missing-contributors.md
@@ -1,0 +1,1 @@
+- Make datasets robust against missing `contributors`.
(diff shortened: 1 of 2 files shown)


mccalluc:
... and as for why vitessce fails: The files it needs are missing, and they are missing because the input to the pipeline had bad filenames:

Matt:
> Ah, this looks relevant:
> ```
> [ruffalo@login000 /hive/hubmap-test/data/consortium/University of Florida TMC/550c5a19866fb79faf17ecb75ea6987f/sprm_outputs]$ ls | grep polygon
> reg001_expr.ome.tiff-cell_polygons_spatial.csv.csv
> ```
> (note the .csv.csv extension)
> Ted Zhang how did this happen, and how do we ensure it doesn't happen again?

... but we could have caught this earlier when there were [no matches for the input glob](https://github.com/hubmapconsortium/portal-containers/blob/master/containers/sprm-to-anndata/context/main.py#L131). Will file.

PR #2479, opened 2022-03-03
multi-dataset notebooks

mccalluc:
@ngehlenborg - You only sketched out what you wanted from multi-dataset notebooks in our conversation: Is this going in the right direction? If it is good, I would like to clean up some copy and paste.

Adds a "Notebook" option under the "Metadata" pulldown on Search.

Returns a notebook that looks like:
> This is how you can gather metadata for HuBMAP datasets:
> ```
> !pip install requests
> ```
> ```
> from csv import DictReader, excel_tab
> from io import StringIO
> import requests
> 
> response = requests.post(
>     'http://localhost:5001/metadata/v0/datasets.tsv',
>     json={'uuids':["84fee104ebfd4acb764b800395393512", "7fc584d526749606d66a8dffc694718a", "2da29bc80379675f605afeaa41484704"]}) # Can be a very long list if there are lots of search results!
> metadata = list(DictReader(StringIO(response.text), dialect=excel_tab))
> ```
> ```
> len(metadata)
> ```
> ```
> metadata[0].keys()
> ```
> ```
> field_defs = dict(zip(*[line.split('\t') for line in response.text.split('\n')[:2]]))
> ```

### DIFF
--- a/context/app/static/js/components/searchPage/MetadataMenu/MetadataMenu.jsx
+++ b/context/app/static/js/components/searchPage/MetadataMenu/MetadataMenu.jsx
@@ -10,39 +10,36 @@ import { useStore } from 'js/shared-styles/dropdowns/DropdownMenuProvider/store'
 import DropdownMenu from 'js/shared-styles/dropdowns/DropdownMenu';
 import { StyledDropdownMenuButton, StyledLink, StyledInfoIcon } from './style';
 
-function makeDownloader({ urlPath, allResultsUUIDs, closeMenu, analyticsCategory }) {
-  async function fetchAndDownload() {
-    const response = await fetch(urlPath, {
-      method: 'POST',
-      body: JSON.stringify({ uuids: allResultsUUIDs }),
-      headers: {
-        'Content-Type': 'application/json',
-      },
-    });
-    if (!response.ok) {
-      console.error('Download failed', response);
-      closeMenu();
-      return;
-    }
-    const results = await response.blob();
-    const name = response.headers.get('content-disposition').split('=')[1];
-    const mime = response.headers.get('content-type');
-
-    const downloadUrl = createDownloadUrl(results, mime);
-    const tempLink = document.createElement('a');
-    tempLink.href = downloadUrl;
-    tempLink.download = name;
-    tempLink.click();
-
-    ReactGA.event({
-      category: analyticsCategory,
-      action: `Download ${mime}`,
-      label: urlPath.split('/').pop(),
-    });
-
+async function fetchAndDownload({ urlPath, allResultsUUIDs, closeMenu, analyticsCategory }) {
+  const response = await fetch(urlPath, {
+    method: 'POST',
+    body: JSON.stringify({ uuids: allResultsUUIDs }),
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  });
+  if (!response.ok) {
+    console.error('Download failed', response);
     closeMenu();
+    return;
   }
-  return fetchAndDownload;
+  const results = await response.blob();
+  const name = response.headers.get('content-disposition').split('=')[1];
+  const mime = response.headers.get('content-type');
+
+  const downloadUrl = createDownloadUrl(results, mime);
+  const tempLink = document.createElement('a');
+  tempLink.href = downloadUrl;
+  tempLink.download = name;
+  tempLink.click();
+
+  ReactGA.event({
+    category: analyticsCategory,
+    action: `Download ${mime}`,
+    label: urlPath.split('/').pop(),
+  });
+
+  closeMenu();
 }
 
 function MetadataMenu({ type, analyticsCategory }) {
@@ -64,12 +61,14 @@ function MetadataMenu({ type, analyticsCategory }) {
           </SecondaryBackgroundTooltip>
         </MenuItem>
         <MenuItem
-          onClick={makeDownloader({
-            urlPath: `/metadata/v0/${lcPluralType}.tsv`,
-            allResultsUUIDs,
-            closeMenu,
-            analyticsCategory,
-          })}
+          onClick={() =>
+            fetchAndDownload({
+              urlPath: `/metadata/v0/${lcPluralType}.tsv`,
+              allResultsUUIDs,
+              closeMenu,
+              analyticsCategory,
+            })
+          }
         >
           Download
           <SecondaryBackgroundTooltip title="Download a TSV of the table metadata." placement="bottom-start">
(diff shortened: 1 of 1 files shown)


ngehlenborg:
Does the list of UUIDs correspond to what the user selected, e.g., all scRNA-seq datasets in a particular organ? Or will the list of datasets correspond to *all* datasets?

In general, I assume/suggest that you are going to annotate the blocks a bit more so that users understand what the outputs are.

ngehlenborg:
We also had discussed a scenario where we would check for a particular assay type at the top of the notebook, e.g., to make a notebook that runs a particular analysis.

Example:
1. from the list of input datasets select the ones that are of type scRNA-seq (salmon) or snRNA-seq (salmon) or CODEX (SPRM), spit out some stats about number of datasets in the input list that were ignored and number that can be used for analysis
1. the aforementioned datasets should all contain information about cells and transcript/antibody quantifications as an anndata Zarr store. Add a block that iterates over all datasets in the filtered list, opens the anndata (it might be in a slightly different place for each assay type, at least for CODEX) and parses out the number of cells for the dataset. 
1. Plot a bar chart with the number of cells for each dataset and maybe group by assay type and tissue (which you could get from your metadata table)

I think this would be a great starting point for anyone to start doing cross-dataset analysis.

ngehlenborg (approved):
Approved with the understand that this we will continue to iterate on the notebook content.

mccalluc:
> Does the list of UUIDs correspond to what the user selected, e.g., all scRNA-seq datasets in a particular organ? Or will the list of datasets correspond to _all_ datasets?

It will correspond to the current found set... but if there are no filters applied, it will be everything.

> In general, I assume/suggest that you are going to annotate the blocks a bit more so that users understand what the outputs are.

Will do.

Will change it back to draft, tweak, and get John to do a code review. Thanks.

mccalluc:
@ngehlenborg -- FYI: I think this is good enough for now. One concern is the unevenness of the data: Depending on the selection, there may just not be files, or there may not be metadata, so the user could plausibly think something is wrong. Options:
- **Don't worry about this!** 👈 My favorite
- Add more explanatory notes... but the caveats will be confusing for users that get everything they want, and not helpful for users who don't.
- Make notebook generation more conditional... but then the notes will vary in ways that won't make sense to end users, and it adds complexity.
- Lean on IEC to make the data more consistent.

ngehlenborg:
In general the first option is fine but if there is no info at all about what went wrong it would confuse the user and give the impression that something is wrong with the notebook.

But as long as there's a note that no files/no metadata were found and that another query could/would work, I'm fine.

mccalluc:
> But as long as there's a note that no files/no metadata were found and that another query could/would work, I'm fine.

@ngehlenborg, No, right now the code in the notebook is not pre-evaluated: If the particular datasets they've selected have no metadata or no files, we are not checking for this at notebook generation time, and are not doing any special messaging in these cases.

If this is not what you want, please put it on the UI agenda: A thread here is not the place to sort this out.

context/app/static/js/components/searchPage/MetadataMenu/MetadataMenu.jsx
```diff
@@ -11,43 +10,47 @@ import { useStore } from 'js/shared-styles/dropdowns/DropdownMenuProvider/store'
...
+    tempLink.download = name;
     tempLink.click();
 
     ReactGA.event({
       category: analyticsCategory,
-      action: `Download Metadata`,
-      label: type,
+      action: `Download ${mime}`,
+      label: urlPath.split('/').pop(),
     });
```
john-conroy, line 42: @tsliaw if this is merged as is the google analytics event action would be changed from 'Download Metadata' to 'Download text/tab-separated-values; charset=utf-8` moving forward. @mccalluc what would the mime for the notebook be?

context/app/notebooks.py
```diff
@@ -0,0 +1,128 @@
+import json
+
+from nbformat.v4 import new_code_cell
+import black
+
```
john-conroy, line 5: A little difficult to read with the formatting as is, is there a way to get more of this formatted as python?

context/app/static/js/components/searchPage/MetadataMenu/MetadataMenu.jsx
```diff
@@ -61,12 +64,35 @@ function MetadataMenu({ type, analyticsCategory }) {
...
             <StyledInfoIcon color="primary" />
           </SecondaryBackgroundTooltip>
         </MenuItem>
+        <MenuItem
+          onClick={makeDownloader({
+            urlPath: `/notebooks/${lcPluralType}.ipynb`,
+            allResultsUUIDs,
+            closeMenu,
+            analyticsCategory,
+          })}
```
john-conroy, line 86: Alternatively
```
() => fetchAndDowload({
            urlPath: `/notebooks/${lcPluralType}.ipynb`,
            allResultsUUIDs,
            closeMenu,
            analyticsCategory,
          })

**Review by john-conroy — COMMENTED** [review]
Not for this PR, but it's worth thinking more about the UX. The menu doesn't close itself until the download returns (a couple of seconds for the tsv, not sure how long for the notebooks) which makes the page almost seem frozen.

File: context/app/notebooks.py (hunk 2 above)
Comment (john-conroy) [reply context/app/notebooks.py:5]: Referring to the f strings in the functions in `notebooks.py`.

File: context/app/static/js/components/searchPage/MetadataMenu/MetadataMenu.jsx [hunk 4]
```diff
@@ -11,43 +10,47 @@ import { useStore } from 'js/shared-styles/dropdowns/DropdownMenuProvider/store'
...
     if (!response.ok) {
-      console.error('Metadata TSV Failed', response);
+      console.error('Download failed', response);
       closeMenu();
       return;
     }
     const results = await response.blob();
+    const headers = Object.fromEntries(Array(...response.headers));
+    const name = headers['content-disposition'].split('=')[1];
+    const mime = headers['content-type'];
```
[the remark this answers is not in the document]
mccalluc: thanks!

context/app/static/js/components/searchPage/MetadataMenu/MetadataMenu.jsx
mccalluc: `application/x-ipynb+json` ... but if that's ugly, we could pass in an additional parameter: I was trying to keep the function signature simple, but if we want the stats to be continuous, it wouldn't be hard to pass in human-readable names. Let me know.

context/app/notebooks.py
mccalluc: I looked a little bit at using python introspection, so we could just define functions, and then inject the definitions of the functions. Didn't go that route because:
- The introspection utils would return the whole function body, and then we'd have to strip the first line.
- Unlike javascript, python won't let you define a multi-line function as an expression... so we'd be defining the functions in one place, and then using them another place.
- The introspection just gets kind of gnarly: I'd worry about debugging it.

So: Not really happy with the big strings, but the alternative felt like it was introducing a lot of complexity which would have been an additional burden.

context/app/static/js/components/searchPage/MetadataMenu/MetadataMenu.jsx
mccalluc: much simpler. Thanks!

context/app/notebooks.py
mccalluc: Another idea: We can check in an actual notebook to store the cell definitions. Not sure whether the variable parts will just be prepended, or if we can use a template syntax without introducing too much escaping.

PR #2489, opened 2022-03-08
Mccalluc/store notebooks as notebooks

### LINKED ISSUE #2486: store notebook definitions in notebooks
> Another idea: We can check in an actual notebook to store the cell definitions. Not sure whether the variable parts will just be prepended, or if we can use a template syntax without introducing too much escaping.

_Originally posted by @mccalluc in https://github.com/hubmapconsortium/portal-ui/pull/2479#discussion_r822101667_

mccalluc:
- Fix #2486

When/if there are user-contributed notebooks, there will be questions about how exactly to insert them, but this demonstrates how it could be done with the notebooks already being generated.

The checked-in notebooks use [`$`-templates](https://docs.python.org/3/library/string.html#string.Template): So they can't be executed as is, but they should be easy to edit in an IDE, and `$` is a sufficiently rare character in python that I don't think we'll need to worry too much about escaping.

Mixed feelings about:
- Notebooks are really easy to use with IDE tooling... but the raw diffs are going to be kind of ugly... but github may have a pretty UI for notebook diffs? not sure.
- I'm still using the SDK for single-dataset notebooks: I think there, it does help us avoid some low-level details... but for the multi-dataset notebooks extracting the information and then re-wrapping it in the SDK objects seems excessive.


john-conroy (approved):
I'm still working to better understand the context here.

mccalluc:
@john-conroy : Let me know if there's anything I can help with. I'll resolve conflicts and merge.

PR #2491, opened 2022-03-09
check syntax of notebook cells

mccalluc:


mccalluc:
I think this was based off the notebooks in an earlier state: It's not worth trying to resolve conflicts, but the [test itself](https://github.com/hubmapconsortium/portal-ui/pull/2491/files#diff-2824321dc8179f2048eb5615ad342edad557a1b07155d45b4e821f9361771c14) could still be used.

PR #2492, opened 2022-03-09
John conroy/storybook mocks

john-conroy:
Configures [msw](https://mswjs.io/docs/), which we already use for test mocks, for storybook. We use the 'server' approach to intercepting requests in our test environment, but need to use the service worker in storybook's browser environment which requires a little more configuration.

Here's a link to [msw-storybook-addon](https://storybook.js.org/addons/msw-storybook-addon) for reference. We needed to add the service worker file to our public dir which required some extra webpack config to ensure that file is only cleared before production builds.


context/package.json
```diff
@@ -109,7 +110,7 @@
     "webpack-cli": "^3.3.11",
     "webpack-dev-server": "^3.11.0",
     "webpack-manifest-plugin": "^2.2.0",
-    "webpack-merge": "^4.2.2",
+    "webpack-merge": "^5.8.0",
```
mccalluc, line 113: This is what prompts these changes? 
```diff
- const merge = require('webpack-merge');
+ const { merge } = require('webpack-merge');
```

context/package.json
```diff
@@ -127,7 +128,7 @@
     "test:coverage": "jest --coverage",
     "lint": "eslint -c .eslintrc.yml --ext .js,.jsx app/static/js/",
     "lint:fix": "eslint -c .eslintrc.yml --ext .js,.jsx --fix app/static/js/",
-    "storybook": "start-storybook -p 6006",
+    "storybook": "start-storybook -p 6006 -s app/static/public/",
```
mccalluc, line 131: This may be [deprecated](https://storybook.js.org/docs/react/api/cli-options#static-dir-deprecation)?
> Starting in 6.4 the -s flag is deprecated. Instead, use a configuration object in your .storybook/main.js file. See the [images and assets documentation](https://storybook.js.org/docs/react/configure/images-and-assets#serving-static-files-via-storybook) for more information.

Also wondering ... if the point of the this is to get the msw, could you use a directory other than `public`, and then have fewer configurations to tweak, like the gitignore?

context/build-utils/webpack.common.js
```diff
@@ -78,7 +77,6 @@ const config = {
     ],
   },
   plugins: [
-    new CleanWebpackPlugin(),
```
mccalluc, line 81: Is this going away because `mockServiceWorker.js` now lives in public, and we don't want to actually clear it out? So it's there in the production build, but nothing references it in that context?

context/build-utils/webpack.dev.js
```diff
@@ -23,6 +24,7 @@ const envConfig = {
       },
     },
   },
+  plugins: [new CleanWebpackPlugin({ cleanOnceBeforeBuildPatterns: ['**/*', '!mockServiceWorker.js'] })],
```
mccalluc, line 27: Is there a way webpack.dev.js and webpack.analyze.js could share this config?

mccalluc (commented):
Big question is whether this could live some place other than `public/`... It feels like having it there entails a lot of other changes.

Secondary question is whether the little bits of duplicated configuration at different points here could be unified... but sometimes, if something is simple enough, it'll be simpler and more clear just to keep it inline.

Could be merged as is, if you're happy with all the choices, but just putting it off for a day or two and looking at it with fresh eyes might also be useful.

.gitignore
```diff
@@ -115,9 +115,9 @@ venv.bak/
 !.vscode/portal-ui.code-snippets
 
 # public
-context/app/static/public
+context/app/static/public/*
 context/app/static/js/maintenance/public
-
+!context/app/static/public/mockServiceWorker.js
```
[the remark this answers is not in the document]
john-conroy: I'll update the `package.json`, but from [what I've read](https://mswjs.io/docs/getting-started/integrate/browser), the service worker needs to be in `/public`.

context/package.json
john-conroy: I needed to use `mergeWithCustomize` and `unique` seen in `webpack.analyze.js` to specifically override the `clean-webpack-plugin`. By default `merge` just concats the plugins array.

.gitignore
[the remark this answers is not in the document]
john-conroy: I can create a `storybook-public` dir just to house the service worker if we want to keep this separate for now and not use `msw` with a service worker elsewhere.

john-conroy:
To summarize changes since review:

- I created a separate `storybook-public` dir to house the service worker.
- The `-s` flag is deprecated as of 6.4 but we still use 6.3. We can take a look at upgrading later.
- No changes to the webpack configs outside of upgrading `webpack-merge`.

Issue #2434, opened 2022-02-16
Move viewconf build


mccalluc:
Now that viewconf build may rely on the network, it's not ideal that it blocks page rendering in the `details` route. Possibilities:
- Try again to move it to index-time.
- Move the a separate route that JS will hit from client-side.

mccalluc:
Decided in meeting to just set up a flask endpoint.

PR #2494, opened 2022-03-10
Mccalluc/new route for view conf

mccalluc:
- Towards #2434 (Turns out there was already a `.vitessce.json` endpoint: This just uses it.)

Filing as draft: Works for some visualizations... but I had forgotten that this code was also involved in vis-lifting. I think the solution is for the JSON endpoint to return a wrapper object, with the vitessce-conf, and the lifted UUID.... Or maybe there's a point in the vitessce object where we could put additional user data?

@john-conroy -- Could you give this a pre-review? I know that you've been concerned about layout shifts... but this is going to make it a lot worse: We just don't know if there is visualization section until the response comes back... and that could take a while... which is precisely why I think page render shouldn't block on it.

Backing up, maybe this is another reason to try again to move this to index-time, so all the information is there in the original JSON?

### DIFF
--- a/CHANGELOG-request-vitessce-conf-from-client.md
+++ b/CHANGELOG-request-vitessce-conf-from-client.md
@@ -1,0 +1,1 @@
+- Request vitessce conf from client-side, rather than blocking page load.
--- a/context/app/routes_browse.py
+++ b/context/app/routes_browse.py
@@ -51,13 +51,9 @@ def details(type, uuid):
         return redirect(url_for('routes_browse.details', type=actual_type, uuid=uuid))
 
     template = 'pages/base_react.html'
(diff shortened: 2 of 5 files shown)


mccalluc:
>  We can just add a skeleton loader similar to what we do for the charts in the cells ui.

But on load we won't know whether there's going to be a visualization or not, so we don't know whether to include a skeleton or not. I might be thinking about this wrong. We _could_ know... but that's moving towards putting more information in the index...

john-conroy:
Maybe it makes sense to add a `has_visualization` bool or similar to the index? I know we've wanted this information elsewhere.

diane-psc:
@john-conroy Could you please review this?

john-conroy:
@diane-psc This PR is blocked and I've removed myself as a reviewer until it's unblocked.

PR #2501, opened 2022-03-11
Mccalluc/outbound links

### LINKED ISSUE #2015: Make outbound icons the default
Tiffany says:
> I would prefer all outbound links (except CCF/Azimuth) to get an icon so the users know they are navigating away from the portal

mccalluc:
Fix  #2015

### DIFF
--- a/CHANGELOG-outbound-links.md
+++ b/CHANGELOG-outbound-links.md
@@ -1,0 +1,1 @@
+- Add icons to links going outside HuBMAP.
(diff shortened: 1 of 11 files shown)


john-conroy (approved):
Feel free to make the default `iconFontSize` prop `1rem`. For these cases, where the typography is either `body1` or `body2`, that works. I've mentioned in the past it might be worth while to map icon font size to the typography size for specific cases like these, but haven't gotten around to it.

PR #2536, opened 2022-03-24
If search looks like "HBM...", put quotes around it

### LINKED ISSUE #2530: Design info box below search, if user wants "HBM..."
- Search still happens, I think... not sure if we want to try to explain what the other matches are about.
- Not sure if we need to pre-check that the HBM ID is actually valid... or just offer the redirect link, and if they've somehow gotten it wrong, that's the user's problem.
- but before investing time, see if #2532 might be easy

### LINKED ISSUE #2531: Implement "HBM..." search info box
Waiting for design: 
- #2530
- but before investing time, see if #2532 might be easy

### LINKED ISSUE #2532: Quote-wrap queries that like look like HBM ID
I can confirm that [q="HBM975.FVCG.922"](https://portal.hubmapconsortium.org/search?q=%22HBM975.FVCG.922%22&entity_type[0]=Dataset) returns only the one result, but without the quotes, there are more.

mccalluc:
- Fix #2532
- Resolve #2531 (not needed)
- Resolve #2530 (not needed)
- Resolve https://github.com/hubmapconsortium/search-api/issues/328 (not needed)

### DIFF
--- a/CHANGELOG-anndata.md
+++ b/CHANGELOG-anndata.md
@@ -1,1 +1,0 @@
-- Stub out anndata section.
--- a/CHANGELOG-default-icon-size.md
+++ b/CHANGELOG-default-icon-size.md
@@ -1,1 +1,0 @@
-- Link icons default to `1rem` if unspecified.
--- a/CHANGELOG-fix-submodule-error.md
+++ b/CHANGELOG-fix-submodule-error.md
@@ -1,0 +1,1 @@
+- Fix error for deleted portal-visualization submodule.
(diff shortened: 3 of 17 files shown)


mccalluc:
>  Error: fatal: No url found for submodule path 'context/portal-visualization' in .gitmodule

This may be my first PR after removing the submodule... There seems to be a loose thread. I'll try to fix this in a separate branch.

mccalluc:
@john-conroy ... This is already up to date with `main`... but you said you had fixed a submodule problem already? Maybe you still need to push a PR... or maybe something else is wrong?

john-conroy (approved):

PR #2540, opened 2022-03-28
John conroy/always show detail subheader

john-conroy:
Follow up in #2544.

- Fixes the entity subheader atop the detail pages.
- Removes intersection observer and store used for the appearance of the entity subheader.
- Adds the es document to the app providers.
- Reorganizes detail page alerts.


mccalluc (approved):
Looks good to me: Let me know if I should look closer at anything in particular.

PR #2545, opened 2022-03-29
John conroy/pin werkzeug

john-conroy:
Pin werkzeug version to 2.03 to fix CI errors seen in https://github.com/hubmapconsortium/portal-ui/runs/5738927912?check_suite_focus=true.

From: https://werkzeug.palletsprojects.com/en/2.1.x/test/#werkzeug.test.Client.open
> Changed in version 2.1: Removed the as_tuple parameter.

### DIFF
--- a/context/requirements.txt
+++ b/context/requirements.txt
@@ -6,5 +6,5 @@ python-datauri==0.2.8
 python-frontmatter==0.5.0
 hubmap-api-py-client==0.0.9
 hubmap-commons==2.0.12
-Werkzeug==2.0.3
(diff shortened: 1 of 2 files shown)


context/requirements.txt
```diff
@@ -6,4 +6,5 @@ python-datauri==0.2.8
 python-frontmatter==0.5.0
 hubmap-api-py-client==0.0.9
 hubmap-commons==2.0.12
+Werkzeug==2.0.3
```
mccalluc, line 9:
```suggestion
Werkzeug==2.0.3  # TODO: This is just a transitive dependency. Remove when https://github.com/hubmapconsortium/portal-ui/issues/2547 is fixed.
```

mccalluc (approved):

PR #2551, opened 2022-03-31
John conroy/searchkit concept

john-conroy:
Adds a basic dataset search page to return results in a table. I'll next work on adding facet components.

The factory function `createFacet` will eventually depend on the type definitions in ingest-validation-tools' `field-types.yaml` to determine which type of facet should be used.


context/app/static/js/components/entity-search/SearchWrapper/utils.js
```diff
@@ -0,0 +1,46 @@
...
+      label: `${labelPrefix}Sex`,
+    },
+    {
+      field: `${pathPrefix}mapped_metadata.race`,
+      label: `${labelPrefix}Race`,
+    },
+  ];
+}
+
+function createFacet({ field, label, ...rest }) {
```
mccalluc, line 19: I think this is fine for now, but if it's helpful, here's some pseudo code for how this might be more OO:
```javascript
class FieldDef {
  static createFieldDef(id) {
    // Factory pattern
    const knownMetadataFields = ...
    if (knownMetadataFields.contains(id)) {
      return new MetadataFieldDef(id)
    } else {
      return new EntityFieldDef(id)
    }
  }
  constructor(id) {
    this.id = id;
  }
}

class MetadataFieldDef extends FieldDef {
  getType() {
    // pull from submodule
  }
  getDescription() {
    // pull from submodule
  }
}

class EntityFieldDef extends FieldDef {
  getType() {
    // hardcode here?
  }
  getDescription() {
    // hardcode here?
  }
}
```

mccalluc (commented):
I'm listening to the talk at the same time, so I'm not really focused, but some thoughts. Thanks.

context/app/static/js/components/entity-search/SearchWrapper/utils.js
john-conroy: Thanks!

context/app/static/js/components/entity-search/ResultsTable/ResultsTable.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+function ResultsTable({ hits }) {
+  const { fields } = useStore();
+
+  return (
+    <StyledTable data-testid="search-results-table">
+      <TableHead>
+        <TableRow>
+          {fields.map(({ label }) => (
+            <TableCell key={label}>{label}</TableCell>
```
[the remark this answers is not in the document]
john-conroy: Fixed to use the field name.

context/app/static/js/components/entity-search/Search/RequestTransporter.js
```diff
@@ -0,0 +1,20 @@
+// Copied from https://www.searchkit.co/docs/reference/searchkit-sdk#network-transport
+// Licensed under https://github.com/searchkit/searchkit/blob/next/LICENCE
```
mccalluc, line 2: Thanks! I think the Apache license requires two more things from us:
- Include the original copyright
- Include the original license

Maybe just paste both into this file? Or maybe reorganize so all the code we're copy-and-pasting from them lives in the same directory?

mccalluc (commented):
I think the Apache license requires a little more from us, but after that, looks good to merge.

context/app/static/js/components/entity-search/Search/RequestTransporter.js
john-conroy: I moved the modified files to a directory and included copies of the license and copyright.

mccalluc (approved):
Thanks!

PR #2552, opened 2022-04-01
Split tests into four chunks

### LINKED ISSUE #2546: split CI
Maybe split into 4 separate files:
- `python-ci.yaml`: Install python dependencies and run python tests and pyflake
- `node-ci.yaml`: Install npm dependencies and run jest tests and eslist
- `cypress-ci.yaml`: Build docker container and run cypress
- `changelog-ci.yaml`: changelog

Using a build matrix, perhaps ranging over `*-test.sh`, would be one way to do this... but I think we could save processing by only doing the setup that a particular test suite needs... Though that means more details are buried in CI configuration. Trade-offs.

I don't think I've run `test.sh` in a very long time, so consider getting rid of it?

I think it's over-complicated for us, but we could [programmatically build the matrix](https://imhoff.blog/posts/parallelizing-jest-with-github-actions)... maybe a good technique for a more mono-lingual project?

mccalluc:
- fix #2546
- I think I'm ok with getting rid of `test.sh`... but it's arguable ... maybe we want to be able to run everything locally with a single command.
- I'm only running tests on PR, not the push... just because there are just so many CI runs otherwise. This does mean that tests won't until the PR is started. Also:
  - possible that push would have failed, but pull passes... But this seems unlikely: maybe a test suite has just been disabled on main for some reason?
  - possible that push would have passed, but pull fails... Because the reason for the failure is a change on main. This does happen. We'll just need to remember that possibility?


.github/workflows/cypress-ci.yml
```diff
@@ -1,6 +1,6 @@
-name: Portal CI
+name: Cypress CI
 
-on: [push, pull_request]
+on: [pull_request]
```
john-conroy, line 3:
```suggestion
on:
  push:
    branches:
      - main
  pull_request:
```

We only want ci to run once for PRs and their updates, but we also want ci to run on pushes to main? With `pull_request` alone CI wouldn't run once PRs are merged to main. This applies to the other actions as well.

john-conroy (changes requested):
Cypress still takes ~13m to finish. Maybe that's just how long it takes to run, but worth looking into later. Thanks!

john-conroy (approved):

PR #2555, opened 2022-04-04
John conroy/searchkit list facet

john-conroy:
- Add search sidebar and layout.
- Add select facet component.

The select components are repurposed from our existing search components.


context/app/static/js/components/entity-search/Search/Search.jsx
```diff
@@ -41,7 +43,14 @@ function Search() {
   const variables = useSearchkitVariables();
   const { results } = useSearchkitSDK(config, variables, transporter, filters);
 
-  return results?.hits ? <ResultsTable hits={results.hits} /> : null;
+  return (
+    <SearchLayout>
+      <SidebarLayout>
+        {results?.facets && results.facets.map((facet) => <SelectFacet key={facet.identifier} facet={facet} />)}
+      </SidebarLayout>
+      <ResultsLayout>{results?.hits && <ResultsTable hits={results.hits} />}</ResultsLayout>
```
mccalluc, line 51: Down the road would these both be ternaries? ... but shouldn't there always be facets, even if no results are returned?

context/app/static/js/components/entity-search/facets/FacetAccordion/style.js
```diff
@@ -0,0 +1,22 @@
...
+import { StyledAccordionSummary } from 'js/components/searchPage/filters/style';
+
+const InnerAccordionDetails = styled(AccordionDetails)`
+  flex-direction: column;
+  padding: 4px 10px 4px 16px;
+`;
+
+const InnerAccordionSummary = styled(StyledAccordionSummary)`
+  justify-content: left;
+  padding: 0px 16px 0px 16px;
```
mccalluc, line 13: My sense is that color discrepancies would get to be annoying, and so we put them in theme, but for distances it doesn't make sense to worry about uniformity: The only question is whether one particular component looks right. Is that a good way of framing it?

context/app/static/js/components/entity-search/facets/select/FacetItemLabelCount/FacetItemLabelCount.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+import { Flex, FormLabelText } from './style';
+
+function FacetItemLabelCount({ label, count, active, ...rest }) {
+  return (
+    <Flex {...rest} $active={active}>
+      <FormLabelText>{label}</FormLabelText>
```
mccalluc, line 9: could use `children`... but it doesn't matter.

mccalluc (approved):
Thanks!

context/app/static/js/components/entity-search/Search/Search.jsx
john-conroy: So long as there isn't an error the results object will always be returned. Hits are the actual 'results' which can be empty.

context/app/static/js/components/entity-search/facets/select/FacetItemLabelCount/FacetItemLabelCount.jsx
john-conroy: There wasn't a clear child in this case, so I opted to use the 'slots' pattern.

PR #2560, opened 2022-04-06
John conroy/group facets

### LINKED ISSUE #2444: Filters should be grouped by category and nested together in accordions. The Filters themselves should also be nested in accordions and collapsed by clicking on their titles.

john-conroy:
Fix #2444. Groups the facets. Facets without a defined `facetGroup` will be grouped into 'Additional Facets'.

The functions used to create the facet objects will change, but the shape should be eventually be something along the lines of `{field, label, type, facetGroup}`.


context/app/static/js/components/entity-search/Search/Search.jsx
```diff
@@ -32,7 +32,7 @@ function Search() {
       query: new MultiMatchQuery({
         fields: ['all_text'],
       }),
-      facets: facets.map((facet) => createFacet(facet)),
+      facets: Object.values(facets).map((facet) => createSearchkitFacet(facet)),
```
mccalluc, line 35: Not new with this PR, but it could be confusing for `facets` to be an Object in some contexts, and an Array in others. Not sure which name should change and which should stay.

context/app/static/js/components/entity-search/SearchWrapper/SearchWrapper.jsx
```diff
@@ -12,9 +12,9 @@ function SearchWrapper({ uniqueFacets, uniqueFields, entityTypeKeyword }) {
   const facets = [
     ...uniqueFacets,
     ...getDonorMetadataFilters(entityTypeKeyword === 'Donor'),
-    { field: 'group_name', label: 'Group' },
-    { field: 'created_by_user_displayname', label: 'Creator' },
-  ];
+    getAffiliationFacet({ field: 'group_name', label: 'Group' }),
+    getAffiliationFacet({ field: 'created_by_user_displayname', label: 'Creator' }),
+  ].reduce((acc, curr) => ({ ...acc, ...curr }), {});
```
mccalluc, line 17: Might make this a helper function, so it can read something like:
```
  const facets = mergeObjects([
    ...
```

context/app/static/js/components/entity-search/SearchWrapper/utils.js
```diff
@@ -1,22 +1,32 @@
 import { RefinementSelectFacet } from '@searchkit/sdk';
 
+function getFacetWithGroup(facetGroup = 'Additional Facets') {
```
mccalluc, line 3:
```suggestion
function getFacetWithGroup(facetGroup) {
```

If nothing uses the default, don't bother with it?

Naming: I tend to reserve `get*` when the value is being pulled from somewhere else... if there's nothing external, may `build*` instead? ... Particularly since this doesn't return facets: It returns functions.

context/app/static/js/components/entity-search/facets/FacetGroupAccordion/FacetGroupAccordion.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+import { StyledExpandMoreIcon } from 'js/components/searchPage/filters/style';
+import { OuterAccordion, OuterAccordionSummary, OuterAccordionDetails } from './style';
+
+function FacetGroupAccordion({ label, isFirst, children }) {
+  return (
+    <OuterAccordion defaultExpanded={isFirst}>
```
mccalluc, line 9: I'd suggest calling it `isExpanded`... Farther down, I wasn't sure why being first mattered. Sometimes, more abstraction is good, but since it's really just doing one thing, I think it'd be more clear to carry that through.

mccalluc (commented):
Some comments, mostly about naming. Tweak if anything resonates with you, and it's good to merge. Thanks!

context/app/static/js/components/entity-search/Search/Search.jsx
john-conroy: The name `facets` here is required by searchkit so the name for facets in our provider will have to change.

context/app/static/js/components/entity-search/SearchWrapper/utils.js
john-conroy: Noted, I'll make the change.

john-conroy: The default will be useful later.

PR #2574, opened 2022-04-11
Two new filters on dev-search

### LINKED ISSUE #2563: Add analyte to dev search
- as a hierarchy above assay types
- Will by used to get feedback from NIH if this will be useful
- Doesn't matter that metadata doesn't propagate... just trying to get any sense if this will be useful

mccalluc:
- Follow-up to  https://github.com/hubmapconsortium/ingest-validation-tools/issues/1062
- Fix #2563

Screen Shot 2022-04-11 at 3 39 13 PM

There are some weird analyte values.... I think the original roadmap was to ask NIH if this would be useful. If it looks like it still could be useful, but we want to understand the data first, do you want to file an issue for that?

### DIFF
--- a/CHANGELOG-dev-search-assay-filters.md
+++ b/CHANGELOG-dev-search-assay-filters.md
@@ -1,0 +1,1 @@
+- Add two hierarchical filters to dev-search; May help us represent assays better.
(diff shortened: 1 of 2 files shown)


mccalluc:
@ngehlenborg : That's what this does: Just haven't expanded to the next level in the screen shot.

PR #2581, opened 2022-04-15
fewer warnings during tests

mccalluc:
This cuts the output from 1213 to 771 lines. I don't think it's really that important, but I think it's good to avoid warnings that we just get used to, though I'm not sure how best to get there, and this PR isn't consistent. Some options:
- Loosen prop-type checks.
- Fill in placeholders so prop-types pass.
- mock console.error so we don't see the warnings.

### DIFF
--- a/CHANGELOG-test-warnings.md
+++ b/CHANGELOG-test-warnings.md
@@ -1,0 +1,1 @@
+- Reduce the number of warnings during tests.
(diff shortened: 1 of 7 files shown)


mccalluc:
Having second thoughts about this... Are any of these changes actually useful? Adding more props just feels like clutter; The DOM nesting doesn't really matter... Maybe it'd be better to just to mock the console on on every test with a warning, but when we do notice a new warning, make a decision about it individually?

mccalluc:
Replaced by #2650

Issue #2596, opened 2022-04-27
Implement customizable search columns and facets for search ui revision


john-conroy:
- [ ] Add ability for users to configure fields with relevant metadata fields.
- [ ] Add data types select to configure search table for dataset pages.
- [ ] Enable users to filter configure search table with text input.
- [ ] Add ability for users to configure facets with relevant metadata fields.
- [ ] Add histogram facet components for numeric fields.

PR #2597, opened 2022-04-27
John conroy/sample donor search pages

john-conroy:
Towards #2596.


context/app/routes_main.py
```diff
@@ -58,9 +58,9 @@ def search():
     )
 
 
-@blueprint.route('/test-search')
-def test_search():
-    title = 'Test Search'
+@blueprint.route('/test-search/<type>')
+def test_search(type):
+    title = f'{type.capitalize()} Test Search'
```
mccalluc, line 63: In the existing search, the type is a query parameter, but they we jump through hoops so it doesn't actually show up for the user -- It looks like you've found a different way to do it? That sounds good!

mccalluc: ... but if someone tries `/test-search/not-a-type`, they'll get a soft 404 because the JS routing will fail? Soft 404s aren't great... but not the end of the world, either.

context/app/static/js/components/Routes/Routes.jsx
```diff
@@ -110,8 +112,23 @@ function Routes(props) {
...
+
+  if (urlPath.startsWith('/test-search/samples')) {
+    return (
+      <Route>
+        <SampleSearch />
+      </Route>
+    );
+  }
+
+  if (urlPath.startsWith('/test-search/datasets')) {
```
mccalluc, line 131: I could imagine moving this switch logic down a level... Doesn't really matter.

context/app/static/js/pages/entity-search/DonorSearch/DonorSearch.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import SearchWrapper from 'js/components/entity-search/SearchWrapper';
+import { getDonorMetadataFields } from 'js/components/entity-search/SearchWrapper/utils';
+
+function DonorSearch() {
+  const entityType = 'donor';
```
mccalluc, line 7: Maybe just inline it?

mccalluc (approved):
All looks good... Some comments as I was reading it, but it's fine as-is. Thanks!

context/app/static/js/pages/entity-search/DonorSearch/DonorSearch.jsx
john-conroy: It's beneficial as it's used in two places.

PR #2617, opened 2022-05-03
John conroy/configure search data types

john-conroy:
Enables users to populate the configure search table on the datasets search page by selecting categories in a form. The categories are 'General', 'Donor Metadata', 'Sample Metadata' and a category for each assay. 

Focusing on functionality. Style polishing to come.

Screen Shot 2022-05-05 at 10 20 54 AM


context/app/static/js/components/entity-search/sidebar/ConfigureSearch/hooks.js
```diff
@@ -5,40 +5,44 @@ import { useStore } from 'js/components/entity-search/SearchWrapper/store';
...
-  return { selectedFields, handleToggleField, setSelectedFields };
+  return { selectedItems, handleToggleItem, setSelectedItems };
 }
 
 function useConfigureSearch() {
   const [dialogIsOpen, setDialogIsOpen] = useState(false);
   const { setFields, fields } = useStore();
-  const { selectedFields, handleToggleField, setSelectedFields } = useSelectedFields(fields);
+  const {
+    selectedItems: selectedFields,
```
mccalluc, line 42: ... and this inconsistency feels awkward. There are good reasons for names not to match... semantics are clear from context in an inner scope so names can be abbreviated... function handles a general case and outer scope has particular specific semantics... and others... but I'm not sure what's driving this case.

context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
```diff
@@ -0,0 +1,96 @@
...
+import { invertKeyToArrayMap as createDataTypesToFieldsMap } from './utils';
+import { useMetadataFieldConfigs } from '../ConfigureSearchTable/hooks';
+
+const dataTypesToFieldsMap = createDataTypesToFieldsMap(fieldsToAssayMap);
+
+function selectedItemsReducer(state, { type, payload }) {
+  const tempState = state;
+  switch (type) {
+    case 'selectItem':
+      return { ...state, [payload]: undefined }; // using an object as a set
```
mccalluc, line 16: Not clear about this... but I'm not clear on the bigger picture either.

context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
```diff
@@ -0,0 +1,96 @@
...
+
+function useDatasetConfigureSearchTable() {
+  const { handleToggleDataType, dataTypesFields } = useSelectedDataTypes();
+  const { groups, handleToggleGroup, selectedGroupFieldConfigs } = useFieldGroups();
+
+  const availableFieldConfigs = { ...selectedGroupFieldConfigs, ...dataTypesFields };
+  return { dataTypesToFieldsMap, handleToggleDataType, groups, handleToggleGroup, availableFieldConfigs };
+}
+
+export { useDatasetConfigureSearchTable };
```
mccalluc, line 96: I'm not really following this... but this might be a case where it make more sense in context: if I was coming in from the top, I might understand better how each of these is being used.

context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/utils.spec.js
```diff
@@ -0,0 +1,14 @@
+import { invertKeyToArrayMap } from './utils';
+
+test('should invert map', () => {
+  const initialMap = {
+    ablation_frequency_value: ['3D Imaging Mass Cytometry', 'Imaging Mass Cytometry'],
```
mccalluc, line 5:
```suggestion
    empty_value_ok: [],
    ablation_frequency_value: ['3D Imaging Mass Cytometry', 'Imaging Mass Cytometry'],
```
I don't know that this is actually a scenario you need to worry about, but it's an edge case.

mccalluc: Where the logic has nothing to do with the particular names, slight preference to just use `'A', 'B', 'C'`: Less to be distracted by,

context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
mccalluc: Is the logic of the hooks complicated enough that they should have some unit tests of their own?

context/app/static/js/components/entity-search/sidebar/ConfigureSearch/hooks.js
```diff
@@ -5,40 +5,44 @@ import { useStore } from 'js/components/entity-search/SearchWrapper/store';
 function selectedFieldReducer(state, { type, payload }) {
   const tempState = state;
   switch (type) {
-    case 'selectField':
+    case 'selectItem':
```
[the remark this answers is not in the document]
john-conroy: I renamed the actions to make this more generic for reuse. I use this reducer in two other places where 'field' doesn't always apply.

context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
john-conroy: A set would make sense here, but updating a set doesn't trigger react updates. There are ways around it by redefining a new set each time, but it adds more code so using an object  here where the values don't matter made sense.

context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
john-conroy: One react practice that is encourage is encapsulating all of your hooks into a single hook which is something I did less early on, but I am working towards now. I have made some unit tests, and there is a room for more, but where we are under a time pressure, I might revisit once the feature is more whole.

PR #2628, opened 2022-05-05
pull in organ icons; more boolean fields

### LINKED ISSUE #2624: Rough out new organ page
Make a quick draft PR to confirm that more information is available: John will take this over and make it conform to UI designs.

mccalluc:
- fix #2624

Filing as draft: When there's a design (#2350), @john-conroy can implement it here.

Surprised that there are a number of organs without icons... Thinking the best thing to do might be a design that is robust against missing data... but also reach out to IU and see if there are any more available now.


mccalluc:
Actually, there are more organ icons... I'll fill them in.

john-conroy (commented):
This is helpful, thanks! This will stay as a draft until there is a design and I can branch off of it, right?

PR #2631, opened 2022-05-06
John conroy/fix source sample access

john-conroy:
As we discussed, `source_sample` inside dataset es documents are arrays and must be handled differently.


mccalluc (approved):
Looks good. I'm trying to remember: Will you also be accessing origin metadata? or do you only need to worry about source metadata?

john-conroy:
`origin_sample` is not an array and can be handled by lodash/get.

Issue #2612, opened 2022-04-28
implement matomo


mccalluc:
- [ ] start 21 day trial
- [ ] add matomo to our codebase
- [ ] get a real account
- [ ] #2651

PR #2637, opened 2022-05-09
Mccalluc/matomo

mccalluc:
Page visits are showing up in the [dashboard](https://hubmapconsortium.matomo.cloud/index.php).
- Towards #2612 
- Suspect I'm not using the hook the best way.
- Do we want to distinguish visits to different instances? Right now, they all go into the same pool.
- This covers all references to `ReactGA.pageview`, but nothing else.

### DIFF
--- a/context/app/static/js/components/Providers.jsx
+++ b/context/app/static/js/components/Providers.jsx
@@ -1,11 +1,12 @@
 import React from 'react';
 import { ThemeProvider } from 'styled-components';
-import GlobalStyles from 'js/components/globalStyles';
+import { MatomoProvider, createInstance } from '@datapunt/matomo-tracker-react';
 
 import PropTypes from 'prop-types';
 import { MuiThemeProvider, StylesProvider, createGenerateClassName } from '@material-ui/core/styles';
 import CssBaseline from '@material-ui/core/CssBaseline';
 
+import GlobalStyles from 'js/components/globalStyles';
 import theme from '../theme';
 import GlobalFonts from '../fonts';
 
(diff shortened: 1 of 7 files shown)


john-conroy (changes requested):
This looks reasonable and I don't see anything wrong with how you've used the hook, but we do want to separate prod/dev instances as we have done with google analytics before merging.

context/app/static/js/components/Providers.jsx
```diff
@@ -14,24 +15,58 @@ const generateClassName = createGenerateClassName({
...
+  // },
+  // linkTracking: false, // optional, default value: true
+  // configurations: { // optional, default value: {}
+  //   // any valid matomo configuration, all below are optional
+  //   disableCookies: true,
+  //   setSecureCookie: true,
+  //   setRequestMethod: 'POST'
+  // }
+});
+
```
john-conroy, line 49: I'd pull `createInstance` and the provider into a separate component and the `siteId` logic into a util with a case switch to avoid the global and `let`.

john-conroy (changes requested):

context/app/static/js/components/Providers.jsx
mccalluc: Good call; Should look better now.

context/app/static/js/components/MatomoProviderWrapper.jsx
```diff
@@ -0,0 +1,47 @@
...
+      return 1;
+    case 'localhost:5001':
+      return 2;
+    default:
+      return 3;
+  }
+}
+
+function MatomoProviderWrapper(props) {
+  const { children } = props;
```
john-conroy, line 21:
```suggestion
function MatomoProviderWrapper({children}) {
```

john-conroy (approved):

PR #2641, opened 2022-05-10
John conroy/remove date metadata fields

### LINKED ISSUE #2630: Map elasticsearch date fields consistently
Currently date fields can only be sorted if `.keyword` is appended to the field name, otherwise it errors. When you apply `.keyword ` I believe the fields are sorted as then sorted as strings which leads to problems since `metadata.metadata.execution_date` is formatted `MM/DD/YYYY` and I can't imagine other fields are consistent. I think we will need to map dates consistently.

john-conroy:
Removes date fields from configure search table while we work to resolve #2630.

### DIFF
--- a/context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
+++ b/context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
@@ -61,6 +61,7 @@ function useFieldGroups() {
 
 function buildFieldConfigs(fieldNames) {
   return fieldNames.reduce((acc, fieldName) => {
+    // Version must be removed before creating field configs because it does not have a type defined in ingest-validation-tools docs.
     const excludedFieldNames = ['version'];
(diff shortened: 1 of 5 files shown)


context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
```diff
@@ -61,6 +61,7 @@ function useFieldGroups() {
 
 function buildFieldConfigs(fieldNames) {
   return fieldNames.reduce((acc, fieldName) => {
+    // Version must be removed before creating field configs because it does not have a type defined in ingest-validation-tools docs.
```
mccalluc, line 64: It's an integer, and I could provide it in the field list... but my sense is that it's really only useful when supplying data: Downstream, it shouldn't matter what schema version was used. Anyway: push for this if that doesn't seem right to you.

mccalluc (commented):
Looks good.

PR #2646, opened 2022-05-13
Matomo everywhere

mccalluc:
Introduce a wrapper function that will call both GA and matomo. (The special react package was awkward, because I kept on violating rules-of-hooks when I tried to use it.)


mccalluc:
Thanks for writing the tests that are failing: they're catching some bugs!

Tests seem to actually hit the analytics... In most places it's not mocked. Would it make sense to add a guard clause on all of the tracker functions?
```
if (process.env.NODE_ENV === 'test') return;
```

It seems that matomo is fussier about their being an analytics category? Should we provide a default? Or make sure it's provided in every case?

context/app/static/js/helpers/trackers.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+const tracker = new MatomoTracker({
+  urlBase: 'https://hubmapconsortium.matomo.cloud/',
+  // eslint-disable-next-line no-restricted-globals
+  siteId: getSiteId(location),
+  // userId: 'UID76903202', // optional, default value: `undefined`.
+  // trackerUrl: 'https://LINK.TO.DOMAIN/tracking.php', // optional, default value: `${urlBase}matomo.php`
+  // srcUrl: 'https://LINK.TO.DOMAIN/tracking.js', // optional, default value: `${urlBase}matomo.js`
+  // disabled: false, // optional, false by default. Makes all tracking calls no-ops if set to true.
```
mccalluc, line 27:
```suggestion
  disabled: process.env.NODE_ENV === 'test', // Makes all tracking calls no-ops if set to true.
```

context/app/static/js/helpers/trackers.js
```diff
@@ -0,0 +1,65 @@
...
+  // linkTracking: false, // optional, default value: true
+  // configurations: { // optional, default value: {}
+  //   // any valid matomo configuration, all below are optional
+  //   disableCookies: true,
+  //   setSecureCookie: true,
+  //   setRequestMethod: 'POST'
+  // }
+});
+
+ReactGA.initialize('UA-133341631-3', { testMode: process.env.NODE_ENV === 'test' });
```
john-conroy, line 41: I think this still needs to be initialized at the `App` level?

john-conroy (changes requested):
I believe the trackers for both Matomo and ReactGa need to be initialized inside `App`.

context/app/static/js/helpers/trackers.js
mccalluc: @john-conroy :
> I think this still needs to be initialized at the App level?

I could be confused, but I don't think so? I can double check, but I think events were getting logged... but even if it does work, that doesn't mean it's the best way to do things...

The two are a little different: We initialize `ReactGA` directly, while we need to create a new instance of `MatomoTracker`. But in either case, their methods are only called inside this module, so I don't think there's any need to expose these internals anywhere else... In fact, the opposite: one point of this PR is to hide these details from the rest of the codebase.

Talk after standup if this doesn't seem right?

mccalluc: We're not absolutely sure about this, but got John on-board to try it out: He points out that usual guidance is to initialize early on... GA may do some logging beyond what we explicitly call for? But the module private consts will be initiated when the module is loaded, which seems early enough.

mccalluc:
@john-conroy , if this approach seems ok to you now, can you approve? Or if there are other things you skipped on the first pass, please do raise them now!

john-conroy (approved):

PR #2648, opened 2022-05-13
Use specialized matomo events

### LINKED ISSUE #2647: Use more precise matomo events
Matomo has some more precise events that GA: In PR #2646 I've got them commented out right now:
```
// function trackLink(href) {
//   tracker.trackLink({
//     href: 'https://link-to-other-website.org',
//   })
// }

// function trackSiteSearch(keyword) {
//   tracker.trackLink({
//     keyword: 'test',
//   })
// }
```

mccalluc:
Fix #2647


context/app/static/js/shared-styles/Links/iconLinks/EmailIconLink/EmailIconLink.jsx
```diff
@@ -1,15 +1,10 @@
...
+import { trackLink } from 'js/helpers/trackers';
 
 function sendEmailEvent(event) {
-  trackEvent({
-    category: 'Email Link',
-    action: 'Clicked',
-    label: event.target.href,
-    nonInteraction: false,
-  });
+  trackLink(event.target.href);
```
john-conroy, line 7: Seems like we're losing value here and no longer categorize our link events?

context/app/static/js/components/searchPage/SearchWrapper.jsx
```diff
@@ -52,11 +52,7 @@ function SearchWrapper(props) {
     const removalFn = searchkit.addResultsListener((results) => {
       const newQueryString = searchkit.state?.q;
       if (![queryString.current, undefined].includes(newQueryString)) {
-        trackEvent({
-          category: analyticsCategory,
-          action: 'Free Text Search',
-          label: newQueryString,
-        });
+        trackSiteSearch(newQueryString);
```
john-conroy, line 55: Not entirely sure what `trackSiteSearch` handles for us, but how can we differentiate 'site searches' as we have in the past with Google Analytics' category and action?

john-conroy (changes requested):

context/app/static/js/shared-styles/Links/iconLinks/EmailIconLink/EmailIconLink.jsx
mccalluc: The Matomo API sees to have a few more functions, for different types of events, but those functions are also less flexible. Looking at the [docs](https://developer.matomo.org/api-reference/tracking-javascript):
> `trackLink( url, linkType )` - Log a click from your own code. url is the full URL which is to be tracked as a click. linkType can either be 'link' for an outlink or 'download' for a download.

In this particular case, maybe the matomo way would be to filter based on the URL?

context/app/static/js/components/searchPage/SearchWrapper.jsx
mccalluc: `trackSiteSearch` has an optional `category` param, which I'm now providing.

context/app/static/js/shared-styles/Links/iconLinks/EmailIconLink/EmailIconLink.jsx
john-conroy: Should we provide a `linkType` arg?

john-conroy (approved):
I think we lose specificity and our tracking won't be 1:1 with GA, but that seems unavoidable?

context/app/static/js/shared-styles/Links/iconLinks/EmailIconLink/EmailIconLink.jsx
mccalluc: I'd want to make the `linkType` "email", but it says that's not allowed? But what happens if we do? I'd like to merge this as-is, but I'll file a follow-up issue.

PR #2656, opened 2022-05-17
Gene Alias

### LINKED ISSUE #2393: In Vitessce, use Hugo IDs rather than Ensembl IDs
The Ensembl IDs just aren't human-readable. Matt says Hugo IDs are in there... but Sean clarifies that it isn't filled out for every dataset. This depends on Vitessce functionality to alias fields from AnnData that aren't indexed. (Sorry if I'm not using the right terms to describe this!)

Vitessce blocker, and more details:
- https://github.com/vitessce/vitessce/issues/1147

ilan-gold:
Fixes #2393

### DIFF
--- a/CHANGELOG-vitessce-upgrade.md
+++ b/CHANGELOG-vitessce-upgrade.md
@@ -1,0 +1,1 @@
+- Upgrade `vitessce` (and dependents) to allow for using an alias for genes.
--- a/context/package-lock.json
+++ b/context/package-lock.json
@@ -2431,9 +2431,9 @@
       }
     },
     "@deck.gl/aggregation-layers": {
-      "version": "8.6.8",
-      "resolved": "https://registry.npmjs.org/@deck.gl/aggregation-layers/-/aggregation-layers-8.6.8.tgz",
-      "integrity": "sha512-1mkkXLKleDlXcoURVAPFj4luXyMzB51TLmQirin6irJbdE1dPsNE43+9IiLzF6vbhzsl0NUEssxSfzklFmEqQQ==",
+      "version": "8.6.9",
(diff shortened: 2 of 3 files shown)


ilan-gold:
I can't run `pip-compile --allow-unsafe --generate-hashes --output-file=context/requirements.txt context/requirements.in` until the `portal-visualization` package is released: https://github.com/hubmapconsortium/portal-visualization/pull/48

mccalluc (approved):
Great! I'm going to move this into "draft" until the dependencies are there, just so we don't accidentally merge a PR that's approved and passing tests. (I wondered about having a check on the consistency of `requirements.txt`... but it's pretty rare that they are out of sync, and it's a bit of work to add to every CI run.)

mccalluc (changes requested):
So I've checked out this branch, and updated dependencies, but I still see `ENSG...` IDs in Vitessce. Is more work needed in vitessce-python to actually make use of the new functionality? Or am I confused? (If it helps to unblock things, this could be merged, but my preference would be to see it working in this branch first.)

ilan-gold:
@mccalluc I would make sure your dependencies are up to date.  Actually, the first one revealed a bug but the second had the Hugo ID's so I think there might be something about your environment....

ilan-gold:
I don't think the first one has a Hugo ID column to  be used.  The second does.

mccalluc:
Context: Ilan's PR to handle on bug is:
- https://github.com/hubmapconsortium/portal-visualization/pull/49

mccalluc:
Update from Ilan on Slack:
> I am not at home at the moment, so I can’t fix the portal-ui PR but I think all that needs to change is a version bump for portal-visualization now.

mccalluc (approved):
And confirmed that http://localhost:5000/browse/dataset/b5ff6c347488c45deafeb6316dc7925b looks good. Thanks!

PR #2661, opened 2022-05-20
John conroy/numeric facets hook

john-conroy:


context/app/static/js/components/entity-search/Search/utils.js
```diff
@@ -15,4 +15,12 @@ function buildSortPairs(fields) {
     .flat();
 }
 
-export { buildSortPairs };
+function getNumericFacetProps(fieldName, numericFacetsProps) {
+  if (!numericFacetsProps?.[fieldName]) {
+    return {};
+  }
+
+  return { range: numericFacetsProps[fieldName] };
```
mccalluc, line 23: Maybe a naming tweak for consistency? Either to the function name, or to this parameter: Right now `numericFacetsProps` might be either the _entire_ returned object, or just the value of `range`.

context/app/static/js/components/entity-search/SearchWrapper/SearchWrapper.jsx
```diff
@@ -35,14 +36,25 @@ function SearchWrapper({ uniqueFacets, uniqueFields, entityType }) {
...
-  const config = { initialFacets, initialFields, facets: initialFacets, fields: initialFields, filters, entityType };
-
+  const numericFacetsProps = useNumericFacetsProps(entityType);
   return (
-    <SearchConfigProvider initialConfig={config}>
-      <SearchkitProvider client={skClient}>
-        <Search />
-      </SearchkitProvider>
-    </SearchConfigProvider>
+    Object.keys(numericFacetsProps).length > 0 && (
```
mccalluc, line 41: Slight preference to pull this out as a guard clause: Something like:
```
if (!Object.keys(numericFacetsProps).length) return;
return (<SearchConfigProvider 
... 
/>);

File: context/app/static/js/components/entity-search/SearchWrapper/store.js [hunk 3]
```diff
@@ -5,16 +5,18 @@ import createContext from 'zustand/context';
...
     initialFields,
     initialFacets,
     fields,
     facets,
     filters,
     entityType,
+    numericFacetsProps,
     setFields: (selectedFields) => set({ fields: selectedFields }),
     setFacets: (selectedFacets) => set({ facets: selectedFacets }),
+    setNumericFacetsProps: (numericFacets) => set({ numericFacetsProps: numericFacets }),
```
mccalluc, line 19: As the list gets longer, would it make sense to pair the setters and the things they set? Or if this is more conventional, ignore.

mccalluc (commented):
Minor comments: The only one that I think matters is the naming, but do what you think best.

context/app/static/js/components/entity-search/Search/utils.js
john-conroy: Renamed the function, thanks!

context/app/static/js/components/entity-search/SearchWrapper/SearchWrapper.jsx
john-conroy: Thanks!

PR #2662, opened 2022-05-20
John conroy/search pagination

john-conroy:
The pagination components do not have the same appearances as the pagination on our existing search ui, but before investing more time I want to confirm the design with Tiffany. I'd like to merge with the default styling and revisit it once Tiffany is back from the conference.


context/app/static/js/components/entity-search/Search/Search.jsx
```diff
@@ -50,7 +52,14 @@ function Search({ numericFacetsProps }) {
   return (
     <SearchLayout>
       <Sidebar results={results} />
-      <ResultsLayout>{results?.hits && <ResultsTable hits={results.hits} />}</ResultsLayout>
+      <ResultsLayout>
+        {results?.hits && (
```
mccalluc, line 56: Down the road, this becomes a ternary with a "no results" message?

mccalluc (approved):

PR #2663, opened 2022-05-23
John conroy/fix configure facet bug

john-conroy:
State would update before results causing an error.

### DIFF
--- a/context/app/static/js/components/entity-search/facets/Facets/hooks.js
+++ b/context/app/static/js/components/entity-search/facets/Facets/hooks.js
@@ -4,7 +4,12 @@ function useGroupedFacets(resultsFacets) {
   const { facets } = useStore();
 
   return resultsFacets.reduce((acc, facet) => {
+    if (!(facet.identifier in facets)) {
+      return acc;
+    }
+
     const { facetGroup } = facets[facet.identifier];
+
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/entity-search/facets/Facets/hooks.js
```diff
@@ -4,7 +4,12 @@ function useGroupedFacets(resultsFacets) {
   const { facets } = useStore();
 
   return resultsFacets.reduce((acc, facet) => {
+    if (!(facet.identifier in facets)) {
+      return acc;
+    }
```
mccalluc, line 9: Is it worth a comment explaining why/when the facet identifier would be missing from the list?

mccalluc (commented):
Maybe add a comment, particularly since it seems like an edge case that was missed the first time through.

context/app/static/js/components/entity-search/facets/Facets/hooks.js
john-conroy: This is only an issue because we don't have a loading state. The facets configuration updates before the api hits. We could reduce over the configured facets instead of the results facets, but lookup to get the correct facet in results facets is more convoluted.

PR #2668, opened 2022-05-25
update matomo host

mccalluc:
Before merging this, we should confirm that it's logging events correctly. To do that, we need to add localhost to the list: I don't have privs. When setting up sites in the dashboard, Matomo only cares about the integer ID: The mapping from hostname to integer ID happens in our code.

- @ngehlenborg, please add websites `2` (localhost) and `3` (other) at https://hubmap.matomo.cloud/index.php?module=SitesManager.
- I think you'll also need to add John, Tiffany, and myself as users on each site. (Permissions go site by site.)
- @john-conroy, when that's done, checkout this branch, load http://localhost:5001 in one window, and look at https://hubmap.matomo.cloud/index.php in the other: You should see events being logged.

### DIFF
--- a/CHANGELOG-update-matomo-id.md
+++ b/CHANGELOG-update-matomo-id.md
@@ -1,0 +1,1 @@
+- Update matomo init to reflect the new, paid, account.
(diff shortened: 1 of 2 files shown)


mccalluc:
Nils has everything set up. Just getting John's thumbs up. (If I'm gone before that happens, please just merge without me!)

PR #2671, opened 2022-05-26
John conroy/search bar

john-conroy:
More styling to come down the road, but the functionality matches the existing search page.

Screen Shot 2022-05-26 at 1 01 18 PM


context/app/static/js/components/entity-search/SearchBar/SearchBar.jsx
```diff
@@ -0,0 +1,44 @@
...
+      onSubmit={(event) => {
+        setQuery(query);
+        api.setQuery(query);
+        api.search();
+        event.preventDefault();
+      }}
+      role="search"
+    >
+      <TextField
+        id={`${entityType}-search`}
```
mccalluc, line 25: I don't think this is referenced by the code in this PR; Referenced from somewhere else?

context/app/static/js/components/entity-search/Search/Search.jsx
```diff
@@ -13,9 +13,27 @@ import Pagination from 'js/components/entity-search/results/Pagination';
...
 import { buildSortPairs, getRangeProps } from './utils';
 
+const query = new CustomQuery({
+  queryFn: (q) => {
+    return {
+      bool: {
+        must: [
+          {
+            simple_query_string: {
+              fields: ['all_text', 'description'],
```
mccalluc, line 27: Particularly if we'll be revisiting the ES configuration, this might be something to keep an eye on. If memory serves, `all_text` is a field we create, but may be possible for ES to combine fields for search without actually duplicating them in the stored document.

mccalluc (approved):

context/app/static/js/components/entity-search/SearchBar/SearchBar.jsx
john-conroy: `TextField` is a composition of multiple elements which all rely on the id.

PR #2678, opened 2022-06-07
John conroy/metadata menu

john-conroy:
The icons are now horizontally aligned. Otherwise, it's the same as the existing search page.

Screen Shot 2022-06-07 at 1 22 11 PM

### DIFF
--- a/CHANGELOG-metadata-menu.md
+++ b/CHANGELOG-metadata-menu.md
@@ -1,0 +1,1 @@
+- Add metadata menu to search revision ui.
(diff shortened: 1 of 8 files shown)


mccalluc (approved):
I think a lot of this is porting over existing code? Looks good, but let me know if there's something particular I should look at.

john-conroy:
> I think a lot of this is porting over existing code? Looks good, but let me know if there's something particular I should look at.

Yes, I should have mentioned that in the main comment.

PR #2696, opened 2022-06-21
John conroy/field descriptions

john-conroy:
Adds description tooltips to configure search table. Creates a reusable `IconTooltipCell` component. I replaced the existing icon tooltip components in `MetadataTable` and `ContributorsTable` and I think those are the only applicable cases, but maybe you remember more of them?

Screen Shot 2022-06-21 at 2 04 59 PM


mccalluc (approved):
Seems like a good simplification. Thanks.

PR #2700, opened 2022-06-22
John conroy/configure search scroll

john-conroy:
- Modify configure search table in search revision ui so the data types select form and table scroll separately.

configure-search-scroll


mccalluc (approved):
Thanks!

PR #2709, opened 2022-06-24
John conroy/scroll shadows

john-conroy:
If this looks good, I'll open an issue to add scroll shadows to our existing tables and lists. 

@tsliaw Do we want horizontal scroll shadows for the main search table when columns are added?

scroll-shadows

### DIFF
--- a/context/app/static/js/shared-styles/scrollShadows.js
+++ b/context/app/static/js/shared-styles/scrollShadows.js
@@ -1,4 +1,5 @@
 import { css } from 'styled-components';
+import { isMobileSafari } from 'react-device-detect';
 
 // See https://css-tricks.com/scroll-shadows-with-javascript/ for reference.
 // The scroll shadows will not work on iOS Safari.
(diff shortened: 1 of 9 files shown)


context/app/static/js/shared-styles/scrollShadows.js
```diff
@@ -0,0 +1,48 @@
+import { css } from 'styled-components';
+import { isMobileSafari } from 'react-device-detect';
+
+// See https://css-tricks.com/scroll-shadows-with-javascript/ for reference.
+// The scroll shadows will not work on iOS Safari.
```
mccalluc, line 5: Maybe add a comment explaining why supporting Safari wasn't workable? (Imagine someone in the future is tasked with adding support: What would you tell them?)

context/app/static/js/shared-styles/scrollShadows.js
```diff
@@ -0,0 +1,48 @@
...
+    background:
+      /* Shadow Cover TOP */ linear-gradient(white 30%, rgba(255, 255, 255, 0)) center
+        ${tableHeadHeightPx},
+      /* Shadow Cover BOTTOM */ linear-gradient(rgba(255, 255, 255, 0), white 70%) center bottom,
+      /* Shadow TOP */ radial-gradient(farthest-side at 50% 0, rgba(0, 0, 0, 0.2), rgba(0, 0, 0, 0)) center top
+        ${tableHeadHeightPx},
+      /* Shadow BOTTOM */ radial-gradient(farthest-side at 50% 100%, rgba(0, 0, 0, 0.2), rgba(0, 0, 0, 0)) center bottom;
+
+    ${sharedStyles};
+  `;
```
mccalluc, line 45: This seems to match what's above, except for `tableHeadHeightPx` instead of `'top'`: Maybe factor it out into a function that takes that as a parameter?

mccalluc (commented):
Maybe factor the repeated CSS into a helper function? But looks good.

tsliaw (approved):

context/app/static/js/shared-styles/scrollShadows.js
john-conroy: As we discussed in standup, I'll defer this until I tackle the horizontal scroll shadows.

john-conroy:
Filed #2724.

PR #2710, opened 2022-06-27
Mccalluc/new package json for cypress

mccalluc:


john-conroy:
Was this generated using node 14?

john-conroy (approved):

mccalluc:
> Was this generated using node 14?

No... I think this was node 16... I should uninstall the versions we're not using to prevent confusion. Sorry / Closing.

PR #2711, opened 2022-06-27
Mccalluc/show workspace api in service list

mccalluc:
On https://portal.hubmapconsortium.org/services adds:
Screen Shot 2022-06-27 at 1 47 01 PM

- @jpuerto-psc : Mostly want to confirm that https://github.com/hubmapconsortium/user_workspaces_server/issues/28 and https://github.com/hubmapconsortium/user_workspaces_server/issues/34 are on your radar. I don't think this PR needs to wait on them... but when that information is available, this is where it will be used.
- @john-conroy : normal review.


jpuerto-psc (approved):
@mccalluc thank, you this is on my radar.

PR #2720, opened 2022-06-29
Mccalluc/UI error if no ws token

### LINKED ISSUE #2718: Friendlier error if workspace auth fails
Workspace auth seems to be flakey: 
- https://github.com/hubmapconsortium/user_workspaces_server/issues/58

We don't want to bother anyone with a big warning at login time if the process has failed, but we should have some messaging on the workspaces page if nothing is going to work.

mccalluc:
- Fix #2718
- @jpuerto-psc : Is there any particular action you would like us to take if we hit this again? We could say "Slack Juan Puerto if you see this message"... but that might not give you any more useful information.
- @john-conroy : I'm ok with two component functions in this case... Nested ternaries seem less readable, and I'm not sure I want to push these details down to a lower level in a separate file... but don't care strongly, if you do.

### DIFF
--- a/context/app/static/js/pages/Workspaces/Workspaces.jsx
+++ b/context/app/static/js/pages/Workspaces/Workspaces.jsx
@@ -1,40 +1,10 @@
 import React, { useContext } from 'react';
 
-import OutboundLink from 'js/shared-styles/Links/OutboundLink';
 import { AppContext } from 'js/components/Providers';
 import SectionHeader from 'js/shared-styles/sections/SectionHeader';
 import Description from 'js/shared-styles/sections/Description';
 import { LightBlueLink } from 'js/shared-styles/Links';
-import WorkspacesList from 'js/components/workspaces/WorkspacesList';
-
-function AuthenticatedWorkspacesDescription() {
-  const { workspacesToken } = useContext(AppContext);
-
-  if (!workspacesToken) {
-    return (
-      <Description padding="20px">
-        The workspaces token was not set when you logged in.
-        <OutboundLink href="https://github.com/hubmapconsortium/user_workspaces_server/issues/58">
-          Github issue
-        </OutboundLink>
-      </Description>
-    );
-  }
-
-  return (
-    <>
-      <Description padding="20px">
-        {/*
-          TODO: Add links below.
-          TODO: Not all of these entry points will be functional on first release. Update text accordingly.
-        */}
-        Workspaces are provided through Jupyter notebooks. Navigate to a dataset, collection, dataset search or My Lists
-        page to begin a new workspace.
-      </Description>
-      <WorkspacesList />
-    </>
-  );
-}
+import AuthenticatedWorkspacesDescription from './AuthenticatedWorkspacesDescription';
 
 function Workspaces() {
   const { isAuthenticated } = useContext(AppContext);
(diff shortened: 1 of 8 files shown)


context/app/static/js/pages/Workspaces/Workspaces.jsx
```diff
@@ -6,6 +6,33 @@ import Description from 'js/shared-styles/sections/Description';
...
 import WorkspacesList from 'js/components/workspaces/WorkspacesList';
 
+function AuthenticatedWorkspaces() {
+  const { workspacesToken } = useContext(AppContext);
+
+  if (!workspacesToken) {
+    return (
+      <Description padding="20px">
+        The workspaces token was not set when you logged in.
+        <a href="https://github.com/hubmapconsortium/user_workspaces_server/issues/58">Github issue</a>
```
john-conroy, line 16: Use `OutboundLink`.

context/app/static/js/pages/Workspaces/Workspaces.jsx
```diff
@@ -6,6 +6,33 @@ import Description from 'js/shared-styles/sections/Description';
 import { LightBlueLink } from 'js/shared-styles/Links';
 import WorkspacesList from 'js/components/workspaces/WorkspacesList';
 
+function AuthenticatedWorkspaces() {
```
john-conroy, line 9:
```suggestion
function AuthenticatedWorkspacesDescription() {
```
`AuthenticatedWorkspaces` gives a sense that the component is more than a simple description.

context/app/static/js/pages/Workspaces/Workspaces.jsx
```diff
@@ -6,6 +6,33 @@ import Description from 'js/shared-styles/sections/Description';
 import { LightBlueLink } from 'js/shared-styles/Links';
 import WorkspacesList from 'js/components/workspaces/WorkspacesList';
```
john-conroy, line 8: I still prefer components to be in separate files if only for consistency now.

john-conroy (changes requested):
I don't share the same disdain for ternaries, but I don't feel strongly enough to push for a change. By nested do you mean inside the `jsx`? Normally I'd think of a nested ternary as a ternary inside of another ternary. I'd like to avoid repeating the `Description`, and if I were to do this I'd likely just use a ternary for the `Description` children.

Ultimately fine with the guard clause, but I think other comments should be addressed.

context/app/static/js/pages/Workspaces/Workspaces.jsx
mccalluc: I changed the name, but then remembered that it does contain `<WorkspacesList />`... So might change back.

mccalluc:
> I'd like to avoid repeating the Description, and if I were to do this I'd likely just use a ternary for the Description children.

I looked at this again... It's a little awkward because the error case doesn't include `<WorkspacesList />` ... so I could have a ternary inside the `<Description>`... but then I'd need a separate check outside, too, which feels awkward.

Right now, I'm going to wait for feedback from @jpuerto-psc , and then ping you again.

jpuerto-psc:
@mccalluc when issues crop up - send a message in the workspaces channel and ping me for now. Once other team members have been trained up we can approach error handling/diagnostics differently. Thanks!

mccalluc:
@john-conroy -- Tidied up; Re-requesting review.

john-conroy (approved):
I thought you decided to revert the rename I mistakenly suggested for `AuthenticatedWorkspaces`. Looks good.

PR #2721, opened 2022-06-29
Fix plurals in workspace section header.

mccalluc:
I think "0 Workspaces" (with the "s") sounds right?

Right now what I see in the header is
> 0 Workspace0

### DIFF
--- a/context/app/static/js/components/workspaces/WorkspacesList.jsx
+++ b/context/app/static/js/components/workspaces/WorkspacesList.jsx
@@ -50,7 +50,7 @@ function WorkspacesList() {
       <SpacedSectionButtonRow
         leftText={
           <Typography variant="subtitle1">
-            {workspacesList.length} Workspace{workspacesList.length && 's'}
+            {workspacesList.length} Workspace{workspacesList.length !== 1 && 's'}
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/workspaces/WorkspacesList.jsx
```diff
@@ -50,7 +50,7 @@ function WorkspacesList() {
       <SpacedSectionButtonRow
         leftText={
           <Typography variant="subtitle1">
-            {workspacesList.length} Workspace{workspacesList.length && 's'}
+            {workspacesList.length} Workspace{workspacesList.length !== 1 && 's'}
```
john-conroy, line 53: We still want it to read `Workspaces` if it's 0, right?

john-conroy (approved):
Thanks! Sorry, I did think of this possibility when making the suggested change, but I guess it escaped me.

context/app/static/js/components/workspaces/WorkspacesList.jsx
mccalluc: Yeah, that seems right to me? But I don't care strongly.

PR #2729, opened 2022-07-06
John conroy/new search tile view

### LINKED ISSUE #2447: Implement results tile view.

john-conroy:
Fix #2447. There is no URL param for `view` yet, but all of the URL/state synchronization will be updated in a subsequent PR.

There are still a few fields incorrectly typed... I'll reach out to the ES team.

```
// These fields are incorrectly typed as text in ES and cannot be treated as numbers.
const fieldsWithIncorrectEsType = new Set([
  'cold_ischemia_time_value',
  'specimen_tumor_distance_value',
  'warm_ischemia_time_value',
]);
```

Screen Shot 2022-08-31 at 1 40 08 PM

### DIFF
--- a/context/app/static/js/components/entity-search/SearchWrapper/SearchWrapper.jsx
+++ b/context/app/static/js/components/entity-search/SearchWrapper/SearchWrapper.jsx
@@ -21,7 +21,7 @@ function SearchWrapper({ uniqueFacets, uniqueFields, entityType }) {
     ...donorFacets,
     ...createAffiliationFacet({ fieldName: 'group_name', label: 'Group', type: 'string' }),
     ...createAffiliationFacet({ fieldName: 'created_by_user_displayname', label: 'Creator', type: 'string' }),
-  }
+  };
 
   const initialFields = {
     ...createField({ fieldName: 'hubmap_id', label: 'HuBMAP ID', type: 'string' }),


context/app/static/js/components/entity-search/SearchWrapper/SearchWrapper.jsx
```diff
@@ -4,28 +4,33 @@ import Search from 'js/components/entity-search/Search';
...
 import { useNumericFacetsProps } from './hooks';
 
 function SearchWrapper({ uniqueFacets, uniqueFields, entityType }) {
-  const initialFacets = mergeObjects([
-    ...uniqueFacets,
-    ...getDonorMetadataFields(entityType),
+  const { tableFields: donorFacets } = buildDonorFields(entityType);
+
+  const initialFacets = Object.assign(
+    uniqueFacets,
```
mccalluc, line 20: Is modifying `uniqueFacets` a desired side effect? If so, do we need `initialFacets` as well? If not, would spread be more clear?

context/app/static/js/components/entity-search/SearchWrapper/SearchWrapper.jsx
```diff
@@ -4,28 +4,33 @@ import Search from 'js/components/entity-search/Search';
...
+  buildTileFields,
 } from './utils';
 import SearchConfigProvider from './provider';
 import { useNumericFacetsProps } from './hooks';
 
 function SearchWrapper({ uniqueFacets, uniqueFields, entityType }) {
-  const initialFacets = mergeObjects([
-    ...uniqueFacets,
-    ...getDonorMetadataFields(entityType),
+  const { tableFields: donorFacets } = buildDonorFields(entityType);
```
mccalluc, line 17: I'm assuming `buildDonorFields` is also used somewhere else, and we just need to get part of it here?

context/app/static/js/components/entity-search/SearchWrapper/utils.js
```diff
@@ -96,31 +96,72 @@ const withFacetGroup = (facetGroup) => (o) =>
...
       entityType,
     }),
-    createDonorFacet({
+    ...createDonorFacet({
       fieldName: 'body_mass_index_value',
       entityType,
     }),
-  ];
+  };
+  return { tableFields: tileFields, tileFields };
```
mccalluc, line 125: Since we've got several places that return objects with similar structures, this might be a good place to introduce a helper class, instead of just making the object on the fly... but whatever works.

context/app/static/js/components/entity-search/results/Results/Results.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+function Results({ results }) {
+  const { view } = useStore();
+
+  return (
+    <ResultsLayout>
+      {results?.hits && (
+        <>
+          {view === 'table' && <ResultsTable hits={results.hits} />}
+          {view === 'tile' && <ResultsTiles hits={results.hits} />}
```
mccalluc, line 17:
```suggestion
  const ResultsComponent = {table: ResultsTable, tile: ResultsTiles}[view];

  return (
    <ResultsLayout>
      {results?.hits && (
        <>
          <ResultsComponent hits={results.hits} />
```

Some logic could be pulled out the JSX... If there were more props being repeated, might make more of a difference. Whatever you think best.

context/app/static/js/pages/entity-search/DatasetSearch/DatasetSearch.jsx
```diff
@@ -1,27 +1,21 @@
 import React from 'react';
 
 import SearchWrapper from 'js/components/entity-search/SearchWrapper';
-import { createDatasetFacet, createField } from 'js/components/entity-search/SearchWrapper/utils';
+import { createDatasetFacet, buildDatasetFields } from 'js/components/entity-search/SearchWrapper/utils';
+
+const { tableFields } = buildDatasetFields();
+
+const facets = Object.assign(
+  createDatasetFacet({ fieldName: 'mapped_data_types', label: 'Data Type', type: 'string' }),
```
mccalluc, line 9: Could `string` be a default?

mccalluc (approved):
I think it's fine as-is. Some comments: If anything resonates with you, great, but don't worry if not.

PR #2730, opened 2022-07-06
Create workspace and redirect

### LINKED ISSUE #2723: New route to to support workspace creation
Regarding testing out workspaces, while the API is still rough… My thinking right now is that `<UUID>.ipynb` will still serve the notebook for download, but GET `<UUID>.ipynb?name=my-workspace` creates a new workspace… and down the road we change the routing to only support POST (because a plain GET shouldn’t create resources).

Proposal has buy-in from Nils.

mccalluc:
- fix #2723

I have seen it work end to end. On the workspaces page you need to start the jupyter job, and then when it's running, connect to it... we'd like there to be fewer hops. 

- There are quirks / bugs:  https://github.com/hubmapconsortium/user_workspaces_server/issues/66


context/app/static/js/components/workspaces/WorkspacesList.jsx
```diff
@@ -67,7 +67,8 @@ function WorkspacesList() {
       />
       <Paper>
         {workspacesList.map((workspace) => (
-          <div key={workspace.id}>
+          /* Inbound links have fragments like "#workspace-123" */
+          <div key={workspace.id} id={`workspace-${workspace.id}`}>
```
john-conroy, line 71: You'll need to offset for the sticky header when using an anchor link as we have for the `Section` component in `js/shared-styles/sections/Section`.

Maybe use `Section` or create a reusable component with shared logic that is more fitting.

john-conroy (approved):

context/app/static/js/components/workspaces/WorkspacesList.jsx
mccalluc: Thanks for reminder! Filed issue for follow-up.

PR #2745, opened 2022-07-11
Mccalluc/link to ws

mccalluc:
In the long run, this will almost entirely go away, but for now, it enables us to click through to workspaces... which may or may not actually work, but that's work on the other side.

### DIFF
--- a/context/app/static/js/components/workspaces/JobDetails.jsx
+++ b/context/app/static/js/components/workspaces/JobDetails.jsx
@@ -2,8 +2,7 @@ import React from 'react';
 
 import { LightBlueLink } from 'js/shared-styles/Links';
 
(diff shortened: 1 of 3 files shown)


context/app/static/js/components/workspaces/JobDetails.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import { LightBlueLink } from 'js/shared-styles/Links';
+
+function JobDetails(props) {
+  const { job } = props;
```
john-conroy, line 6:
```suggestion
function JobDetails({ job }) {
```

john-conroy (approved):

PR #2753, opened 2022-07-13
Address `npm ci` slowdown

mccalluc:
File as draft while I gather information.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -24,6 +24,8 @@ jobs:
         cache: 'npm'
         cache-dependency-path: context/package-lock.json
    
+    - run: npm --version
+
     - name: Install node dependencies
       run: npm ci --prefix context
 
--- a/CHANGELOG-npm-slow.md
+++ b/CHANGELOG-npm-slow.md
@@ -1,0 +1,1 @@
+- Address the `npm ci` slowdow.


mccalluc:
Hitting node-sass / gyp problem now.

mccalluc:
dev-start working locally now. With any luck tests will pass, and complete in less time...

mccalluc:
`RUN npm ci` fails with peer dependency conflicts... because we don't `COPY` the conf file...

mccalluc:
Cypress passes! Installs seem to work, but Node tests fail early. There's a lot of upgrading here, and maybe the behavior of something we rely on has changed? 
```
  /home/runner/work/portal-ui/portal-ui/context/node_modules/jsdom/lib/jsdom/living/xhr/XMLHttpRequest-impl.js:111
        referrer: _ownerDocument.URL,
                                 ^
...
      at /home/runner/work/portal-ui/portal-ui/context/app/static/js/hooks/useImmediateDescendantProv.js:131:32
      at Array.map (<anonymous>)
      at _callee$ (/home/runner/work/portal-ui/portal-ui/context/app/static/js/hooks/useImmediateDescendantProv.js:126:57)
      at tryCatch (/home/runner/work/portal-ui/portal-ui/context/node_modules/regenerator-runtime/runtime.js:45:40)

File: end-to-end/.npmrc [hunk 1]
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
mccalluc, line 1: This should be another symlink: The real copy is kept in the docker context.

PR #2757, opened 2022-07-15
add columns to datasets TSV

### LINKED ISSUE #2755: add fields to datasets.tsv endpoint
The metadata table should include:

- [ ] Publication Date
- [ ] Modification Date
- [ ] Creation Date (might not exist if published?)
- [ ] Status
- [ ] Access
- [ ] Version
- [ ] Consortium
- [ ] Affiliation - Group
- [ ] Affiliation - Registered By

In essence, I am looking for the information that is shown at the top and bottom of each dataset page:

image

image

mccalluc:
(Basing PR on `mccalluc/npm-upgrade` because I don't want to revert my local state.)

- Fix #2755 

Filing as draft: Questions:
- Should these also be added to Donor and Sample?
- Do we need human-readable formatted times for all three? That will be additional work in search-api.
- Do these need descriptions, like we have for the other columns?
- Is it fine for these columns to be sorted alphabetically in with the other columns?
- Version is not available, and it would take substantial work to enable it.


ngehlenborg:
> Should these also be added to Donor and Sample?
Yes, that would be useful.

> Do we need human-readable formatted times for all three? That will be additional work in search-api.
No, downstream consumers should be able to easily convert the dates into whatever date string format they want.

> Do these need descriptions, like we have for the other columns?
Not necessary.

> Is it fine for these columns to be sorted alphabetically in with the other columns?
Yes.

> Version is not available, and it would take substantial work to enable it.
Ok, it is not required at the moment.

ngehlenborg (approved):
If my comments are already addressed by this PR, you can go ahead and merge.

PR #2758, opened 2022-07-15
John conroy/organs page redesign

### LINKED ISSUE #2660: Implement organ redesign
Design is in #2350. If dataset count information takes too much time to implement, push that to a second version and leave the footer as a regular purple footer with nothing in it.

john-conroy:
Fix #2660.

Using the [compound pattern](https://www.patterns.dev/posts/compound-pattern/) here which I haven't done in this project, but it seems to fit this use case.

@mccalluc I need to update stories. I also think this is good starting point to extend our end to end testing. I'll do both in follow-up PRs so this does not get too bloated.

Screen Shot 2022-07-15 at 3 36 49 PM


context/app/static/js/pages/Organs/hooks.js
```diff
@@ -0,0 +1,49 @@
...
+    return acc;
+  }, organs);
+}
+
+function useOrgansDatasetCounts(organs) {
+  const { searchData, isLoading } = useSearchData(query);
+
+  if (isLoading) {
+    return { isLoading };
+  }
```
mccalluc, line 39: Could you drop this? Or does something go wrong with `addDatasetCountsToOrgans`. Just trying to minimize branches...

context/app/static/js/pages/Organs/hooks.js
```diff
@@ -0,0 +1,49 @@
...
+    'origin_sample.mapped_organ': { terms: { field: 'origin_sample.mapped_organ.keyword', size: 10000 } },
+  },
+};
+
+function buildOrganToCountMap(aggsBuckets) {
+  return aggsBuckets.reduce((acc, { key, doc_count }) => {
+    acc[key] = doc_count;
+    return acc;
+  }, {});
+}
```
mccalluc, line 20: You use `reduce` more fluently than me. I'd probably write it like this, trying to make it analogous to a dict comprehension in python, but really, whatever you're more comfortable with.

```suggestion
function buildOrganToCountMap(aggsBuckets) {
  return Object.fromEntries(
    aggsBuckets.map(
      ({key, doc_count}) => [key, doc_count]
    )
  ) 
}
```

mccalluc (commented):
Scattered comments: The only part that concerns me is `path` vs. `name`. Looks good once you check that.

context/app/static/js/pages/Organs/hooks.js
john-conroy: `addDatasetCountsToOrgans` errors without valid args, but I could change it to handle an empty array for `aggsBuckets` if preferable.

context/app/static/js/pages/Organs/Organs.jsx
```diff
@@ -1,40 +1,32 @@
...
-  const { organs } = props;
+import TileGrid from 'js/shared-styles/tiles/TileGrid';
+import { StyledTypography } from './style';
+import { useOrgansDatasetCounts } from './hooks';
+
+function Organs({ organs }) {
+  const { isLoading, organsWithDatasetCounts } = useOrgansDatasetCounts(organs);
 
+  if (isLoading) {
+    return null;
```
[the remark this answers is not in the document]
john-conroy: I think you'd believe so.

context/app/static/js/components/organ/OrganTile/OrganTile.jsx
```diff
@@ -0,0 +1,31 @@
...
+    <Tile
+      href={`/organ/${name}`}
+      icon={
+        <StyledDiv>
+          <ColoredImage icon={icon} role="img" aria-label={`Icon for ${name}`} />
+        </StyledDiv>
+      }
+      bodyContent={
+        <>
+          <Tile.Title>{name}</Tile.Title>
```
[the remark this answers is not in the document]
john-conroy: Thanks, I fixed the `href`.

PR #2767, opened 2022-07-22
change to "marker"... or "Select gene to query"

### LINKED ISSUE #2766: "Marker" instead of "Substring"
In the cells UI

mccalluc:
- Fix #2766

I'm not confident this is the right choice... would it be better to say `gene or protein` or either `gene` or `protein` depending on the context?

(Whatever the name is here, I think keeping the same name internally makes sense: Devs are just going to think of it as a substring.)

Screen Shot 2022-07-22 at 5 24 57 PM

### DIFF
--- a/CHANGELOG-not-substring.md
+++ b/CHANGELOG-not-substring.md
@@ -1,0 +1,1 @@
+- In Cells UI, change from "substring" to something domain specific.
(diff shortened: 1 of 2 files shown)


tsliaw:
This was the original design. No need to add the additional subtext in this PR, but change the text to "Select gene" or "Select protein" depending on query type.
Screen Shot 2022-07-25 at 9 17 56 AM

mccalluc:
This is what it looks like with this PR: The label is used as the hint text, and then when data-entry begins, it moves up to the border.

Screen Shot 2022-07-25 at 3 17 58 PM
Screen Shot 2022-07-25 at 3 18 10 PM

If there's more that is needed, I'll leave it to you, @tsliaw , to say whether that should be in this PR or a separate one.

tsliaw (approved):

PR #2771, opened 2022-07-25
cells title is just "Cells | HuBMAP" now

### LINKED ISSUE #2770: Fix Cells UI page title
Not clear exactly what Nils wants, so make a PR and ask him for review.

mccalluc:
@ngehlenborg -- There was some question as to whether this was all you wanted, or if you were thinking something else is needed. This string is concatenated with ` | HuBMAP` for the html title. The on-page title is still just "Cells".

Fix #2770.

### DIFF
--- a/CHANGELOG-cells-title.md
+++ b/CHANGELOG-cells-title.md
@@ -1,0 +1,1 @@
+- Page title had been 'Cells API Demo' ... but we're moving into production, so shorten it to 'Cells'.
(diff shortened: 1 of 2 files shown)


mccalluc:
In #2789, the on-page title is changed to "Datasets"... so "Cells" probably isn't what you want here... but I'm not sure what you do want. Changing to draft.

mccalluc:
Per Nils, It should be "Datasets: Molecular Data Queries" in both contexts

mccalluc:
Replaced by #2844.

PR #2789, opened 2022-07-29
change cells title; add description

### LINKED ISSUE #2774: Top of page description (Tiffany will author text)
Text should say:
Screen Shot 2022-07-26 at 9 39 01 AM

"Refine datasets to discover genomic and proteomic information including expression distribution and cluster membership. To begin your search, select a query type and the relevant parameters."

mccalluc:
- Fix #2774


john-conroy (approved):
I would perhaps keep it as `Cells` until the two pages are merged, but your call. Thanks!

PR #2790, opened 2022-07-29
Reuse component for home description?

mccalluc:
- Inspired by #2789

No worries if this isn't a direction you want to go. When I'm looking at new UI work, I usually try to find a similar element elsewhere on the site... so I first looked at the home page... and saw a lot of complicated styling that I didn't think was relevant.

Not a high priority, but cleanups like this make ongoing development and maintenance a little easier, and help with style consistency. Thoughts?


john-conroy (approved):
Thanks, I do think it's worthwhile revisiting older areas - especially those that predate our adoption of `storybook`.

PR #2793, opened 2022-07-29
Mccalluc/workspaces join sections

### LINKED ISSUE #2785: Use workspaces API to put jobs under workspaces
Right now they are two separate sections: Make use of https://github.com/hubmapconsortium/user_workspaces_server/issues/53 to join them.

mccalluc:
- Fix #2785

It seemed best to me to bury the logic for merging the two API responses inside one hook, but you may have thoughts about it.

### DIFF
--- a/context/app/static/js/components/workspaces/utils.js
+++ b/context/app/static/js/components/workspaces/utils.js
@@ -41,4 +41,22 @@ async function startJob({ workspaceId, workspacesEndpoint, workspacesToken }) {
   });
 }
 
-export { createNotebookWorkspace, startJob };
+function mergeJobsIntoWorkspaces(jobs, workspaces) {
+  const wsIdToJobs = {};
+  jobs.forEach((job) => {
+    const { workspace_id } = job;
+    if (!(workspace_id in wsIdToJobs)) {
+      wsIdToJobs[workspace_id] = [];
+    }
+    wsIdToJobs[workspace_id].push(job);
+  });
+
+  workspaces.forEach((workspace) => {
+    // eslint-disable-next-line no-param-reassign
+    workspace.jobs = wsIdToJobs?.[workspace.id] || [];
+  });
+
+  return workspaces;
+}
+
+export { createNotebookWorkspace, startJob, mergeJobsIntoWorkspaces };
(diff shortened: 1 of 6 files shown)


mccalluc:
(Moving back to draft... let me add a little unit test.)

context/app/static/js/components/workspaces/utils.js
```diff
@@ -41,4 +41,22 @@ async function startJob({ workspaceId, workspacesEndpoint, workspacesToken }) {
...
+function mergeJobsIntoWorkspaces(jobs, workspaces) {
+  const wsIdToJobs = {};
+  jobs.forEach((job) => {
+    const { workspace_id } = job;
+    if (!(workspace_id in wsIdToJobs)) {
+      wsIdToJobs[workspace_id] = [];
+    }
+    wsIdToJobs[workspace_id].push(job);
+  });
+
```
john-conroy, line 53:
```suggestion
  jobs.reduce((acc, job) => {
      const { workspace_id } = job;
      if(!(workspace_id in acc)){
          acc[workspace_id] = [];
      }
      acc[workspace_id] = [...acc[workspace_id], job];
      return acc;
      }, {});

```

john-conroy: A slight preference for reduce, but up to you. The indentation on the suggestion seems slightly off.

john-conroy (approved):

context/app/static/js/components/workspaces/utils.js
mccalluc: I'd need to do something with the result of the `reduce`, right? It's not my usual idiom, but I'm wondering if there are arguments for it beyond:
- No external scratch variable.
- Favoring functional style over procedural.

Also, wondering about the array construction... the point is to favor creating new objects over mutating existing? I worry that it hurts readability.

mccalluc: (Going to fix indentation and merge as is, but happy to learn more about your motivation.)

mccalluc: ... and I realize I don't see the problem with the indentation, either. Sorry not to get it.

PR #2794, opened 2022-08-01
John conroy/cells expansion improvements

### LINKED ISSUE #2706: Design and implement change to cells table fold-down
Nils:
> We need to change the design of the dataset expansion in the cell API results. I always (literally every single time I go back to the the UI) click on the dataset name instead of the little triangle on the right.
>
>I think that putting the triangle on the left and making it more pronounced might help

john-conroy:
Fix #2706.

- The entire row is now clickable. The expand icon is now just an icon and not a button.

Trade-offs

- The MUI tooltip doesn't work with table rows. For now, the tooltip still only applies to the icon.
- As the `cursor: pointer` applies to the whole row and the ID link, it can be hard to differentiate when you're hovering over the link.
- The MUI `TableRow` doesn't have a `disabled` implementation. The icon will be the disabled gray and the hover effects for the row will be removed.


mccalluc (approved):
Thanks!

PR #2795, opened 2022-08-01
John conroy/use mui for expandable row

john-conroy:


context/app/static/js/shared-styles/tables/ExpandableRow/ExpandableRow.jsx
```diff
@@ -1,18 +1,14 @@
-import React, { useRef } from 'react';
-
-import { animated } from 'react-spring';
```
mccalluc, line 3: Across the whole project it looks like we have 7 uses of `react-spring` -- Would it be worth making a house-keeping issue to follow up and replace those others as well? Or is it still the best tool for some jobs?

mccalluc (approved):
Thanks!

context/app/static/js/shared-styles/tables/ExpandableRow/ExpandableRow.jsx
john-conroy: Still the best tool for some jobs.

Issue #2772, opened 2022-07-26
`KeyError: 'files'`


mccalluc:
Spotted in logs: (at least) two published production datasets produce this:

https://portal.hubmapconsortium.org/browse/dataset/c6a254b2dc2ed46b002500ade163a7cc
https://portal.hubmapconsortium.org/browse/dataset/eaad5ed4b3aa2186bd87e05ff28b593e

```
File "/Users/chuck/github/hubmap/portal-ui/context/app/routes_browse.py", line 60, in details
conf_cells_uuid = client.get_vitessce_conf_cells_and_lifted_uuid(entity, marker=marker)
File "/Users/chuck/github/hubmap/portal-ui/context/app/api/client.py", line 166, in get_vitessce_conf_cells_and_lifted_uuid
derived_entity['files'] = derived_entity['metadata']['files']
```

mccalluc:
Bill says: "nothing has changed on PROD for quite a while."

> Publication Date
> 2020-08-22
> Modification Date
> 2022-02-11

Could revert portal code to earlier state and see how it behaves.

Looking at the JSON: There is `immediate_descendants[1].metadata.files`. Should this be vis-lifted?

mccalluc:
Trying to revert: 
- Same error at [v0.47.1](https://github.com/hubmapconsortium/portal-ui/commit/83aaffaa1c4f894135cd576f82b68122ec2f7ea2)
- Same error at [v0.37.2](https://github.com/hubmapconsortium/portal-ui/commit/043fd1feb9043edf85ad112a7a178f81c4fcc2f5) (February 9; before the dataset's modification date.)

Looking at similar datasets: There are 9 published `seqFISH` all from CalTech, three each of small intestine, spleen and heart. Vis-lifted visualization works for these, except 2 from heart tissue: The same two that we saw in the log.

Next: Compare these two to the heart dataset that does work to understand the difference.

mccalluc:
Bill says it's in their court:
>  I think this is related to a problem that was reported previously, we’ve got a fix in for it (but hasn’t been released yet pending this whole other search-api reorg that we’re doing).

mccalluc:
This heart seqFish works: https://portal.hubmapconsortium.org/browse/dataset/8f0becaf2d0eea9df67e45b8d66d69ac.json
- Two `descendants`: `seqFish_lab_processed` and `image_pyramid`+`seqFish`
- The latter has `metadata.files`

[...]

PR #2796, opened 2022-08-01
Mccalluc/audit vis bugs

mccalluc:
- Inspired by #2772 ... Are there other similar errors?

Turns out, there are two more: Added a note to the issue.

TODO:
- This is network bound: it could probably be sped up significantly with some async code
- How should this be used? Maybe add a to a CI task that would run it for every new tag?

### DIFF
--- a/CHANGELOG-find-vis-bugs.md
+++ b/CHANGELOG-find-vis-bugs.md
@@ -1,0 +1,1 @@
+- Add script to find vis bugs before users hit them.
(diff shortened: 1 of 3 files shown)


john-conroy (approved):
Thanks. 

- Do we only care to do this for public datasets?
- The number of project root level scripts keeps growing. Should we think about better organization?

mccalluc:
Filed an issue for follow-up on organization. For unpublished datasets, we know there are tons of errors... I left a note on slack for their curation folks, pointing out that this exists, but I'm not sure about how it should fit into the bigger picture.

PR #2800, opened 2022-08-03
Mccalluc/refactor panel list

mccalluc:
I started off trying to reuse my-lists components, but they didn't seem to actually be appropriate here.

- I could be wrong, but I don't think we want a fixed height div here: It makes sense if there are multiple scrolling divs on the page, but here, since it's the only one, it just makes scrolling trickier.
- The panel there is really specific to the my-lists content, and there isn't much left if it is generalized enough to work for this case.

I have not touched the content in the list -- I want to approach that separately, since there need to be more changes to the structure that I want Juan's buy-in on.
Screen Shot 2022-08-03 at 3 02 11 PM

### DIFF
--- a/context/app/static/js/components/workspaces/style.js
+++ b/context/app/static/js/components/workspaces/style.js
@@ -8,11 +8,4 @@ const StyledButton = styled(WhiteBackgroundIconButton)`
   height: 36px;
 `;
 
-const Left = styled.div`
-  flex-shrink: 0;
-`;
-const Right = styled.div`
-  flex-shrink: 0;
-`;
-
-export { StyledButton, Left, Right };
+export { StyledButton };
--- a/context/app/static/js/components/workspaces/style.js
+++ b/context/app/static/js/components/workspaces/style.js
@@ -8,4 +8,11 @@ const StyledButton = styled(WhiteBackgroundIconButton)`
   height: 36px;
 `;
 
-export { StyledButton };
+const Left = styled.div`
+  flex-shrink: 0;
+`;
+const Right = styled.div`
+  flex-shrink: 0;
+`;
+
+export { StyledButton, Left, Right };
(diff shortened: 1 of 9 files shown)


context/app/static/js/components/workspaces/style.js
```diff
@@ -8,4 +8,11 @@ const StyledButton = styled(WhiteBackgroundIconButton)`
...
 `;
 
-export { StyledButton };
+const Left = styled.div`
+  flex-shrink: 0;
+`;
+const Right = styled.div`
+  flex-shrink: 0;
+`;
+
```
john-conroy, line 17: If the names are useful, I'd create a single styled component and then create `Left` and `Right` from that.

context/app/static/js/components/workspaces/style.js
```diff
@@ -8,4 +8,11 @@ const StyledButton = styled(WhiteBackgroundIconButton)`
   height: 36px;
 `;
 
-export { StyledButton };
+const Left = styled.div`
+  flex-shrink: 0;
```
john-conroy, line 12: What does this look like without `flex-shrink: 0`? Struggling to visualize the impact.

john-conroy (commented):
I think in previous workspace reviews, I've been considering the PRs as prototyping and have been less scrutinizing. Are we now moving out of that phase?

context/app/static/js/components/workspaces/style.js
mccalluc: Actually, plain divs work; Replaced.

mccalluc:
> Are we now moving out of that phase?

I'd say moving out, but not out yet! I've filed a follow-up (#2803) that changes the information we show, and I want to get Juan's buy-in on that before going too deep on the UI polishing.

PR #2803, opened 2022-08-03
Mccalluc/only show one status

### LINKED ISSUE #2801: Map job status terms
API uses [these](https://drive.google.com/file/d/1kte4lZArTykWX6vRVIW9YOcCtNC0Qduw/view) / Design (agreed to by Juan and Nils) uses [these](https://www.figma.com/file/NgBjvs0jRbDHhiyn5nicDm/Workspace?node-id=126%3A3544)

### LINKED ISSUE #2802: Only show one job status

mccalluc:
The modelling of jobs in the API is rather different from the design from Tiffany envisions. UIs exist to simplify things, so this isn't necessarily a problem... but I wanted to to get buy-in from @jpuerto-psc , so we're not trying to sort this out a month from now, when more has been built on top.

Notable differences:

- **This fixes #2801** 

The design uses a different set of terms from what the API returns: some don't correspond to states the API provides (ie, there is no "Deleting" state from the API) and other distinctions the API makes don't matter in the UI, so we're just left with 'Active', 'Activating', and 'Inactive'.

- **This fixes #2802**

A workspace can have multiple jobs... but the UI design just tells the user about a single job. To implement that, I scan the jobs, and just display the highest status I can find.

- **We don't display any job metadata**

Only if the job is "Active" do I show a link. Otherwise, the UI just has "Status: Activating" or "Status: Inactive", and nothing else.

@jpuerto-psc , if this is fine with you, (and @pdblood), great: Just approve. If it's not, I'd suggest:
- Review this, and request changes.
- Get in touch with Tiffany, and explain your usecase, and see if the UI design can change.
- If you can't get on the same page, no worries: Just document what the differences are, and we elevate to Nils and Phil.

Thanks!

### DIFF
--- a/CHANGELOG-clean-status.md
+++ b/CHANGELOG-clean-status.md
@@ -1,0 +1,1 @@
+- Change the display of workspace status to match Tiffany's spec.
--- a/CHANGELOG-only-offer-start-job-if-no-active-pending.md
+++ b/CHANGELOG-only-offer-start-job-if-no-active-pending.md
@@ -1,0 +1,1 @@
+- Only offer to start jupyter is there is nothing active or pending.
--- a/context/app/static/js/components/workspaces/JobDetails.jsx
+++ b/context/app/static/js/components/workspaces/JobDetails.jsx
@@ -3,24 +3,44 @@ import React from 'react';
 import { LightBlueLink } from 'js/shared-styles/Links';
 
 const ACTIVE = 'Active';
+const ACTIVATING = 'Activating';
+const INACTIVE = 'Inactive';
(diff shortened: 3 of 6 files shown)


jpuerto-psc (approved):
The only concern I have is the one job per workspace constraint. There may be cases (in the future) where a user may want to run different jobs (RStudio and JupyterLab) for the same workspace at the same time. I am fine with doing this for now (one job per workspace) for the MVP, as long as this is something that we're aware of for potential future development.

tsliaw (approved):

PR #2813, opened 2022-08-04
Mccalluc/pass groups token

### LINKED ISSUE #2812: Pass globus group when creating workspace.
```
{
  "workspace_name": "string",
  "description": "string",
  "workspace_details": {
    "globus_groups_token": "string",
    "symlinks": [
      {
        "name": "string",
        "path": "string",
        "dataset_uuid": "string"
      }
    ],
    "files": [
      {
        "name": "string",
        "content": "string"
      }
    ]
  }
}
```

mccalluc:
- Fix #2812

It doesn't interfere with login right now, but let me know when the logic that would actually check it is in place, so I can confirm that it's working correctly.

### DIFF
--- a/CHANGELOG-group-token.md
+++ b/CHANGELOG-group-token.md
@@ -1,0 +1,1 @@
+- Add globus token to workspace creation call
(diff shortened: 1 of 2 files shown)


jpuerto-psc:
@mccalluc one quick comment - the globus groups id is only "required" if a dataset_uuid is being passed. So if you are just passing files and symlinks without any dataset_uuids you don't "have" to include it. That being said, if you do include that token when there are no dataset_uuids, there won't be any issues either.

mccalluc:
Assuming this as Juan's approval.

PR #2818, opened 2022-08-08
fix script path

mccalluc:
- Follow-up to #2811.

Correct path for new location.

### DIFF
--- a/scripts/dev/organ-utils/make_organ_dir.py
+++ b/scripts/dev/organ-utils/make_organ_dir.py
@@ -17,7 +17,7 @@ import json
 
 
 def main():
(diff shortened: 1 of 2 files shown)


scripts/dev/organ-utils/make_organ_dir.py
```diff
@@ -17,7 +17,7 @@
 
 
 def main():
-    repo_path = Path(__file__).resolve().parent.parent
+    repo_path = Path(__file__).resolve().parent.parent.parent.parent
```
john-conroy, line 20: Is there a more reliable way to get the project root dir? If not, this works.

john-conroy (approved):

Issue #2816, opened 2022-08-05
Handle empty files array in search api v3 when generating Vitessce confs


john-conroy:
From Bill:

> We found an issue with Vitesse and the new https://search.api.hubmapconsortium.org/v3/ indices.  These new indices are now providing an empty array for files everywhere instead of just no filed at all or and empty object (this is a fix for something else that hadn’t been released yet that we discussed a while ago)-- It looks like Vitesse is trying to display even though there are no files.

mccalluc:
More context:

Bill:

> Right, but check https://portal.test.hubmapconsortium.org/browse/dataset/906ce8025bd07f04fe6a33d1a4daf400, which is connected to the new, v3 version of the indices.

John

> Gotcha, we look for files in the top level of the dataset document. In the TEST link, there is the empty files array.
> ```
>         elif 'files' not in entity or 'data_types' not in entity:
>             vitessce_conf = ConfCells(None, None)
> ```
> where entity is the dataset document returned from search-api.

PR #2819, opened 2022-08-08
Mccalluc/vis qa to es

### LINKED ISSUE #2817: Run find-vis-bugs.py against `v3` index
Currently takes a portal hostname as endpoint, rather that going directly to the index.

mccalluc:
- Add `--portal_index_path` argument
- Break code into functions
- Use the portal's internal client, rather than straight HTTP.
- Clean up output.
- Fix #2817

... except, from 
-  #2816 
> Right, but check https://portal.test.hubmapconsortium.org/browse/dataset/906ce8025bd07f04fe6a33d1a4daf400, which is connected to the new, v3 version of the indices.

I was expecting to see an error on that UUID, and I don't:
```
$ scripts/qa/find-vis-bugs.py --portal_index_path /portal/v3
...
d55e751748bd094ff5b0b55befb08d41	Image pyramid assay with uuid 0f3894d45b772804de43137fa8c353ed has no matching files
c6a254b2dc2ed46b002500ade163a7cc	seqFish assay with uuid 9db61adfc017670a196ea9b3ca1852a0 has no matching files
eaad5ed4b3aa2186bd87e05ff28b593e	seqFish assay with uuid b5707bbde0050121ca7291e30054d548 has no matching files
d3e05111d84be552b98f5fe2b18e3054	Image pyramid assay with uuid 3af85031cea5b0c772212855550cc3b1 has no matching files
```

(Same errors as on the default.)

Filing as draft: Will add arg to run a single UUID.

### DIFF
--- a/CHANGELOG-better-vis-scan.md
+++ b/CHANGELOG-better-vis-scan.md
@@ -1,0 +1,1 @@
+- Add another parameter to the vis-scan script, so we can target other indices.
--- a/CHANGELOG-find-git-root.md
+++ b/CHANGELOG-find-git-root.md
@@ -1,0 +1,1 @@
+- Use Trevor's idiom for finding the git root.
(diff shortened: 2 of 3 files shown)


mccalluc:
Oh, the example given is unpublished, so this is good, as far as it goes.

john-conroy (approved):

PR #2820, opened 2022-08-08
Use index v3

### LINKED ISSUE #2815: Support "v3" index
I think this may make more sense as something that we keep in version control, rather than something we get from the environment.

mccalluc:
- Fix #2815
- I don't believe there are new errors in published datasets: See #2819
- Blocking on #2816... but I think these are unpublished?
- Question: Should the CCF also be using `v3`?


mccalluc:
Replaced by #2822. (Might as well avoid merge conflicts.)

PR #2821, opened 2022-08-08
Mccalluc/find git root

mccalluc:
Follow up to your earlier comment, John.


john-conroy (approved):
Thanks! Python CI failing, but looks good once that passes. Out of curiosity do you know if for submodules `.git` is always a file?

mccalluc:
I think that if this repo were a submodule, then, yes there would be a `.git`... but if you wanted the top-level, you could iterate in the reverse order. I'll fix the linting and merge.

PR #2822, opened 2022-08-09
handle v3

### LINKED ISSUE #2815: Support "v3" index

### LINKED ISSUE #2816: Handle empty files array in search api v3 when generating Vitessce confs

mccalluc:
- Fix #2815
- Fix #2816
- Replace #2820

I've confirmed that http://localhost:5001/browse/dataset/906ce8025bd07f04fe6a33d1a4daf400 works. I have not run find-vis-bugs.py on everything.

### DIFF
--- a/CHANGELOG-fix-v3-files.md
+++ b/CHANGELOG-fix-v3-files.md
@@ -1,0 +1,1 @@
+- Handle the empty files array provided by the v3 index correctly.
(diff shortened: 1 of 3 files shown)


john-conroy (commented):
Thanks! The calls to `intercept` in cypress tests need their search urls updated.

Issue #2746, opened 2022-07-12
Design loading screen for workspaces


tsliaw:
Workspaces need a 'launch' screen to tell users of wait time.

mccalluc:
Just to clarify, not a whole screen... We'll have a list of the users workspaces, and an individual entry there may be in a "pending" state.

mccalluc:
Closing: I don't think anything is needed here.

mccalluc:
Reopening -- I had forgotten what this was about: We'll need this after a job has been requested, but before there's a Jupyter URL we can redirect to.

PR #2824, opened 2022-08-09
Bring the Workspaces closer to the design

mccalluc:
Present a uniform UI for workspaces, regardless of whether Jupyter is already running, or if it needs to be started from scratch. We're getting into territory where we're disguising how the system works behind the scenes, and that worries me a little.

@tsliaw :
- Is this acceptable, as far as it goes?
- How should we handle workspaces with no active jobs?
- This reminded me what #2746 was for. I'll branch from this and implement a stop-gap.
Screen Shot 2022-08-09 at 4 18 48 PM

### DIFF
--- a/context/app/static/js/components/workspaces/JobLink.jsx
+++ b/context/app/static/js/components/workspaces/JobLink.jsx
@@ -21,7 +21,7 @@ function JobLink({ workspace, job, children }) {
     const handleStart = createHandleStart(workspace.id);
     return (
       <OutboundIconLink>
-        <button onClick={handleStart} type="submit" style={{ all: 'unset' }}>
+        <button onClick={handleStart} type="submit" style={{ all: 'unset', cursor: 'pointer' }}>
           {children}
         </button>
       </OutboundIconLink>
--- a/context/app/static/js/components/workspaces/JobLink.jsx
+++ b/context/app/static/js/components/workspaces/JobLink.jsx
@@ -1,0 +1,36 @@
+import React, { useContext } from 'react';
+
+import { AppContext } from 'js/components/Providers';
+import OutboundIconLink from 'js/shared-styles/Links/iconLinks/OutboundIconLink';
+import { startJob } from './utils';
+
+function JobLink({ workspace, job, children }) {
+  const { workspacesEndpoint, workspacesToken } = useContext(AppContext);
+
+  function createHandleStart(workspaceId) {
+    async function handleStart() {
+      startJob({ workspaceId, workspacesEndpoint, workspacesToken });
+      // TODO: Open new tab
+      // eslint-disable-next-line no-alert
+      alert('TODO: Open a new tab that will poll until the job is started.');
+    }
+    return handleStart;
+  }
+
+  if (job.allowNew) {
+    const handleStart = createHandleStart(workspace.id);
+    return (
+      <OutboundIconLink>
+        <button onClick={handleStart} type="submit" style={{ all: 'unset' }}>
+          {children}
+        </button>
+      </OutboundIconLink>
+    );
+  }
+  if (job?.url) {
+    return <OutboundIconLink href={job.url}>{children}</OutboundIconLink>;
+  }
+  return children;
+}
+
+export default JobLink;
--- a/context/app/static/js/components/workspaces/JobLink.jsx
+++ b/context/app/static/js/components/workspaces/JobLink.jsx
@@ -3,6 +3,7 @@ import React, { useContext } from 'react';
 import { AppContext } from 'js/components/Providers';
 import OutboundIconLink from 'js/shared-styles/Links/iconLinks/OutboundIconLink';
 import { startJob } from './utils';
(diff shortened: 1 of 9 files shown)


context/app/static/js/shared-styles/Links/index.js
```diff
@@ -6,6 +6,7 @@ const LinkWithoutUnderline = (props) => <Link {...props} underline="none" />;
 
 const LightBlueLink = styled(LinkWithoutUnderline)`
   color: ${(props) => props.theme.palette.info.main};
+  cursor: pointer;
```
john-conroy, line 9: This shouldn't be necessary.

context/app/static/js/components/workspaces/WorkspaceDetails.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { condenseJobs } from './utils';
+import JobLink from './JobLink';
+import JobStatus from './JobStatus';
+
+function WorkspaceDetails({ workspace }) {
+  const job = condenseJobs(workspace.jobs);
+
+  return (
+    <b>
```
john-conroy, line 11: From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

> However, you should not use <b> for styling text; instead, you should use the CSS [font-weight](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight) property to create boldface text, or the [<strong>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong) element to indicate that text is of special importance.

john-conroy: Should use some from of `Typography`.

context/app/static/js/components/workspaces/JobLink.jsx
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  if (job.allowNew) {
+    const handleStart = createHandleStart(workspace.id);
+    return (
+      <OutboundIconLink>
+        <button onClick={handleStart} type="submit" style={{ all: 'unset' }}>
+          {children}
+        </button>
+      </OutboundIconLink>
```
john-conroy, line 27: Also should not use in-line styles.

john-conroy, line 27:
```suggestion
    return (
      <OutboundIconLink  component="button" onClick={() => handleStart(workspace.id)} type="submit" style={{ all: 'unset' }}>
          {children}
      </OutboundIconLink>
```

john-conroy (changes requested):
Could you include me as a review on non-trivial UI focused PRs?

context/app/static/js/components/workspaces/JobLink.jsx
mccalluc: I think I've applied all the other suggestions above, so I have
```
      <OutboundIconLink>
        <LinkButton onClick={createHandleStart(workspace.id)}>{children}</LinkButton>
      </OutboundIconLink>
```
but `component={LinkButton}` doesn't produce the same result.

john-conroy: It should be `component={button}`.

john-conroy: There should only be one component which is `OutboundIconLink`.

john-conroy: If that doesn't work, feel free to merge with whatever you think is best and I can pick it up later.

context/app/static/js/components/workspaces/JobLink.jsx
```diff
@@ -0,0 +1,34 @@
...
+    return handleStart;
+  }
+
+  if (job.allowNew) {
+    return (
+      <OutboundIconLink>
+        <LinkButton onClick={createHandleStart(workspace.id)}>{children}</LinkButton>
+      </OutboundIconLink>
+    );
+  }
```
mccalluc, line 27:
```diff
     return (
-      <OutboundIconLink>
-        <LinkButton onClick={createHandleStart(workspace.id)}>{children}</LinkButton>
+      <OutboundIconLink component="button" onClick={createHandleStart(workspace.id)}>
+        {children}
       </OutboundIconLink>
     );
```
Baseline of text changes, along with weight. Fixing the weight is easy, but not sure how to align baseline.
Screen Shot 2022-08-10 at 9 44 50 AM

--------

```diff
     return (
-      <OutboundIconLink>
-        <LinkButton onClick={createHandleStart(workspace.id)}>{children}</LinkButton>
+      <OutboundIconLink component={LinkButton} onClick={createHandleStart(workspace.id)}>
+        {children}
       </OutboundIconLink>
     );
```
Now the text baseline is good, but the icon baseline has moved.

Screen Shot 2022-08-10 at 10 02 39 AM

mccalluc:
John volunteers to look at this post-merge. Tiffany suggested displaying "No Jobs". I'll go ahead and merge.

PR #2827, opened 2022-08-10
users do not supply name when creating workspace

mccalluc:
RIght now, it looks for a query parameter with the name of the workspace ... but my understanding of the planned UI is that there will be no prompt for a name, and the workspaces will be assigned names like "HBM989.VVXT.224 Workspace". If that is correct, this is a small step in the right direction.

### DIFF
--- a/CHANGELOG-ws-no-name.md
+++ b/CHANGELOG-ws-no-name.md
@@ -1,0 +1,1 @@
+- Users will not supply a name when creating workspaces from datasets.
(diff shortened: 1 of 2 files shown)


mccalluc:
Looks like docker had trouble pulling an image?
```
 failed to do request: Head "https://registry-1.docker.io/v2/tiangolo/uwsgi-nginx-flask/manifests/python3.9": EOF
```
restarted CI

tsliaw (approved):

PR #2829, opened 2022-08-10
John conroy/link button

### LINKED ISSUE #2826: Tidy up component nesting in Workspaces
Ideally, we'd just use `component`... but that introduces other display problems.

```diff
     return (
-      <OutboundIconLink>
-        <LinkButton onClick={createHandleStart(workspace.id)}>{children}</LinkButton>
+      <OutboundIconLink component="button" onClick={createHandleStart(workspace.id)}>
+        {children}
       </OutboundIconLink>
     );
```
Baseline of text changes, along with weight. Fixing the weight is easy, but not sure how to align baseline.
Screen Shot 2022-08-10 at 9 44 50 AM

--------

```diff
     return (
-      <OutboundIconLink>
-        <LinkButton onClick={createHandleStart(workspace.id)}>{children}</LinkButton>
+      <OutboundIconLink component={LinkButton} onClick={createHandleStart(workspace.id)}>
+        {children}
       </OutboundIconLink>
     );
```
Now the text baseline is good, but the icon baseline has moved.

Screen Shot 2022-08-10 at 10 02 39 AM

_Originally posted by @mccalluc in https://github.com/hubmapconsortium/portal-ui/pull/2824#discussion_r942496830_

john-conroy:
Fix #2826.

Screen Shot 2022-08-10 at 4 45 22 PM


mccalluc (changes requested):
At least when I check out this branch locally, there seem to be problems.
- display: The space on the left of the pipe is collapsing. I've tried moving it up and between the components, but then it collapses on both sides, and I don't think you'd like a solution involving `&nbsp;`
- prop-types: `Failed prop type: Invalid prop `linkComponent` supplied to `LinkButton`, expected a ReactNode`
- functionality: I don't think that `handleStart` is being called, or at least I don't see the `alert('TODO: Open a new tab...')`.

At the start of this, I think you were aiming for a reusable component that if provided an href would function as a link, and if provided a handler, would use that instead... I'm not sure how much effort we should put here, if we don't get something general purpose out of it.

.... Thinking a little more, maybe this can be simplified: maybe we use event handlers in either case... Where there's a link now, we replace it with a little code that just resets the `location`... Wouldn't normally do this, but it would mean that we wouldn't have to worry with any conditional display logic: The only part that matters is the call back.

Maybe talk about this before investing more time?

john-conroy:
```.... Thinking a little more, maybe this can be simplified: maybe we use event handlers in either case... Where there's a link now, we replace it with a little code that just resets the location... Wouldn't normally do this, but it would mean that we wouldn't have to worry with any conditional display logic: The only part that matters is the call back.```

I think having a workaround like that is bad for accessibility.

File: context/app/static/js/components/workspaces/WorkspaceDetails/WorkspaceDetails.jsx [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+function WorkspaceDetails({ workspace }) {
+  const job = condenseJobs(workspace.jobs);
+
+  return (
+    <Flex>
+      <JobLink workspace={workspace} job={job} typographyVariant={typographyVariant}>
+        {workspace.name}
+      </JobLink>
+      <Typography variant={typographyVariant}>
+        <span>&nbsp;</span>|<span>&nbsp;</span>
```
mccalluc, line 20:
```suggestion
        &nbsp;|&nbsp;
```
Seems to work w/o spans?

context/app/static/js/components/workspaces/WorkspacesList.jsx
```diff
@@ -8,10 +8,10 @@ import { DeleteIcon, AddIcon } from 'js/shared-styles/icons';
 import { SpacedSectionButtonRow } from 'js/shared-styles/sections/SectionButtonRow';
 import { PanelWrapper } from 'js/shared-styles/panels';
 
+import WorkspaceDetails from 'js/components/workspaces/WorkspaceDetails';
```
mccalluc, line 11:
```suggestion
import WorkspaceDetails from './WorkspaceDetails';
```

When the IDE fills in imports, it doesn't take account of the ones that could be relative. Is this worth fixing by hand?

mccalluc (approved):
Thanks! Comments you can take or leave.

PR #2834, opened 2022-08-11
John conroy/fix log script

john-conroy:
Same approach as #2821. You may have an idea for moving the `sys.path.append` code block to a utils file (or whatever the Python convention is...) for reuse between scripts. Alternatively, I could move the  `portal-logs-errors` output dir down to `etc/qa`.


mccalluc (approved):
This is fine with me... but moving the output dir down would be fine too.

PR #2839, opened 2022-08-12
John conroy/fix empty cells charts

john-conroy:
Like we discussed in stand-up, I believe we were unintentionally relying on a component dismount to reset the store. The recent refactors disrupted the store reset.

### DIFF
--- a/context/app/static/js/components/cells/CellsCharts/CellsCharts.jsx
+++ b/context/app/static/js/components/cells/CellsCharts/CellsCharts.jsx
@@ -19,33 +19,31 @@ function CellsCharts({ uuid, cellVariableName, minExpression, queryType, isExpan
   const clusterData = 'clusterData' in cellsData ? cellsData.clusterData : {};
 
   return (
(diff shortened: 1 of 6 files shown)


context/app/static/js/components/cells/CellsCharts/CellsCharts.jsx
```diff
@@ -19,33 +19,31 @@ function CellsCharts({ uuid, cellVariableName, minExpression, queryType, isExpan
...
-                diagnosticInfo.numCells
-              } cells.`
-            ) : (
-              <Skeleton />
-            )}
-          </StyledTypography>
-        </>
-      )}
-    </PaddedDiv>
+    (isLoading || Object.keys(cellsData).length > 0) && (
```
mccalluc, line 22: The change here is to pull the check outside the `<PaddedDiv>`, I think?

Maybe take it one step further and pull this out of the return entirely, and instead of have a little guard clause just above?

mccalluc (approved):

PR #2841, opened 2022-08-15
Mccalluc/ws job please wait

mccalluc:
- Add a "please wait" page that polls until a Jupyter URL is available.
- From the workspace list, if the workspace doesn't have any active jobs, start a new one and open a new tab on the please-wait page.
- From the `.ws.ipynb` route, start a new job, and redirect to the please wait page. (Making this a new tab will be handled when there is a UI exposure of this functionality.)

Will ping Tiffany and Juan about design... If there were progress updates from the API, that would be nice.


context/app/static/js/pages/WorkspacePleaseWait/WorkspacePleaseWait.jsx
```diff
@@ -0,0 +1,32 @@
...
+    const { current_job_details } = responseJson.data.jobs[0].job_details;
+    setStatus(current_job_details.message);
+    const { connection_details } = current_job_details;
+    if (connection_details) {
+      const { url_path, url_domain } = connection_details;
+      document.location = `${url_domain}${url_path}`;
+    }
+  }
+
+  setInterval(loadIfUp, 1000);
```
john-conroy, line 27: I'm unfamiliar with the performance of the workspaces API, but it could be problematic if the request takes longer than the interval to return.

https://developer.mozilla.org/en-US/docs/Web/API/setInterval#ensure_that_execution_duration_is_shorter_than_interval_frequency

john-conroy, line 27: I would move all of this into a `useWorkspacesPleaseWait` hook.

mccalluc: That's reasonable... but I'm realizing that I'm going to be redoing most of this. So leaving as is, but will pull it out when it's redone.

PR #2853, opened 2022-08-18
Mccalluc/beta workspaces users

mccalluc:
The goal of this PR is to provide UI exposure for workspace creation with the least changes to existing code... That may not be the right set of goals.

- It's not a priority for me right now to polish the UI -- I think the unpolished state is actually a good reminder that this is not public functionality.
- @john-conroy -- If we were to polish it, I'd need the three circle icon... I think you said you had it in your scratch work somewhere?
- @john-conroy -- One next step is to set up the routing for POST. (GET should not prompt resource creation, like it does here.) Would a plain HTML form with just one submit button be ok, or is there something better? 
- @tsliaw -- Currently, not all datasets, much less samples and donors, have notebooks. Should they? For now, I'm disabling the button if there's no notebook... maybe we'd do something similar even when it is published?
- I've forgetting if there are any situations where `shouldDisplaySection.visualization !== hasNotebook`... could be that `hasNotebook` is redundant? Need to look into that.


john-conroy:
I think it'll be easiest for me to branch off this to get you the icon.

mccalluc:
> Currently, not all datasets, much less samples and donors, have notebooks. Should they?

yes they should. (says Tiffany)

context/app/static/js/components/Providers.jsx
```diff
@@ -18,13 +18,18 @@ const AppContext = React.createContext({});
 
 function Providers(props) {
   const { endpoints, groupsToken, isAuthenticated, userEmail, children, workspacesToken } = props;
+  // TODO: Delete this when workspaces are publicly released.
+  // If we stay in limbo for a long time, this configuration could be moved out of code.
+  const workspacesUsers = ['chuck_mccallum@hms.harvard.edu'];
```
jpuerto-psc, line 23: 'pdblood@andrew.cmu.edu', 'jpuerto@andrew.cmu.edu'

john-conroy, line 23:
```suggestion
  const workspacesUsers = ['chuck_mccallum@hms.harvard.edu', 'pdblood@andrew.cmu.edu', 'jpuerto@andrew.cmu.edu'];
```

jpuerto-psc: 'dbetancur@cmu.edu'

This is fine for the initial list, we are working on getting a few more users together, but this does not have to wait for that list to be complete. Once I hear back w/ RE to that, I will let the team know.

mccalluc:
@john-conroy -- I'll take "comment" as indicating "ok to merge with changes" ... so I'll double check this locally and then do that. The three-circles PR can come any time. Thanks!

john-conroy:
@mccalluc That PR already exists. #2860.

PR #2854, opened 2022-08-19
rename component, change text

mccalluc:


context/app/static/js/components/workspaces/WorkspacesAuthenticated.jsx
```diff
@@ -5,30 +5,23 @@ import { AppContext } from 'js/components/Providers';
...
-          Github issue
-        </OutboundIconLink>
-      </Description>
-    );
+    return <Description padding="20px">The workspaces token request failed at login.</Description>;
   }
 
   return (
     <>
       <Description padding="20px">
```
john-conroy, line 17: I think it'd be more fitting to use an `Alert` as we have for other warnings, but since this is temporary I don't have a strong preference.

context/app/static/js/components/workspaces/WorkspacesAuthenticated.jsx
```diff
@@ -5,30 +5,23 @@ import { AppContext } from 'js/components/Providers';
...
-        <strong>This is not yet ready for production use! {contact}</strong>{' '}
+        <strong>
```
john-conroy, line 18: Again, I'd lean towards using `Typography` to keep things consistent... but good for now. Thanks!

john-conroy (approved):

context/app/static/js/components/workspaces/WorkspacesAuthenticated.jsx
mccalluc: As I'm thinking about it now, I think I just want to throw an exception: This is unrecoverable, and it's not a user error.

PR #2860, opened 2022-08-22
John conroy/workspaces icon

john-conroy:
Hope this helps. I'm not sure why you used `type='submit'` before?

@tsliaw FYI the default disabled state for icon buttons is a transparent background and a grey icon. If you'd like it to be something else, let me know.


mccalluc (approved):
Looks good; Thanks! (`type="submit"` was there because of an accessibility warning.)

PR #2861, opened 2022-08-22
Mccalluc/sticky cookie

### LINKED ISSUE #2859: Sticky cookie to track hubmap users
In the analytics, we want to be able to distinguish hubmap users from outsiders, even when we're not logged in. To do that, we'll set a long-lived cookie on log-in, and check it when logging user actions.

mccalluc:
- Fix #2859 
- I started to use a "custom variable", but it looks like "custom dimension" is preferred now.
- Dimensions need to be pre-registered in the matomo UI, and then referenced by ID in the API.
- I added this dimension for each of our three sites (local, prod, other).
- Docs confused me: [this](https://developer.matomo.org/api-reference/tracking-javascript) made me think it was a method on the tracker object, but that didn't work, and [this](https://github.com/jonkoops/matomo-tracker/pull/585/files#diff-bcd475bc7ec4b99349047ee03e3358d7e713e576ac0c4b710b88abdeca9cd266R13) does.
- Or at least it doesn't error? I haven't seen data show up in the dash board, but maybe the non-real-time options take time to aggregate?

### DIFF
--- a/context/app/static/js/helpers/trackers.js
+++ b/context/app/static/js/helpers/trackers.js
@@ -1,5 +1,6 @@
 import MatomoTracker from '@datapunt/matomo-tracker-js';
 import ReactGA from 'react-ga';
+import { readCookie } from 'js/helpers/functions';
 
 function getSiteId(location) {
   const { host } = location;
@@ -15,6 +16,11 @@ function getSiteId(location) {
     default:
       return 3;
   }
+}
+
+function isHubmapUser() {
+  // Set in routes_auth.py:
+  return Boolean(readCookie('last_login'));
 }
 
 const tracker = new MatomoTracker({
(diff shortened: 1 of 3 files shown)


tsliaw (approved):

context/app/static/js/helpers/trackers.js
```diff
@@ -36,6 +42,7 @@ const tracker = new MatomoTracker({
   //   setSecureCookie: true,
   //   setRequestMethod: 'POST'
   // }
+  configurations: { setCustomDimension: [1 /* user_type */, getUserType()] },
```
john-conroy, line 45: So the ID for `user_type` is 1 since this is our first custom dimension?

john-conroy (approved):
Thanks!

context/app/static/js/helpers/trackers.js
mccalluc: yep.

PR #2862, opened 2022-08-23
add workspaces link to menu

mccalluc:
- Should `userEmail` also be pulled from context, since it's there? Or is there an argument for passing things explicitly, sometimes?
- I think the plan for the future would be for this not to be conditional: If they aren't logged in, the workspaces page doesn't do much, but we'd still link to it?


john-conroy (approved):
Thanks! At some point I should consolidate the app context into a single hook import.

PR #2863, opened 2022-08-23
Mccalluc/big workspace icon

mccalluc:
- The height is based on the height for h1 headers... Maybe I should take this from the theme?
- The vertical-align just makes it look more like the mock-up, but I'm not sure how well it would work for other icons.

Screen Shot 2022-08-23 at 12 30 25 PM

### DIFF
--- a/CHANGELOG-workspace-header-icon.md
+++ b/CHANGELOG-workspace-header-icon.md
@@ -1,0 +1,1 @@
+- Add an icon in the header title.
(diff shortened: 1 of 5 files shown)


john-conroy:
I'd use a flex wrapper with `display: flex; align-items: center;` to align the text and icon. You could also pass the 'variant' used for the `Typography` to the icon as well to keep the font sizes sync with something like:

```
const StyledIcon = styled(SvgIcon)` font-size: ${(props) => props.theme.typography[props.$variant].fontSize}`;
```

john-conroy (changes requested):
Let me know if my comment above is unclear and I can take a swing at this.

Issue #2856, opened 2022-08-22
Twitter on homepage is taking full height


mccalluc:
Both on production and stage. Maybe something in their widget changed? Or weird CSS side-effect from some unrelated change on our side.

john-conroy:
https://twittercommunity.com/t/new-timeline-embed-is-not-respecting-documented-parameters/175729

john-conroy:
https://twittercommunity.com/t/suggestions-for-the-new-twitter-embed/175690/5

mccalluc:
Home page looks ok with the last PR, but the list isn't scrollable.

PR #2864, opened 2022-08-23
John conroy/fix homepage spacing

john-conroy:
Towards #2856.

Fixes overflow, but scrolling no longer works. Waiting for more dust to settle with Twitter's widget changes before investing more time.


mccalluc (approved):
Thanks! I don't think we should invest much time in this... hopefully someone upstream will have a fix.

PR #2873, opened 2022-08-29
Mccalluc/tsv and lineup

### LINKED ISSUE #2866: Lineup is broken
Error is different on Donors/Samples (JS: something is null) and for Datasets (500 from server), so probably two separate bugs.

### LINKED ISSUE #2871: Unconstrained dataset TSV download fails
Believe the root cause is the same as
- #2866

mccalluc:
- Fix #2871 
- Fix #2866

TSV: Excluding `files` because without this, the response size is too big, and even if we window the search results, we get a column with json blobs.

Lineup: I think lineup was blocked at both points, but I could be mistaken. Could file it upstream, but my general sense is that they're satisfied with the behavior.

### DIFF
--- a/context/app/static/js/pages/LineUpPage/LineUpPage.jsx
+++ b/context/app/static/js/pages/LineUpPage/LineUpPage.jsx
@@ -13,7 +13,7 @@ import metadataFieldTypes from 'metadata-field-types';
 
 function LineUpPage({ entities }) {
   const cleanEntities = entities.map((entity) =>
-    Object.fromEntries(Object.entries(entity).filter((pair) => pair[1] !== null)),
+    Object.fromEntries(Object.entries(entity).filter(([,v]) => v !== null)),
   );
   const firstRow = cleanEntities[0];
   const notEnumFields = new Set([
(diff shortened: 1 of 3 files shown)


context/app/static/js/pages/LineUpPage/LineUpPage.jsx
```diff
@@ -11,9 +11,11 @@ import Paper from '@material-ui/core/Paper';
 import SectionHeader from 'js/shared-styles/sections/SectionHeader';
 import metadataFieldTypes from 'metadata-field-types';
 
-function LineUpPage(props) {
-  const { entities } = props;
-  const firstRow = entities[0];
+function LineUpPage({ entities }) {
+  const cleanEntities = entities.map((entity) =>
+    Object.fromEntries(Object.entries(entity).filter((pair) => pair[1] !== null)),
```
john-conroy, line 16:
```suggestion
    Object.fromEntries(Object.entries(entity).filter(([,v]) => v !== null)),
```
A little more clear to me, but not a big deal.

john-conroy (approved):
As it is now, `cleanEntities` will be updated on every render. If it seems slow given the number of entities and fields ,you could consider memoizing so it only runs once on mount.

mccalluc:
> As it is now, `cleanEntities` will be updated on every render. If it seems slow given the number of entities and fields ,you could consider memoizing so it only runs once on mount.

Unless I'm missing something, Lineup renders just once: successive interactions with Lineup are not re-renders. (I've just put a console.warn in there to confirm.) Memoizing seems like a premature optimization?

PR #2876, opened 2022-08-30
Only show "active" workspaces

mccalluc:


jpuerto-psc:
Is there any desire to show "idle" workspaces so that users can start a new job on them? @mccalluc

jpuerto-psc (approved):

PR #2877, opened 2022-08-30
Mccalluc/pick nb to open

### LINKED ISSUE #2869: Pre-open a particular notebook in JupyterLab
[This can be done in the URL.](https://jupyterlab.readthedocs.io/en/stable/user/urls.html#combining-url-functions) 

Probably something like: 
> <span>https</span>://workspaces-pt.hubmapconsortium.org/passthrough/l002.hive.psc.edu/119/<b>lab/tree/HBM989.VVXT.224.ipynb</b>

but (1) Can we agree that the notebook to pre-open will always be `notebook.ipynb`, or something like that? Otherwise, if we support multiple notebook workspaces, we need some other channel to indicate which is the right notebook to open.

(2) It doesn't open the notebook on load... it seems to take a couple seconds. For something less distracting, could use an old-fashioned notebook interface, which will go straight there: 
> <span>https</span>://workspaces-pt.hubmapconsortium.org/passthrough/l002.hive.psc.edu/119/<b>tree/HBM989.VVXT.224.ipynb</b> (no "lab")

mccalluc:
- Fix #2869

When generating the list of workspaces, try to find a notebook to open for the workspace, and append it as the hash of the please-wait page's URL.

When Jupyter is available, take the notebook path from the URL hash and insert it into the Jupyter URL.

- @jpuerto-psc : The API response includes a `current_workspace_details`... but it's empty. Would it be possible to populate it? and when done, nudge me.
- @ngehlenborg : Given the caveat that this might not be the _current_ state of the workspace, is this good enough for now?

(`request_workspace_details` includes the full content of the original creation call... I'm a little concerned about this size of this response. Obviously, the file names are useful, but I'm not sure that the file content needs to be in this API response.)

### DIFF
--- a/context/app/static/js/components/workspaces/utils.js
+++ b/context/app/static/js/components/workspaces/utils.js
@@ -70,11 +70,7 @@ async function startJob({ workspaceId, workspacesEndpoint, workspacesToken, setM
 function getNotebookPath(workspace) {
   // TODO: Replace with current_workspace_details, when available.
   const { files } = workspace.workspace_details.request_workspace_details;
-  const nbFiles = files.filter(({ name }) => name.endsWith('.ipynb'));
-  if (nbFiles.length !== 1) {
-    return '';
-  }
-  return nbFiles[0].name;
+ const files.find({name} => name.endsWith('.ipynb'))?.name || '';
 }
 
 function mergeJobsIntoWorkspaces(jobs, workspaces) {
--- a/context/app/static/js/components/workspaces/utils.js
+++ b/context/app/static/js/components/workspaces/utils.js
@@ -67,6 +67,16 @@ async function startJob({ workspaceId, workspacesEndpoint, workspacesToken, setM
   setMessage(start.message);
 }
 
+function getNotebookPath(workspace) {
+  // TODO: Replace with current_workspace_details, when available.
+  const { files } = workspace.workspace_details.request_workspace_details;
+  const nbFiles = files.filter(({ name }) => name.endsWith('.ipynb'));
+  if (nbFiles.length !== 1) {
+    return '';
+  }
+  return nbFiles[0].name;
+}
+
 function mergeJobsIntoWorkspaces(jobs, workspaces) {
   const activeWorkspaces = workspaces.filter(({ status }) => ['active', 'idle'].includes(status));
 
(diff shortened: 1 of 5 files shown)


jpuerto-psc:
@mccalluc - question here, I haven't implemented the current_workspace_details b/c I wasn't sure how useful it would be. What are we thinking would be available from this? Just a tree of the current files/symlinks?

ngehlenborg:
Can you clarify what you mean by "current state" of the workspace?

mccalluc:
If the user renamed the notebook after creation, then the open-nb-on-launch would not work because the API doesn't tell us what the _current_ files in the workspace are, only what they were at creation.

> Just a tree of the current files/symlinks?

I think just a list of the current files/directories would be useful. Not sure if users can rename the symlinks? If they can, then yes, that would be useful, too. Flat list would be easier to work with than tree structure in JSON.

jpuerto-psc:
@mccalluc - sounds good, I will create an issue for this in workspaces. will let you know once it is available.

context/app/static/js/components/workspaces/utils.js
```diff
@@ -67,6 +67,16 @@ async function startJob({ workspaceId, workspacesEndpoint, workspacesToken, setM
...
 
+function getNotebookPath(workspace) {
+  // TODO: Replace with current_workspace_details, when available.
+  const { files } = workspace.workspace_details.request_workspace_details;
+  const nbFiles = files.filter(({ name }) => name.endsWith('.ipynb'));
+  if (nbFiles.length !== 1) {
+    return '';
+  }
+  return nbFiles[0].name;
+}
```
john-conroy, line 78: Could also use [Array.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find).

context/app/static/js/components/workspaces/utils.js
```diff
@@ -67,6 +67,16 @@ async function startJob({ workspaceId, workspacesEndpoint, workspacesToken, setM
...
 }
 
+function getNotebookPath(workspace) {
+  // TODO: Replace with current_workspace_details, when available.
+  const { files } = workspace.workspace_details.request_workspace_details;
+  const nbFiles = files.filter(({ name }) => name.endsWith('.ipynb'));
+  if (nbFiles.length !== 1) {
+    return '';
+  }
+  return nbFiles[0].name;
```
john-conroy, line 77:
```suggestion
 const files.find({name} => name.endsWith('.ipynb'))?.name || '';
```

john-conroy (approved):

mccalluc:
... waiting to see workspace come up... There may be an API problem right now.

mccalluc:
There's an infinite loop on workspace startup right now... not sure that this is problem, but move back to draft until fixed.

mccalluc:
The API problem that caused the infinite loop is fixed. (It was a mis-capitalized status code.) Jupyter starts up, and a notebook is loaded. Code was approved by John, so I'll merge.

PR #2878, opened 2022-08-31
stub out new preview

mccalluc:
I was mistaken when I said there wasn't machinery for private previews; It's the other way around -- new previews need to be explicitly added to the list.

I've taken `multimodal-molecular-imaging-data` as a template for the text: Up to you whether to stick to it or not. From here, I think it's in your court.


ngehlenborg (approved):

mccalluc:
I would prefer not to merge this until all required information has been provided. Keeping as draft. Will request re-review when more info is available.

PR #2881, opened 2022-08-31
Mccalluc/two step ws deletion

### LINKED ISSUE #2875: Two-stage workspace deletion UI
- For background: #2848

Our UI needs to change. I don’t think stop-jobs followed by delete polling (which we have now) is a good solution because it relies on the front-end to keep track of the is-being-deleted state… If the user leaves the page, then nothing remembers that the workspace is in the process of being deleted.

mccalluc:
- Fix #2875

Deleting workspaces is a 2 step process at the API level:
- Stop running jobs
- Delete the workspace

(Obviously, things would be easier on my end if there were an `rm -rf`, but there's not, and Juan's made a good argument that these are separate concerns.)

Currently, the portal sends the stop job requests, and then tries to delete the workspace... and tries again until it works. Juan has said that it could take some time to stop running jobs. The problem with this is that the only the UI knows that we're in the middle of a deletion: If you leave the page, the retries stop, and the workspace will not be deleted.

This PR gives the user two separate buttons for the two steps in the process. I'm not proposing this as the final, polished UI, but just something that can work until a new design is available.

@ngehlenborg -- Approval indicates that you accept that the UI will need to change to accommodate the API. If you can not approve, then you need to negotiate with Phil and Juan to provide a richer API.

@jpuerto-psc -- I want to check with you that the in-between states are being handled appropriately. I'll post more questions line-by-line in the PR.

@john-conroy -- You've asked to be on UI PRs... but I think this is more about hashing out what the UI should be... For now, I'm happy with the plain buttons; Feel free to remove yourself from the reviewers list.

### DIFF
--- a/context/app/static/js/components/workspaces/WorkspacesList.jsx
+++ b/context/app/static/js/components/workspaces/WorkspacesList.jsx
@@ -73,7 +73,7 @@ function WorkspacesList() {
                 Created {workspace.datetime_created.slice(0, 10)}
                 <button
                   type="submit"
-                  disabled={workspace.jobs.length > 0 /* TODO: And/or check workspace status? */}
+                  disabled={workspace.jobs.length > 0 || workspace.status === 'deleting'}
                   onClick={() => {
                     handleDelete(workspace.id);
                   }}
--- a/context/app/static/js/components/workspaces/WorkspacesList.jsx
+++ b/context/app/static/js/components/workspaces/WorkspacesList.jsx
@@ -71,7 +71,6 @@ function WorkspacesList() {
               <WorkspaceDetails workspace={workspace} />
               <div>
                 Created {workspace.datetime_created.slice(0, 10)}
-                {}
                 <button
                   type="submit"
                   disabled={workspace.jobs.length > 0 /* TODO: And/or check workspace status? */}
@@ -83,7 +82,7 @@ function WorkspacesList() {
                 </button>
                 <button
                   type="submit"
-                  disabled={workspace.jobs.length === 0 /* TODO: And/or check workspace status? */}
+                  disabled={workspace.jobs.length === 0}
                   onClick={() => {
                     handleStop(workspace.id);
                   }}
--- a/context/app/static/js/components/workspaces/WorkspacesList.jsx
+++ b/context/app/static/js/components/workspaces/WorkspacesList.jsx
@@ -9,7 +9,7 @@ import { SpacedSectionButtonRow } from 'js/shared-styles/sections/SectionButtonR
 import { PanelWrapper } from 'js/shared-styles/panels';
 
 import WorkspaceDetails from 'js/components/workspaces/WorkspaceDetails';
-import { createEmptyWorkspace, deleteWorkspace } from './utils';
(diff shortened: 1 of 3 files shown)


context/app/static/js/components/workspaces/WorkspacesList.jsx
```diff
@@ -69,12 +73,21 @@ function WorkspacesList() {
                 Created {workspace.datetime_created.slice(0, 10)}
                 <button
                   type="submit"
+                  disabled={workspace.jobs.length > 0 /* TODO: And/or check workspace status? */}
```
mccalluc, line 76: @jpuerto-psc -- This list of jobs is already filtered down; See below. This should probably also be disabled if the workspace is already in the process of deletion... Do any other cases come to mind? The API docs don't list the possible statuses... is it just `deleting`?

context/app/static/js/components/workspaces/utils.js
```diff
@@ -21,33 +21,36 @@ async function createEmptyWorkspace({ workspacesEndpoint, workspacesToken, works
...
+}
+
+async function stopJobs({ workspaceId, workspacesEndpoint, workspacesToken }) {
   const headers = getWorkspacesApiHeaders(workspacesToken);
   const jobsResponse = await fetch(`${workspacesEndpoint}/jobs`, { headers });
   const jobsResults = await jobsResponse.json();
   const { jobs } = jobsResults.data;
   jobs.forEach((job) => {
-    if (String(job.workspace_id) === workspaceId) {
+    if (String(job.workspace_id) === String(workspaceId)) {
```
mccalluc, line 50: Maybe the API was returning strings at some point? In any event, the types are mismatched right now, and this will be safest.

context/app/static/js/components/workspaces/utils.js
```diff
@@ -72,7 +75,10 @@ function mergeJobsIntoWorkspaces(jobs, workspaces) {
 
   const wsIdToJobs = {};
   jobs.forEach((job) => {
-    const { workspace_id } = job;
+    const { status, workspace_id } = job;
+    if (status === 'complete') {
```
mccalluc, line 79: Are there other job statuses that will not block workspace deletion? For example, if it errored out?

context/app/static/js/components/workspaces/WorkspacesList.jsx
jpuerto-psc: Correct, the workspace status can be one of the following:

- idle
- active
- deleting

I would, for now, check for job length as this will be more accurate than checking for a workspace's status. So far I have seen these two match up 100% but since they are updated asynchronously it is possible that they might be de-synced.

context/app/static/js/components/workspaces/utils.js
jpuerto-psc: failed would be the other job status here, pending and running both block workspace deletion.

ngehlenborg (approved):
I approve with the understanding that the additional button will be temporary. The API should be able to report the state to the UI.

mccalluc:
Juan is correct; This is necessary for the API that we have. Nils, if this is not acceptable to you, then
> you need to negotiate with Phil and Juan to provide a richer API.

But this negotiation should not happen here.

PR #2882, opened 2022-08-31
Bound Vitessce

### LINKED ISSUE #2409: When vitessce components are expanded, Vitessce grows, but the page doesn't
If Vitessce components are stretched vertically, or stacked, the Vitessce also expands so nothing overflows. This might or might not be the right behavior, but right now, when Vitessce expands vertically, it will overlap the footer, and that doesn't seem right. On the preview page:
Screen Shot 2022-02-07 at 12 19 40 PM

The text in the footer is completely covered: The white bars to the left and right remain.

(@ngehlenborg -- I think Ajay may have described this issue to you? Or something like it?)

mccalluc:
- Fix #2409

Use new vitessce functionality to prevent components being dragged off the bottom... well, it doesn't prevent dragging off the bottom, it just reverts the drag if it's off the bottom, but during the drag, we can still have issues:
Screen Shot 2022-09-06 at 11 43 16 AM

I took a couple minutes to explore whether wrapping it in 
```
<div style={{ maxHeight: vitessceFixedHeight, overflow: 'hidden' }}>
```
would change anything, but it doesn't... and that's probably a bad idea regardless.

I'd suggest merging this, unless you have a better idea, and leaving a note on the agenda to explain the limitations... but hoping not to do more work in this area.

### DIFF
--- a/CHANGELOG-bound-vitessce.md
+++ b/CHANGELOG-bound-vitessce.md
@@ -1,0 +1,1 @@
+- Use new Vitessce functionality so components can't be dragged off the bottom.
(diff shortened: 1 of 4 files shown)


john-conroy (commented):
I don't have an immediate idea, but I'll make note to look into it later. Is this something that should be deferred until after the kickoff meeting?

john-conroy:
Filed #2892.

PR #2885, opened 2022-09-01
Mccalluc/tidy mermaid

mccalluc:


john-conroy (approved):
Maybe darken the font for the green nodes? It's a little hard to read.

PR #2886, opened 2022-09-02
Update preview

keller-mark:
This PR updates the Vitessce view configs for the preview and adds the new preview to the dropdown menu. I called it "Liver Map" but this may need to change based on the information we receive from the PSU/Columbia group.


mccalluc (approved):
Thanks! Looks great to me...
- HuBMAP has emphasized that it's only for human tissue: Does including mouse tissue muddle this?
- On the 3D Human Liver, the "Data Set" pane is empty... is something missing?

mccalluc:
(Normally, we're leaving it to the original author to merge... but since this is just going to a feature branch, I'll go ahead and merge it myself: Hope that's not stepping on your toes.

Issue #2799, opened 2022-08-03
workspaces count down list


mccalluc:
[workspace designs](https://www.figma.com/file/NgBjvs0jRbDHhiyn5nicDm/Workspace?node-id=0%3A1)

MVP:

- [x] #2875
- [x] #2801
- [x] #2802
- [ ] Polish UI
- [ ] #2746
- [x] #2831
- [ ] Handle portal session expiration -- If session expires, should we close the workspace list?
- [x] #2846

MVP, but Chuck blocked:

- [ ] Resolve Widget rendering bug: https://github.com/vitessce/vitessce-python/issues/198
- [x] #2850
- [x] #2832
- [x] #2812
- [x] Get both teams on the same page WRT security: See https://github.com/hubmapconsortium/user_workspaces_server/issues/55
- [ ] Use workspace deletion when available: See https://github.com/hubmapconsortium/user_workspaces_server/issues/76
- [ ] #2633: Waiting on https://github.com/hubmapconsortium/user_workspaces_server/issues/28 and https://github.com/hubmapconsortium/gateway/issues/224

MVP, but save for last:

- [x] #2869
- [ ] Implement UI for workspace creation from dataset
- [ ] Add workspaces to the user profile menu

Nice to have:

- [x] #2883
- [ ] From workspaces list, support empty workspace creation
- [x] Update workspace page (polling?)
- [x] #2808
- [ ] ~~Create multiple notebooks in dataset workspace -- Get Nils sign off on this as alternative to forcing early choice on users.~~ Talked, and Nils is unpersuaded.
- [ ] Figure out what a collection notebook looks like
- [ ] Figure out what a search notebook looks like
- [ ] Figure out what a my-lists notebook looks like
- [ ] Notebooks for all entity types, so it's not conditional 
- [ ] scroll shadows... only if it's a fixed height div?
- [ ] Design and implement status icons
- [ ] Design and implement please wait page
- [x] #2848

Definitely out of scope:

- [ ] #2804

PR #2889, opened 2022-09-06
Mccalluc/workspaces message

mccalluc:
I believe that both sides have strong ideas about what the text should be, but I can not read your minds. This will need to be revisited as it becomes more functionally complete / leaves beta.

### DIFF
--- a/CHANGELOG-ws-message.md
+++ b/CHANGELOG-ws-message.md
@@ -1,0 +1,1 @@
+- Update workspaces messaging to better explain what's working, and what it should be used for.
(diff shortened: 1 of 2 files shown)


jpuerto-psc:
@pdblood - tagging for review

pdblood (approved):
My primary concern was that the previous text didn't match the current functionality, so as long as this text addresses that concern, I approve.

ngehlenborg (changes requested):
My preferred message is (@pdblood please agree or disagree): 

> HuBMAP Workspaces are in [beta testing](https://github.com/hubmapconsortium/portal-ui/issues/2799) and are provided as Jupyter Notebooks that may be instantiated with code blocks to perform basic and advanced operations on user-selected HuBMAP data. Instantiated workspaces can be launched on data sets, collections, and visualizations. Blank workspaces can be launched directly from this page.

pdblood:
@ngehlenborg your message looks good as long as "workspaces can be launched on data sets, collections, and visualizations". I thought that was not yet the case. If not, then please update that sentence so it describes the current state of workspaces. As additional functionality is added, the message can be updated to reflect that.

ngehlenborg (approved):

PR #2893, opened 2022-09-07
Mccalluc/fix home prop types

mccalluc:
`false` as a child was failing the check, but `null` is fine... Is this ok, or would it be better to change the proptype?


john-conroy (commented):
I'd just remove `children` from `Dropdown`'s prop-types.

PR #2902, opened 2022-09-11
John conroy/update redeploy

john-conroy:


yuanzhou (approved):

yuanzhou:
Recently we upgraded a new AWS VM with new OS and switched to a new user `hive` for the redeployment. Also due to the new VM, I've switched to reuse the old Elastic IPs. So you'll need to delete the old entries for DEV/TEST/STAGE in `.ssh/known_host`. I believe @john-conroy has verified.

yuanzhou:
PROD will have the same VM and OS upgrade later this week or early next week, and would not affect the redeployment since I handle the PROD release differently without using this script.

mccalluc (approved):

PR #2906, opened 2022-09-15
dev-search is 404 away from localhost

### LINKED ISSUE #2903: dev-search is just a tool for developers
If other folks start looking at information there before it's ready, things will get confused. Suggest adding an boolean envvar to determine whether it should it be displayed.

mccalluc:
- Fix #2903

This is partly a policy question, but dev-search has been a place to demonstrate that sometimes fields value _do not_ line up with their labels, and it's been very useful in that regard. If endusers start using it, then we need to ask what they think they're doing, and make the needed changes to support that usecase.

But, with this PR, it will just return 404 for any user not on localhost. (Devs can see the new behavior by looking at http://127.0.0.1:5001/dev-search.)

### DIFF
--- a/CHANGELOG-devsearch-is-for-devs.md
+++ b/CHANGELOG-devsearch-is-for-devs.md
@@ -1,0 +1,1 @@
+- To avoid confusion, /dev-search is now 404 if host is anything other than localhost.
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -1,3 +1,5 @@
+from urllib.parse import urlparse
+
 from flask import (render_template, current_app, abort,
                    session, request)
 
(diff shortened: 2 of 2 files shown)


john-conroy (commented):
I think some of the IEC folks use `dev-search`, would it be better to restrict it to some of the non-prod instances of the portal? I believe only devs are on their IP allow lists.

ngehlenborg (changes requested):
I think that locking this down to a few selected users is creating more issues than it solves. If there is concern about this being used incorrectly, we can include a message at the top of the page:

> The data and features provided on this page are for **debugging purposes only** and no guarantees are made for correctness or completeness of the data shown.

mccalluc:
Replaced by #2915

PR #2909, opened 2022-09-19
John conroy/fix url sync

### LINKED ISSUE #2899: Updated search route to state sometimes does not work
When visiting the new search page with state included in the url params roughly 80% of the time the state will update based on the url params (e.g. facets, page size), but sometimes the page view does not reflect the params.

- The correct ES request is being made, but the results returned from `useSearch` do not reflect the ultimate request. 
- Modifying the state (e.g. selecting a facet) will include both the new state and the previously 'missing' state.

john-conroy:
Fix #2899.


context/app/static/js/components/entity-search/searchkit-modifications/useSearchkitSDK.js
```diff
@@ -1,35 +1,48 @@
...
+          hits: {
+            size: variables.page.size,
+            from: variables.page.from,
+          },
+        });
+        setLoading(false);
+        setResponse(response);
+      } catch (error) {
+        if (error.name === 'AbortError') {
+          // Aborting throws an error and will prevent state from being updated.
```
mccalluc, line 36:
```suggestion
        if (error.name !== 'AbortError') {
          // Aborting throws an error and will prevent state from being updated.
          // Other errors should still be thrown
          throw error;
```
I'm not sure about the bigger picture, but right now it's a no-op.

mccalluc (commented):
Not sure about one `catch`, but otherwise looks good.

I can't reproduce the original bug on this branch, but there is some other weird behavior... not necessarily bugs, but does intersect with URL state.

- If you had a numeric range filter, and clear it from the snack bar, the UI in the sidebar doesn't update: It still shows a range being selected.
- http://localhost:5001/test-search/donors?sort=bad: No results, and then the url updates to `?size=18`... I think what I worry about here is the url update clobbering the original, bad, url that got us here.
- Should the size be in the URL at all?
- Conversely, tile mode is current captured in the URL, but it's not in the new search.

I don't think any of these should block this PR, but I wanted to note the behavior on the branch where I observed it.

context/app/static/js/components/entity-search/searchkit-modifications/useSearchkitSDK.js
john-conroy: It's a required no-op. You need a `catch` with `try`.

john-conroy:
Thanks. I should have been more clear, this wasn't meant to be a cure-all for the URL synchronization, but a fix for the specific bug described in the issue which was blocking other progress.

PR #2927, opened 2022-09-26
version index for ccf

mccalluc:
- Follow-up to #2924

For now, I assume it makes sense for all of these to be using the same version, even thought they are all accessing different resources under `v3`. This is mostly a FYI.

### DIFF
--- a/CHANGELOG-bump-ccf-path.md
+++ b/CHANGELOG-bump-ccf-path.md
@@ -1,0 +1,1 @@
+- The CCF uses a separate index, but it should also be bumped to v3.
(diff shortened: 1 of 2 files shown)


john-conroy (approved):

yuanzhou (approved):
Absolutely. @bherr2 according to Bill, you've tested the CCF against the `/v3/` search-api. We should be able to shut down the old v2 instance once the portal-ui migration completes and gets released to PROD.

bherr2:
Yep! :heavy_check_mark:

PR #2930, opened 2022-09-27
Limit organ-info datasets to HRA-mapped HuBMAP data only

bherr2:
See related issue in CCF-UI: <https://github.com/hubmapconsortium/ccf-ui/issues/948>

### DIFF
--- a/CHANGELOG-organ-info-data.md
+++ b/CHANGELOG-organ-info-data.md
@@ -1,0 +1,1 @@
+- Limit organ-info datasets to HRA-mapped HuBMAP data only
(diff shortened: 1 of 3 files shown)


mccalluc (changes requested):
I've fixed the linting error... but the component just doesn't seem to work: I've looked at a couple organs (small intestine, pancreas) and the ccf component just shows all zeros. Also: I'm not seeing any requests to `https://ccf-api.hubmapconsortium.org/v1/hubmap/rui_locations.jsonld` in the network tab.

I'd suggest trying out the component locally to confirm that it works for you. Maybe you've released a newer version we need to catch up to.

mccalluc:
Bruce will update version numbers in context/app/templates/special-pages/organ.html and Chuck will confirm that it works on localhost.

(The counts in the CCF viewer may not equal the total number of datasets if not all of them are RUI located.)

mccalluc (approved):
works now

PR #2940, opened 2022-09-29
(chore): upgrade vitessce to 1.2.2

ilan-gold:
This upgrade features both a speedup to the heatmap (testable on big CODEX datasets) and the upgrade of AnnData `io` capabilities to 0.8.0 of the AnnData package..

PR #2950, opened 2022-10-03
Mccalluc/vitessce 1.2.2

mccalluc:
Merge the latest changes from main, in particular
- #2920

into 
- #2940

With this, I am able to confirm that the heatmap speeds up: Now the spatial component is the last to load, and the overall load time is down by ~5s. 

I still suspect I just have a local environment problem, so this isn't strictly necessary, if John can get the PR to work without this.


ilan-gold:
I don't think this can hurt though?  Going to merge and see if CI passes then.

Issue #2939, opened 2022-09-29
incorrect language used for "protein abundance" (?) in molecular query UI


ngehlenborg:
The use of "gene expression" or "expression" is not correct for protein queries. Awaiting input from @mruffalo on how we should refer to this quantity.

<img width="625" alt="image" src="https://user-images.githubusercontent.com/1216518/193122602-28afb7db-1323-4bc2-80dc-3d539719f18f.png">

PR #2953, opened 2022-10-03
Mccalluc/cells message

### LINKED ISSUE #2938: Disclaimer at the top of cells
From Nils:

> Cells API next steps:
> - PIs agreed that a beta release to a targeted audience would be the most reasonable next step
> - **On our end, we need to update the UI with “disclaimers” at the top of the page**
> - We should explore if we can integrate a “walk through” feature like the one that we have on the standard dataset search page (but don’t need to implement this)
> - There will be a broader discussion about proper mapping of entities in the data (proteins/image channels, cell types) to appropriate ontologies or controlled vocabularies 

we need the exact text for the disclaimer

mccalluc:
- Fix #2938 
- filing as draft until questions resolved.

Screen Shot 2022-10-03 at 3 32 34 PM

@ngehlenborg 
- [ ] Is the display ok, or does tiffany need to be involved?
- [ ] Original plans called for integration with search. I feel like that is not required right now, and that a header link is more appropriate... confirm?
- [ ] Is limit to logged in users ok, or do you need an even smaller subset? If a smaller subset, can it be the same subset as the workspace beta users?
- [ ] There aren't actually any `cells` `bugs`: Is linking to all `cells` issues ok?
- [ ] Who will provide the second link?

@mruffalo (carried over from issue):
- [ ] Can you help with the language for the cut off? I am still not exactly sure what the meaning of that variable is and how one should use it.
- [ ] I assume the cut off has no unit. Is that correct?
- [ ] Also, what is the quantity that is being measured for the protein queries? (right now it says "expression", which we will need to fix (see #2939).

### DIFF
--- a/CHANGELOG-cells-text.md
+++ b/CHANGELOG-cells-text.md
@@ -1,0 +1,1 @@
+- Add a lot of text at the top of the cells page before beta release.
--- a/context/app/static/js/components/Header/OtherLinks/OtherLinks.jsx
+++ b/context/app/static/js/components/Header/OtherLinks/OtherLinks.jsx
@@ -1,9 +1,13 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 
+import { AppContext } from 'js/components/Providers';
 import DropdownLink from '../DropdownLink';
 
 function OtherLinks({ isIndented }) {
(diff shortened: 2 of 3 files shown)


john-conroy:
Closing in favor of #3001.

PR #2955, opened 2022-10-04
John conroy/sync search cols

john-conroy:
Adds search fields to URL state. Fixes bug when the default sort of `mapped_last_modified_timestamp` was not included as a field.

I'd recommend checking this out locally. I've opted to include the url param as `field`, but maybe column would be more fitting?

Follow ups in

- #2956 to reuse the available fields added to the search store in other hooks.
- #2962 to remove redundant `routeToState` and page size from url params.

More follow ups to come in addressing search facets/filters and fixing sort when the existing sort field is removed as a field.


mccalluc (approved):
Checked it out locally, and it seemed to work for me.

PR #2956, opened 2022-10-04
John conroy/reuse available fields

john-conroy:
Follow up to #2955 to reuse the available fields added to the search store in other hooks.


context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
```diff
@@ -25,21 +24,24 @@ function selectedItemsReducer(state, { type, payload }) {
...
+          [metadataFieldName]: metadataFieldConfig,
+        };
+        return draft;
+      });
+    },
+    {},
+  );
+  // Return only sample and donor metadata
+  const { 'Dataset Metadata': datasetMetadata, ...rest } = groupedFieldConfigs;
+  return rest;
```
mccalluc, line 44:
```suggestion
  delete groupedFieldConfigs.datasetMetadata;
  return groupedFieldConfigs;
```
Would this do the same thing? But whatever you prefer.

mccalluc (approved):
Yay for simplifications!

context/app/static/js/components/entity-search/sidebar/DatasetConfigureSearchTable/hooks.js
john-conroy: I opted for this at first, but it was erroring out for a reason I've now forgotten.

PR #2960, opened 2022-10-06
add nils

### LINKED ISSUE #2959: add @ngehlenborg to workspaces beta tester list

mccalluc:
@ngehlenborg , can you confirm that this is the address you see in the upper right when you log in, and then approve?
- Fix #2959
- Would you like Cells to also be limited to this set of users? Comment on #2953 if so.


john-conroy:
Closing in favor of #2972.

PR #2962, opened 2022-10-06
John conroy/remove size from route state

john-conroy:
Follow up to #2955 to remove page size from url state. One difference from the existing search page is that the results window is tracked in the route with `from`, for example `?from=18`, instead of `p` for page. I don't feel overly strongly towards either direction, but maybe you do?


context/app/static/js/components/entity-search/searchkit-modifications/useSearchkitSDK.js
```diff
@@ -25,7 +25,7 @@ const useSearchkitSDK = ({ config, variables, filters, defaultSort }) => {
         const response = await request.execute({
           facets: true,
           hits: {
-            size: variables.page.size,
+            size: 18,
```
mccalluc, line 28: This might be a good thing to pull from a higher-level configuration? But your call.

context/app/static/js/components/entity-search/searchkit-modifications/withSearchkitRouting.js
```diff
@@ -36,7 +36,6 @@ export const stateToRouteFn = (searchState, fields) => {
     sort: searchState.sortBy,
     filters: searchState.filters,
     fields,
-    size: parseInt(searchState.page?.size),
```
mccalluc, line 39: Have you been tracking the modifications made to the SK code we've forked? If that's not what you've been doing, probably not worth bothering about it now, but I'd have a slight preference for that.

mccalluc (approved):

Issue #2975, opened 2022-10-27
Organ pages not showing HRA info for some organs


bherr2:
Looks like paired organs like Kidney and Lung are not showing HRA info on organ pages ([example](https://portal.hubmapconsortium.org/organ/kidney)). We clearly have these. What information is needed to add these back in (they were there previously)?

bherr2:
I think [this commit](https://github.com/hubmapconsortium/portal-ui/commit/a0bd351f162e20b1138ebe8c8938541a7235b051#diff-3686587d94d2059037f25d932e7dc58a4898be42c83a8e55d56411a6bf6425aaR74-R95) might have triggered the error.

PR #2984, opened 2022-11-03
Updated the make_organ_dir.py logic to capture more HRA Reference Organs for display

bherr2:
This update fixes the logic for displaying HRA Reference Organs on the Organ pages. There are still a few unmapped 'organs' that I need to confer with @katyb on. Namely:

* Blood http://purl.obolibrary.org/obo/UBERON_0000178
* Bone Marrow http://purl.obolibrary.org/obo/UBERON_0002371
* Bronchus http://purl.obolibrary.org/obo/UBERON_0002185
* Lymph Vasculature http://purl.obolibrary.org/obo/UBERON_0004536
* Peripheral Nervous System http://purl.obolibrary.org/obo/UBERON_0000010
* Trachea http://purl.obolibrary.org/obo/UBERON_0003126

I'll submit a follow-up PR when I find a solution for these (if there is a change that is needed).


bherr2:
> Does this fix #2975?

Yep!

PR #3021, opened 2023-01-17
John conroy/fix organ page toc names

### LINKED ISSUE #3019: Match Organ Table of Contents to the Section Names
Current Organ Page Table of Contents:
Screen Shot 2023-01-10 at 9 05 26 AM

Table of content headers should match up to the section names. If the section name are too long, this a mockup from the detail pages of how the headers should behave. 
Screen Shot 2023-01-10 at 9 06 13 AM

john-conroy:
Fix #3019. 

Screen Shot 2023-01-17 at 2 40 20 PM

### DIFF
--- a/CHANGELOG-fix-organ-page-toc-names.md
+++ b/CHANGELOG-fix-organ-page-toc-names.md
@@ -1,0 +1,1 @@
+- Fix organ page table of contents to reflect actual section names.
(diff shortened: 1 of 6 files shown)


tsliaw:
More of a note to myself, but I'm not sure if users will be confused if the second line is actually just a new line, so I might need to a hyphen on the first line to indicate that there is a second line. Nothing blocking, but something for me to look out for in terms of user feedback.

tsliaw (approved):

PR #3095, opened 2023-05-03
john-conroy/start workspaces on workspace page HMP-7

john-conroy:
Workspaces are now started when launching on the workspace page or in actions where a workspace is created and launched immediately. Previously, workspaces were started on the loading page which caused workspaces you may have stopped to start again unexpectedly.

Introduces [swr](https://swr.vercel.app/docs/getting-started) to better manage external data. The workspaces data used to generate the workspaces list is revalidated when `mutate` is called and, among other situations, when a tab is refocused. Workspace lists across tabs should now reflect each workspace's true status.

### DIFF
--- a/context/app/static/js/components/workspaces/hooks.js
+++ b/context/app/static/js/components/workspaces/hooks.js
@@ -3,7 +3,7 @@ import useSWR from 'swr';
 
 import { AppContext } from 'js/components/Providers';
 
-import { mergeJobsIntoWorkspaces, createWorkspaceAndNotebook, deleteWorkspace, stopJobs } from './utils';
+import { mergeJobsIntoWorkspaces, createWorkspaceAndNotebook, deleteWorkspace, stopJobs, startJob } from './utils';
 
 async function fetchWorkspaces(workspacesEndpoint, workspacesToken) {
   const fetchOpts = {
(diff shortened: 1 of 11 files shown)


context/app/static/js/components/workspaces/hooks.js
```diff
@@ -1,61 +1,78 @@
...
+  const { data: workspacesList, mutate } = useSWR(
+    ['workspaces', workspacesToken],
+    ([, token]) => fetchWorkspaces(workspacesEndpoint, token),
+    { fallbackData: [] },
+  );
 
   async function handleDeleteWorkspace(workspaceId) {
     await deleteWorkspace({ workspaceId, workspacesEndpoint, workspacesToken });
-    getAndSetWorkspaces();
+    mutate();
```
NickAkhmetov, line 45: If we know how each of these actions will affect the local store of workspaces, we could optimistically update the local data and reflect the change a tiny bit faster, e.g. https://swr.vercel.app/docs/mutation#bound-mutate

The optimistic update would still be reconciled with the server data once it's received, but if there are any UI elements that are visible to the user when this fires, then the update would be reflected as soon as the awaited action before it returns instead of needing to wait for another round trip to the servers, and would appear that much more responsive to users.

NickAkhmetov (commented):
I wonder if we can make it even snappier with optimistic updates; otherwise, this looks good to me!

context/app/static/js/components/workspaces/hooks.js
john-conroy: Thanks! The workspaces UI is very much in beta and we expect it to change substantially in the coming months. I don't want to over-optimize while it's in this state of flux, but I hope to introduce optimistic updates in the future.

NickAkhmetov (approved):

PR #3099, opened 2023-05-24
NickAkhmetov/HMP-19 Improve perceived performance of vignette accordion on publication pages

NickAkhmetov:
This PR adjusts the behavior of the vitessce visualizations on the publication page:
- All accordion contents are initially rendered to prefetch vitessce conf
- Vitessce conf fetching is disabled when a conf already exists for that vignette
- Stale requests for vitessce conf data are aborted if effect re-triggers before it completes
- The `Vitessce` component is only mounted once the accordion it's in finishes expanding

In essence, this PR should make it so that the animation of switching from one vignette to another goes smoothly in more cases; however, this still has limitations. If a user tries to go to another vignette while the current one is still loading, the event does not immediately register; this is because Vitessce fires a lot of effects etc when it's first mounted which introduces a lot of actions into the main thread, and React 16 is not able to interrupt renders that are in progress or defer updates. In the future, once we update our React version, it should be able to interrupt the render cycle and execute the transition faster.

https://github.com/hubmapconsortium/portal-ui/assets/19957804/231db720-ca0b-4dc2-ba41-5a6160e012a7

In the demo above, we can see the initial animation from Vitessce filling in the available space; prior to these changes, React would attempt to do both this and the accordion animation simultaneously, hit lag from the Vitessce initialization, pause for a second, then display the expanded vignette with loading already in progress, skipping the accordion animation.

### DIFF
--- a/context/app/static/js/components/publications/PublicationVignette/PublicationVignette.jsx
+++ b/context/app/static/js/components/publications/PublicationVignette/PublicationVignette.jsx
@@ -6,7 +6,7 @@ import VisualizationWrapper from 'js/components/detailPage/visualization/Visuali
 
 import { fillUrls } from './utils';
 
-async function fetchVitessceConf({ assetsEndpoint, uuid, filePath, groupsToken, vignetteDirName }) {
+async function fetchVitessceConf({ assetsEndpoint, uuid, filePath, groupsToken, vignetteDirName, signal }) {
   const urlHandler = (url, isZarr) => {
     return `${url.replace('{{ base_url }}', `${assetsEndpoint}/${uuid}/data`)}${isZarr ? '' : `?token=${groupsToken}`}`;
   };
@@ -19,6 +19,7 @@ async function fetchVitessceConf({ assetsEndpoint, uuid, filePath, groupsToken, 
   const response = await fetch(
     `${assetsEndpoint}/${uuid}/vignettes/${vignetteDirName}/${filePath}?token=${groupsToken}`,
     {
+      signal,
       headers: {
         'Content-Type': 'application/json',
       },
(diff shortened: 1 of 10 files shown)


context/app/static/js/components/publications/PublicationVisualizationsSection/PublicationsVisualizationSection.jsx
```diff
@@ -1,45 +1,54 @@
...
+            </StyledAccordionDetails>
+          </Accordion>
+        );
+      })}
     </DetailPageSection>
   );
 }
 
-export default PublicationsVisualizationSection;
+export default memo(PublicationsVisualizationSection);
```
john-conroy, line 54:
```suggestion
export default memo(PublicationsVisualizationSection);
```
How does this impact the render performance in this case?

john-conroy (commented):
Thanks! A few comments I'm curious about. I am somewhat weary of blanket optimization, adding `useMemo`, `memo` and `useCallback` in situations where it's not wholly necessary, but seems to make sense here.

context/app/static/js/components/publications/PublicationVignette/PublicationVignette.jsx
```diff
@@ -33,22 +34,46 @@ async function fetchVitessceConf({ assetsEndpoint, uuid, filePath, groupsToken,
...
+  // Workaround to make the visualization render only after the accordion section has been expanded while
+  // still letting the prerequisites for the visualizations prefetch
+  const [hasBeenMounted, setHasBeenMounted] = useState(false);
   useEffect(() => {
+    if (mounted) {
+      setHasBeenMounted(true);
+    }
+  }, [mounted]);
+
+  useEffect(() => {
```
john-conroy, line 51: Also, I think I like the pattern of moving all hooks into a single hook per component. How do you feel about this?

context/app/static/js/components/publications/PublicationVisualizationsSection/PublicationsVisualizationSection.jsx
NickAkhmetov: This was a relic of a previous attempt that I forgot to revert before committing 😅 dropping it now

context/app/static/js/components/publications/PublicationVignette/PublicationVignette.jsx
NickAkhmetov: That actually sounds great to me, as long as we keep the hooks logically organized and focused on one purpose per hook. e.g, in this case, I'll turn the logic in this component into a `useVitessceConf` hook and a `useHasBeenMounted` hook, as opposed to trying to squeeze both sets of logic into one hook.

PR #3101, opened 2023-05-26
NickAkhmetov / HMP-97 Add Jupyter Notebook download button to expanded visualization

NickAkhmetov:
This PR adds the Jupyter Notebook download button to the expanded visualization view.

- vizNotebookId / setVizNotebookId states have been added to the Visualization store
- When a visualization with a notebook mounts, the vizNotebookId is set to the uuid of the visualization; if the visualization does not have a notebook, the value is set to null.
- The `EntityHeaderContent` component accesses the `vizNotebookId` value in the store and conditionally inserts the download button if it's not null.

**Screenshot** with tab-focused button visible at the top right:
image

### DIFF
--- a/CHANGELOG-hmp-97-add-download-notebook-to-expanded-view.md
+++ b/CHANGELOG-hmp-97-add-download-notebook-to-expanded-view.md
@@ -1,1 +1,1 @@
-- Add Jupyter Notebook download button to expanded visualization view
+- Add Jupyter Notebook download button to expanded visualization view.
--- a/CHANGELOG-hmp-97-add-download-notebook-to-expanded-view.md
+++ b/CHANGELOG-hmp-97-add-download-notebook-to-expanded-view.md
@@ -1,0 +1,1 @@
+- Add Jupyter Notebook download button to expanded visualization view
--- a/context/app/static/js/components/detailPage/visualization/Visualization/Visualization.jsx
+++ b/context/app/static/js/components/detailPage/visualization/Visualization/Visualization.jsx
@@ -75,13 +75,12 @@ function Visualization({ vitData, uuid, hasNotebook, shouldDisplayHeader, should
   } = useVisualizationStore(visualizationStoreSelector);
 
   // Propagate UUID to the store if there is a notebook so we can display the download button when the visualization is expanded
+  // Reruns every time vizIsFullscreen changes to ensure the proper notebook's UUID is used
   useEffect(() => {
     if (hasNotebook) {
       setVizNotebookId(uuid);
-    } else {
-      setVizNotebookId(null);
     }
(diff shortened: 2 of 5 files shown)


CHANGELOG-hmp-97-add-download-notebook-to-expanded-view.md
```diff
@@ -0,0 +1 @@
+- Add Jupyter Notebook download button to expanded visualization view
```
john-conroy, line 1:
```suggestion
- Add Jupyter Notebook download button to expanded visualization view.
```

john-conroy: Sorry to be annoying!

context/app/static/js/components/detailPage/visualization/Visualization/Visualization.jsx
```diff
@@ -70,8 +71,18 @@ function Visualization({ vitData, uuid, hasNotebook, shouldDisplayHeader, should
...
     setOnCopyUrlSnackbarOpen,
+    setVizNotebookId,
   } = useVisualizationStore(visualizationStoreSelector);
 
+  // Propagate UUID to the store if there is a notebook so we can display the download button when the visualization is expanded
+  useEffect(() => {
+    if (hasNotebook) {
+      setVizNotebookId(uuid);
+    } else {
+      setVizNotebookId(null);
```
john-conroy, line 82: So I can better understand this probably is not necessary now, but would be necessary once client state isn't refreshed on navigation? Not suggesting we remove it.

john-conroy (approved):

CHANGELOG-hmp-97-add-download-notebook-to-expanded-view.md
NickAkhmetov: Not annoying at all, I appreciate the correction when I miss details like this 👍

context/app/static/js/components/detailPage/visualization/Visualization/Visualization.jsx
NickAkhmetov: I was thinking of this as a way to clean up the state in case we ever have cases where there's multiple visualizations on the same page, but this probably wouldn't even accomplish that on second thought. I'll drop this line to avoid the unnecessary state update

NickAkhmetov: I've removed the null setter in lieu of rerunning it every time a visualization's expansion status changes - this should make it so that if we eventually have a page with multiple vitessce instances that have notebooks (publications currently don't have notebooks) the proper visualization's ID is used.

Issue #3088, opened 2023-04-19
Pass flask data to downstream components with context


john-conroy:
The prop drilling from flask to app to child components is cluttered and inconsistent. Everything from flask should be passed by context.

PR #3104, opened 2023-05-30
John conroy/add file browser to publication page HMP-9

john-conroy:
Adds files section to publication pages. While we do have pipeline support for publications with files, no publications with files exist now.

@tsliaw The mockup doesn't include Globus links, but the typical files section includes Globus links for both the main entity and 'support' entity.

Screen Shot 2023-05-30 at 11 23 06 AM

### DIFF
--- a/context/app/static/js/components/Routes/Routes.jsx
+++ b/context/app/static/js/components/Routes/Routes.jsx
@@ -101,7 +101,7 @@ function Routes({ flaskData }) {
   if (urlPath.startsWith('/browse/publication/')) {
     return (
       <Route>
-        <Publication publication={entity} vignette_json={vignette_json} />
+        <Publication publication={entity} vignette_json={vignette_json} visLiftedUUID={vis_lifted_uuid} />
       </Route>
     );
   }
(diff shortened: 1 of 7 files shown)


tsliaw:
@john-conroy Does publications have Globus links?

context/app/static/js/components/Routes/Routes.jsx
```diff
@@ -101,7 +101,7 @@ function Routes({ flaskData }) {
   if (urlPath.startsWith('/browse/publication/')) {
     return (
       <Route>
-        <Publication publication={entity} vignette_json={vignette_json} />
+        <Publication publication={entity} vignette_json={vignette_json} visLiftedUUID={vis_lifted_uuid} />
```
NickAkhmetov, line 104: Should we consider adding a context to avoid needing to pass this prop down? 
- The `Publication` component itself doesn't use this value except passing it down to `Files`
- The `VisLIftedUUID` value does not change on the page, and would cause a rerender of the whole `Publication` even if it did;
- Creating a context and a corresponding `use` hook for it would help keep the variable visible only where it's used

lchoy: Issue regarding passing flask data by context https://github.com/hubmapconsortium/portal-ui/issues/3088

lchoy (approved):

PR #3106, opened 2023-05-30
NickAkhmetov / HMP-100 Improve vitessce conf generation errors

NickAkhmetov:
This PR

- Fixes support page entity header so the application doesn't crash to a white screen when scrolling down on Support entities' pages
- Adds handling for cases where the image pyramid descendants are missing file metadata
- Add more informative error messages on vitessce conf gen failures

image

### DIFF
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -177,18 +177,17 @@ class ApiClient():
             # TODO: Entity structure will change in the future to be consistent
             # about "files". Bill confirms that when the new structure comes in
             # there will be a period of backward compatibility to allow us to migrate.
-            
 
             metadata = derived_entity.get('metadata', {})
 
             if (metadata.get('files')):
                 derived_entity['files'] = metadata.get('files', [])
                 vitessce_conf = self.get_vitessce_conf_cells_and_lifted_uuid(
-                    derived_entity, marker=marker, wrap_error=wrap_error
-                ).vitessce_conf
-                vis_lifted_uuid = derived_entity['uuid'] 
-            else: # no files
-                vitessce_conf = _create_vitessce_error('No files found in lifted image pyramid descendants metadata.')
+                    derived_entity, marker=marker, wrap_error=wrap_error).vitessce_conf
+                vis_lifted_uuid = derived_entity['uuid']
+            else:  # no files
+                vitessce_conf = _create_vitessce_error(
+                    'No files found in lifted image pyramid descendants metadata.')
 
         elif not entity.get('files') or not entity.get('data_types'):
             vitessce_conf = ConfCells(None, None)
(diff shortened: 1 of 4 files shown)


NickAkhmetov:
@tsliaw Could you review the phrasing in the related error message to make sure it works for you? The current error is `Error while generating the Vitessce configuration: No files found in lifted image pyramid descendants metadata.`

NickAkhmetov:
I've also set the default suggested formatter in VS Code to `autopep8` to try to avoid PEP errors from stopping the build going forward 🙂

tsliaw:
@NickAkhmetov Message looks good!

context/app/static/js/components/detailPage/entityHeader/EntityHeaderContent/EntityHeaderContent.jsx
```diff
@@ -54,9 +55,11 @@ function EntityHeaderContent({ assayMetadata, shouldDisplayHeader, vizIsFullscre
...
               <StyledSvgIcon component={entityIconMap[entity_type]} />
               <EntityHeaderItem text={hubmap_id} />
-              {Object.entries(entityToFieldsMap[entity_type]).map(([label, fn]) => (
-                <EntityHeaderItem text={fn(assayMetadata) || `undefined ${label}`} key={label} />
-              ))}
+              {entity_type in entityToFieldsMap
+                ? Object.entries(entityToFieldsMap[entity_type]).map(([label, fn]) => (
+                    <EntityHeaderItem text={fn(assayMetadata) || `undefined ${label}`} key={label} />
+                  ))
+                : `No fields defined for entity type ${entity_type}`}
```
john-conroy, line 62: I would prefer not to show anything for support entities.

context/app/api/client.py
```diff
@@ -177,11 +177,17 @@ def get_vitessce_conf_cells_and_lifted_uuid(self, entity, marker=None, wrap_erro
...
+            metadata = derived_entity.get('metadata', {})
+
+            if (metadata.get('files')):
+                derived_entity['files'] = metadata.get('files', [])
+                vitessce_conf = self.get_vitessce_conf_cells_and_lifted_uuid(
+                    derived_entity, marker=marker, wrap_error=wrap_error).vitessce_conf
+                vis_lifted_uuid = derived_entity['uuid']
+            else:  # no files
+                vitessce_conf = _create_vitessce_error(
+                    'No files found in lifted image pyramid descendants metadata.')
```
lchoy, line 190: Do end users know what "lifted image pyramid descendants" are? 

If we're showing an error message to the user, should this also be logged? Perhaps at Info level, not Error.

context/app/static/js/components/detailPage/entityHeader/EntityHeaderContent/EntityHeaderContent.jsx
NickAkhmetov: I agree! I've added an empty object as the value for the `Support` key in `entityToFieldsMap` above on line 32. Since `Object.keys({})` returns an empty array, no items are displayed in the header for support entities. I also added the `else` case here to be type safe, since `Object.keys(undefined)` throws `TypeError: can't convert undefined to object`, which is currently causing Support pages on prod to crash to a white screen if you scroll down, e.g.: https://portal.hubmapconsortium.org/browse/support/d04a1abdf3546caf1896f0d9923027db

I'll add a comment above the `Support: {}` to indicate we are explicitly not showing anything for this entity.

context/app/api/client.py
tsliaw: > 'Related entity {uuid} is missing files information in its metadata'

This is a lot better message for end users. (delete the 's' after 'files' for grammar, but optional since I know you have the 's' for the `files` key), but my only question is for people debugging the issue, is there a way for them to know from that message (or any other similar fast ways) that the `lifted_image_pyramid`is missing metadata?

NickAkhmetov: It makes sense to conform to proper grammar so things don't look wrong to end users - I get put off when I see incorrect grammar on a site. We should also be able to further increase specificity.

Fresh adjustment: `'Related image entity {uuid} is missing information (no "files" key found in its metadata).'`

context/app/static/js/components/detailPage/entityHeader/EntityHeaderContent/EntityHeaderContent.jsx
john-conroy: You could also return null or use a short-circuit, right?

john-conroy (approved):

PR #3109, opened 2023-06-01
John conroy/upgrade ga4 HMP-112

john-conroy:
Upgrades Google Analytics from Universal Analytics to GA4. @tsliaw there are new features in GA4 to evaluate and additional configuration needed in the console.


NickAkhmetov (approved):
Looks good to me 👍 Seems to match the instructions in the react-ga4 readme.

PR #3112, opened 2023-06-05
John conroy/fix publication duplication HMP-140

john-conroy:
Fix duplication of publications as both published and pre-print.

We should fetch the publications data differently instead of making one big request. Ideally we'd make an aggregations request to populate the counts published/pre-print tabs, and then separate scrolling requests for each tab. We really only expect a handful of publications in the short to medium term so I've filed an issue and will table it for now.

### DIFF
--- a/CHANGELOG-fix-publication-duplication.md
+++ b/CHANGELOG-fix-publication-duplication.md
@@ -1,0 +1,1 @@
+- Fix duplication of publications as both published and pre-print.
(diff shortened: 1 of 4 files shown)


NickAkhmetov (approved):
Looks good to me! It's smart to add unit tests to prevent this from regressing.

I agree that the two lists can/should be fetched separately in the future. I could picture this being done with the [parallel routing](https://nextjs.org/docs/app/building-your-application/routing/parallel-routes) feature of the Next.js app router once we migrate; in the meantime, we could get a similar effect with swr and prefetching preprints only when the appropriate tab is hovered.

For ease of future reference, allow me to link to the issue you made so a link is made: https://hms-dbmi.atlassian.net/browse/HMP-138

PR #3114, opened 2023-06-05
Nickakhmetov/HMP-103 Cypress tests for publication pages

NickAkhmetov:
This PR:

- updates Cypress from v6 to v12
- Adds test cases for the publications landing page
- Adds test cases for individual publication pages (using a real publication page from the dev/test environment)

Additional changes I made while implementing the above:
- added `data-testid` props to various components to aid in testing
- fixed publication `reduce` function to properly split published/preprint entities
  - My change ~~conflicts with #3112 but can be~~ **has been reverted in favor of the work in that branch**
- added safe fallbacks for `undefined` contributors lists 
- Changed development source map type from `inline-source-map` to `eval-source-map` to get the proper component name and line number to be displayed in error stack traces
  - Per [webpack devtool config docs](https://webpack.js.org/configuration/devtool/), this is the "recommended choice for development builds with high quality SourceMaps" and does not offer any downsides relative to `inline-source-map`
  - This will not affect production, which uses an entirely separate webpack config.

There are still cases to add for:
- Authors section (current publications on dev environment are missing contributors information, similar to the issue we were having on prod last week I think)
- Provenance section (prov API is throwing 404's during local dev)
- Files section (seeing a `Files are not available through the Globus Research Management system` message for all available publications on dev)

image
image


end-to-end/cypress/e2e/portal/publication-page.cy.js
```diff
@@ -0,0 +1,110 @@
...
+      cy.findAllByTestId("donor-row").should("have.length", 1);
+
+      // Samples tab needs to be clicked to activate
+      cy.findByTestId("samples-tab")
+        .should("exist")
+        .and("contain", "Samples")
+        .and("contain", "(3)")
+        .click();
+      cy.wait(100);
+      cy.findAllByTestId("sample-row").should("have.length", 3);
```
john-conroy, line 44: `find` will retry if the element isn't present in the DOM with a default timeout of [4000ms](https://docs.cypress.io/guides/references/configuration#Timeouts). Is `wait` necessary here? If so you can extend the timeout of `find`.

end-to-end/cypress/e2e/portal/publications.cy.js
```diff
@@ -0,0 +1,42 @@
...
+    it("has clickable publication links", () => {
+      const publishedTab = cy.findByTestId("publication-tab-published");
+      const preprintTab = cy.findByTestId("publication-tab-preprint");
+      publishedTab.should("exist").then(($pubTab) => {
+        if ($pubTab.text().includes("(0)")) {
+          preprintTab.click();
+          cy.wait(1000);
+        }
+      });
+      cy.findAllByTestId("panel-title").should("exist");
```
john-conroy, line 39: Maybe use `.should('have.attr', 'href')`?

end-to-end/cypress/e2e/portal/publication-page.cy.js
```diff
@@ -0,0 +1,110 @@
+// This test was written against a existing publication in the dev/test env and will need to be updated if that publication is deleted
+
+const publicationId = "cd03fa15322492f4173cdbdc30c74a62";
```
john-conroy, line 3: The docker container uses the endpoints defined in `example-app.conf`. We should request that Mark's publication be added to `test` as it will be more stable and better reflect what we expect a publication to be.

john-conroy (commented):
Thanks! A few comments.

end-to-end/cypress/e2e/portal/publication-page.cy.js
NickAkhmetov: When I was running this locally, any tests which relied on the page state updating from clicking on a tab failed unless I explicitly added a short wait period. I tried to work around this by playing with the timeout, by chaining with `.then`, and trying the `cypress-real-events` library which previously solved a similar issue for me. Unfortunately, the only approach which worked for me at the time was adding this `.wait`.

end-to-end/cypress/e2e/portal/publications.cy.js
NickAkhmetov: I'll add it as a chain, good point:
```suggestion
      cy.findAllByTestId("panel-title").should("exist").and.should("have.attr", "href)
```

end-to-end/cypress/e2e/portal/publication-page.cy.js
NickAkhmetov: Agreed, that would be ideal; it would allow us to complete the rest of the test cases for the publication page as well - who can we discuss this with to get it done?

john-conroy:
We should probably add https://www.npmjs.com/package/eslint-plugin-cypress if you haven't done it here.

end-to-end/cypress/e2e/portal/publication-page.cy.js
john-conroy: I've reached out in the `feature-publicationpages` channel.

kevinyooky (approved):
Not sure if I am the best person to provide input on this, but from what I saw, it looks good to me.

NickAkhmetov:
As discussed with @john-conroy offline, in order to keep this PR from getting staler, we will merge the work that has been done so far; this will also confirm whether the updated Cypress config is fully functional. 

Once the Vitessce publication is up on TEST, I will continue work on this ticket to add more publication page test cases in another PR. The ESLint plugin will also be added to our lint configuration as part of HMP-194.

PR #3115, opened 2023-06-05
Nickakhmetov/HMP-113 Fix provenance table sample calculation logic

NickAkhmetov:
The previous `isSampleSibling` logic was using the `Sample` entity type string as the source array for `sample_category` comparison. This led to a typerror being thrown when there were more samples than the length of this string.

This PR fixes the logic to compare the current sample's `sample_category` to the previous one by using the sorted `items` from the `.map` method signature to access the previous item.

To verify fix:
- attempt to access this dataset  on prod: https://portal.hubmapconsortium.org/browse/dataset/d15a20b5c07c638e758a25bb320091e1
- attempt to access this dataset on this branch: http://localhost:5001/browse/dataset/d15a20b5c07c638e758a25bb320091e1

### DIFF
--- a/CHANGELOG-hmp-113-fix-provenance-table.md
+++ b/CHANGELOG-hmp-113-fix-provenance-table.md
@@ -1,0 +1,1 @@
+- Fix provenance table sample sorting calculation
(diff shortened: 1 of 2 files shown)


john-conroy (approved):
Nice catch, thanks!

PR #3116, opened 2023-06-06
Nickakhmetov/HMP-144 Upgrade React and other front-end dependencies [DEFUNCT/REFERENCE ONLY] 

NickAkhmetov:
This got a bit out of hand; apologies for the PR size 😅 

image

I initially only updated the React/React-DOM libraries, but once I started making the adjustments necessary to make things work again, it became clear that updating all the dependencies as far as they could go with our current node settings was the easiest way to eliminate the errors that were being thrown by various dependencies experiencing version mismatches. 

These changes also involved updating our linting libraries to take into account the new API's/changes in behavior that came with React 17/18. After updating our existing lint rule sets appropriately and setting up VS code to enforce them by default, I went through the 200+ lint errors and revised them appropriately. 

Recurring patterns which were changed to comply with new rulesets' default recommendations:
- Named components are now defined as `function Component() { ... }` notation instead of `const Component = () => { ... }` to automatically set the `displayName`
- Functions that are passed as props to other components are now wrapped in `useCallback` to maintain referential equality when their dependencies are unchanged
  - Even if our internal code doesn't use `PureComponent`s, this is still important to ensure that effect execution is only triggered as needed and any pure components in imported code work as expected.
- Context values are wrapped in a `useMemo` for the same reason as above; context values in tests are defined as constants outside of components.
- Unit tests' uses of `toBeNull` have been changed to `not.toBeInDocument` and `getByText` calls derived from the `render` function have been replaced with `screen.getByText`
- Components that were defined inside of other components are now defined at the top level and get the appropriate data passed in via props instead.
  - Same concept as with `useCallback` but even more severe - each time the wrapping component rerendered, a brand new version of the underlying React component was generated; without equality on the `$$typeof` prop, the component needed to be unmounted and recreated from scratch with each update.

Remaining TODOs:
- [x] Storybook - updated and applied relevant codemods
- [ ] Jest
  - [x] Tests failing due to `testEnviroment` config not being set to `jsdom` / missing `jest-environment-jsdom` dev dependency - fixed
  - [x] Tests failing due to `InterVariable` font handling - changed to import the font in `index.jsx` and reference by name; also updated to use actively maintained `fontsource` project repo instead of deprecated `typeface` repo. 
  - [ ] Tests failing due to [MUI v4 incompatibility](https://github.com/mui/material-ui/issues/30953), now working on v5 upgrade to resolve Jest test issues

Files with manual changes:
- context/app/static/js/components/home/ImageCarouselContainer/style.js - The homepage carousel broke after the updates to MUI, the `slider` container had either width 0 (when the CTA was set to flex shrink 0) or ~10,000 px wide when the CTA container was allowed to shrink. I extended the styles for the flex container to constrain the carousel to 2/3 of the total available space for this breakpoint in order to restore the previous behavior.


NickAkhmetov:
Per our discussion on today's strategy call, this branch will not be merged, and will be used as a reference for the piece by piece upgrades noted in this document: https://hms-dbmi.atlassian.net/wiki/spaces/GL/pages/3009871888/Tech+Updates?atlOrigin=eyJpIjoiYzU3MzFkNDY1OGJjNDQzNThlMmU0NGM2MjcxNmQ4MGYiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ

PR #3118, opened 2023-06-08
Kevin yoo/refactor app context

kevinyooky:
Went through all files using App Context and refactored to use UseAppFactor from Context.js


NickAkhmetov (approved):
Looks good to me, just the one issue from duplicate imports:

PR #3120, opened 2023-06-09
John conroy/remove origin sample dependencies HMP-169

john-conroy:
We need to move from `origin_sample`, an object, to `origin_samples` which is an array of objects so we can remove `origin_sample` from the ES index and unblock publications. This PR and the follow-up PR #3122 are meant to change to using `origin_samples` quickly with little refactoring. This PR also moves from `source_sample` to `source_samples`, but `source_sample` was already an array of objects. 

Ideally we can get this in, perform some extensive QA, and address refactoring/updates in future PRs. In some areas of our UI (nested ES docs, icons), it is less straightforward to handle multiple `origin_samples`. I've used the first sample in the array in those cases and marked those areas with TODOs and have filed issues in an [epic](https://hms-dbmi.atlassian.net/browse/HMP-163).

A little background.

We use `origin_sample` to determine the organ for a sample or dataset. As of now all but a few of our datasets have only a single origin sample. The datasets with multiple `origin_samples` can be viewed [here](https://portal.hubmapconsortium.org/search?mapped_data_types[0]=LC-MS%20Bottom%20Up&origin_sample.mapped_organ[0]=Small%20Intestine&origin_sample.mapped_organ[1]=Large%20Intestine&entity_type[0]=Dataset). We expect more datasets with multiple `origin_samples` when we begin to ingest multi-assay datasets.

A few questions that have arisen.

- The existing datasets with multiple `origin_samples` with different organs still only have a single donor. Should we expect to handle datasets with multiple donors in the future?
- Will the `origin_samples` always be organs?
- How can we determine which datasets will have multiple organs? Assay type?

Once merged, I'll make a release and get these changes on to dev/test. Joe will then remove `origin_sample` and `source_sample` from the index and we can perform QA. 

A few areas to pay attention to when performing QA.

- The datasets table in the cells UI results.
- Both the dataset and samples search pages on the primary routes and `/test-search`. Check both the table and tiles views.
- Organs landing page, `/organ`, and organ detail pages. The aggregations counts will change slightly for small and large intestines, but otherwise the pages should remain the same. Check that once you follow the links to the search page from the organ page the correct filters are shown.
- Dataset and sample detail pages.

### DIFF
--- a/CHANGELOG-remove-origin-sample-dependencies.md
+++ b/CHANGELOG-remove-origin-sample-dependencies.md
@@ -1,0 +1,1 @@
+- Use `origin_samples` field instead of `origin_sample` throughout.
--- a/context/app/routes_api.py
+++ b/context/app/routes_api.py
@@ -110,7 +110,7 @@ def _get_entities(entity_type, constraints={}, uuids=None):
         'created_by_user_email',
     ]
     if entity_type in ['samples', 'datasets']:
-        extra_fields += ['donor.hubmap_id', 'origin_sample.mapped_organ']
+        extra_fields += ['donor.hubmap_id', 'origin_samples_unique_mapped_organs']
     if entity_type in ['samples']:
         extra_fields += ['sample_category']
     entities = client.get_entities(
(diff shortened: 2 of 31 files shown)


NickAkhmetov (approved):
LGTM (though this review will get dismissed when the other branch gets merged in 😄)

PR #3121, opened 2023-06-09
Nickakhmetov/HMP-154 Upgrade to Webpack 5/Storybook 7

NickAkhmetov:
This PR
- Upgrades webpack 4 -> 5
- Replaces babel loader with swc-loader and url-loader with asset/inline
- Updates storybook 6 -> 7 for compatibility
- Drop-in replaces jest with @swc/jest
- Fixes compatibility errors from webpack updates (e.g. destructuring package.json is no longer allowed)
- Fixes the chart-related storybook stories which were previously failing
- Removes the large eslint-ignored block in the entry `index.jsx` and adds global variable definitions to satisfy eslint instead
- Adjusts source map settings to generate consistent (cheap) sourcemaps in dev/prod

This PR looks big, but almost all of the line changes are in the `package-lock.json` file.

Note I have not removed babel-related packages yet, as our current eslint config relies on those packages; those will be taken care of when I reach that step of the tech upgrade.


kevinyooky (dismissed):
Went through the files, and there wasn't anything that stood out. Looks good to me!

NickAkhmetov:
(resolved merge conflict)

context/package.json
```diff
@@ -96,7 +101,7 @@
     "babel-plugin-styled-components": "^1.11.1",
     "chromatic": "^5.9.2",
     "clean-webpack-plugin": "^3.0.0",
```
john-conroy, line 103: Can we uninstall `clean-webpack-plugin` after the changes to our webpack config files?

context/.storybook/main.js
```diff
@@ -1,26 +1,28 @@
...
-    '../app/static/js/components/**/*.stories.js'
-  ],
-  "addons": [
-    "@storybook/addon-links",
-    "@storybook/addon-essentials"
-  ],
-  "webpackFinal": async (config) => {
+  "stories": ['../app/static/js/shared-styles/**/*.stories.js', '../app/static/js/components/**/*.stories.js'],
+  "addons": ["@storybook/addon-links", "@storybook/addon-essentials"],
+  "webpackFinal": async config => {
```
john-conroy, line 8: Like we discussed on the call, seems like we're getting a lot of formatting changes here and elsewhere throughout the PR. Do you know what's causing this?

context/.swcrc
```diff
@@ -0,0 +1,28 @@
...
+            "numericSeparator": false,
+            "classPrivateProperty": false,
+            "privateMethod": false,
+            "classProperty": false,
+            "functionBind": false,
+            "decorators": false,
+            "decoratorsBeforeExport": false
+        },
+        "target": "es2019",
+        "loose": false,
```
john-conroy, line 15: Is this specific config coming from somewhere?

john-conroy (dismissed):
Thanks! A few comments, but looks good to me.

context/.storybook/main.js
NickAkhmetov: To my knowledge, these changes were automatically performed by Storybook's script, while the automatic changes in this commit: https://github.com/hubmapconsortium/portal-ui/pull/3121/commits/e4a74f8589bd60309053c5ebfacf915336ee9cff were performed because they started throwing errors after the webpack 5 update. I haven't set up anything new in terms of lint config, so any applied rules should be coming from the existing `.eslintrc.yml` (and by extension, `.prettierrc.yml`).

context/.swcrc
NickAkhmetov: I adapted it from the default config here: https://swc.rs/docs/configuration/swcrc

I needed to remove the `$schema` key because the provided schema did not accept the `experimental` key, which is necessary for styled-components plugin support.

NickAkhmetov:
Build is passing here but failing to build the docker image locally with `npm run build` resulting in `Killed` while building. Based on research, this seems to be caused by some sort of memory issue; will need to keep digging. May potentially be related to the two polyfills I added in order to support `xml2js`'s dependencies.

NickAkhmetov:
I've confirmed that the cypress tests are passing for me locally again; I was probably experiencing some weirdness from hopping between environments/branches. 

The changes I've made since the last review were 
- promoting the new polyfills from dev deps to main deps
- removed the deprecated clean-webpack-plugin
- restored the build:analyze script by switching from `webpack-visualizer-plugin` to its fork `webpack-visualizer-plugin2` since the original plugin is no longer maintained and not compatible with webpack 5

john-conroy (approved):
Great! 

I believe `xml2js` was introduced in @kevinyooky's work with the biomarkers and only needed to work with one of the external APIs. Given that we're hoping to get all biomarker information from `ubkg`, we should be keen to remember to remove these polyfills once `xml2js` is removed.

NickAkhmetov:
> we should be keen to remember to remove these polyfills once `xml2js` is removed.

Agreed, I've created a task to track this: https://hms-dbmi.atlassian.net/browse/HMP-196

PR #3122, opened 2023-06-12
John conroy/remove origin sample dependencies followup

john-conroy:
Follow up to #3120. Didn't want the original to PR to get too big. There is no logical divide between the two.

### DIFF
--- a/context/app/routes_api.py
+++ b/context/app/routes_api.py
@@ -110,7 +110,7 @@ def _get_entities(entity_type, constraints={}, uuids=None):
         'created_by_user_email',
     ]
     if entity_type in ['samples', 'datasets']:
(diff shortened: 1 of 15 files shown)


NickAkhmetov (commented):
LGTM but there are some other remaining instances of `origin_sample` in various test fixtures:

end-to-end/artillery/queries/datasets.json
end-to-end/artillery/queries/ccf.json
end-to-end/cypress/fixtures/dataset-search/default.js
end-to-end/cypress/fixtures/dataset-search/heart-only-5-or-less-data-types.js

Should those be updated to match the change in schema/should we open another task for that update?

john-conroy:
Thanks Nick! I filed an issue related to the instances in artillery earlier today. It's not something we've used in quite some time so I've tabled it for now.

NickAkhmetov (approved):

PR #3123, opened 2023-06-13
Kevin yoo/get gene common name

kevinyooky:


context/app/static/js/pages/Genes/Genes.jsx
```diff
@@ -25,7 +29,7 @@ function Genes() {
           Gene
         </Typography>
         <Typography variant="h1" component="h2">
-          CD4
+          {`CD4 (${commonName})`}
```
NickAkhmetov, line 32: If I'm not mistaken, `CD4` was an example of a gene name/test data and can be removed here:

```suggestion
          {commonName}
```

context/app/static/js/pages/Genes/Genes.jsx
```diff
@@ -25,7 +29,7 @@ function Genes() {
           Gene
         </Typography>
         <Typography variant="h1" component="h2">
```
NickAkhmetov, line 31: I guess this was in the scope of the last PR rather than this one, but I'm a bit skeptical of the HTML semantics here (lines 24 and 27); making `Gene` the top level heading of the page (i.e. making it an `h1`) while making the gene's common name an `h2` seems a bit like making `Blog` the top level heading of a specific blog page rather than the title of the article. Maybe this should be reversed? (i.e. with Gene having `component="h2"` and commonName having `component="h1"`)

context/app/static/js/pages/Genes/hooks.js
```diff
@@ -0,0 +1,36 @@
...
+    },
+  );
+
+  useEffect(() => {
+    if (geneCommonName && geneCommonName.response && geneCommonName.response.docs[0]) {
+      setCommonName(geneCommonName.response.docs[0].name);
+    }
+  }, [geneCommonName]);
+
+  return commonName;
```
NickAkhmetov, line 33: - This `useState` and `useEffect` are unnecessary since the set value is entirely derived from the already-stateful `useSWR` hook. This [react doc article](https://react.dev/learn/you-might-not-need-an-effect) goes into more detail, but essentially this is similar to the `firstName`/`lastName`/`fullName` example described there, where fullName is better represented as a derived const than as its own state.
- We should be able to simplify it by using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to drill into potentially undefined values.
- We should also hook into `useSWR`'s `isLoading` boolean to provide an accurate loading state
- The `fallbackData` provided to `useSWR` should correspond to the shape of the expected response from the fetcher function. 
- With all of the above in mind, the fetcher function could match its name more closely by extracting the common name itself, which should make the `useState` and `useEffect` hook even more unnecessary :)

```suggestion
async function fetchGeneCommonName(geneNamesURL) {
  const response = await fetch(geneNamesURL, { headers: { Accept: 'application/json' } });

  if (!response.ok) {
    console.error('Genenames API failed.', response);
  }

  const results = await response.json();
  
  const commonName = results?.response?.docs?[0]?.name;

  return commonName;
}

function useGeneCommonName(geneSymbol) {

  const { data: commonName, isLoading } = useSWR(
    `https://rest.genenames.org/fetch/symbol/${geneSymbol}`,
    fetchGeneCommonName,
    {
      fallbackData: '',
    },
  );

  return {commonName, isLoading};
```

context/app/static/js/components/detailPage/genes/Summary/Summary.jsx
```diff
@@ -0,0 +1,26 @@
...
+Summary.defaultProps = {
+  created_timestamp: undefined,
+  last_modified_timestamp: undefined,
+  description: '',
+  status: '',
+  mapped_data_access_level: '',
+  entityCanBeSaved: true,
+  collectionName: '',
+  mapped_external_group_name: undefined,
+};
```
NickAkhmetov, line 24: Are these prop types here to support future functionalities in the designs? If some of these are values are going to be passed from flask (e.g. the `variables_with_underscores`), should they be in the flask context instead?

NickAkhmetov (commented):
I could only find the code for the calls to the `genenames.org` API - is there some work related to the `NIH` XML api that didn't get swept up?

context/app/static/js/pages/Genes/hooks.js
NickAkhmetov: There's also the question of useSWR's default behavior being to revalidate the query every time the window is refocused: https://swr.vercel.app/docs/revalidation

I'm not sure if that's necessary for data this static - we could probably quiet down our API calls by adding `revalidateOnFocus: false` to the useSWR config.

context/app/static/js/pages/Genes/Genes.jsx
kevinyooky: I agree with this.

context/app/static/js/pages/Genes/hooks.js
kevinyooky: Thanks for this, learning more about swr

context/app/static/js/pages/Genes/Genes.jsx
```diff
@@ -17,15 +21,19 @@ function Genes() {
...
+
   return (
     <FlexRow>
       <TableOfContents items={[...sections.values()]} />
       <Content>
-        <Typography variant="subtitle1" component="h1" color="primary">
+        <Typography variant="subtitle1" component="h2" color="primary">
           Gene
         </Typography>
         <Typography variant="h1" component="h2">
```
NickAkhmetov, line 35: One final tiny fix to go along with the revision to use `h2` for the `gene` subtitle:

```suggestion
        <Typography variant="h1" component="h1">
```

kevinyooky: great catch!

NickAkhmetov (approved):
🚀

PR #3124, opened 2023-06-14
Kevin yoo/summary description api v2

kevinyooky:


context/app/static/js/pages/Genes/hooks.js
```diff
@@ -26,4 +26,55 @@ function useGeneCommonName(geneSymbol) {
...
+    console.error('Gene data API failed.', response);
+  }
+
+  const responseText = await response.text();
+  const result = await parseStringPromise(responseText);
+  const geneSummary = result['Entrezgene-Set'].Entrezgene[0].Entrezgene_summary[0];
+  return geneSummary;
+}
+
+const useGeneData = (geneID) => {
```
NickAkhmetov, line 42: - The only thing we currently need the `geneId` for is to provide it to the `useGeneData` hook
 - To start fetching `geneSummary` we need `geneId` to be available, but `useGeneData` currently doesn't do anything to prevent an attempted fetch when `geneId` is not yet available

My first suggestion is to modify the `useGeneData` hook to prevent it from trying to fetch data when it's being provided with a blank `geneId` ( since `geneID` is `''` until the first hook finishes fetching). The [Conditional Fetching](https://swr.vercel.app/docs/conditional-fetching) section of the SWR documentation should be helpful for this.

Additionally, we should consider using the `useGeneId` hook inside the `useGeneData` hook and making the `useGeneData` hook take in the `geneSymbol` instead. The `geneID` would still be internally cached by SWR for any other queries which may rely on it in the future, and the `useGeneData` hook would be a bit easier to reuse

kevinyooky: This all makes sense, thanks so much for the feedback!

context/app/static/js/pages/Genes/Genes.jsx
```diff
@@ -5,12 +5,14 @@ import TableOfContents from 'js/shared-styles/sections/TableOfContents';
...
 import { FlexRow, Content } from './style';
-import { useGeneCommonName } from './hooks';
+import { useGeneCommonName, useGeneData, useGeneId } from './hooks';
 
 const summaryId = 'Summary';
 
 function Genes({ geneSymbol }) {
   const { geneCommonName, isLoading } = useGeneCommonName(geneSymbol);
+  const geneSummary = useGeneData(geneSymbol);
+  const geneId = useGeneId(geneSymbol);
```
NickAkhmetov, line 15: - Since these hooks' values are just getting passed down to `SummaryBody`, we could pass down `geneSymbol` instead and call these hooks from `SummaryBody`; being able to use hooks colocated with where the data is used without having to worry about duplicate requests is one of the benefits of using SWR and makes future troubleshooting easier!
  - I think it's also worth considering creating a context to hold the `geneSymbol` value so you don't have to [prop drill](https://www.geeksforgeeks.org/what-is-prop-drilling-and-how-to-avoid-it/) the `geneSymbol` down to every component on the page; since the `geneSymbol` should change after page routing, so keeping it in Context shouldn't cause any performance concerns and cleans up some visual noise in components where the symbol is not used, e.g. the `<Summary>` component.
- To support that revision (moving the hooks down to the `SummaryBody`), the `isLoading` check on line 26 could be revised so that it only blocks the `geneCommonName` display in the title.

Lines 26-28 would be replaced with:
```
const title = `${geneSymbol.toUpperCase()}} ${ isLoading ? `(${geneCommonName})` : null }`
```

Then the `title` would be used in line 38:
```
<Typography variant="h1" component="h1">
  {title}
</Typography>
```

This way, the fetch of the `geneCommonName` shouldn't block any fetch operations that don't depend on the name being present, which should make the page load faster for users (with requests firing in parallel instead of serially)

context/app/static/js/shared-styles/sections/LabelledSectionText/LabelledSectionText.jsx
```diff
@@ -38,12 +39,14 @@ LabelledSectionText.propTypes = {
...
   bottomSpacing: PropTypes.number,
   childContainerComponent: PropTypes.elementType,
+  fontWeight: PropTypes.number,
 };
 
 LabelledSectionText.defaultProps = {
   iconTooltipText: undefined,
   bottomSpacing: undefined,
   childContainerComponent: undefined,
+  fontWeight: undefined,
```
NickAkhmetov, line 49: This should match the default value defined in line 16
```suggestion
  fontWeight: 400,
```

NickAkhmetov (commented):
Great job! I just have a couple of small notes/ideas to further enhance the implementation, then we should be ready to merge this IMO.

context/app/static/js/pages/Genes/Genes.jsx
kevinyooky: As mentioned before, I was working on the wrong branch! I reverted my changes and pushed. I will make sure to still apply your comments on the branch.

context/app/static/js/shared-styles/sections/LabelledSectionText/LabelledSectionText.jsx
kevinyooky: As mentioned before, I was working on the wrong branch! I reverted my changes and pushed. I will make sure to still apply your comments on the branch.

context/app/static/js/pages/Genes/Genes.jsx
john-conroy: You shouldn't need context with SWR. SWR caches results, right?

context/app/static/js/shared-styles/sections/LabelledSectionText/LabelledSectionText.jsx
john-conroy: Reference the theme?

context/app/static/js/pages/Genes/Genes.jsx
NickAkhmetov: Yup, no need for an extra context with SWR after v1, it has a global cache by default now.

I’m referring to moving the “geneSymbol” for the page into its own context, not the products of the fetches that use it.

NickAkhmetov (approved):
Great work!

PR #3126, opened 2023-06-15
Upgrade to Vitessce v3 and restore metadata for samples and datasets

lchoy:
This PR essentially rolls back to tag 0.72.2 and cherry-picks the changes to remove dependencies on origin_sample and upgrades the version of Vitessce to 3.0.0.

### DIFF
--- a/CHANGELOG-fix-metadata.md
+++ b/CHANGELOG-fix-metadata.md
@@ -1,0 +1,1 @@
+- Reverts changes to add biomarker pages and app context refactorings


context/app/static/js/pages/Donor/Donor.jsx
```diff
@@ -8,28 +8,31 @@ import Protocol from 'js/components/detailPage/Protocol';
...
+    hubmap_id,
+    entity_type,
+    created_timestamp,
+    last_modified_timestamp,
+    description,
+    mapped_metadata = {},
+    // As data comes in from other consortia, we won't be able
+    // to rely on donor metadata always being available.
+    // Unpublished HuBMAP data may also be missing donor metadata.
+  } = assayMetadata;
```
NickAkhmetov, line 34: This could be a potential source of the issues we're seeing - this was getting deconstructed from assayMetadata before the changes to `useFlaskDataContext`, and was being deconstructed from `entity` afterwards; let me try changing this on main to see if it's a potential culprit so we don't have to roll back/redo as many changes

NickAkhmetov: Never mind, that wasn't it 👎

PR #3135, opened 2023-06-21
kevin-yoo/hmp-20-trim-publication-title

kevinyooky:


NickAkhmetov:
To clarify the feedback above: rather than trimming the text programatically, applying these styles to the `Typography` component that wraps it should 
```
    max-width: 100ch;
    overflow: hidden;
    display: -webkit-box;
    -webkit-box-orient: vertical;
    -webkit-line-clamp: 1;
```
Useful context: 
- Good breakdown of how this works in css-tricks: https://css-tricks.com/almanac/properties/l/line-clamp/
- MDN docs for a more formal definition: https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp

kevinyooky:
Hey thanks for providing me the css code, really appreciate it. I applied it, but it is not working, there are potential confounding variable as to why this is the case, my guess is that MUI Typography has its own styling that is interfering with it or it could be something else. 

For the sake of simplicity, I will go with a javascript one-liner.

NickAkhmetov:
I opened up #3138 with a demo of the CSS solution - the `100ch` max-width was too large in my original suggestion.

kevinyooky:
As discussed in our call, I will go ahead with this approach, ready to be merged!

NickAkhmetov (approved):
Ready for liftoff 🚀

PR #3137, opened 2023-06-21
John conroy/remove support files from publications HMP-175

john-conroy:
Remove globus link for lifted publication ancillary entities on publication pages.


NickAkhmetov (approved):
Makes sense to me 👍

PR #3138, opened 2023-06-22
[Suggestion] Trim header entity names with CSS

NickAkhmetov:
This PR is meant to demonstrate that the text trim behavior is possible with CSS.

image

My original suggestion in the other PR with `max-width: 100ch` set the upper boundary of the width too high since the `ch` unit's size is calculated from the width of the `0` character from the font in use; with non-monospaced fonts (i.e. if you're not using a coding or typewriter font), this means that spaces/narrow letters/punctuation don't "count" towards that 100 character width as much as they should.

We should still keep the `truncateText` utility/the corresponding spec, there may be other uses for it throughout the site.


NickAkhmetov:
As discussed on the call today, given the triviality of the performance difference between the JS and CSS approaches and the JS version more closely matching the original specifications in the ticket, we'll go with the original JS approach.

PR #3140, opened 2023-06-26
NickAkhmetov / Fix error page overflow

NickAkhmetov:
This was a change made while diagnosing HMP-151 (which has since been traced to an internal vitessce issue, so any further changes related to this will be made there and will be accompanied with a dependency version bump here)

One of the issues identified for this ticket was that the error message overflows its container due to the length of the URL, so this PR is to correct this issue.

image

### DIFF
--- a/context/app/static/js/pages/Error/style.js
+++ b/context/app/static/js/pages/Error/style.js
@@ -15,8 +15,9 @@ const Background = styled.div`
 const StyledPaper = styled(Paper)`
   width: 100%;
   max-width: 880px;
-  margin: 0px 16px;
-  padding: 16px;
+  margin: 0 1rem;
+  padding: 1rem;
+  word-break: break-all;
 `;
 
 const StyledTypography = styled(Typography)`
--- a/context/app/static/js/pages/Error/style.js
+++ b/context/app/static/js/pages/Error/style.js
@@ -15,9 +15,9 @@ const Background = styled.div`
 const StyledPaper = styled(Paper)`
   width: 100%;
   max-width: 880px;
- ${({ theme: { spacing } }) => css`
-   margin: 0 ${spacing(2)}px;
-   padding: ${spacing(2)}px;
+  ${({ theme: { spacing } }) => `
+    margin: 0 ${spacing(2)}px;
+    padding: ${spacing(2)}px;
   `}
   word-break: break-all;
 `;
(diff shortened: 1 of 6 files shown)


lchoy (commented):
I like the overlay: 
hmp-151-error

When I dismiss this, I see the same rendering prior to the fix. Could the error be confined within the "Visualization" section instead so that the remainder of the page could be rendered?

NickAkhmetov:
Thanks for the review!

The overlay is from the webpack dev server; it shows up for compiler errors and uncaught runtime errors by default, and is not visible in prod. https://webpack.js.org/configuration/dev-server/#overlay

I will make a second pass at this and add an error boundary around the visualization as suggested, so the rest of the page is still displayed.

context/app/static/js/pages/Error/style.js
```diff
@@ -15,8 +15,9 @@ const Background = styled.div`
 const StyledPaper = styled(Paper)`
   width: 100%;
   max-width: 880px;
-  margin: 0px 16px;
-  padding: 16px;
+  margin: 0 1rem;
+  padding: 1rem;
```
john-conroy, line 19: Prefer to use spacing from the theme for now which uses `px`. We can consider using `rem` when we overhaul styling down the road.

NickAkhmetov: Makes sense; I've reverted this to use `px`.

context/app/static/js/pages/Error/style.js
```diff
@@ -15,8 +15,9 @@ const Background = styled.div`
 const StyledPaper = styled(Paper)`
   width: 100%;
   max-width: 880px;
-  margin: 0px 16px;
+  margin: 0 16px;
   padding: 16px;
```
john-conroy, line 19:
```suggestion
 ${({ theme: { typography, palette, spacing } }) => css`
   margin: 0 ${spacing(2)}px;
   padding: ${spacing(2)}px;
  `}
```

NickAkhmetov:
I've updated the styles to use values from the theme as John suggested and added a scrollable error boundary around the Visualization:
image


https://github.com/hubmapconsortium/portal-ui/assets/19957804/9d2be693-fbfe-4317-bae9-813f4a2cb91d

john-conroy (approved):
Looks good to me. @tsliaw  might have some input on future styling, but that shouldn't block.

tsliaw:
@NickAkhmetov Can you file a Jira issue for me to look at the styling at another time so I don't forget about this issue? Thanks!

NickAkhmetov:
I've made https://hms-dbmi.atlassian.net/browse/HMP-209 to track the above concern 👍 

Given that Lisa's out for the rest of the week, this should be ready to merge now with John's approval above.

john-conroy (approved):

PR #3143, opened 2023-06-28
Nickakhmetov/HMP-194 Apply fixes from lint config

NickAkhmetov:
I organized the commits by the rules they pertained to where appropriate; some edits were made to the configs as well (which I'll be documenting)

This PR is a draft for now as I'm still in the process of getting tests to pass; I will request reviews as soon as that's done. In the meantime, the changes I've made so far are now visible.


context/build-utils/webpack.common.js
```diff
@@ -14,19 +13,20 @@ const config = {
...
     },
   },
   resolve: {
-    extensions: ['.js', '.jsx', '.css', '.woff', '.woff2', '.svg', '.yaml', '.yml', '.json'],
+    extensions: ['.js', '.jsx', '.ts', '.tsx', '.css', '.woff', '.woff2', '.svg', '.yaml', '.yml', '.json'],
     fallback: {
       // Now necessary because webpack 5 doesn't include these polyfills by default
       timers: require.resolve('timers-browserify'),
       stream: require.resolve('stream-browserify'),
+      buffer: require.resolve('buffer'),
```
john-conroy, line 29: What required introducing the polyfill?

context/app/static/js/shared-styles/charts/VerticalStackedBarChart/VerticalStackedBarChart.jsx
```diff
@@ -10,6 +10,20 @@ import { trimStringWithMiddleEllipsis } from 'js/shared-styles/charts/utils';
...
+      return (
+        <Text onMouseEnter={handleMouseEnter({ key: formattedValue })} onMouseLeave={handleMouseLeave} {...tickProps}>
+          {trimStringWithMiddleEllipsis(formattedValue)}
+        </Text>
+      );
+    },
+    [handleMouseEnter, handleMouseLeave],
+  );
+}
+
```
john-conroy, line 26: This pattern seems strange, was this the outcome of an eslint suggestion? Which rule?

context/app/static/js/components/detailPage/provenance/ProvVis/__tests__/ProvVis.spec.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { render, unmountComponentAtNode } from 'react-dom';
+// eslint-disable-next-line import/no-unresolved
+import { render } from 'test-utils/functions';
```
john-conroy, line 3: Again with module name mapper, this shouldn't be necessary? Are the aliases not working?

context/app/static/js/components/detailPage/provenance/ProvVis/ProvVis.jsx
```diff
@@ -15,7 +15,7 @@ export default function ProvVis({ provData, getNameForActivity, getNameForEntity
   const { steps, setSteps } = useProvenanceStore(useProvenanceStoreSelector);
 
   useEffect(() => {
-    setSteps(new ProvData(provData, getNameForActivity, getNameForEntity, entity_type).toCwl());
+    setSteps(new ProvData(provData, entity_type, getNameForActivity, getNameForEntity).toCwl());
```
john-conroy, line 18: What prompted the argument reordering?

context/app/static/js/components/detailPage/provenance/ProvGraph/ProvGraph.jsx
```diff
@@ -23,69 +70,33 @@ function ProvGraph({ provData, entity_type, uuid }) {
...
-    // Fallback to ID, if needed. https://github.com/hubmapconsortium/prov-vis/issues/15
-    return entity ? `${entity[typeKey]} - ${entity[idKey]}` : id;
-  }
+  const getNameForActivity = useCallback(
+    function getNameForActivity(id, prov) {
+      const activity = prov.activity[id];
+      return `${activity['hubmap:creation_action']} - ${activity[idKey]}`;
+    },
+    [idKey],
+  );
```
john-conroy, line 79: Were the additions of callback here prompted by eslint?

john-conroy (commented):
Thankfully most of it was swapping functional components from arrow functions to the traditional function syntax. Were all of these changes prompted by eslint? It seems like there were a few changes that might not have been and it's a little difficult for me to discern.

context/app/static/js/components/detailPage/files/FileBrowser/FileBrowser.spec.js
```diff
@@ -1,7 +1,6 @@
-/* eslint-disable import/no-unresolved */
 import React from 'react';
 import userEvent from '@testing-library/user-event';
-import { render, screen } from 'test-utils/functions';
+import { render, screen } from '@/../../test-utils/functions';
```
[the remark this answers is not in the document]
NickAkhmetov: Yup, this was a mistake on my part, once I realized this pattern was present throughout all the other specs I stopped doing this. I meant to go back and fix this, thanks for catching it!

context/build-utils/webpack.common.js
NickAkhmetov: Same xml2js library as before started throwing errors during docker build until I included this. I added this to the related ticket so it's removed alongside the other ones.

context/app/static/js/components/detailPage/provenance/ProvVis/__tests__/ProvVis.spec.js
NickAkhmetov: To be reverted 👍

context/app/static/js/components/detailPage/provenance/ProvVis/ProvVis.jsx
NickAkhmetov: https://eslint.org/docs/latest/rules/default-param-last

Since `ProvData` has default arguments defined for `getNameForActivity` and `getNameForEntity`, the convention is to put them last so they can be omitted entirely; otherwise, there's no way to define `entity_type` without also providing values for the other two arguments.

context/app/static/js/components/detailPage/provenance/ProvGraph/ProvGraph.jsx
NickAkhmetov: Removed the useCallbacks here, I think I added those before we agreed on our current heuristic.

context/app/static/js/components/entity-search/SearchWrapper/utils.js
```diff
@@ -42,7 +42,7 @@ function buildFieldConfig({
   type,
   facetGroup = 'Additional Facets',
   configureGroup = 'General',
-  entityType: pageEntityType,
+  entityType: _pageEntityType,
```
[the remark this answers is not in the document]
NickAkhmetov: Since this prop is part of a spread deconstruction operator, removing it altogether would implicitly add `entityType` to the `rest` object whenever  `entityType` is ever provided as part of the input object. This would lead to the `entityType` key being propagated into the field config returned since `...rest` is part of the returned fields, and may cause unforeseen bugs to pop up. My understanding of this key being present in the deconstructed props but unused otherwise is that `entityType` is being purposely excluded from the returned object (assuming that the related search functionality is working as expected)

context/app/static/js/shared-styles/buttons/index.js
```diff
@@ -29,7 +29,14 @@ const WhiteBackgroundToggleButton = styled(ToggleButton)`
   }
 `;
 
-function TooltipToggleButton({ children, tooltipComponent, tooltipTitle, buttonComponent, id, ...rest }) {
+function TooltipToggleButton({
+  children,
+  tooltipComponent,
+  tooltipTitle,
+  buttonComponent: _buttonComponent,
```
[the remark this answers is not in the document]
NickAkhmetov: Same as in the other case; if this is ever called with a `buttonComponent` prop defined, it would be implicitly propagated to `WhiteBackgroundToggleButton`, so I marked it as unused with the underscore prefix to be safe.

context/app/static/js/components/detailPage/files/FileBrowser/FileBrowser.spec.js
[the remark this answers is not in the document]
NickAkhmetov: Aha! There is a plugin to get this to play nice: https://www.npmjs.com/package/eslint-import-resolver-jest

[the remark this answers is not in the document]
NickAkhmetov: Thanks for pointing my attention back to this, I was able to remove all the eslint ignores for unresolved imports from the test files 🙌

context/app/static/js/components/detailPage/provenance/ProvVis/ProvVis.jsx
NickAkhmetov: Alternatively - we could change the constructor of `ProvData` to take in an object with these args instead of having them be positional (i.e. so the constructor is `new ProvData({provData, entity_type, getNameForActivity, getNameForEntity})`, we can have any of them be optional without worrying about their order; we'd still need to modify all the calls to `new ProvData` accordingly, but this idea seems a bit more maintainable in the long run.

john-conroy: Preference to use an object, but shouldn't block here.

NickAkhmetov: I've made this adjustment, I prefer that alternative approach as well. The only downside is needing to make sure that the `prov` prop is actually being passed in as `prov` (which I've adjusted accordingly where needed)

john-conroy (approved):

PR #3155, opened 2023-07-06
Nickakhmetov/HMP-144 React 18 Upgrade

NickAkhmetov:
This PR
- Bumps React to v18.2.0
- Bumps Storybook to the latest minor version
- Bumps `@hms-dbmi-bgm/react-workflow-viz` to the latest minor version
- Updates visx, react-intersection-observer, and styled-components to newer versions that support React 18

Additional unrelated small changes made along the way:
- I added an alias to `dev-start` called just `start` - I was tired of tab-complete not working when typing in `./etc/dev/d` due to the `dev-start` script sharing a first letter with the `docker` script. This doesn't actually add any code as it's just a shortcut/reference. Now tab-complete should work from `./etc/dev/s` (full command = `./etc/dev/start`)
- I added the `code actions on save` config to the `optional` keys in `default.settings.json` after seeing that the `eslint format on save` setting was not enough to make Kevin's env autofix prettier issues on save.
- I prevented the entity count tooltip from showing up before the entity counts were loaded (before, they were showing e.g. `undefined Donors`).
- I needed to add a try/catch block to the search page fetch in order to catch runtime errors from it firing a cleanup function while navigating to a new page - I didn't update `searchkit` or any other elastic UI pieces 

**It looks big but 99% of it is just the package.lock!** I purposely limited the scope of my work to minimize manual reconfigurations (which I've created tickets for below). I spread out the updates for the codependencies into different commits to monitor for regressions after upgrades and reran the test suite after each one.

---

We are still using `render`, so functionally, this update only brings us up to speed with React 17. In order to switch to `createRoot`, our next steps are to update remaining dependencies that rely on React 16 which involved bigger revisions:
- Update MUI https://hms-dbmi.atlassian.net/browse/HMP-228
- ~~Update React-PDF https://hms-dbmi.atlassian.net/browse/HMP-229~~ Done, merged into this PR
- ~~Update or replace nuka-carousel https://hms-dbmi.atlassian.net/browse/HMP-230~~ Done, pushed into this PR

After that's done:
- Add React Strict Mode to identify any incompatibilities with React 18 in our own code https://hms-dbmi.atlassian.net/browse/HMP-231
- And finally, switch to `createRoot` and resolve any remaining issues which were not surfaced by StrictMode https://hms-dbmi.atlassian.net/browse/HMP-232


NickAkhmetov:
image

lchoy (dismissed):

etc/dev/start
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
john-conroy, line 1: How did you create the alias? Showing up here as an entirely new file.

john-conroy (approved):
Looks good to me. Curious about the alias, but not blocking.

etc/dev/start
NickAkhmetov: I used the `ln` command, but I forgot to use the `-s` flag to make sure it creates a symlink as opposed to a hardlink. 

I removed and re-created the alias with the `-s` flag, and this appears to be fixed now.

john-conroy (approved):

PR #3157, opened 2023-07-07
HMP-229 Upgrade React-PDF

NickAkhmetov:
Separate PR for the react-pdf upgrade - seems to work fine after testing, but since it involved a bit of reconfiguration I figured it makes sense to split it off from the rest of the upgrade for now. Since this is related to the React 18 upgrade, I based this PR on that branch.

The perceivable benefits of the version bump are some simplification of Jest config (no more special handling for react-pdf import) and the offloading of PDF processing from the main thread, meaning there's less of a freeze-up when the PDF loads.

Screenshot 2023-07-07 at 11 29 50 AM

### DIFF
--- a/CHANGELOG-hmp-229.md
+++ b/CHANGELOG-hmp-229.md
@@ -1,0 +1,1 @@
+- Upgrade React-PDF to use worker thread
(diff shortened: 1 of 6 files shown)


john-conroy (approved):
Looks good to me, thanks!

PR #3159, opened 2023-07-10
NickAkhmetov / HMP-235 Add support for text-only publication vignettes

NickAkhmetov:
This PR adjusts the Vitessce vignette rendering and fetching logic:
- The lack of a vitessce conf no longer blocks the rendering of the vignette description, which allows text-only vignettes to be displayed correctly
- Vitessce conf fetching has been updated
- I've started to gradually include more TypeScript annotations where appropriate
- Improved tsconfig so that `js/*` paths are properly recognized in TS files

I've tested this locally; if there are no conf files provided, the accordion only displays the markdown description.

The markdown description is also instantly visible as the accordion expands/Vitessce loads, instead of popping in at the same time Vitessce finishes loading. You can see it flash by in the recorded demo:

https://github.com/hubmapconsortium/portal-ui/assets/19957804/7893091a-5a89-4bc3-8c48-0ac5ff37c661

---

Future Improvements

I ran into some problems with Storybook failing to parse `Contexts.tsx` after I tried to convert that file; I'll revisit it at another time (created HMP-250), and have worked around it by defining the types in JS comments.


context/app/static/js/components/publications/PublicationVignette/hooks.ts
```diff
@@ -0,0 +1,47 @@
+import useSWR from 'swr';
+import { useAppContext } from 'js/components/Contexts';
+import { fillUrls } from './utils';
+import { PublicationVignette } from '../types';
+
+// Fetcher function that lets SWR fetch multiple urls at once
+const multiFetcher = (...urls: string[]) => {
+  const f = (url: string) => fetch(url).then((response) => response.json());
+  return Promise.all(urls.map((url) => f(url)));
+};
```
john-conroy, line 10: Would this be worth pulling out for use elsewhere?

context/app/static/js/components/publications/PublicationVignette/hooks.ts
```diff
@@ -0,0 +1,47 @@
...
+type PublicationVignetteConfsInput = {
+  uuid: string;
+  vignette: PublicationVignette;
+  vignetteDirName: string;
+};
+
+export function usePublicationVignetteConfs({ uuid, vignetteDirName, vignette }: PublicationVignetteConfsInput) {
+  const { assetsEndpoint, groupsToken } = useAppContext();
+  // Extract file paths from the vignette object to form the urls to fetch for this vignette
+  const urls = vignette.figures.map(
```
john-conroy, line 21: In this case, figures could be undefined.

john-conroy (commented):
Thanks! I'm curious if this would handle undefined figures in a vignette.

context/app/static/js/components/publications/PublicationVignette/hooks.ts
NickAkhmetov: I'll move this to a shared utils folder so we can reuse it elsewhere, good call!

context/app/static/js/components/publications/PublicationVignette/hooks.ts
NickAkhmetov: Good call 👍 I'm operating under the assumption that `figures` would be an empty list if none are provided as opposed to being null/undefined, but it's better to be safe than sorry.

context/app/static/js/components/publications/PublicationVignette/hooks.ts
NickAkhmetov: This should be fixed now 👍

context/app/static/js/components/publications/PublicationVignette/hooks.ts
NickAkhmetov: All set; added a `?.` gate before the `map` call and a fallback to an empty list.

context/app/static/js/components/publications/PublicationVignette/hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { assetsEndpoint, groupsToken } = useAppContext();
+  // Extract file paths from the vignette object to form the urls to fetch for this vignette
+  const urls =
+    vignette.figures?.map(
+      ({ file }) => `${assetsEndpoint}/${uuid}/vignettes/${vignetteDirName}/${file}?token=${groupsToken}`,
+    ) ?? [];
+  const { data } = useSWR(urls, multiFetcher, {
+    revalidateOnFocus: false,
+    revalidateOnReconnect: false,
+  });
```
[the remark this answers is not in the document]
NickAkhmetov: The nullish coalescing fallback on line 19 ensures that `urls` is always at least an empty list.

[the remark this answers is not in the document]
john-conroy: Ah sorry didn't catch that!

john-conroy (approved):

context/app/static/js/components/publications/PublicationVignette/hooks.ts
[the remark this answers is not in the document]
NickAkhmetov: Actually, on second thought, that `?? []` shouldn't be there since an empty array is truthy. One moment.

john-conroy (approved):

PR #3166, opened 2023-07-12
John conroy/collection dataset query HMP-257

john-conroy:
- Get collection datasets from separate search-api query instead of the collection es document.
- Use `RelatedEntitiesTable` for display.
- Fix display of organ column.

Screen Shot 2023-07-12 at 12 26 58 PM


NickAkhmetov (approved):
Looks good to me 👍

PR #3170, opened 2023-07-13
John conroy/scroll prov HMP-258

john-conroy:
Feels a bit hacky. I had a slightly less hacky version before the [commit to introduce the dom selector](https://github.com/hubmapconsortium/portal-ui/commit/13475bd7d1c001e34e94fd8228e00eda236c4d6f), but the detail pane would be inside the scrollbox without the changes introduced in the commit.

Initial position without detail pane.
Screen Shot 2023-07-13 at 4 18 34 PM

Initial position with detail pane outside of scrollbox.
Screen Shot 2023-07-13 at 4 18 26 PM

Graph without scroll.
Screen Shot 2023-07-13 at 4 29 10 PM

### DIFF
--- a/CHANGELOG-scroll-prov.md
+++ b/CHANGELOG-scroll-prov.md
@@ -1,0 +1,2 @@
+- Set max vertical height and enable vertical scrolling for provenance graph.
+- Set initial vertical scroll position in provenance graph to halfway.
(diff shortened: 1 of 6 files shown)


john-conroy:
Fixing the tests...

NickAkhmetov (commented):
The code looks good! I think you should be able to mock the `scroll` function out with Jest to fix the failing tests?

NickAkhmetov (approved):
👍 works for me, just ran the tests on local to confirm.

PR #3171, opened 2023-07-14
Nickakhmetov/HMP-269 Implement time-gated publication slide

NickAkhmetov:
This PR adds a flag sent from the server-side which is enabled after the unix timestamp passes 2023-07-19 at 15:00:00 UTC (the end of the publication embargo). This behavior can be overridden by adding the `ENABLE_PUBLICATIONS` key to `app.conf` and setting it to `True` for ease of testing (e.g. in `dev`/`test`)

The time gate works by running a comparison of the current unix timestamp against the timestamp of 2023-07-19 at 15:00:00 UTC (1689778800). I cross-checked the timestamp from `datetime.now().timestamp()` with https://www.unixtimestamp.com/ (which I used to derive the embargo expiration timestamp) to confirm this method works for all time zones.

The flag enables the new hubmap publications slide, which links to `/publications` with its get started button and pulls the proper image from s3. When the flag is enabled, the randomization of the starting slide index is also disabled.

image

### DIFF
--- a/CHANGELOG-hmp-269-implement-publication-slide.md
+++ b/CHANGELOG-hmp-269-implement-publication-slide.md
@@ -1,0 +1,1 @@
+- Add publication slide to homepage.
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -1,7 +1,7 @@
 from flask import (render_template, current_app, abort,
                    session, request)
 
-from .utils import get_default_flask_data, make_blueprint, get_organs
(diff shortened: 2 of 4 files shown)


john-conroy (approved):
Looks good to me. Thanks!

john-conroy:
Maybe file an issue for cleanup?

NickAkhmetov:
> Maybe file an issue for cleanup?

👍 Done!

https://hms-dbmi.atlassian.net/browse/HMP-271

PR #3172, opened 2023-07-14
John conroy/publication collection aggs HMP-270

john-conroy:
There is room for some refactoring to remove the additional request when an associated collection is not present and to push the datasets mapping inside the SWR fetcher so it's only run once, but I wanted to get this out sooner.


NickAkhmetov (approved):
Let's make a ticket for the refactorings/todos you mentioned, but otherwise this looks fine to me. Thanks for figuring this out on such short notice!

PR #3180, opened 2023-07-17
Handle entity api failure for globus link

john-conroy:
There's lots of room for refactoring and we've filed https://hms-dbmi.atlassian.net/jira/software/projects/HMP/issues/HMP-279 to do so, but this felt like the quickest way to get this in given the deadline.

Screen Shot 2023-07-17 at 1 19 51 PM


NickAkhmetov (approved):
👍 This should work, good call. I do think the status being `undefined` to indicate loading works as well, but this is more explicit/easier to follow.

PR #3183, opened 2023-07-17
Kevin yoo/hmp 123 convert flask data to context for sample page

kevinyooky:


context/app/static/js/components/detailPage/Attribution/Attribution.jsx
```diff
@@ -1,11 +1,15 @@
 import React from 'react';
 import SectionHeader from 'js/shared-styles/sections/SectionHeader';
+import { useFlaskDataContext } from 'js/components/Contexts';
 import { DetailPageSection } from 'js/components/detailPage/style';
 import EmailIconLink from 'js/shared-styles/Links/iconLinks/EmailIconLink';
 import { FlexPaper } from './style';
 import SectionItem from '../SectionItem';
 
-function Attribution({ group_name, created_by_user_displayname, created_by_user_email }) {
+function Attribution() {
```
NickAkhmetov, line 9: This `Attribution` component is also used on the Donor, Dataset, and Preview pages (last 3 links under `Resources` in the header) - does this work correctly on all of them?

context/app/static/js/components/detailPage/summary/Summary/Summary.jsx
```diff
@@ -23,10 +18,13 @@ function Summary({
   collectionName,
   mapped_external_group_name,
 }) {
+  const {
```
NickAkhmetov, line 21: This `Summary` component is also used on a lot of other pages (Collection, Dataset, and Donor), we should double check functionality there as well

NickAkhmetov (commented):
We should verify that the changes to components shared with other pages don't introduce regressions on those pages

context/app/static/js/components/detailPage/Attribution/Attribution.jsx
kevinyooky: Yeah, they are all coming from flaskdata context, so they are working properly. Double checked again to make sure.

context/app/static/js/components/detailPage/summary/Summary/Summary.jsx
kevinyooky: Checked, all are working!

kevinyooky:
I am having a really difficult time with this build error

NickAkhmetov:
@kevinyooky The fix for the failing specs will be wrapping them in a provider that has the expected data in the exact format it's asking for.
As an example, check out the FileBrowser.spec.js: the various contexts used by the FileBrowser have mock data defined for them on lines 10-16, a wrapper component which provides this data is defined on line 18-26, and every `render` call wraps the tested component in these mock providers.

NickAkhmetov (commented):
I'll fire this up locally tomorrow and take a closer look to double check, but the changes look good to me overall 👍

NickAkhmetov (approved):
👍 Great job!

PR #3187, opened 2023-07-18
NickAkhmetov/HMP-294 - No header for unauthenticated users

NickAkhmetov:
This PR disables the auth header/groups token inclusion when the current user is not logged in in order to enable access to publication vignettes in Globus after the embargo is up.

Screenshot taken with env pointed at the stage endpoints, note the `Log In` prompt at the top right as evidence of being unauthed:
image

Original slack discussion for posterity: https://hubmapconsortium.slack.com/archives/C011RF3NQ23/p1689702903588899

### DIFF
--- a/CHANGELOG-no-auth-for-unauthed.md
+++ b/CHANGELOG-no-auth-for-unauthed.md
@@ -1,0 +1,1 @@
+- Disable auth header/token inclusion for unauthenticated users.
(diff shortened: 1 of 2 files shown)


yuanzhou (approved):
@NickAkhmetov thanks for the quick fix!

PR #3188, opened 2023-07-18
NickAkhmetov/HMP-295 Add args to dev-start script / create `tasks.json` file / disable flaky entity header test

### LINKED ISSUE #3107: Add bash argument to opt-out of pip/npm install lines in dev-start.sh
While testing out with dependencies locally, such as for testing Vitessce release candidates, I typically need to comment out the `npm install` line in https://github.com/hubmapconsortium/portal-ui/blob/c389ddd0d013f2e82694a16489376e091c4aa4ac/etc/dev/dev-start.sh#L33 since otherwise my local copy that I want to test will be re-installed with the NPM version

It would be great to be able to run `./etc/dev/dev-start.sh --no-npm-install` or `./etc/dev/dev-start.sh --no-pip-install` using command-line arguments rather than commenting out lines manually

NickAkhmetov:
This PR adds options to the `dev-start` script which allow us to skip pip/npm installs if necessary; this change was motivated by my other current ticket (upgrading the vitessce-python version in portal-visualization). The script still runs as expected without args, and the `start` alias also accepts these args.

Fixes #3107 

---

`-h` shows help text
`-n` and `--no-npm-install` skip the `npm install` line
`-p` and `--no-pip-install` skip the `pip install` line
`-s` and `--skip-install` skips both installs

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -61,7 +61,7 @@
             "label": "Install dependencies",
             "group": {
                 "kind": "build",
-                "isDefault": true
+                "isDefault": false
             },
             "dependsOn": [
                 "Init submodules",
@@ -165,6 +165,10 @@
                 "Install dependencies",
                 "dev-start (no install)"
             ],
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            },
             "dependsOrder": "sequence",
             "problemMatcher": [
                 "$eslint-stylish"
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -175,7 +175,7 @@
             ]
         },
         {
-            "label": "dev-start (no storybook)",
+            "label": "dev-start (no storybook, no install)",
             "dependsOn": [
                 "Flask Server",
                 "React Dev Server (+ lint)"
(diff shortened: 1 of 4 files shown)


NickAkhmetov:
I added VS Code tasks to further improve DX!

- To run: 
  - command+shift+P on mac or ctrl+shift+P on others
  - "Run tasks..."
  - Select configuration to run
- `dev-start` - runs through full dev start process
- Additional variants: `no install`, `no storybook`, `no storybook no install`, `no storybook no install no lint`
- Each server/task starts in separate terminal pane
- Important tasks (installs, lint, storybook) get separate terminal panes for easy reference if debugging
  - Lint problems prevent the tasks that depend on lint, but don't affect the rest of the server startup process
  - Lint warnings/errors display status icon
- Individual servers can be restarted at will (e.g. running `Flask Server` will kill the current instance of the task, copy a fresh app.conf, and restart just the flask server, leaving the React server running)
- Servers shut down when terminal pane is closed; additional `Kill project ports` task

NickAkhmetov:
> Following up on the cypress failures in CI, I made revisions in this PR after several runs failed:
> 
>     * 50ms -> it couldn't find entityheader whatsoever
> 
>     * 100ms -> entityheader exists but has height 0 so the test still fails
> 
>     * 200ms -> entityheader animation completed, test passes

I've disabled this test altogether for the time being; we should test the entity header scroll appearance functionality via unit tests and test the contents of the header in the publication integration test instead.

john-conroy (approved):
Thanks! This should be a big help, especially `kill project ports`. No more spamming Ctrl-C!

PR #3189, opened 2023-07-18
NickAkhmetov/HMP-296 Fix seqfish display

NickAkhmetov:
WIP; depends on updated portal-vis vitessce dependency from this PR: https://github.com/hubmapconsortium/portal-visualization/pull/81 (`pip install -e path/to/portal-visualization`)

After updating portal-visualization so it returns 1.0.16 schema inputs and attaching unique UIDs for each one on the front-end, I'm still having the same issue viewing the other slides from [c9799ce59c4518bd243555c92c4edc21](https://portal-prod.stage.hubmapconsortium.org/browse/dataset/c9799ce59c4518bd243555c92c4edc21) on local; I've confirmed the `uid` is different for each config and that it's not being stripped by the visualization.


NickAkhmetov:
Closing this PR in favor of https://github.com/vitessce/vitessce/pull/1617; once that change is merged in and released, we can bump the Vitessce version to resolve this issue.

PR #3193, opened 2023-07-20
John conroy/fix detail links to organ page HMP-249

john-conroy:


NickAkhmetov (approved):
👍 Very solid solution!

PR #3194, opened 2023-07-20
John conroy/reverse toc nowrap

john-conroy:
Reversing styles recently introduced which cause overflow in the table of contents.

Before

Screen Shot 2023-07-20 at 5 00 08 PM

After

Screen Shot 2023-07-20 at 5 00 29 PM
Screen Shot 2023-07-20 at 5 00 42 PM

### DIFF
--- a/CHANGELOG-reverse-toc-nowrap.md
+++ b/CHANGELOG-reverse-toc-nowrap.md
@@ -1,0 +1,1 @@
+- Reverse no-wrap styles causing table of contents to overflow.
--- a/context/app/static/js/shared-styles/sections/TableOfContents/style.js
+++ b/context/app/static/js/shared-styles/sections/TableOfContents/style.js
@@ -21,7 +21,6 @@ const StyledItemLink = styled(Link)`
   padding-left: 4px;
   border-left: 3px solid transparent;
   margin-bottom: ${(props) => props.theme.spacing(0.25)}px;
-  white-space: nowrap;
 
   &:hover {
     border-left: 3px solid #c4c4c4; // TODO: Move to theme.


NickAkhmetov:
The wrapped links look like two separate links at first glance; I suggest we address that by decreasing the line height and compensate for the decreased space between each link by increasing the padding by an equal amount. This would maintain the visual space between individual links while compacting the wrapped links to make them more apparently part of the same element. I believe increasing padding instead of margin would be the safer option to avoid decreasing accessibility by shrinking the clickable/tappable area, though it does increase the size of the selection indicator (we can work around that by making that bar a `::before` pseudo-element instead of a border if need be).

Here's what it looks like with .5rem padding on the top/bottom and a line-height of 1; what do you think?

Not hovered:
image
image


Hovered/active:
image
image

CC @tsliaw

tsliaw:
@NickAkhmetov I agree with your concerns since that was one of my concerns when I proposed this ad-hoc change to fix multi-line links in the toc, and I put this concern on the backburner until it was brought up again or I had time to design a better solution. The reason I did not go with the designs changes you proposed initially is because of the "scrunched-up" look of the link and I prefer to have more consistent spacing for readability. 

Since this has been bought up again, I propose not fixing this in this PR, but if you can file an issue in Jira I can work on this within this sprint to come up with the exact spacing or to add dividers between the links or another solution.

NickAkhmetov:
@tsliaw That works for me. I created https://hms-dbmi.atlassian.net/browse/HMP-299 to track this concern.

NickAkhmetov (approved):

PR #3169, opened 2023-07-13
Nickakhmetov/HMP-215 Add tooltips to publication page

NickAkhmetov:
This PR adds explanatory tooltips to the publication page. I used my other PR as the base branch to avoid conflicts since both edit the publication summary.
 

https://github.com/hubmapconsortium/portal-ui/assets/19957804/6aaecd52-e182-476b-a226-9e25d4fb49a9

PR #3195, opened 2023-07-21
Kevin yoo/hmp 80 implement revisions

kevinyooky:


context/app/static/js/pages/search/style.js
```diff
@@ -1,8 +1,20 @@
...
   margin-bottom: ${(props) => props.theme.spacing(2)}px;
 `;
 
-export { SearchHeader };
+const StyledSvgIcon = styled(SvgIcon)`
+  font-size: 2.5rem;
+  margin-right: ${(props) => props.theme.spacing(0.5)}px;
+`;
+
+const EntityHeader = styled.div`
```
NickAkhmetov, line 14: Another duplicate

john-conroy:
I'm reviewing this now, but going forward can you try to break up your work across a few smaller PRs? You don't need to complete a ticket in a single PR. Smaller PRs are much easier to review.

If you could add screenshots as well - those always help!

NickAkhmetov (changes requested):
Hey Kevin! Great work on getting all of these changes in!

I have some suggestions to help improve the long-term maintainability of these changes which might also help with future development. We can go over them after the call today if there's anything you'd like me to clarify, but I wrote out my full thoughts below. Apologies for the gargantuan comment 😅 

---

There are a lot of duplicated stylings/mappings which we should try to shrink down to a "single source of truth" - the entity icon mappings and `StyledEntityHeader` in particular stood out to me. It's considered a best practice to avoid this sort of duplication because it means that any future changes to these components will need to touch a lot more files;.
- For example, if we want to increase the space between the header text and the SVG in the future, we will need to edit `StyledEntityHeader` in seven different files. This introduces the potential for one of those components to be missed during development and during review, and for the spacing to end up being off for one section as a result.
- Ditto with the entity icon mapping object; while we'd be able to handle updates to the individual icons by updating e.g. the DonorIcon in the icons file that it's being imported from, what if the keys being requested from the mapping change? 
  - For example if we have a section that's currently only on the Sample and Donor pages (and thus only has those two icons in its local mapping), that needs to be added to the Support page, we now need to add an extra case to that mapping; whereas if we have a unified mapping that has the icons for all the entity types which we can just import, then that case will already be handled.

My suggestions for fixing each set of comments:

### Extend SectionHeader to include conditional icon styles when `iconTooltipText` is provided instead of defining duplicate components
https://github.com/hubmapconsortium/portal-ui/blob/35ff03cfae202142fae97957143e97982c2226f6/context/app/static/js/shared-styles/sections/SectionHeader.jsx#L12

  - For the `SectionHeader` duplication my suggested fix would be to conditionally apply the styling related to positioning the SVG icon to the `StyledTypography` component in `SectionHeader.jsx`, and make use of the `iconTooltipText` prop I added instead of importing `SecondaryBackgroundTooltip` and `InfoIcon` separately in each instance where they're needed. 
  - The same way we can use `props` inside of styled components to access the theme, we can use it to access any other prop passed in the JSX.
    - So if we add the prop `$hasTooltip={Boolean(iconTooltipText)}` to the `<StyledTypography variant="h4" component="h2" {...rest}>` on line 14 of `SectionHeader.jsx`, we can access the `$hasTooltip` boolean in the `styled` component above via `${(props) => props.$hasTooltip && ... ` and conditionally apply the set of styles you're applying in all the duplicate `StyledSectionHeader` functions using the `css` export from styled-components
  - This pattern can be seen in action in the `FaceItemLabelCount`'s `Flex` component, which has an `$active` prop that is used in this manner:
     - Definition: https://github.com/hubmapconsortium/portal-ui/blob/1eb27f38669b9f8d647b599a266d7466dbf00463/context/app/static/js/components/entity-search/facets/select/FacetItemLabelCount/style.js#L4
     - In use: https://github.com/hubmapconsortium/portal-ui/blob/1eb27f38669b9f8d647b599a266d7466dbf00463/context/app/static/js/components/entity-search/facets/select/FacetItemLabelCount/FacetItemLabelCount.jsx#L8
 - Whenever I find myself copying the same pattern into the same component in multiple places (i.e. adding the secondary background tooltip/info icon inside all these SectionHeaders), I try to approach the situation by extending the component with an optional prop. Since we had a lot of SectionHeaders that needed tooltips, adding the `iconTooltipText` prop was the approach I took in #3169. 

### Use a shared `entityIcons` mapping

- Right now, this is being defined in 4 separate places
- There's actually already an `entityIconMap` in `shared-styles/icons` with all of these icons: 
  - https://github.com/hubmapconsortium/portal-ui/blob/195f30b65a2abffbffcffd59345efbc5aec9b5d5/context/app/static/js/shared-styles/icons/entityIconMap.js#L3
- But in general, when you find yourself defining more or less the same thing in multiple places, you can put it in its own file, export it, and import it where needed - it's like copy-pasting without needing to actually copy paste!

### Object maps vs. Template Literals

- Object maps are really useful for organizing cases where the different keys in the object have very different values
- When the only thing that changes in a string is one word, and that same word is used as the key, template literals are a good way to interpolate that word into the string.

For example:

```
  const tooltipText = `This tooltip is for ${entity_type.toLowerCase()} entities.`;
```
vs.
```
  const tooltipTexts = {
    Donor: "This tooltip is for donor entities.",
    Sample: "This tooltip is for sample entities.",
    Dataset: "This tooltip is for dataset entities.",
    Publication: "This tooltip is for publication entities.",
    Support: "This tooltip is for support entities."
  };
  const tooltipText = tooltipText[entity_type];
```

### `EntityHeader`
Besides deduplicating the definitions for this, I strongly suggest we avoid using this specific name since it's the name we already use for the header that appears as we scroll down the page: https://github.com/hubmapconsortium/portal-ui/tree/kevin-yoo/HMP-80-Implement-revisions/context/app/static/js/components/detailPage/entityHeader 

Renaming it to `TableEntityHeader` should work; I'd also suggest we find a logical shared directory to put it in so we don't have separate definitions.

### `!important`
The only time I recommend using this escape hatch is to override library styles you otherwise have no control over. as is the case in the other situations where it's currently in use in the codebase; otherwise, with how unstoppable/unoverrideable `!important` is, with heavy use you can end up with situations where the fix for one edge case breaks another edge case and you get stuck in a catch-22 where the only way out is to redo the styling altogether.

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
```diff
@@ -5,19 +5,36 @@ import { useTheme } from '@material-ui/core/styles';
...
+    },
+    {
+      icon: SampleIcon,
+      name: 'Sample',
+    },
+    {
+      icon: DatasetIcon,
+      name: 'Dataset',
+    },
+  ];
```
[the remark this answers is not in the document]
john-conroy: It already exists, you can use `import { entityIconMap } from 'js/shared-styles/icons/entityIconMap';`.

context/app/static/js/components/Header/HeaderContent/style.js
```diff
@@ -1,14 +1,18 @@
...
   flex-grow: 1;
 `;
 
 const HeaderButton = styled(Button)`
-  margin-left: 10px;
   color: ${(props) => props.theme.palette.white.main};
+
+  button {
+    padding: 0px;
+  }
```
john-conroy, line 15: I'm not sure this css element selector is having an effect. Since we are specifying `component={Link}` the element is actually an `a` and there is a `span` wrapping the text. I would suggest inspecting the elements in dev tools.

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
```diff
@@ -27,10 +44,15 @@ function HeaderContent({ anchorRef }) {
       {!shouldDisplayMenu && (
         <>
           <FlexNoWrap>
-            {['Donor', 'Sample', 'Dataset'].map((type) => (
-              <HeaderButton key={type} href={`/search?entity_type[0]=${type}`} component={Link}>
-                {`${type}s`}
-              </HeaderButton>
+            {types.map((type) => (
+              <HeaderType>
+                <div>
```
john-conroy, line 49: If you remove this `div` and the `margin-top` in `StyledSvgIcon` the appearance is the same.

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
```diff
@@ -27,10 +44,15 @@ function HeaderContent({ anchorRef }) {
...
+                  <StyledSvgIcon component={type.icon} />
+                </div>
+                <HeaderButton key={type.name} href={`/search?entity_type[0]=${type.name}`} component={Link}>
```
john-conroy, line 52: As it is now, just the link only includes the text.

@tsliaw do we want the icons to be included in each link (clickable)?

Screen Shot 2023-07-21 at 12 15 17 PM

context/app/static/js/components/detailPage/provenance/ProvSection/ProvSection.jsx
```diff
@@ -1,12 +1,15 @@
...
 import useProvData from 'js/hooks/useProvData';
 import { Alert } from 'js/shared-styles/alerts';
-import SectionHeader from 'js/shared-styles/sections/SectionHeader';
+import { SecondaryBackgroundTooltip } from 'js/shared-styles/tooltips';
 import { DetailPageSection } from 'js/components/detailPage/style';
+import { StyledSectionHeader } from './style';
 import ProvTabs from '../ProvTabs';
 
-function ProvSection({ iconTooltipText }) {
+function ProvSection() {
```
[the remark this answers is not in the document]
tsliaw: @NickAkhmetov I agree with you, this is one of the tooltips I had trouble coming up with the language for. The main reason for this tooltip is because some users didn't know the definition of provenance when they were on that section. Do you have any suggestions for improving this tooltip's language?

context/app/static/js/components/detailPage/Protocol/Protocol.jsx
```diff
@@ -22,6 +24,16 @@ function ProtocolLink({ title, resolverHostnameAndDOI }) {
 }
 
 function Protocol({ protocol_url }) {
+  const {
+    entity: { entity_type },
+  } = useFlaskDataContext();
+
+  const tooltipTexts = {
```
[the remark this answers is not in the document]
john-conroy: @tsliaw

context/app/static/js/components/detailPage/provenance/ProvSection/ProvSection.jsx
```diff
@@ -16,14 +19,24 @@ function ProvSection({ iconTooltipText }) {
   if (isLoading) {
     return (
       <DetailPageSection id="provenance">
-        <SectionHeader>Provenance</SectionHeader>
+        <StyledSectionHeader>
+          Provenance
+          <SecondaryBackgroundTooltip title="The provenance shows the sequence of events and actions that led to this page creation.">
```
john-conroy, line 24:
```suggestion
          <SecondaryBackgroundTooltip title="The provenance shows the sequence of events and actions that led to this page's creation.">
```
@tsliaw ?

kevinyooky:
Okay, will break these down into smaller pull requests moving forward!

kevinyooky:
Thanks Nik always for taking the time to do this!

context/app/static/js/components/detailPage/provenance/ProvTable/ProvTable.jsx
```diff
@@ -21,13 +24,22 @@ function ProvTable() {
     { Donor: [], Sample: [], Dataset: [] },
   );
 
+  const entiitesIcons = {
```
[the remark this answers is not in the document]
john-conroy: `import { entityIconMap } from 'js/shared-styles/icons/entityIconMap';`

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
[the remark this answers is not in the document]
kevinyooky: Oh I see, okay. WIll do.

context/app/static/js/components/detailPage/provenance/ProvTableDerivedLink/style.js
```diff
@@ -6,8 +6,6 @@ const LinkButton = styled.a`
   background-color: ${(props) => props.theme.palette.primary.main};
   font-size: ${(props) => props.theme.typography.subtitle2.fontSize};
   font-weight: ${(props) => props.theme.typography.subtitle2.fontWeight};
-  margin-top: ${(props) => props.theme.spacing(1)}px;
-  width: 175px;
```
john-conroy, line 10: What motivated this change? I'm somewhat worried about unintended consequences of removing a hard coded width.

context/app/static/js/shared-styles/tabs/Tab/style.js
```diff
@@ -3,6 +3,14 @@ import Tab from '@material-ui/core/Tab';
 
 const StyledTab = styled(Tab)`
   min-height: 45px;
+
+  span {
+    flex-direction: row;
+  }
+
+  svg {
+    margin-bottom: 0 !important;
```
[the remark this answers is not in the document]
john-conroy: It's worth noting that the components in `shared-styles` are used repeatedly through the application for different purposes. Did you check to make sure that these changes work for every instance of `Tab`?

CHANGELOG-HMP-80-implement-revisions.md
```diff
@@ -0,0 +1 @@
+- Added tooltips and icons throughout the project.
```
john-conroy, line 1:
```suggestion
- Add tooltips and icons throughout the project. 
```

john-conroy (changes requested):
I agree with Nick's comments. Smaller PRs please. It is easier for us to provide feedback and, in turn, you don't have to deal with 20+ comments at once.

I'm not sure what happened, but a majority of the child tasks in [HMP-80](https://hms-dbmi.atlassian.net/jira/software/projects/HMP/boards/66?selectedIssue=HMP-80&text=hmp-80) were marked as done. These tasks shouldn't be done until they are released to production.

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
tsliaw: @john-conroy Yes that entire area, including the icons, should be clickable

context/app/static/js/components/detailPage/provenance/ProvTableDerivedLink/style.js
kevinyooky: The button size was fixed width which was affecting the alignment with the text above it.

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
john-conroy: @tsliaw do you want margin between the links? In this implementation padding is used so there wouldn't be any 'un-clickable' space between the links.

kevinyooky:
Hey guys, regarding what happened with the PR. 

I did not think that creating a PR for every subtask was necessary as some of them were very similar to each other. It would have been excessive to create a PR for each of the dozen tasks. 

At the same time, now that you mention it John, I was supposed to follow the convention of treating each HMP as a task, and not simply marking them as done. I am realizing now that I should have merged some of the subtasks together into 1. That way I would have ended up with 4 PR's which is more reasonable than 12 or so. 

I think I started the task and fell into the rabbit hole of completing everything in the same branch. I will make sure to do that in the future. My apologies.

context/app/static/js/components/detailPage/provenance/ProvSection/ProvSection.jsx
tsliaw: Yes please add the apostrophe. Will fix in the Figma.

context/app/static/js/shared-styles/tabs/Tab/style.js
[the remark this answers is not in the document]
kevinyooky: That's a really good point John, this was a bit careless on my end. Whether it works in this case or not, I did not take that into consideration and check for other instances of tab. Will be more careful with styles that are used in other places.

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
kevinyooky: That's a good point, I think margin might be more appropriate in this case. Going to make that change unless Tiffany says otherwise

kevinyooky:
The main takeaways I got from your comments that I will try to implement better in the future: 

1.) DO NOT CREATE DUPLICATE CODE
- I will do my best to see if I could create a shared style that could be reused throughout the codebase

2.) DOUBLE CHECK IF A SIMILAR COMPONENT STYLE ALREADY EXISTS TO PREVENT CREATING DUPLICATES
- Before I begin created a new styled component, I will make sure to double check if something similar already exists. Otherwise, simply cluttering the codebase. 

3.) BE MINDFUL OF SIDE EFFECTS OF SHARED STYLES
- This one was actually really careless, and a less forgiving mistake. I will make sure to double check if there are any side effects from styles that are shared among multiple files. 

4.) EVERY HMP SHOULD BE TREATED AS ITS OWN PR. DO NOT MARK AS DONE
- Will not simply mark as done. Will gauge subtasks first and merge them together if necessary before starting them off.

context/app/static/js/components/detailPage/Protocol/Protocol.jsx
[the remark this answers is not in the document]
tsliaw: @NickAkhmetov I used the word `provided` since I was thinking of how the data submitter "provides" the protocols to protocols.io, but I see the confusion now with how I worded it in this tooltip. I'm not sure about the word `hosted` either. Does `Protocols provided for the given {entity_type}` or `Protocols from protocols.io uploaded for the given {entity_type} or another wording work better?

PR #3198, opened 2023-07-26
kevin-yoo/HMP-302-Implement-icons-for-entity-headers-and-tables-Swedlow-feedback-response

kevinyooky:
The following tasks were in the story for this PR:

1. Add icon to entity header
 
  This task had already been completed in the past so there is no additional code added. 
  image

2. Add entity icons to header links to search pages
  
  As mentioned in the previous pull request (https://github.com/hubmapconsortium/portal-ui/pull/3195) Tiffany requested that the entire space be clickable so padding is used for the div around the icon and text instead of margin. 

  image

3. Add entity icons to provenance table tab for entity headers
  image

4. Add sample and dataset icon to tabs in "Derived Sample and Datasets"

I added additional comment at the bottom in response to Nik's suggestion. 

I understand that it is not ideal to use !important, but the custom styling I was giving was getting overwritten by MUI. If there is a better way to handle this please let me know!

image
image

Before:

image

After: 

image

### DIFF
--- a/context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
+++ b/context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
@@ -4,6 +4,7 @@ import Link from '@material-ui/core/Link';
 import { useTheme } from '@material-ui/core/styles';
 import useMediaQuery from '@material-ui/core/useMediaQuery';
 
+import { entityIconMap } from 'js/shared-styles/icons/entityIconMap';
 import { useAppContext } from 'js/components/Contexts';
 import Menu from '../Menu';
 import ResourceLinks from '../ResourceLinks';
@@ -11,13 +12,14 @@ import Dropdown from '../Dropdown';
 import UserLinks from '../UserLinks';
 import AtlasToolsLinks from '../AtlasToolsLinks';
 import OtherLinks from '../OtherLinks';
-import { Spacer, HeaderButton, FlexNoWrap } from './style';
+import { Spacer, HeaderButton, FlexNoWrap, StyledSvgIcon, HeaderType } from './style';
 import HubmapLogo from '../HubmapLogo';
 
 function HeaderContent({ anchorRef }) {
   const theme = useTheme();
   const shouldDisplayMenu = !useMediaQuery(theme.breakpoints.up('md'));
   const { isAuthenticated, userEmail } = useAppContext();
+
   return (
     <>
       {shouldDisplayMenu && <Menu anchorRef={anchorRef} />}
@@ -28,9 +30,12 @@ function HeaderContent({ anchorRef }) {
         <>
           <FlexNoWrap>
             {['Donor', 'Sample', 'Dataset'].map((type) => (
-              <HeaderButton key={type} href={`/search?entity_type[0]=${type}`} component={Link}>
-                {`${type}s`}
-              </HeaderButton>
+              <HeaderType>
+                <StyledSvgIcon component={entityIconMap[type]} />
+                <HeaderButton key={type} href={`/search?entity_type[0]=${type}`} component={Link}>
+                  {`${type}s`}
+                </HeaderButton>
+              </HeaderType>
             ))}
             <Dropdown title="Other">
               <OtherLinks />
--- a/context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
+++ b/context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
@@ -12,7 +12,7 @@ import Dropdown from '../Dropdown';
 import UserLinks from '../UserLinks';
 import AtlasToolsLinks from '../AtlasToolsLinks';
 import OtherLinks from '../OtherLinks';
-import { Spacer, HeaderButton, FlexNoWrap, StyledSvgIcon, HeaderType } from './style';
+import { Spacer, HeaderButton, FlexNoWrap, StyledSvgIcon } from './style';
 import HubmapLogo from '../HubmapLogo';
 
 function HeaderContent({ anchorRef }) {
@@ -30,12 +30,14 @@ function HeaderContent({ anchorRef }) {
         <>
           <FlexNoWrap>
             {['Donor', 'Sample', 'Dataset'].map((type) => (
-              <HeaderType>
-                <StyledSvgIcon component={entityIconMap[type]} />
-                <HeaderButton key={type} href={`/search?entity_type[0]=${type}`} component={Link}>
-                  {`${type}s`}
-                </HeaderButton>
-              </HeaderType>
+              <HeaderButton
+                key={type}
+                href={`/search?entity_type[0]=${type}`}
+                component={Link}
+                startIcon={<StyledSvgIcon component={entityIconMap[type]} />}
+              >
+                {`${type}s`}
+              </HeaderButton>
             ))}
             <Dropdown title="Other">
               <OtherLinks />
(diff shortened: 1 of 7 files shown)


kevinyooky:
image

kevinyooky:
Entity icons added to provenance table headers

image

kevinyooky:
I am going to double check one more time before I convert this from draft to pull request.

kevinyooky:
image
image

I understand that it is not ideal to use !important, but the custom styling I was giving was getting overwritten by MUI. If there is a better way to handle this please let me know!

NickAkhmetov:
Conflicting CSS rules are weighted by specificity - ID's (#icon, #icon-example) > Classes, attributes, and pseudoclasses (.icon, [type="radio"], :first-child) > Elements and pseudoelements (div, span, ::before)

https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

Your rule is turned into the following CSS selector: `.csypIA svg` (which has a weight of 0-1-1 - there's one class and one element type).

The MUI rule is `.portal-MuiTab-labelIcon-221 .portal-MuiTab-wrapper-229 > :first-child`, which has a weight of 0-3-0 as it contains two classes and a pseudoclass; so, it takes precedence over the rule you defined.

It looks like MUI offers a few ways to work around these styling conflicts without introducing `!important`: https://v4.mui.com/customization/components/#overriding-styles-with-classes

kevinyooky:
Thanks Nik, will take a look at this.

kevinyooky:
I've been trying multiple methods with no success. This my most current implementation to avoid using !important. Nik do you see something that is off in the current code that is prevent the styles from applying? 

image
image

NickAkhmetov:
@kevinyooky I think you may need to apply the necessary styles to the Tab component rather than the SVGIcon, since the offending rules are applied by a rule targeting the .MuiTab-wrapper and .MuiTab-labelIcon classes.

john-conroy:
Not sure if it's the most exact approach, but you can do what styled-components suggests [to increase specificity](https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity).

Instead of introducing a new span wrapper, I would style the tab component directly. You can then remove the wrapper component and change `Tab` to `StyledIconTab`.

In `style.js`:
```
const StyledIconTab = styled(Tab)`
  span {
    flex-direction: row;
  }

  &&& svg {
    margin-bottom: 0px;
  }
`;
```
In `RelatedEntitiesTabs`:

```
<StyledTabs value={openIndex} onChange={handleChange} aria-label={ariaLabel}>
        {entities.map((entity, i) => (
          <StyledIconTab
            label={`${entity.tabLabel} (${entity.data.length})`}
            index={i}
            key={entity.tabLabel}
            data-testid={`${entity.tabLabel.toLowerCase()}-tab`}
            icon={
              <StyledSvgIcon
                component={entityIconMap[`${entity.tabLabel.slice(0, entity.tabLabel.length - 1)}`]}
                color="white"
              />
            }
          />
        ))}
      </StyledTabs>
```

john-conroy:
The link also recommends a plugin we could use to increase styled-components specificity globally by automatically using a set number of these `&` selectors.

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
```diff
@@ -28,9 +30,12 @@ function HeaderContent({ anchorRef }) {
...
             {['Donor', 'Sample', 'Dataset'].map((type) => (
-              <HeaderButton key={type} href={`/search?entity_type[0]=${type}`} component={Link}>
-                {`${type}s`}
-              </HeaderButton>
+              <HeaderType>
+                <StyledSvgIcon component={entityIconMap[type]} />
+                <HeaderButton key={type} href={`/search?entity_type[0]=${type}`} component={Link}>
+                  {`${type}s`}
+                </HeaderButton>
+              </HeaderType>
```
john-conroy, line 38:
```suggestion
              <HeaderButton
                key={type}
                href={`/search?entity_type[0]=${type}`}
                component={Link}
                startIcon={<StyledSvgIcon component={entityIconMap[type]} />}
              >
                {`${type}s`}
              </HeaderButton>
```

john-conroy: There is a `startIcon` available to https://v4.mui.com/api/button/. You can then remove the `HeaderType` styles.

context/app/static/js/components/detailPage/related-entities/RelatedEntitiesTabs/RelatedEntitiesTabs.jsx
```diff
@@ -11,16 +12,24 @@ function RelatedEntitiesTabs({ entities, openIndex, setOpenIndex, ariaLabel, ren
...
+        <StyledTabs value={openIndex} onChange={handleChange} aria-label={ariaLabel}>
+          {entities.map((entity, i) => (
+            <Tab
+              label={`${entity.tabLabel} (${entity.data.length})`}
+              index={i}
+              key={entity.tabLabel}
+              data-testid={`${entity.tabLabel.toLowerCase()}-tab`}
+              icon={
+                <StyledSvgIcon
+                  component={entityIconMap[`${entity.tabLabel.slice(0, entity.tabLabel.length - 1)}`]}
```
john-conroy, line 25:
```suggestion
                  component={entityIconMap[entity.entityType]}
```
I believe `entityType` should exist. Make sure to double check the instances for this component.

john-conroy (changes requested):
Looks good! A few areas to cleanup.

kevinyooky:
@NickAkhmetov, I actually tried that very thing as well to no avail at multiple levels of the parent component, but will check one more time.

NickAkhmetov:
@kevinyooky I think John's idea is worth trying - if there's a built-in mechanism to override specificity suggested by styled-components, that should work (though I never knew that including the same class as part of the selector multiple times works like that!)

context/app/static/js/components/Header/HeaderContent/HeaderContent.jsx
kevinyooky: That makes it much simpler, implemented the change. Double checked and it appears this is the only instance I need to make that change. Thanks!

context/app/static/js/components/detailPage/related-entities/RelatedEntitiesTabs/RelatedEntitiesTabs.jsx
kevinyooky: Ah I see, I still had access to entityType...

kevinyooky:
It worked... Thanks to both of you for the input.

context/app/static/js/components/detailPage/related-entities/RelatedEntitiesTabs/style.js
```diff
@@ -11,11 +13,24 @@ const StyledTabPanel = styled(TabPanel)`
...
 const StyledTabs = styled(Tabs)`
   flex: none;
+
+  span {
+    display: contents;
+  }
+
+  &&& svg {
+    margin-bottom: 0;
+  }
```
john-conroy, line 23: This is slightly different to what I suggested, but that's okay. 

I generally prefer to style components directly (creating separate styled components as necessary) instead of leveraging css selectors. If we do have to use css selectors, like we do here, I try to make sure we limit the scope of the css as much as possible.

In this case, we are trying to apply styles to span and svg elements inside each `Tab`. I would apply the styles in a new styled `Tab`.

john-conroy: Here's a brief discussion about the topic https://www.robinwieruch.de/styled-components/#singlemultiple-styled-components.

kevinyooky: It worked :)

NickAkhmetov (approved):
👍 Looks good to me! Nice work.

john-conroy (approved):

PR #3200, opened 2023-07-26
John conroy/fix protocols errors HMP-306

john-conroy:
Fixes the handling of loading and errors for protocols.io requests in protocols section. Previously the non-200s were not being treated as errors and the resulting display matches what is shown in https://hms-dbmi.atlassian.net/browse/HMP-306. That being said, the requests were only failing locally due to our out of date `app.conf` files. Requests on PROD should work.

I'm not sure if the mismatched key passed to `protocolsContext` in `Providers` would have caused any issues ( it does not seem like we're using this context anywhere?), but I fixed it to match the types defined in `ProctocolAPIContext`. Should typescript have picked this up?

I've also added `HelpLink` to be used throughout the application. I've filed https://hms-dbmi.atlassian.net/browse/HMP-307.

200

Screen Shot 2023-07-26 at 11 21 26 AM

Non-200

Screen Shot 2023-07-26 at 11 21 52 AM

### DIFF
--- a/CHANGELOG-fix-protocols-errors.md
+++ b/CHANGELOG-fix-protocols-errors.md
@@ -1,0 +1,1 @@
+- Display message when protocols.io requests fail in protocols section.
--- a/context/app/static/js/components/Providers.jsx
+++ b/context/app/static/js/components/Providers.jsx
@@ -40,7 +40,7 @@ function Providers({
   );
 
   const protocolsContext = useMemo(
-    () => ({ protocolsClientId: flaskData?.protocolsClientId, clientAuthToken: flaskData?.protocolsClientToken }),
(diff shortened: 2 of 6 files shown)


NickAkhmetov:
Great work, I like this approach!

My only idea for potentially improving this further would be to also allow a custom error message function to be passed which would allow us to provide more in-depth error messages depending on the received response; e.g., it might be more informative to the user to display different error messages for 401/403/404s, or handle custom error messages inside response bodies when APIs don’t conform to standard status codes (protocols doesn’t throw 403s for unauthorized access, for example.) However, we can add that functionality at a later time, so it’s not a blocking issue whatsoever.

PR #3202, opened 2023-07-27
John conroy/fix storybook typescript

john-conroy:
Storybook was failing to load as it was treating typescript files as javascript since it compiles with babel by default. Configuring storybook to use `swc` with `storybook-addon-swc` seems to fix it.


NickAkhmetov (approved):
👀 good catch!

lchoy (approved):
Thanks!

PR #3206, opened 2023-07-31
Kevin yoo/hmp 241 improve error messaging for deleted user saved list

kevinyooky:
I changed the error message.

image


NickAkhmetov (changes requested):
1. With a large try/catch block, we lose clarity as to what is causing this error when reading the code.
 - We also include a lot more in the scope of the try/catch than intended: if a child component throws an entirely unrelated error, the `This list does not exist` error message will be shown instead, which will make troubleshooting that unrelated error much more difficult.
 - I suggest testing for the condition which causes the original error before the error is thrown in the code and `throw`ing the custom error message there, similarly to how you've displayed loading states in other PR's.
2. The second half of the error message is not present.

kevinyooky:
Hi Nik, thanks for that feedback! I think this implementation addresses all the points you made :)

image
image

NickAkhmetov (dismissed):
That's exactly the approach I had in mind, well done! 🎉 

---

The second half of the error message that @tsliaw indicated in the ticket isn't there, but the text that is available already has the email link and is pretty similar overall (`if this problem persists, submit a bug report` vs `If this list should exist, please message`). We could consider extending the error page logic to allow that text to be edited but that may be somewhat tricky given that throwing errors only lets us send one message up to the error boundary with the way things are currently implemented. 

I'll leave the importance of that second line of copy up to her discretion.

tsliaw:
No need to include the second part of the message since it's already there!

NickAkhmetov:
@lchoy I agree, that would be a good idea. Implementing it may be a bit tricky since `savedLists` is in a `zustand` store.

I think it should be possible to mock the `useSavedEntitiesStore` hook and make it return `savedLists` as an empty object, but if that doesn't work as expected we may need to mock the creation of the store itself, which may require some jest config changes to get working properly: https://docs.pmnd.rs/zustand/guides/testing#jest

PR #3212, opened 2023-08-02
Nickakhmetov/HMP-317 HMP-318 fix font sizes

NickAkhmetov:
This PR restores the previous global default font sizes to address sizing issues on components which were previously using these default styles. It also updates the Diversity page to use the global PageTitle component

Fixed SRA experiment sizes: 
image

Fixed diversity header:
image

### DIFF
--- a/CHANGELOG-mui-qa-fixes-1.md
+++ b/CHANGELOG-mui-qa-fixes-1.md
@@ -1,1 +1,0 @@
-- Restore MUI v4 font size.
--- a/CHANGELOG-mui-qa-fixes-1.md
+++ b/CHANGELOG-mui-qa-fixes-1.md
@@ -1,0 +1,1 @@
+- Restore MUI v4 font size.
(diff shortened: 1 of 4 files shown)


john-conroy (commented):
I'm a little confused. Isn't the default size for mui v4 typography `body1`?

NickAkhmetov:
@john-conroy Per [MUI docs](https://mui.com/material-ui/migration/v5-component-changes/#update-body-font-size)

> The body font size has changed from theme.typography.body2 (0.875rem) to theme.typography.body1 (1rem). To return to the previous size, you can override it in the theme

Since our `body2` size is .8 rem instead of .875, I stuck with that one.

john-conroy (approved):
Thanks. I guess I'm unclear on how `body` is used by MUI, but that shouldn't block.

NickAkhmetov:
>I'm unclear on how body is used by MUI, but that shouldn't block.

These styles are applied by MUI's `CssBaseline` component in our `Providers`; essentially, they're used by any text that's not inside of a `Typography` component, such as the `Here is additional documentation` part of the SRA experiment link and the right side of search panels (I had addressed that case by adding the style myself during the merge before I was aware of what's going on)

PR #3215, opened 2023-08-04
Kevin yoo/hmp 157 add error messages to workspace creation dialog 10 datasets protected datasets selected

kevinyooky:
**This branch is the main branch for this feature. I have chained other branches from this one to prevent PRs from getting too large.** 

image


context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
```diff
@@ -10,16 +10,19 @@ function CreateWorkspaceDialog({ handleCreateWorkspace, buttonComponent: ButtonC
...
       <DialogModal
-        title="Create New Workspace"
+        title="Launch New Workspace"
+        // title={`Launch New Workspace: `}
         isOpen={dialogIsOpen}
         handleClose={handleClose}
         maxWidth="md"
         content={
           <form onSubmit={handleSubmit(onSubmit)}>
+            <br />
```
kevinyooky, line 25: Would adding a break here be acceptable?

NickAkhmetov: I prefer to avoid using HTML for styling, if that makes sense - if the goal is to get more of a margin on the top part of the farm, a margin-top on the form would preferable.

The MUI Box would work well here, in my opinion: https://mui.com/material-ui/react-box/

Using the `component` prop, you could make the Box act as a form and apply some margin or padding using the `sx` prop.

context/app/static/js/shared-styles/DialogModal/DialogModal.jsx
```diff
@@ -19,6 +21,7 @@ function DialogModal({
   ...props
 }) {
   const DialogContent = DialogContentComponent || MUIDialogContent;
+  const { selectedRows } = useStore();
```
NickAkhmetov, line 24: Rather than adding case-specific code to the shared `DialogModal`, what if we add an `error` prop and display the `Alert` if it's not null? This would allow us to handle other error cases the same way in the future and would keep this component more generic/reusable. It would also let you e.g. disable submission of the create workspace form if the `selectedRows.size` condition is met.

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
kevinyooky: Thanks for that input!

context/app/static/js/shared-styles/DialogModal/DialogModal.jsx
kevinyooky: Done

context/app/static/js/components/workspaces/CreateWorkspaceInput/CreateWorkspaceInput.jsx
kevinyooky, line 1: The additional space in between the quotes was creating an empty space in the dialog.

context/app/static/js/pages/SavedList/SavedList.jsx
kevinyooky, line 1: This was to fix the build error which required the provider wrapping around savedList in order to access the store.

context/app/static/js/shared-styles/DialogModal/DialogModal.jsx
```diff
@@ -28,6 +30,11 @@ function DialogModal({
         </Typography>
       </StyledDialogTitle>
       <DialogContent>
+        {errorMessage && (
+          <Alert $marginBottom={20} severity="error">
```
kevinyooky, line 34: Passed in marginBottom prop for ALert

context/app/static/js/pages/SavedList/SavedList.jsx
NickAkhmetov: The `SelectableTableProvider` component should be usable in place of the raw `Provider` and `createStore` calls: https://github.com/hubmapconsortium/portal-ui/blob/main/context/app/static/js/shared-styles/tables/SelectableTableProvider/SelectableTableProvider.jsx#L6

In general though, I'm pretty sure this should be the same store instance that the selected list is saved in; otherwise, you have two instances of the state that may not necessarily be synchronized. 

The instance of the zustand store that the saved entities table uses is created as part of the default export: https://github.com/hubmapconsortium/portal-ui/tree/main/context/app/static/js/components/savedLists/SavedEntitiesTable/SavedEntitiesTable.jsx#L130

I think it would make sense to import the version of `SavedEntitiesTable` that doesn't have that provider exported with it in this file so that you can guarantee the list and error logic are using the same store instance.

NickAkhmetov (commented):
Great progress! A few minor notes on my end.

context/app/static/js/shared-styles/DialogModal/DialogModal.jsx
kevinyooky: That makes sense.

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
```diff
@@ -32,6 +42,11 @@ function CreateWorkspaceDialog({ handleCreateWorkspace, buttonComponent: ButtonC
             </label>
           </>
         }
+        errorMessage={
+          selectedRows.size > 10
+            ? `You have selected ${selectedRows.size} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`
+            : null
```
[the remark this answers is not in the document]
kevinyooky: I SEE

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
```diff
@@ -1,28 +1,39 @@
...
+import { useStore } from 'js/shared-styles/tables/SelectableTableProvider/store';
 import { useCreateWorkspace } from './hooks';
 
 function CreateWorkspaceDialog({ handleCreateWorkspace, buttonComponent: ButtonComponent, defaultName, ...rest }) {
   const { dialogIsOpen, setDialogIsOpen, handleSubmit, handleClose, control, errors, onSubmit } = useCreateWorkspace({
     handleCreateWorkspace,
     defaultName,
   });
+  const { selectedRows } = useStore();
+  const errorMessage = selectedRows.size > 10 ? 'you have selected ...' : null;
```
john-conroy, line 16: I think you removed the full error message when adopting Nick's suggestion.

kevinyooky: That was a bit careless, thank you for catching that.

kevinyooky:
image

This was the error I was getting for http://localhost:5001/workspaces because there was no provider wrapping around workspaces. 

Once I wrapped the Provider around the workspaces content, the issue was resolved. 

image

kevinyooky:
Hm okay.

kevinyooky:
@NickAkhmetov, applied your previous comment so that I am not using a generic Provider :)

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
```diff
@@ -1,28 +1,42 @@
...
   const { dialogIsOpen, setDialogIsOpen, handleSubmit, handleClose, control, errors, onSubmit } = useCreateWorkspace({
     handleCreateWorkspace,
     defaultName,
   });
+  const { selectedRows } = useStore();
+  const errorMessage =
+    selectedRows.size > 10
+      ? `You have selected ${selectedRows.size} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`
+      : null;
+
```
john-conroy, line 20: We don't expect there to be a `SelectableTableProvider` parent to every `CreateWorkspaceDialog` you need to access the store in `MetadataMenu` and pass the requisite information as props.

kevinyooky:
Here are some observations I made:

1.) Access level column
2.) Dropdown toggle
3.) Dropdown alignment
4.) persist selected row state
5.) workspace dialog when not logged in
6.) Create workspace when nothing is selected

https://www.loom.com/share/e59bc6888b5f487ba0dd8015b145ee56?sid=831bc03a-cfc8-4d55-b69e-536af398a024

NickAkhmetov (approved):
I think this is good to go! Great work 👏

kevinyooky:
Thanks @NickAkhmetov! However since this branch is the main branch for this feature, I will need to merge the other chained branched into this one first. I should have made that more clear. Updated the description for clarity!

NickAkhmetov:
@kevinyooky Understood! I realized this was the case after I saw the other two extended this PR. I'll dismiss my own approval for now, then 👍

NickAkhmetov:
# Updated demos

Create Workspace dialog with both errors visible:

image

Dismissing the errors reactivates the workspace name field and the launch button:

https://github.com/hubmapconsortium/portal-ui/assets/19957804/efe1ee1f-2f4c-4b88-a5b0-6677144d1b7d

Storybook demo for the four snackbar variants, which can be triggered from anywhere on the site:

https://github.com/hubmapconsortium/portal-ui/assets/19957804/05fec760-73ff-43f2-87f4-2ceb8539d1bc

Selecting a single protected dataset properly handles the non-plural case in the error message:

image

NickAkhmetov:
TODO: Switch over vitessce visualization toasts from sharing config and from viewing vitessce in firefox

context/.storybook/main.js
```diff
@@ -1,28 +1,31 @@
...
+const { alias } = require('../build-utils/alias');
 module.exports = {
-  "stories": ['../app/static/js/shared-styles/**/*.stories.js', '../app/static/js/components/**/*.stories.js'],
-  "addons": ["@storybook/addon-links", "@storybook/addon-essentials", 'storybook-addon-swc'],
-  "webpackFinal": async config => {
+  stories: [
+    '../app/static/js/shared-styles/**/*.stories.js',
+    '../app/static/js/shared-styles/**/*.stories.tsx',
+    '../app/static/js/components/**/*.stories.js',
+    '../app/static/js/components/**/*.stories.tsx',
```
john-conroy, line 8:
```suggestion
    '../app/static/js/**/*.stories.@(js|jsx|ts|tsx)'
```
Perhaps this could simplify it?

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -37,11 +38,38 @@ function NotebookMenuItem(props) {
...
 }
 
-function MetadataMenu({ entityType }) {
+const errorHelper = {
+  datasets: (rowCount) =>
+    `You have selected ${rowCount} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`,
+  protectedDataset: (rows) =>
+    `You have selected a protected dataset (${rows[0]._source.hubmap_id}). Workspaces currently only supports published public datasets. To remove the protected dataset from workspace creation, click the “Remove Protected Datasets” button below or return to the previous screen to manually remove this dataset.`,
+  protectedDatasets: (rows) =>
+    `You have selected ${rows.length} protected datasets. Workspaces currently only supports published public datasets. To remove protected datasets from this workspace creation, click the “Remove Protected Datasets” button below or return to the previous screen to manually remove those datasets.`,
```
john-conroy, line 47: Should we just pass rows for consistency? Both `datasets` and `protectedDatasets` depend on the length of the rows array.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -37,11 +38,38 @@ function NotebookMenuItem(props) {
...
+    errorMessages.push(errorHelper.datasets(selectedRows.size));
+  }
+
+  if (containsProtectedDataset) {
+    if (protectedRows.length === 1) {
+      errorMessages.push(errorHelper.protectedDataset(protectedRows));
+    } else {
+      errorMessages.push(errorHelper.protectedDatasets(protectedRows));
+    }
+  }
```
john-conroy, line 71: We could add a wrapper component around `CreateWorkspaceDialog` and push all of that logic there.

context/app/static/js/components/entity-search/SearchWrapper/utils.js
```diff
@@ -220,6 +220,8 @@ function createSearchkitFacet({ field, identifier, label, type, ...rest }) {
   });
 }
 
+const defaultFields = createField({ fieldName: 'mapped_data_access_level', label: 'mapped_data_access_level' });
```
john-conroy, line 223: Do we still need this? I remember Kevin and I introduced this at some point, but later moved towards making a request with `useDatasetAccessLevel`.

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
```diff
@@ -1,35 +1,92 @@
...
+  const protectedHubmapIds = protectedRows?.map((row) => row._source.hubmap_id).join(', ');
+  const { toastSuccess } = useSnackbarStore();
+  const { deselectRows } = useSelectableTableStore();
+
+  const removeProtectedDatasets = () => {
+    deselectRows(protectedRows.map((r) => r._id));
+    toastSuccess('Protected datasets successfully removed from selection.');
+  };
+
```
john-conroy, line 32: `CreateWorkspaceDialog` is used outside of the search page on both the workspaces page and datasets pages. I don't believe this component will have access to the `SelectableTableProvider` store. We should move this to a component around `CreateWorkspaceDialog` for use on the search page.

john-conroy (changes requested):
Thanks for picking this up! The only blocking comment is the one made in `CreateWorkspaceDialog` - everything else you can take or leave.

NickAkhmetov:
Following the discussion during yesterday's planning meeting, I have converted the Vitessce snackbars to use the new toast mechanism for consistency. I was not able to trigger the `long URL` condition for the vitessce copy functionality (to my understanding, modern browsers all support 8k+ URL lengths and this functionality dates back to IE only allowing 2048 characters) but the logic in the code is set to trigger a warning if a long URL is detected.

I've also set the Vitessce Firefox performance warning to only pop up once for each user via a localStorage key. This functionality was particularly annoying on publication pages where it would pop up every time the expanded vignette changed. 

https://github.com/hubmapconsortium/portal-ui/assets/19957804/2e5ec498-2741-4613-ab8f-aa2ab350367e

I'll start work on addressing John's feedback.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
NickAkhmetov: I decided to pass the size here initially because `selectedRows` is a set, while `protectedRows` is an array; I didn't like `rows.size` in one case and `rows.length` in the other.

I've applied this revision but modified the arg names and added an explanatory comment to ensure clarity re: why this difference is here.

context/app/static/js/components/entity-search/SearchWrapper/utils.js
NickAkhmetov: Thanks for the context - I initially left this intact since I wasn't familiar with what this was doing, and have removed it after testing that it doesn't impact functionality.

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
NickAkhmetov: I've addressed this and the MetadataMenu comment accordingly; I've created a `CreateWorkspaceWithDatasetsDialog` component which wraps the `CreateWorkspaceDialog` component, performs the error checking logic necessary,

context/app/static/js/shared-styles/snackbars/store.ts
```diff
@@ -52,4 +52,14 @@ const createStore = () =>
...
-export { Provider as SnackbarProvider, useStore as useSnackbarStore, createStore };
+const useSnackbarActions = () => {
+  return useStore((store) => ({
+    toastError: store.toastError,
+    toastInfo: store.toastInfo,
+    toastWarning: store.toastWarning,
+    toastSuccess: store.toastSuccess,
+    closeSnackbar: store.closeSnackbar,
+  }));
+};
```
john-conroy, line 63: I lean towards not selecting the entire state by default as each component will update when any of the variables change even if the component does not depend on them. We could implement something like https://docs.pmnd.rs/zustand/guides/auto-generating-selectors if writing selectors ourselves is too much boilerplate.

context/app/static/js/components/detailPage/visualization/Visualization/Visualization.jsx
```diff
@@ -82,27 +67,33 @@ function Visualization({ vitData, uuid, hasNotebook, shouldDisplayHeader, should
...
+      localStorage.setItem(localStorageFirefoxWarningKey, true);
+    }
+  }, [toastError]);
+
+  // Show instructions for exiting full screen mode when the user enters full screen mode.
+  useEffect(() => {
+    if (vizIsFullscreen) {
+      toastInfo('Press [esc] to exit full window.');
+    }
+  }, [vizIsFullscreen, toastInfo]);
```
john-conroy, line 83: I think I prefer to pull related hooks together into a single hook, but it's entirely a preference. Do you prefer to have everything in a single file?

context/app/static/js/shared-styles/snackbars/store.ts
NickAkhmetov: Thanks for the heads up! I agree completely about not selecting the entire state (since that would be more or less identical to using context). The auto generating selectors functionality seems very useful, and would be a solid improvement for us to make once we're on zustand v4.

**TL;DR:** - I can take this helper out and replace it with specific selectors where needed, but I'm not sure how much benefit that would bring in this case since these are just functions.

To my understanding, since the keys accessed/returned by this helper are functions, they shouldn't be updating at all; the idea was that this helper would let us skip making a separate selector for each case where we just need to be able to send a snackbar message without necessarily needing to read it. Additionally, if zustand functions _are_ regenerated on each state update, wouldn't they all be regenerated regardless of whether they're accessed separately or all together in a helper like this?

context/app/static/js/components/detailPage/visualization/Visualization/Visualization.jsx
NickAkhmetov: It's on a case-by-case basis for me, but I lean towards keeping hooks focused on handling a single functionality/behavior:
- If there's an interdependent process where separating it leads to the output of one function being fed into another function, I do like to unify those into one on at least some level (e.g. if the SWR output from one fetch is used to set the params of a second fetch, it's helpful to have a single hook which abstracts that detail out from the component, but still keep the other two SWR calls as their own hooks)
- If it's two different conditions that trigger similar effects, I still keep them separated in order to keep the code maintainable and prevent unnecessary computations
  - RE: `maintainable` - as time goes on, combining more and more business logic into one hook/into one function leads to big fragile monoliths where 
  - RE: `unnecessary computations` - in this case, combining these two `useEffect`s into one would lead to extra `sniffBrowser` executions/potential extra `localStorage` reads whenever the visualization is full-screened; these are trivial operations, but in other circumstances these could be e.g. lookups from an API.

john-conroy (dismissed):

NickAkhmetov:
Note: the fixes in this branch coincide with the changes in #3258 as I initially didn't realize the lint issues came from upstream. If this branch is merged in, we can close that PR.

john-conroy (dismissed):

NickAkhmetov (approved):

PR #3218, opened 2023-08-07
Kevin yoo/implement error message for protected datasets v2

kevinyooky:
This branch is chained from https://github.com/hubmapconsortium/portal-ui/pull/3215 (hmp 157 kevin-yoo/add-error-message). Alignment has also been fixed. 

image

Second scenario is when we have selected protected datasets

image

Third scenario when both error conditions are met

image

### DIFF
--- a/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
+++ b/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
@@ -1,7 +1,8 @@
 import React from 'react';
 // import { trackEvent } from 'js/helpers/trackers';
 
-import SelectableTableProvider, { useStore } from 'js/shared-styles/tables/SelectableTableProvider/store';
+import SelectableTableProvider from 'js/shared-styles/tables/SelectableTableProvider/SelectableTableProvider';
+import { useStore } from 'js/shared-styles/tables/SelectableTableProvider/store';
 import { SecondaryBackgroundTooltip } from 'js/shared-styles/tooltips';
 import withDropdownMenuProvider from 'js/shared-styles/dropdowns/DropdownMenuProvider/withDropdownMenuProvider';
 import DropdownMenu from 'js/shared-styles/dropdowns/DropdownMenu';
--- a/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
+++ b/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
@@ -58,6 +58,7 @@ function MetadataMenu({ entityType, results }) {
   if (containsProtectedDataset) {
     errorMessages.push(
       'You have selected protected datasets. Workspaces currently only supports published public datasets. Please unselect protected datasets.',
+      // The commented out messaging is for version 2 when the protected datasets are displayed along with the error messages for Protected Datasets.
       // 'You have selected protected datasets. Workspaces currently only supports published public datasets. Selected protected datasets are shown below.',
     );
   }
--- a/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
+++ b/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
@@ -45,11 +45,9 @@ function MetadataMenu({ entityType, results }) {
   const menuID = 'metadata-menu';
 
   const { selectedRows } = useStore();
-  const selectedRowsArray = useDatasetsAccessLevel(selectedRows.size > 0 ? [...selectedRows] : []).datasets;
+  const protectedRows = useDatasetsAccessLevel(selectedRows.size > 0 ? [...selectedRows] : []).datasets;
 
-  const containsProtectedDataset = selectedRowsArray.some((item) => {
-    return item._source.mapped_data_access_level === 'Protected'; //eslint-disable-line
-  });
+  const containsProtectedDataset = protectedRows.length > 0;
 
   let errorMessage;
 
--- a/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
+++ b/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
@@ -45,6 +45,7 @@ function MetadataMenu({ entityType }) {
   return (
     <>
       <StyledDropdownMenuButton menuID={menuID}>Metadata</StyledDropdownMenuButton>
+
       <DropdownMenu id={menuID}>
         <StyledMenuItem>
           <StyledLink href={`/lineup/${lcPluralType}`}>Visualize</StyledLink>
@@ -52,6 +53,7 @@ function MetadataMenu({ entityType }) {
             <StyledInfoIcon color="primary" />
           </SecondaryBackgroundTooltip>
         </StyledMenuItem>
+
         <StyledMenuItem
           onClick={() =>
             fetchAndDownload({
@@ -66,6 +68,7 @@ function MetadataMenu({ entityType }) {
             <StyledInfoIcon color="primary" />
           </SecondaryBackgroundTooltip>
         </StyledMenuItem>
+
         <CreateWorkspaceDialog handleCreateWorkspace={createNotebook} buttonComponent={NotebookMenuItem} />
       </DropdownMenu>
     </>
--- a/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
+++ b/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
@@ -49,15 +49,18 @@ function MetadataMenu({ entityType, results }) {
 
   const containsProtectedDataset = protectedRows.length > 0;
 
-  let errorMessage;
+  const errorMessages = [];
 
   if (selectedRows.size > 10) {
-    errorMessage = `You have selected ${selectedRows.size} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`;
-  } else if (containsProtectedDataset) {
-    errorMessage =
-      'You have selected protected datasets. Workspaces currently only supports published public datasets. Selected protected datasets are shown below.';
-  } else {
-    errorMessage = null;
+    errorMessages.push(
+      `You have selected ${selectedRows.size} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`,
+    );
+  }
+
+  if (containsProtectedDataset) {
+    errorMessages.push(
+      'You have selected protected datasets. Workspaces currently only supports published public datasets. Selected protected datasets are shown below.',
+    );
   }
 
   return (
@@ -87,7 +90,7 @@ function MetadataMenu({ entityType, results }) {
         <CreateWorkspaceDialog
           handleCreateWorkspace={createNotebook}
           buttonComponent={NotebookMenuItem}
-          errorMessage={errorMessage}
+          errorMessages={errorMessages}
           results={results}
         />
       </DropdownMenu>
--- a/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
+++ b/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
@@ -46,9 +46,7 @@ function MetadataMenu({ entityType, results }) {
 
   const { selectedRows } = useStore();
   const protectedRows = useDatasetsAccessLevel(selectedRows.size > 0 ? [...selectedRows] : []).datasets;
-
   const containsProtectedDataset = protectedRows.length > 0;
-
   const errorMessages = [];
 
   if (selectedRows.size > 10) {
@@ -59,7 +57,8 @@ function MetadataMenu({ entityType, results }) {
 
   if (containsProtectedDataset) {
     errorMessages.push(
-      'You have selected protected datasets. Workspaces currently only supports published public datasets. Selected protected datasets are shown below.',
+      'You have selected protected datasets. Workspaces currently only supports published public datasets. Please unselect protected datasets.',
+      // 'You have selected protected datasets. Workspaces currently only supports published public datasets. Selected protected datasets are shown below.',
     );
   }
 
--- a/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
+++ b/context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
@@ -37,7 +37,7 @@ function NotebookMenuItem(props) {
   );
 }
 
-function MetadataMenu({ entityType }) {
+function MetadataMenu({ entityType, results }) {
   const lcPluralType = `${entityType.toLowerCase()}s`;
   const { selectedHits, createNotebook, closeMenu } = useMetadataMenu(lcPluralType);
   const menuID = 'metadata-menu';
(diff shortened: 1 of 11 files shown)


context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
kevinyooky, line 1: The algorithm needs to check many to many for all the ids of selectedRows against all the ids of filteredDatasetIds (the array that contains ids of protected datasets). 

Since we are dealing with two arrays, this is a double for loop, an n * m operation, n being the number of selected Rows, and m being the number of rows that are displayed on the current pagination when the table renders. 

However, the number of rows that are selected has an upper bound of 10, because as soon as we selected more than 10 items, the other error message will be displayed. The time complexity comes out to n * 10 operations. So this is actually O(n), a linear operation.

kevinyooky: I also included screenshot in the scenario where more than 10 protected datasets are selected. 

image

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -38,16 +39,34 @@ function NotebookMenuItem(props) {
...
+  let errorMessage;
+
+  if (selectedRows.size > 10) {
+    errorMessage = `You have selected ${selectedRows.size} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`;
+  } else if (containsProtectedDataset) {
+    errorMessage =
+      'You have selected protected datasets. Workspaces currently only supports published public datasets. Selected protected datasets are shown below.';
+  } else {
+    errorMessage = null;
+  }
```
kevinyooky, line 69: The algorithm needs to check many to many for all the ids of selectedRows against all the ids of filteredDatasetIds (the array that contains ids of protected datasets). 

Since we are dealing with two arrays, this is a double for loop, an n * m operation, n being the number of selected Rows, and m being the number of rows that are displayed on the current pagination when the table renders. 

However, n (the number of rows that are selected) has an upper bound of 10, because as soon as we selected more than 10 items, the other error message will be displayed. Moreover, m (the number of rows that are returned from the results/current pagination) is also bound by 18, since only 18 shows up at a time. 

So in the end, in the worst case scenario, the big O notation comes out to 10 * 18 operations, which is constant time. I thought this might be helpful since a double for loop was being implemented and the operation looks expensive. 

I am also including a screenshot in the scenario where more than 10 protected datasets are selected. 

image

kevinyooky:
Here are some observations I made:

1.) Access level column
2.) Dropdown toggle
3.) Dropdown alignment
4.) persist selected row state
5.) workspace dialog when not logged in
6.) Create workspace when nothing is selected

https://www.loom.com/share/e59bc6888b5f487ba0dd8015b145ee56?sid=831bc03a-cfc8-4d55-b69e-536af398a024

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -38,16 +39,34 @@ function NotebookMenuItem(props) {
...
+  // array that contains all the datasets from current pagination where mapped_data_access_level === 'Protected'.
+  const filteredDataset = results?.hits?.items?.filter(
+    (item) => item?.fields?.mapped_data_access_level === 'Protected',
+  );
+
+  const containsProtectedDataset = filteredDataset?.some((item) => {
+    const filteredDatasetIds = item?.id;
+    // spreading out the selectedRows Set to convert to an array then checking to see if any contains mapped_data_access_level === 'Protected'.
+    return [...selectedRows].some((selectedRowId) => filteredDatasetIds.includes(selectedRowId));
+  });
```
john-conroy, line 58: Do the `results` only reflect the datasets shown on the current page? If so, we wouldn't be able to detect protected datasets selected on a different page of results. I think we need to move towards making a request to get the `protected_access_level` for the datasets in `selectedHits`.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -38,16 +39,34 @@ function NotebookMenuItem(props) {
...
+
+  let errorMessage;
```
john-conroy, line 60: Do we want to show both messages if both cases were true?

context/app/static/js/shared-styles/alerts/index.js
```diff
@@ -26,6 +26,11 @@ const StyledAlert = styled(OutlinedAlert)`
     color: ${(props) => props.theme.palette.text.primary};
   }
   margin-bottom: ${(props) => props.$marginBottom || 0}px;
+
+  > div {
+    display: flex;
+    align-items: center;
+  }
```
john-conroy, line 33: What necessitated this? We should style the instance of `Alert` being used, not the shared component.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
john-conroy: We could use an [aggs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html) query to avoid having to loop over the results that would be returned from this new request.

john-conroy (changes requested):
I believe we will have to make an additional request here, but I think this was a good exercise.

context/app/static/js/shared-styles/alerts/index.js
kevinyooky: This is for the alignment
image

The alignment was off for all of the shared styles so this will fix all instances.

kevinyooky:
I am now using the search API to check for access level of datasets.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
kevinyooky: @tsliaw

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
kevinyooky: Done

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -38,16 +39,28 @@ function NotebookMenuItem(props) {
...
   const lcPluralType = `${entityType.toLowerCase()}s`;
   const { selectedHits, createNotebook, closeMenu } = useMetadataMenu(lcPluralType);
   const menuID = 'metadata-menu';
 
   const { selectedRows } = useStore();
-  const errorMessage =
-    selectedRows.size > 10
-      ? `You have selected ${selectedRows.size} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`
-      : null;
+  const selectedRowsArray = useDatasetsAccessLevel(selectedRows.size > 0 ? [...selectedRows] : []).datasets;
```
NickAkhmetov, line 48: Could you clarify the use of the spread operator here? `useDatasetsAccessLevel` doesn't seem to modify the passed array in any way so I'm not sure what benefit cloning it brings.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -38,16 +39,28 @@ function NotebookMenuItem(props) {
...
+
+  const containsProtectedDataset = selectedRowsArray.some((item) => {
+    return item._source.mapped_data_access_level === 'Protected'; //eslint-disable-line
```
NickAkhmetov, line 51: Let's try to avoid doing full `eslint-disables` and stick to scoped ones in order to avoid disabling more rules than we intend to. You can see what rule is being broken by mousing over the error in VS code; you can also click through on the link to see more info about the error and why it's being thrown:

image

I also recommend using `eslint-disable-next-line` instead of the same-line one for readability.

```suggestion
    //eslint-disable-next-line no-underscore-dangle
    return item._source.mapped_data_access_level === 'Protected'; 
```

context/app/static/js/shared-styles/DialogModal/DialogModal.jsx
```diff
@@ -30,10 +31,14 @@ function DialogModal({
...
-            {errorMessage}
-          </Alert>
+        {errorMessages && (
+          <Box sx={{ marginBottom: 3 }}>
+            {errorMessages.map((errorMessage) => (
+              <Alert $marginBottom={10} severity="error">
+                {errorMessage}
+              </Alert>
+            ))}
+          </Box>
```
kevinyooky, line 41: I created a separate Box container around the error messages for proper spacing. 

image

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -38,16 +39,29 @@ function NotebookMenuItem(props) {
...
+    errorMessages.push(
+      `You have selected ${selectedRows.size} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`,
+    );
+  }
+
+  if (containsProtectedDataset) {
+    errorMessages.push(
+      'You have selected protected datasets. Workspaces currently only supports published public datasets. Selected protected datasets are shown below.',
+    );
+  }
```
kevinyooky, line 64: Converted errorMessage string to errorMessages array and removed "if", "else if" conditions to store more than one errorMessages and map it out so that the errors are stacked.

image

context/app/static/js/components/entity-search/MetadataMenu/hooks.js
```diff
@@ -23,4 +25,18 @@ function useMetadataMenu(lcPluralType) {
...
+      },
+    },
+    _source: ['mapped_data_access_level'],
+    size: ids.length,
+  };
+  const { searchHits: datasets } = useSearchHits(query);
+  return { datasets };
+}
+
+export { useMetadataMenu, useDatasetsAccessLevel };
```
kevinyooky, line 42: Configured api response to remove unnecessary iteration of datasets for finding Protected datasets. This will directly give the aggregated count of Protected datasets.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
kevinyooky: @NickAkhmetov  I agree with your comment during our group call, and this was a good catch. Even though there will be a second version for this, we should still keep the messaging consistent with the current version to uphold the standard and account for potential developmental edge cases (we may not even get to version 2 or forget to change the messaging). I am commenting out V2 messaging and including this temporary one for now.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
```diff
@@ -38,16 +39,28 @@ function NotebookMenuItem(props) {
...
+      `You have selected ${selectedRows.size} datasets. Workspaces currently only supports up to 10 datasets. Please unselect datasets.`,
+    );
+  }
+
+  if (containsProtectedDataset) {
+    errorMessages.push(
+      'You have selected protected datasets. Workspaces currently only supports published public datasets. Please unselect protected datasets.',
+      // 'You have selected protected datasets. Workspaces currently only supports published public datasets. Selected protected datasets are shown below.',
+    );
+  }
```
kevinyooky, line 63: @NickAkhmetov I agree with your comment during our group call, and this was a good catch. Even though there will be a second version for this, we should still keep the messaging consistent with the current version to uphold the standard and account for potential developmental edge cases (we may not even get to version 2 or forget to change the messaging). I am commenting out V2 messaging and including this temporary one for now.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
kevinyooky: Sure! Selected Rows is a set, which is an object, so I need to convert it to an array as useDatasetsAccessLevel takes in an array as an argument. By using the square brackets, I am creating an array, and spreading the contents of the set into it.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
kevinyooky: Both messages will show now.

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
kevinyooky: With the new configuration of the API results, this issue has been resolved.

context/app/static/js/shared-styles/DialogModal/DialogModal.jsx
NickAkhmetov: This looks pretty good! A few minor things:

 - Components rendered in a `map` need a `key` to be provided for React to be able to tell apart the different instances. I suggest using the error message itself as the key.
 - This is completely optional, but I wonder if adding `display: 'grid' and `gap: 1.25` (i.e. 1.25 * 8 = 10px) to the Box and removing `$marginBottom={10}` would also work to manage the spacing? https://mui.com/system/getting-started/the-sx-prop/#grid

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
```diff
@@ -46,7 +51,7 @@ function CreateWorkspaceDialog({
             </label>
```
NickAkhmetov, line 51: This concern is really about the button component on the line above it, but:
 - let's add `disabled={errorMessages.length > 0}` so users can't submit the form if any of the error conditions are met.
 - I really don't like how the label wraps the button and the button is rendered as a span - this is really not great for accessibility and the intended effect is achievable in a semantically correct way.
   - Add an ID to the `form` box above (e.g. `create-workspace-form`)
   - Change the button props to `type="submit" htmlFor="create-workspace-form"`
   - Remove the label wrapping the button and the hidden submit input on line 38

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
NickAkhmetov: Thanks for making this adjustment! 🙌

context/app/static/js/components/entity-search/MetadataMenu/MetadataMenu.jsx
NickAkhmetov: That makes sense, thanks for clarifying! I missed that `selectedRows` was a set (I should've been able to tell by the use of `.size` 😄)

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
kevinyooky: This is a really good catch.

kevinyooky: Not exactly sure what the label was for, removed all those extra lines and still works fine.

kevinyooky: Let me know if I am missing something, applied the suggestion, but now it is no longer working: 

image

NickAkhmetov: My mistake - instead of htmlFor on the button, "form=" should work https://www.impressivewebs.com/html5-form-attribute/

kevinyooky: Works now, thanks :)

PR #3219, opened 2023-08-08
NickAkhmetov/Fix maintenance webpack configs

NickAkhmetov:
This PR fixes the webpack config for the maintenance page.

image


john-conroy (commented):
The changelog is in `/context` but otherwise looks good to me.

NickAkhmetov:
I launched the maintenance page then used `touch` to create the file without going back up to the root level, my bad! It should be resolved now.

john-conroy (approved):

PR #3220, opened 2023-08-09
Kevin yoo/hmp 320 wrong blue for link on organ page

kevinyooky:
Changed color of link on organ page to be consistent with other links on the page. 

Before:
image

After:
image

### DIFF
--- a/context/app/static/js/components/organ/Azimuth/Azimuth.jsx
+++ b/context/app/static/js/components/organ/Azimuth/Azimuth.jsx
@@ -9,7 +9,7 @@ import LabelledSectionText from 'js/shared-styles/sections/LabelledSectionText';
 import { SpacedSectionButtonRow } from 'js/shared-styles/sections/SectionButtonRow';
 
 import { Flex, StyledInfoIcon } from '../style';
-import { StyledPaper } from './style';
+import { StyledPaper, StyledLink } from './style';
 
 function Azimuth({ config }) {
   const dataRefHtml = marked.parseInline(config.dataref);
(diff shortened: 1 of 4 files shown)


context/app/static/js/components/organ/Azimuth/style.js
```diff
@@ -5,4 +5,10 @@ const StyledPaper = styled(Paper)`
   padding: 20px 40px 20px 40px;
 `;
 
-export { StyledPaper };
+const StyledLink = styled.span`
+  a {
+    color: rgb(42, 111, 184);
```
john-conroy, line 10: You can use the color from the theme. ` color: ${(props) => props.theme.palette.info.main};`

john-conroy (changes requested):

context/app/static/js/components/organ/Azimuth/Azimuth.jsx
```diff
@@ -32,7 +32,7 @@ function Azimuth({ config }) {
         <LabelledSectionText label="Nuclei in reference">{config.nunit}</LabelledSectionText>
         {/* eslint-disable react/no-danger */}
         <LabelledSectionText label="Reference dataset">
-          <span dangerouslySetInnerHTML={{ __html: dataRefHtml }} />
+          <StyledLink dangerouslySetInnerHTML={{ __html: dataRefHtml }} />
```
NickAkhmetov, line 35: Rather than introducing another single-use styled component, would `Box` + `sx` work here?

```suggestion
          <Box component="span" sx={(theme) => ({'&>a': { color: theme.palette.link.main })} dangerouslySetInnerHTML={{ __html: dataRefHtml }} />
```

context/app/static/js/components/organ/Azimuth/style.js
NickAkhmetov: I agree; I just double checked, and this would match the figma designs more closely as well, since the links are defined as the same shade of blue that's in our theme (#2A6FB8), while the currently defined color is a little bit different (#2A3B6F).

image

john-conroy (approved):
Nice updates!

PR #3221, opened 2023-08-10
HMP-330 Fix autocomplete/selects on Molecular Data Query page

NickAkhmetov:
This PR
- Fixes the centering of the protein vs. gene select
- Fixes the display of the autocomplete search results

Gene and Protein options are properly centered:
image

Gene lookup results are displayed as expected:
image

I also added a loading state so that the autocomplete doesn't flash "No results" while there's a pending lookup:
image


john-conroy (approved):
Looks good to me. Nice use of a ref!

PR #3223, opened 2023-08-11
NickAkhmetov/HMP-309 Fix organ links

NickAkhmetov:
This PR:
- Makes routing to individual organs' routes case insensitive
  - Organ file names are all lowercase, but some data in the portal has capitalization in organ names; it's much easier to handle that problem by processing the incoming organ names than by applying formatting to every separate link to organs pages.
  - Before: https://portal.hubmapconsortium.org/organ/Placenta -> 404
  - After (note URL capitalization)
image
image
- Also handles spaces and underscores in organ names and removes any leading/trailing whitespace
  - No harm in doing this, and it will make routing to organs a bit less strict for future cases/will make it require a bit less formatting on the front-end
  - e.g. instead of requiring `blood-vasculature` to link to the corresponding organ page, you can now use the literal organ name (`Blood vasculature`) as part of the URL
image

I added more cases to the related test to indicate the updated functionalities.


john-conroy (commented):
Thanks for picking this up! Could you make the organs page a fallback?

NickAkhmetov:
@john-conroy I've added the redirect fallback; it's distinguished from the other redirect type by using code 308 (permanent redirect) instead of 302 (found). I've also indicated this redirect by adding a `redirected_from` var to the flask data.

If `redirected_from` is not null, a redirect alert is shown at the top of the page in order to keep user clarity. I've confirmed @tsliaw's approval on Slack and added a close button:

image


https://github.com/hubmapconsortium/portal-ui/assets/19957804/079edd05-1021-448f-a13d-c98a5fe29dc8


If the user was not redirected to `/organ` but went there themselves, the alert is not displayed.

image

In the future, this component can be reused anywhere where we redirect a user to a potentially unexpected page.

---

I took the opportunity to update the global context providers to use TypeScript since I was already touching the FlaskDataContext and wanted to create the new component as a `.tsx` file.

I have defined the types so ESLint/VS Code allows us to continue using unknown keys in `.js` files (i.e. `.js` files shouldn't be affected by this change), but strongly encourages us to start adding any values we use in `.ts` files to these types. I figured we could populate this as we go along, but can create a ticket specifically to track this if we want to tackle it in one fell swoop.

I also added some context creation helper functions to make context creation more straightforward and remove the possibility of getting an `undefined` context value by calling the hook outside of the scope of the context's provider, since the return type of the default React `useContext` hook is potentially undefined due to that possibility.

john-conroy (approved):
Comprehensive as always! My only thought, and not something that should be addressed here, but we could combine the redirect alert with the alert from `App` into some global alerts component to leverage the redirect alert for all of our routes.

NickAkhmetov:
@john-conroy As tempted as I am to squeeze that into this PR, I agree that we should further enhance this down the line instead :) 

I made a ticket: https://hms-dbmi.atlassian.net/browse/HMP-336

PR #3227, opened 2023-08-15
Kevin yoo/remove protected list button and snackbar

kevinyooky:
Protected Datasets button functionality has been built. Clicking this button will deselect all protected datasets. 

image

Clicking on button will automatically remove the protected datasets as well as remove error.

image

Snackbar functionality has been updated to handle multiple error messages as well as reseting timer for auto hide duration. 

image


context/app/static/js/components/entity-search/Search/Search.jsx
```diff
@@ -39,10 +39,12 @@ function Search() {
         <Results results={results} allResultsUUIDs={allResultsUUIDs} />
       </Flex>
       <SuccessSnackbar
+        // Line 43: This key forces a re-render when the message changes and resets the timer for autoHideDuration
```
NickAkhmetov, line 42: This is a very good use of `key`! When the `key` of a component changes, it's a signal for React to remount the component from scratch, which has the effect of resetting the autoHideDuration in this case. Nice job!

context/app/static/js/components/entity-search/Search/Search.jsx
```diff
@@ -39,10 +39,12 @@ function Search() {
...
+        key={messages[0]}
         open={snackbarOpen}
         onClose={closeSnackbar}
-        message={message}
-        autoHideDuration={2000}
+        message={messages[0]}
+        autoHideDuration={3000}
```
NickAkhmetov, line 47: This might be a bit too brief for accessibility - screen readers and similar tools are unfortunately not quite as fast as our eyes, so they'd start reading the notification then have it disappear halfway through. Would it be possible to set this higher, e.g. 5 or 10 seconds?

cc @tsliaw

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
```diff
@@ -2,23 +2,39 @@ import React from 'react';
...
     handleCreateWorkspace,
     defaultName,
   });
 
+  // eslint-disable-next-line no-underscore-dangle
+  const protectedHubmapIds = protectedRows.map((row) => row._source.hubmap_id).join(', ');
+  const { openSnackbar } = useSnackbarStore();
+  const { deselectRow } = useSelectableTableStore();
+
+  const removeProctedDatasets = () => {
```
NickAkhmetov, line 29:
```suggestion
  const removeProtectedDatasets = () => {
```

NickAkhmetov (commented):
Very nicely done! I appreciate your taking the time to explain the refactors you made to the snackbar store during our earlier discussion.

context/app/static/js/components/entity-search/Search/Search.jsx
tsliaw: I agree, 3 seconds is a bit short. I noted in the handoff doc that the default timer should be 6 seconds for all snackbars. If future snackbars need to be longer or we get feedback saying otherwise, we can adjust the timing then to be longer.

NickAkhmetov: The updated version uses 6 seconds as the default timer and can also be closed early if desired:

image

context/app/static/js/components/workspaces/CreateWorkspaceDialog/CreateWorkspaceDialog.jsx
```diff
@@ -2,23 +2,39 @@ import React from 'react';
...
+    protectedRows.forEach((row) => {
+      // eslint-disable-next-line no-underscore-dangle
+      deselectRow(row._id);
+    });
```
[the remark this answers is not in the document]
NickAkhmetov: I've made this adjustment

NickAkhmetov:
- I addressed the revisions I suggested yesterday
- I made snackbar toasts globally usable with helper methods for the different levels of severity
- I added an ESLint rule to allow leading underscores for names used by ES (_id and _source)

TODOs for tomorrow:
- I seem to have broken the selectabletableprovider tests, will fix accordingly tomorrow
- I will write a storybook story for the snackbars

Once these are done, I'll merge this changeset into the next branch, resolve any remaining concerns from comments there, then merge that into the root (hmp-157) PR for John's review once he's back.

NickAkhmetov:
I've added the elements from my TODOs and will continue with the rest of my planned approach.

PR #3229, opened 2023-08-17
Nickakhmetov/HMP-337 Add Data Products to file browser

NickAkhmetov:
This PR adds controls to the file browser to filter the displayed files to only the data products, and adds a `Data Product` indicator for any files that are tagged as such. If a file is both QA and a data product, both chips are shown (confirmed in tests, but hasn't happened yet with any real datasets). All file type chips are currently rendered as part of the same column in the table.

Real dataset on `test`:

https://github.com/hubmapconsortium/portal-ui/assets/19957804/7c0e53fa-081d-43a4-ac87-e5de5936f4bc

Dataset with no data products:
image

Screenshot where I've set every data product file to also be a QA file to verify styling:

image

I also:
- Removed duplicate `FilesConditionalLink` and `FileBrowserDUA` components which existed in both `Files` and `BulkDataTransfer` (HMP-290)
  - Favored the ones in `BulkDataTransfer`, but this may need a revisit now
- Updated all FileBrowser components (and other components as needed) to use TypeScript
- Updated all FileBrowser styles to `import { styled } from @mui/styles` and removed some single-use styles with no semantic meaning in favor of `Box` where possible
- Rewrote the specs for FileBrowser components
  - Using a getter object for all `screen` commands to improve readability of expect statements
  - Using a single `Test<ComponentName>` component to pass in props
  - Mostly following this [medium guide](https://archive.is/2j9P4), but did not convert specs to `.ts` for the time being


context/app/static/js/components/detailPage/files/FileBrowser/FileBrowser.spec.js
lchoy, line 1: Nice job using shared fixtures here! Would a parameterized test work for the chip toggle tests now?  
https://jestjs.io/docs/api#testeachtablename-fn-timeout

Not that you have to change what you've got.

NickAkhmetov: Interesting! Thank you for introducing me to this helper, I hadn't seen this before.

I think that'd be a reasonable potential improvement since those two tests have such a similar structure, I'll try to apply this.

NickAkhmetov: Done, that's a really handy pattern, thanks for the advice!

john-conroy (approved):
Thanks for sharing the testing patterns, it looks a lot more concise. 

Completely understandable with the cleanup and conversion to typescript that the PR got quite large, but as we discussed, if there is a way we can divide work of similar scope into smaller PRs that would be immensely helpful!

NickAkhmetov:
> as we discussed, if there is a way we can divide work of similar scope into smaller PRs that would be immensely helpful!

I agree completely; I definitely let this one get out of hand. Going forward, my approach will be to open new PR's for all major refactors outside of the scope of the original task.

PR #3231, opened 2023-08-18
NickAkhmetov/HMP-322 Fix table of contents link to derived samples and datasets section

NickAkhmetov:
This PR adjusts the ID used to reference the Derived Samples and Datasets section on the donor page so that it is properly translated by the ID -> title translation function for the table of contents. 

This PR also extends the logic for this function to no longer capitalize `['of', 'to', 'and', 'the', 'a']` if they are not the first word in the title and adds test cases to document this functionality.

https://github.com/hubmapconsortium/portal-ui/assets/19957804/0dc55a5c-7f2f-428f-8ff3-645081b70867

### DIFF
--- a/CHANGELOG-hmp-322-fix-donor-toc.md
+++ b/CHANGELOG-hmp-322-fix-donor-toc.md
@@ -1,0 +1,1 @@
+- Fix Donor page table of contents link to derived samples and datasets section.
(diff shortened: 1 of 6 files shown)


john-conroy (approved):
Thanks! These table of contents utils are getting complex and I'm wondering if it would make more sense to have fixed ids and provide an id/label map in the page component.

PR #3232, opened 2023-08-18
HMP-321 Add sort headers to samples tables on organ pages

NickAkhmetov:
This PR adds sorting to headers on the organ table.

Existing sort-related logic used searchkit, so I implemented my own little sort management hook; I am open to reimplementing this some other way if need be!

In order to prevent the table from flashing "0 samples" and no results whenever the sort changes, I'm using a ref to store the previous value and displaying the ref value whenever results are pending.

I also noticed some serious lag on organ pages whenever the HRA visualization would load in; the entire rest of the page was unresponsive for ~10 seconds each time I loaded it. This may be remediated by the React 18 createRoot upgrade since we'd be able to isolate the HRA visualization inside another root.

https://github.com/hubmapconsortium/portal-ui/assets/19957804/6ee34a72-b654-4655-a7bf-22a4d4438df6

### DIFF
--- a/context/app/static/js/components/organ/Samples/Samples.jsx
+++ b/context/app/static/js/components/organ/Samples/Samples.jsx
@@ -1,5 +1,5 @@
 /* eslint-disable no-underscore-dangle */
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo, useRef } from 'react';
 import format from 'date-fns/format';
 
 import Table from '@mui/material/Table';
@@ -23,21 +23,31 @@ import { useStore } from 'js/shared-styles/tables/SelectableTableProvider/store'
 import AddItemsToListDialog from 'js/components/savedLists/AddItemsToListDialog';
 import { getDonorAgeString } from 'js/helpers/functions';
 
+import { OrderIcon } from 'js/components/searchPage/SortingTableHead/SortingTableHead';
+import { useSortState } from 'js/hooks/useSortState';
+import { LinearProgress } from '@mui/material';
 import { StyledSectionHeader } from './style';
 import { getSearchURL } from '../utils';
 
 const columns = [
-  { id: 'hubmap_id', label: 'Sample' },
-  { id: 'donor.mapped_metadata.age_value', label: 'Donor Age' },
-  { id: 'donor.mapped_metadata.sex', label: 'Donor Sex' },
-  { id: 'donor.mapped_metadata.race', label: 'Donor Race' },
-  { id: 'descendant_counts.entity_type.Dataset', label: 'Derived Dataset Count' },
-  { id: 'last_modified_timestamp', label: 'Last Modified' },
+  { id: 'hubmap_id', label: 'Sample', sortTerm: 'hubmap_id.keyword' },
+  { id: 'donor.mapped_metadata.age_value', label: 'Donor Age', sortTerm: 'donor.mapped_metadata.age_value' },
+  { id: 'donor.mapped_metadata.sex', label: 'Donor Sex', sortTerm: 'donor.mapped_metadata.sex.keyword' },
+  { id: 'donor.mapped_metadata.race', label: 'Donor Race', sortTerm: 'donor.mapped_metadata.race.keyword' },
+  {
+    id: 'descendant_counts.entity_type.Dataset',
+    label: 'Derived Dataset Count',
+    sortTerm: 'descendant_counts.entity_type.Dataset',
+  },
+  { id: 'last_modified_timestamp', label: 'Last Modified', sortTerm: 'last_modified_timestamp' },
 ];
+
+const columnMap = columns.reduce((acc, column) => ({ ...acc, [column.id]: column }), {});
 
 function Samples({ organTerms }) {
   const { selectedRows, deselectHeaderAndRows } = useStore();
   const searchUrl = getSearchURL({ entityType: 'Sample', organTerms });
+  const { sortState, sort } = useSortState();
   const query = useMemo(
     () => ({
       post_filter: {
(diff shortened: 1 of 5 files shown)


context/app/static/js/components/organ/Samples/Samples.jsx
```diff
@@ -60,19 +70,31 @@ function Samples({ organTerms }) {
...
+          : undefined,
     }),
-    [organTerms],
+    [organTerms, sortState],
   );
 
-  const { searchHits } = useSearchHits(query);
+  const previousSearchHits = useRef([]);
+
+  const { searchHits, isLoading } = useSearchHits(query);
```
john-conroy, line 83: We could potentially introduce something like https://swr.vercel.app/docs/middleware.en-US#keep-previous-result so we could leverage this elsewhere.

context/app/static/js/hooks/useSortState.ts
```diff
@@ -0,0 +1,55 @@
...
+    case 'reset':
+      return initialSortState;
+    default:
+      console.warn('Unrecognized action type', action);
+      return state;
+  }
+}
+
+export const useSortState = () => {
+  const [sortState, dispatch] = useReducer(sortReducer, initialSortState);
```
john-conroy, line 44: Nice! What led you to use a reducer?

john-conroy (approved):
Thanks! A few comments you can take or leave. It could improve the readability to pull some of the intertwined hooks together.

context/app/static/js/hooks/useSortState.ts
NickAkhmetov: I like to use reducers when there are interdependent state values, particularly when they share a control or are derived from a previous state. Using this approach lets us define the valid transitions explicitly to avoid accidentally causing an unintended state transition.

In this case, the new sort order is derived from a combination of the user-selected next state and the previous state:
- if the selected column is the same as the previous selected column, flip the existing sort order;
- if the selected column is not the same as the previous selected column, set the sort order to "desc"

Explicitly defining the actions that can be used to modify this sort state ensures we don't accidentally cause an unexpected invalid state with a `setState` call that violates those rules, and provides a more readable interface than a comparable `setState(previousState => ....)` call which would define this behavior inline.

context/app/static/js/components/organ/Samples/Samples.jsx
NickAkhmetov: Great find! I didn't know this functionality could be encapsulated into middleware, that makes a lot of sense. I'll make this revision!

context/app/static/js/components/organ/Samples/Samples.jsx
```diff
@@ -60,19 +70,31 @@ function Samples({ organTerms }) {
       },
       _source: [...columns.map((column) => column.id), 'donor.mapped_metadata.age_unit'],
       size: 10000,
+      sort:
+        sortState?.columnId && columnMap[sortState.columnId].sortTerm
+          ? [{ [columnMap[sortState.columnId].sortTerm]: sortState.direction }]
+          : undefined,
```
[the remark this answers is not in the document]
NickAkhmetov: Implemented in https://github.com/hubmapconsortium/portal-ui/pull/3232/commits/e8746552087d98fb392b5ce7c833bf4571353252

context/app/static/js/components/organ/Samples/Samples.jsx
NickAkhmetov: Implemented in [`e8746552087d98fb392b5ce7c833bf4571353252..4da59e3d8eda41c512f98aca92a99fcbfd5e0660`](https://github.com/hubmapconsortium/portal-ui/pull/3232/files/e8746552087d98fb392b5ce7c833bf4571353252..4da59e3d8eda41c512f98aca92a99fcbfd5e0660)

NickAkhmetov:
The `middleware` feedback led to a bit of a domino effect:
- First I needed to get the actual middleware functionality working as expected; the approach in the docs no longer worked exactly as anticipated (extending the hook's return didn't work) but the main idea of making the previous data "sticky" still worked
- I then realized I needed to extend the `useSearchData` hook to accept middleware, but the existing approach used positional arguments; if I extended it as it was, then I would need to pass in three preceding arguments that weren't relevant. As such, I extended the `useSearchData` and `useSearchHits` components to allow passing in custom SWRConfig values, fetchers, and `useDefaultQuery` in a single object.
- This, of course, broke a lot of the existing usages of this hook, so I needed to fix the places where that broke - this is now fully functional.

In parallel, I noticed that the header buttons' hover effect looked very wonky, so I adjusted their rendering accordingly; my changes have resulted in consistent-looking header cells which are
- fully clickable/accessible with tab
- always have enough space for the column label and sort icon

I think we could realistically reuse this component throughout the site to get a consistent look/feel for all of our tables.


https://github.com/hubmapconsortium/portal-ui/assets/19957804/962a18b2-4d93-489e-9539-15a68f45d3cb

john-conroy (approved):
> I think we could realistically reuse this component throughout the site to get a consistent look/feel for all of our tables.

Thanks. We do have the `EntitiesTable` in `shared-styles/tables` which is used in the `RelatedEntitiesTable`. There are minor cosmetic differences between the `RelatedEntitiesTable` and the one used here on the organ page, but perhaps we could adapt the two?

NickAkhmetov:
> We do have the EntitiesTable in shared-styles/tables which is used in the RelatedEntitiesTable. There are minor cosmetic differences between the RelatedEntitiesTable and the one used here on the organ page, but perhaps we could adapt the two?

I created [HMP-357](https://hms-dbmi.atlassian.net/browse/HMP-357) to track this idea, good call!

Issue #411, opened 2020-06-08
Re-do routing to two separate indexes


mccalluc:
Closed #408 after talking with Bill: They will not change the envvars, so more of this needs to be inside our code.

Issue #312, opened 2020-05-18
theming issue affecting Vitessce instances with Layer Controller view


ngehlenborg:
Problem: the theme from the portal is partially applied to the Vitessce instance.

Example: https://portal.test.hubmapconsortium.org/browse/dataset/9e04dbf8d911c2c875fc961d5b25286a

<img width="1416" alt="image" src="https://user-images.githubusercontent.com/1216518/82237593-e722f180-9903-11ea-821d-cc96ec7b6dc8.png">

ilan-gold:
Just saw this too.  Thanks for opening the issue.  It does not seem to be an issue on my local instance so it might be something specific to our production build.
<img width="1662" alt="Screen Shot 2020-05-18 at 12 37 08 PM" src="https://user-images.githubusercontent.com/43999641/82237867-54368700-9904-11ea-9796-8eaff71f82e6.png">

keller-mark:
If i may add: in vitessce we are now using the option `disableGlobal: true`. You may also want to try this for the parts of the portal that are also using material UI, but set a different seed value https://material-ui.com/styles/api/#creategenerateclassname-options-class-name-generator

keller-mark:
Note: at that Material UI page, they also document something called `options.productionPrefix` but when I tried that in vitessce I could not figure out what difference it was making - it did not seem to affect the classnames in the production css bundle

john-conroy:
I'll give that a try for the portal, thanks!

PR #3233, opened 2023-08-21
John conroy/prod sentry

john-conroy:
Towards #312. Introduces sentry and grafana labs' faro to compare monitoring tools. The `conf` has been updated with new vars in each of the `portal-ui` envs.

Once merged we will need to upgrade our sentry account to team as our trial has expired.

Filed [#411](https://hms-dbmi.atlassian.net/jira/software/projects/HMP/issues/HMP-411) to further explore sentry source maps.


context/app/static/js/components/detailPage/visualization/VisualizationWrapper/VisualizationError.jsx
```diff
@@ -1,49 +1,48 @@
...
+function VisualizationErrorBoundary({ children, ...rest }) {
+  return (
+    <SentryErrorBoundary
+      beforeCapture={(scope) => {
+        scope.setTag('location', 'vitessce');
+      }}
+      onError={(error) => {
+        faro.logError('VisualizationErrorBoundary', { error });
+      }}
+      fallback={VisualizationFallback({ ...rest })}
```
NickAkhmetov, line 41: I wonder if we could combine the props on lines 11 and 13 then change this line to
```
fallback={error => <VisualizationFallback {...error} {...rest} />}
```

This seems like it should resolve the concern we discussed the other week re: how `useCallback` doesn't actually resolve the underlying unstable component issue that lint throws.

We can revisit this in a separate ticket instead, though.

NickAkhmetov (dismissed):
Looks good to me! I had an idea re: how we could potentially resolve the unstable components issue we discussed the other week, but it can definitely wait.

NickAkhmetov (approved):
I think we're ready!

PR #3234, opened 2023-08-22
NickAkhmetov/HMP-350 Matomo tracking improvements

NickAkhmetov:
This PR adjusts Matomo tracking and user login behaviors:
- Non-HuBMAP users are now able to log in with their Globus credentials. They are granted the same level of access as anonymous users as they do not belong to the `HuBMAP-Read` group. 
  - This adjustment allows us to track the globus groups of external HuBMAP users.
  - This change will not reveal any data that is uniquely accessible to internal HuBMAP users, as data permissions are handled by Globus group membership.
- `internal` vs `external` user status is now determined by whether the logged in user is a member of any internal HuBMAP groups. I added a [list of groups sourced from the `hubmap_commons` repo](https://github.com/hubmapconsortium/commons/blob/main/hubmap_commons/21f293b0-globus-groups.json) for this purpose.
  - I was not able to use this file through the `hubmap_commons` dependency; the version of boto3 introduced into `hubmap_commons` dependencies in v2.1.5 is incompatible with the version requested by a Vitessce dependency, and the groups file was last updated in v2.1.7.
  - Bill is checking if there is a specific reason to use the specific version of boto3 that commons is locked to; in the meantime I made a local copy to keep our work unblocked. I've created HMP-359 to track this concern, and we should be able to remove the duplicate file when Bill confirms this dependency can be changed to a `>=` clause instead of just `==`.
- All the names of the globus groups a user belongs to are now stored in a `user_groups` cookie on login, the contents of which are fed into a new custom dimension in Matomo. 
  - This currently uses the group `name` for ease of parsing (i.e. to remove the extra step of looking up what group corresponds to a given UUID), but we can change it to use the `uuid` of the group if preferred.
  - If a user belongs to more than one group, each value is separated by a pipe (`|`) for easy parsing of individual values.
  - If a user belongs to no groups, the cookie is set to `none`, which is also the default for unauthenticated users.
  - Sample `user_groups` value for my account: `HuBMAP-Testing-Group|HuBMAP-Workspaces-Beta|HuBMAP-WashU-TMC|HuBMAP-Read`
- I also added functionalities to further increase the accuracy of the `internal` vs. `external` determination on the front-end:
  - The `last_login` and `user_groups` cookies are now cleared when the user explicitly logs out in order to avoid cases where an internal user/external user are asynchronously using a shared device (e.g. a shared machine in a lab or library). 
    - The `getUserType` helper function now also uses the `isAuthenticated` global value passed by Flask as part of the condition to determine that it's an internal user in order to handle the case where an internal user has lost authentication status and has not yet logged back in, since there's no clean way to clear the `last_login` cookie on authentication timeout.
    - Previously, the `last_login` cookie persisted until the browser's local storage was cleared; if a HuBMAP user ever logged in to a browser, all subsequent traffic was marked as `internal`. These two adjustments should avoid this pitfall.
  - The `last_login` cookie is cleared when an external user logs in to handle the case when the internal user has lost authentication status due to a timeout and the next login is by an external user. 
    - This was not previously a concern since login was previously restricted to internal users.


NickAkhmetov:
Update: I tried updating to hubmap-commons 2.1.10, but still experienced the incompatibility issue:
```
Could not find a version that matches botocore<1.31.18,<1.32.0,<2.0a.0,>=1.12.36,>=1.31.17,>=1.31.33 (from boto3==1.28.33->hubmap-commons==2.1.10->-r context/requirements.in (line 11))
---
There are incompatible versions in the resolved dependencies:
  botocore<1.32.0,>=1.31.33 (from boto3==1.28.33->hubmap-commons==2.1.10->-r context/requirements.in (line 11))
  botocore<1.31.18,>=1.31.17 (from aiobotocore==2.5.4->s3fs==2023.6.0->fsspec[s3]==2023.6.0->ome-zarr==0.2.1->vitessce==3.0.7->portal-visualization==0.0.13->-r context/requirements.in (line 14))
  botocore<2.0a.0,>=1.12.36 (from s3transfer==0.6.2->boto3==1.28.33->hubmap-commons==2.1.10->-r context/requirements.in (line 11))
```

i.e. the botocore version that ome-zarr 0.2.1 requests is 1.31.17, but the botocore version that hubmap-commons wants is 1.31.33, and the patch version mismatch is enough to prevent the upgrade; we'll have to keep the json copy for now and upgrade the vitessce-python dependency to try to resolve this issue.

NickAkhmetov:
I've updated to hubmap-commons 2.1.10 successfully after locking the boto3 version to 1.28.17 in portal-ui.

I'm now troubleshooting the newly failing python tests which were previously passing - maybe they relied on something that changed in the hubmap-commons package

NickAkhmetov:
The failing tests were due to a change in Flask 2.1.0 to use relative paths in `location` headers for 302 redirects on the same site: https://github.com/pallets/flask/pull/4496

My first approach was to strip the `http://localhost` prefix, but that didn't handle external redirects (e.g. for docs), so I edited the test parameters instead.

NickAkhmetov:
I browsed around localhost with one browser window logged in to a non-HuBMAP globus account and one window not logged in, and confirmed the access granted in both cases was identical; given that Nils and Bill have both been made aware of this change coming down the pipeline, I think this should be ready for next steps (i.e. review etc)

I explicitly bumped the flask version in `requirements.in` to reflect the version of the library required by the `hubmap_commons` dependency.

john-conroy (approved):

PR #3235, opened 2023-08-23
Nickakhmetov/HMP-339 data products section

NickAkhmetov:
This PR implements the DataProducts panel at the top of the Dataset page.

- I also added `@types/react` and `@types/react-dom` to our dev dependencies, as those were resolving to React 16.9 types previously (which blocked usage of `ElementRef`)
- I've pulled the instantiation of the FilesContext (i.e. the context that handles the DUA agreement information) into its own component since the exact same logic was present in the `Files` and `BulkDataTransfer` sections + I needed the context to be available in the `DataProducts` section as well.
  - I've noted the purposeful fallthrough to the `public` DUA on the publications page in order to avoid running into any further regressions related to this.
- The `Additional Details` toggle to hide the description/data generation entries has been turned into its own component
  - Given that the control looked more similar to a default `details` component than a MUI accordion, I used the former as the base for this element
  - Storybook entries have been added under `Accordions/StyledDetails`; I ported over the storybook main.js edits from #3215. 
- As noted in the description of the ticket, the `Download All` button is currently for display only 

https://github.com/hubmapconsortium/portal-ui/assets/19957804/de87f605-7e90-4a3f-a811-5c9aec8bc8bc

---

Misc Notes:
- **The majority of the added lines are just a test fixture containing a list of files and dag provenance metadata. Without that file, this PR shrinks by 700 lines.** 
- This PR was originally based on the data products PR's branch, so the PDFViewer commit from that branch appears in the git history here due to the squash
- We should be able to further enhance the StyledDetails component in the future; e.g. by allowing us to pass in custom expand indicators to replace the default `⌃` caret
- All the vitessce visualizations/file links on `test` don't work


context/app/static/js/components/detailPage/files/DataProducts/FileSize.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Typography, { TypographyProps } from '@mui/material/Typography';
+
+type FileSizeProps = {
+  size: number;
+  variant?: TypographyProps['variant'];
+};
+
+export function FileSize({ size, variant = 'body2' }: FileSizeProps) {
+  return (
+    <Typography component="p" variant={variant} color="#00000099">
```
john-conroy, line 12: Did this color come from the design? If this is purposeful, we should introduce it to the theme.

context/app/static/js/shared-styles/accordions/StyledDetails/StyledDetails.tsx
```diff
@@ -0,0 +1,137 @@
+import React, { PropsWithChildren, useRef, ElementRef, AnimationEvent, MouseEvent, useCallback } from 'react';
+import Box from '@mui/material/Box';
+
+type StyledDetailsProps = PropsWithChildren<{
+  summary: React.ReactNode;
+  summaryBackgroundColor?: string;
+}>;
+
+function StyledDetails({ summary, children, summaryBackgroundColor = 'white' }: StyledDetailsProps) {
```
john-conroy, line 9: Did some customization of the MUI `Accordion` components not fit this design? This is great, but using MUI shields us from maintaining the complexity and provides somewhat consistent accessibility (but this is using native html elements so it could be better!).

john-conroy:
Also, should we use absolute imports with the webpack aliases when we're importing from 2+ directories away?

context/app/static/js/components/detailPage/files/DataProducts/FileSize.tsx
NickAkhmetov: This color _is_ in the design, but we don't have it in our palette yet. 

It probably won't hurt to add it to the theme; I was planning to do a full sweep to update it to match Tiffany's palette for Material Design 3 in Figma and make our theme extensions typesafe, but adding an extra palette item wouldn't significantly affect the scope of that task.

context/app/static/js/shared-styles/accordions/StyledDetails/StyledDetails.tsx
NickAkhmetov: That's a valid concern! My reasoning for implementing a fully custom component for this was that I'd essentially need to override every single style that the MUI accordion comes with in order to match the design, but after seeing how tricky reproducing its strengths with semantic HTML was, I'm inclined to believe that creating a `DetailAccordion` variant of some sort would be a more maintainable approach.

This was still a really good exercise for learning how to style semantic HTML components/work with attributes, but in hindsight, the maintainability concerns were a major oversight on my part.

NickAkhmetov:
Converting this back to a draft while I revise:
- I will remove the `StyledDetails` component in favor of a styled MUI accordion in the interest of maintainability
- I noticed there were elements in the Figma designs which I had missed during the initial implementation (e.g. tooltips) - I will go back and add them
- I will change deeply nested relative imports over to use absolute imports.

context/app/static/js/shared-styles/accordions/DetailsAccordion/DetailsAccordion.tsx
```diff
@@ -0,0 +1,33 @@
...
+function DetailAccordion({ summary, children, summaryProps, detailsProps, ...accordionProps }: DetailAccordionProps) {
+  // wrap summary in Typography if it's a string to apply basic styling
+  const wrappedSummary =
+    typeof summary === 'string' ? (
+      <Typography variant="body2" component="span" display="inline-block">
+        {summary}
+      </Typography>
+    ) : (
+      summary
+    );
```
john-conroy, line 24: Pull into a function?

NickAkhmetov:
I've implemented the issues that were identified by John and listed in my previous comment; this PR should be ready for rereview.

context/app/static/js/components/detailPage/files/DataProducts/hooks.ts
```diff
@@ -0,0 +1,47 @@
...
+    if ('origin' in dag) {
+      acc.origin = dag.origin;
+    }
+    if ('name' in dag) {
+      acc.name = dag.name;
+    }
+    return acc;
+  }, defaultPipeline);
+  return pipelineInfo;
+}
```
john-conroy, line 34: Could you explain the logic behind this?

john-conroy (approved):
Looks good, thanks for the updates!

context/app/static/js/components/detailPage/files/DataProducts/hooks.ts
NickAkhmetov: Sure thing!

When I was working with the dag provenance info, I noticed that some of the list items only had a pipeline name or a pipeline link, instead of having both, something like this:
```
[
  { 
     name: 'pipeline name',
  },
  {
     origin: 'https://github.com/link/to/pipeline'
  }, 

  { 
     name: 'more specific pipeline name',
  },
  {
     origin: 'https://github.com/link/to/pipeline-two'
  }, 
]
```

For example, see `metadata.dag_provenance_list` here: https://portal.test.hubmapconsortium.org/browse/dataset/277152f17b5a2f308820ab4d85c5a426.json

I put together this `.reduce` function to handle cases like this.

context/app/static/js/shared-styles/accordions/DetailsAccordion/DetailsAccordion.tsx
NickAkhmetov: I pulled this out into a helper component in the same file and added some basic documentation 👍🏻

NickAkhmetov:
> Since the "Download All" button is for display only and non-functional, could we remove that until the implementation is done as part of https://hms-dbmi.atlassian.net/browse/HMP-340 to remove any confusion should we deploy a build to an environment?

Makes sense 👍🏻 Done!

lchoy (approved):

PR #3241, opened 2023-08-30
Nickakhmetov/HMP-371 MUI Theme cleanup/conversion to TS

NickAkhmetov:
This PR is still a WIP - I haven't yet verified that the site styling is fully functional, and still need to populate the specific colors for the new variants. Once I've run the application and confirmed everything looks correct, I'll open this for review and request reviews.

This PR updates our theme file to use TypeScript, which will allow us to use our custom variables in `style.ts` files. I hadn't converted style.js to style.ts in any cases where we currently use any of our custom variables due to TypeScript not recognizing them; I've augmented the MUI theme types to make this possible.

I moved some of our non-default one-off colors (i.e. `halfShadow` and `link`) to `palette.common` to follow MUI best practices - e.g., `palette.link.main` is now `palette.common.link`.

I added the `white` and `black` variants to the `palette` instead, since `common` already contained values for `black` and `white`. I also added the `provenance` colors to the palette as their own group, e.g. `provenance.step`, `provenance.default`, etc.

I included `lowEmphasis` for `primary` as well, and added support for the already extant `primary.hover`.

I added `active indicator` as `action.active`, which seemed like the most closely matching piece of the palette.


context/app/static/js/components/entity-search/ResultsTable/style.js
```diff
@@ -6,7 +6,7 @@ const StyledTableRow = styled(TableRow)`
   // What looks white is actually transparent and brightness() has no effect.
   background-color: ${(props) => props.theme.palette.white.main};
   :hover {
-    filter: ${(props) => props.theme.palette.white.hover};
+    background-color: ${(props) => props.theme.palette.white.hover};
```
john-conroy, line 9: Was the change from `filter` to `background-color` intentional?

context/app/static/js/components/searchPage/ResultsTable/style.js
```diff
@@ -15,7 +15,7 @@ const StyledTableBody = styled(TableBody)`
   background-color: ${(props) => props.theme.palette.white.main};
 
   :hover {
-    filter: ${(props) => props.theme.palette.white.hover};
+    background-color: ${(props) => props.theme.palette.white.hover};
```
john-conroy, line 18: Same as above.

context/app/static/js/shared-styles/tables/ClickableRow/style.js
```diff
@@ -13,7 +13,7 @@ const StyledRow = styled(TableRow)`
       : css`
           cursor: pointer;
           &:hover {
-            filter: ${props.theme.palette.white.hover};
+            background-color: ${props.theme.palette.white.hover};
```
john-conroy, line 16: As I see this has been repeated in a few places, It seems intentional. Does this have the same visual effect?

context/app/static/js/theme/theme.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+import type { Meta, StoryObj } from '@storybook/react';
+import { Box, Typography, Table, TableBody, TableCell, TableHead, TableRow } from '@mui/material';
+import theme from './theme';
+
+type PaletteColorProps = {
+  color: (typeof theme.palette)[keyof typeof theme.palette];
+  name: string;
+};
+
+function hex2rgb(c: string) {
```
john-conroy, line 11: Did you pull these utils from somewhere?

john-conroy (approved):
Looks good to me!

context/app/static/js/shared-styles/tables/ClickableRow/style.js
NickAkhmetov: I swapped these over from filters to colors with the goal of having only colors in the palette, but given that this didn't work as expected upon closer review, I'll revert it and add a note to the theme to indicate that the `hover` palette members should be `filter` values instead of `color` values.

context/app/static/js/theme/theme.stories.tsx
NickAkhmetov: I based the `hex2rgb` solution on a StackOverflow answer which only showed the basic `parseInt` piece: https://stackoverflow.com/a/69353003

Since that logic only worked with 7-character hex colors (i.e. `#RRGGBB`) and I wanted to be sure that every edge case was covered, I extended it to be index-based in order to allow handling of short hex colors (`#rgb` like `#fff` white) and RGBA values (though the alpha parsing logic may need some minor revisions after looking at it with fresh eyes). I'll fix that piece of the logic and add crediting for the base to make sure all the tracks are covered.

The `rgb2hex` solution was easier to write since I could reliably pull out all the numeric digits from the original string and turn them into hexadecimals.

john-conroy (approved):

john-conroy (approved):

PR #3243, opened 2023-09-01
Nickakhmetov/hmp 372 faro integration

NickAkhmetov:
This PR adds Faro tracing on top of the Sentry tracing work.

I did not include the `@grafana/faro-react` package in this PR since all of its functionalities (besides the profiler) are not useful in the current state of the app:
- The router integration expects `React-Router`, which we don't use
- The SSR compatibility is irrelevant to us as well
- The ErrorBoundary would clash with the sentry error boundary

If we need to remove Sentry in the future, the ErrorBoundary it provides should be replaceable with the error boundary from `@grafana/faro-react`.

To test it out:
- Go to `index.jsx` and comment out the `paused: !enableTracking` line
- Browse the site, note page views increase
- To view the logs, go to https://hidive.grafana.net/a/grafana-kowalski-app/apps/240/overview?from=now-30m&to=now&var-loki=grafanacloud-logs

I also noticed we're getting a 400 error from Faro in response to the React `.render() is deprecated` warning with a response of `json: cannot unmarshal number into Go struct field Event.events.attributes of type string`, and the error does not get populated in the dashboard. This looks like an internal issue on their end.


context/app/static/js/components/detailPage/visualization/VisualizationWrapper/VisualizationError.jsx
```diff
@@ -1,25 +1,26 @@
...
           />
           <VisualizationErrorBoundaryBackground>
             <div>The Vitessce visualization encountered an error. Please try again or contact support.</div>
-            <details>
-              <summary>Click to expand error details</summary>
-              <div>{componentStack}</div>
-            </details>
+            <DetailsAccordion summary="Click to expand error details">
+              {message} {componentStack}
+            </DetailsAccordion>
```
NickAkhmetov, line 23: I switched this to use the new details-like accordion component we created for the data products section. I also added the error message itself, since that was previously getting lost.

john-conroy (approved):

PR #3244, opened 2023-09-05
Nickakhmetov/fix ci issues

NickAkhmetov:
This PR resolves the issues with building the Python side of the application by ensuring the legacy PIP dependency resolver is used. This prevents the build from attempting to access dask's wheel for `fsspec`, which does not work with `require-hashes` mode.


john-conroy (approved):
Great! What are you using to format the yaml file? Should we conform to a standard as to not pick up code style changes on each PR?

NickAkhmetov:
I forgot that I have my editor's default .yaml formatter set to `prettier`. I think having it as a default would make it a bit easier to be consistent going forward - I can add it to the `.vscode` as part of the changes made in HMP-390.

PR #3247, opened 2023-09-05
John conroy/lint ts files

john-conroy:
Did this slip through somehow? Removing the `ext` arg and replacing the path with `app/static/js/**/*.{js,jsx,ts,tsx}` doesn't seem to work for the `lint` commands, but as is, this picks up a few missed prettier errors in our typescript files.


NickAkhmetov (approved):
Great catch!

PR #3248, opened 2023-09-06
John conroy/infinite scroll

john-conroy:
Towards https://hms-dbmi.atlassian.net/jira/software/projects/HMP/issues/HMP-16.

Introduces infinite scroll for the organ page samples table. After feedback here I'll move towards conforming our other entities tables to using these hooks and convert the touched files to typescript.

After trying out `react-virtuoso` and dealing with the resize observer issues we discussed, then `react-window` and `react-window-infinite-scroll` which seemed convoluted for tables, I've settled on using `react-virtual` for virtualization. The included approach is largely adapted from https://tanstack.com/table/v8/docs/examples/react/virtualized-infinite-scrolling.

I've also fixed an issue seen in production where the checkbox in the header isn't sticky and removed our custom `keepPreviousData` in favor of the config option of the same name available in the SWR config.


context/app/static/js/hooks/useSortState.ts
```diff
@@ -35,7 +33,7 @@ function sortReducer(state: SortState, action: SortAction): SortState {
         direction: 'desc',
       } as SortState;
     case 'reset':
-      return initialSortState;
+      return action.payload as SortState;
```
NickAkhmetov, line 36: Adding `payload: SortState` to the `type: "reset"` member of the `SortAction` union type should let you avoid the `as SortState`

context/app/static/js/components/organ/Samples/Samples.jsx
```diff
@@ -79,10 +77,32 @@ function SampleHeaderCell({ column, setSort, sortState }) {
   );
 }
 
+function SampleRow({ virtualRow, index, hits }) {
+  const {
+    _id: uuid,
+    _source: { hubmap_id, donor, descendant_counts, last_modified_timestamp },
+  } = hits[virtualRow.index];
+  return (
+    <TableRow sx={{ height: virtualRow.size }} data-testid={index}>
```
NickAkhmetov, line 86: We could make the test ID more descriptive, e.g. `sample-table-row-${index}` or something along those lines.

NickAkhmetov (commented):
Great work, thank you for persisting through the virtualization library issues!

I only have a few small suggestions for improvements; I also noticed the dividers between the table rows look a bit faint compared to prod:

image

image

context/app/static/js/components/organ/Samples/Samples.jsx
john-conroy: Ah sorry this was a relic from debugging, thanks for noticing this!

NickAkhmetov (approved):
Looks great! Thanks for making those adjustments.

I made a ticket to track my idea to display some sort of loading state in the `tableBodyPadding` cells:
https://hms-dbmi.atlassian.net/browse/HMP-412

PR #3249, opened 2023-09-06
Nickakhmetov/HMP-390 Update linting packages

NickAkhmetov:
This PR:
- Updates Typescript-ESLint parser/estree to v6
  - This adds formal support for Typescript 5.1+ and should suppress the "unsupported TypeScript version" warning.
- Updates Prettier to v3
- Adds Prettier as the default formatter for `.yaml` files for the project
- Applies fixes for TS rule updates

### Types vs. Interfaces
- Types produce more complicated error messages than interfaces
- Types have more flexibility during definition (i.e. type aliases, tuple types, unions, etc)
- Given this, the new rule is to prefer interfaces except when using type-specific functionalities.
- The linter is able to automatically adjust this so developer experience should not be affected.

### Unsafe assignment of `any` value
- `styled` utilities are not annotated with a return type by default
- Adding the `as typeof OriginalComponentType` after the definition of the styles resolves this and informs the compiler that this component has the same props.
  - [This is the approach recommended in the MUI docs.](https://mui.com/material-ui/guides/typescript/#complications-with-the-component-prop)
- This also can pop up when using e.g. `Object.entries`, since there are no types provided for the returned pairs, so additional typechecking is required for those cases.


context/app/static/js/shared-styles/scrollShadows.js
NickAkhmetov, line 1: These changes were brought in by Prettier v3's added support for CSS parsing.

john-conroy (approved):

PR #3250, opened 2023-09-06
NickAkhmetov/HMP-390 Apply `prettier` formatting to `md`, `json` and `yaml` files

NickAkhmetov:
This PR includes a large amount of the formatting changes applied by ESLint/Prettier to avoid bloating the other HMP-390 PR and making it unreadable.


context/package.json
```diff
@@ -158,8 +158,8 @@
     "test:watch": "jest --silent --watch",
     "test:coverage": "jest --silent --coverage",
     "test:clearCache": "jest --silent --clearCache",
-    "lint": "eslint -c .eslintrc.yml --ext .js,.jsx,.ts,.tsx,.yaml,.yml,.md,.json app/static/js/",
-    "lint:fix": "eslint -c .eslintrc.yml --ext .js,.jsx,.ts,.tsx,.yaml,.yml,.md,.json --fix app/static/js/",
+    "lint": "eslint -c .eslintrc.yml --ext .js,.jsx,.ts,.tsx app/static/js/",
+    "lint:fix": "eslint -c .eslintrc.yml --ext .js,.jsx,.ts,.tsx --fix app/static/js/",
```
john-conroy, line 162: Was this unnecessary?

NickAkhmetov: When I tried this out with those added extensions, ESLint attempted to apply various JS/TS rules to those files instead of ignoring them and delegating to prettier; I'll see if I can get this to work given it's a new day 😄

NickAkhmetov: After digging around some more, I've found that ESLint hits parsing issues whenever it touches a yml/md/json file as it expects them to follow JS conventions. The good news is that there are ESLint plugins intended specifically for resolving this, so I'm trying this approach now.

I initially went with a hacky approach where I just added on an explicit `prettier` call at the end of `lint` and `lint:fix`, but that defeated the purpose of having the formatter and linter integrated with each other.

NickAkhmetov: - I installed the YAML/MD/JSON plugins for ESLint and verified that errors in files that are inside the `app/static/js` directory are properly reported.
- After consulting the docs, the file extensions argument is no longer necessary since we indicate which files we want to scan using the `overrides`. I verified that errors still get picked up in non-JS files with this approach.
- I tried to set things up so that the `npm run lint:fix` command performs the prettier/eslint changes globally instead of just in the app/static/js directory, but ran into issues where it failed to recognize cypress globals despite them being configured; I think this may be related to running ESLint from a parent directory, since VS Code still properly handles those globals.
- I ran `prettier --write .` from the root directory to apply the `prettier` changes globally ad hoc.
- .prettierignore is set to be a symlink to the .eslintignore file; both ignore files are colocated with their respective configs, and are set to ignore the template HTML files, ingest validation tools, and organ YAMLs.

PR #3254, opened 2023-09-11
Nickakhmetov/HMP-350 Update Matomo tracking/Update Flask/enable non-HuBMAP globus logins

NickAkhmetov:
This PR reintroduces the Matomo tracking/login/Flask updates from #3234. I've corrected the login flow so that the `AuthHelper`'s `AuthCache` is initialized before cross-referencing the Globus Groups list. 

This issue did not pop up during previous development due to the cached auth behavior persisting; for some reason, the `routes_auth` file in particular has some odd behaviors that keep it from properly hot reloading, which may have resulted in the login regression slipping through during my previous local tests.


.vscode/tasks.json
NickAkhmetov, line 1: This change (and the corresponding edit to `dev-start`) is made to address Flask's changes to hot reloading/debug mode; as of 2.3, `FLASK_ENV=development` is removed, `FLASK_DEBUG` is discouraged, and the `--debug` flag is recommended in order to ensure that this value does not change while the application is running; see [Flask 2.3 config docs](https://flask.palletsprojects.com/en/2.3.x/config/#debug-mode) for more details.

john-conroy (approved):

PR #3256, opened 2023-09-12
NickAkhmetov/HMP-348 Bump `node` to v20.7, upgrade dependencies

NickAkhmetov:
This PR increments the project `node` version to 16.20.2 (latest version of node 16 LTS) in order to unblock dependency updates and enable tools that require 16.17+.

I've opened this separately from all other work in order to make sure that CI does not break.


NickAkhmetov:
I missed this announcement last year: https://nodejs.org/en/blog/announcements/nodejs16-eol

Given that node 16 has reached EOL as of yesterday, I propose we upgrade to node 18, which is the new LTS.

john-conroy (dismissed):

NickAkhmetov:
@john-conroy Yes! The update to Jest in this PR should resolve the CI problems in main.

PR #3257, opened 2023-09-12
Nickakhmetov/HMP-348 Replace `ajv` and `yup` with `zod`

NickAkhmetov:
This PR replaces the current usages of `ajv` and `yup` with `zod`.

- All three do the same thing at the core - runtime object schema validation
- We currently have ajv and yup installed in our dependencies, and zod is used by Vitessce to validate configurations, so we technically have zod in our application already as well (it's in our package-lock, at least!)
- yup was only in use for the CreateWorkspaceDialog form hook (together with the react-hook-form resolver)
- ajv was only used once in the ProvData class, to validate against a JSON schema file

Bundlephobia for latest versions of each package:
- [ajv = 35.2kB gzipped, 119.6kB minified](https://bundlephobia.com/package/ajv@8.12.0)
- [yup = 12.2kB gzipped, 40kB minified](https://bundlephobia.com/package/yup@1.2.0)
- [zod = 13.2kB gzipped, 57kB minified](https://bundlephobia.com/package/zod@3.22.2)
- ajv seems way too big for what we're using it for; yup/zod seem to be more or less identical size-wise

Given these findings, I've removed `ajv` and `yup` to standardize our validation around `zod` with Vitessce.


context/app/static/js/components/detailPage/provenance/ProvVis/ProvSchema.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from 'zod';
+
+// Provenance JSON Schema converted to zod schema
+// Original schema came from https://github.com/hubmapconsortium-graveyard/prov-vis
+
+const stringToAnything = z.string().or(z.boolean()).or(z.number()).or(z.object({}).passthrough());
+const stringToStringToAnything = z.record(stringToAnything);
+
+export const provSchema = z
```
john-conroy, line 9: Just commenting to note that we discussed you are copying the schema exactly as it was, but perhaps we could revisit this to make it more strict?

NickAkhmetov: I agree, the schema would be much more useful if it validated the actual types.
I created https://hms-dbmi.atlassian.net/browse/HMP-419 to handle this. 👍🏻

PR #3262, opened 2023-09-14
NickAkhmetov/Improve push script specificity

NickAkhmetov:
This PR adjusts the specificity of the push script to only target the files that are intended to be committed during the release process.

### DIFF
--- a/etc/build/push.sh
+++ b/etc/build/push.sh
@@ -34,7 +34,7 @@ echo "Version: $VERSION"
 
 ./grab-dependencies.sh
 
-git add .
+git add context/package*.json
 git commit -m "Version bump to $VERSION"
 
 if ls CHANGELOG-*.md; then
@@ -49,7 +49,7 @@ if ls CHANGELOG-*.md; then
   ) > CHANGELOG.md.new
   mv CHANGELOG.md.new context/app/markdown/CHANGELOG.md
   git rm CHANGELOG-*.md
-  git add .
+  git add context/app/markdown/CHANGELOG.md
   git commit -m 'Update CHANGELOG'
 fi
 
--- a/CHANGELOG-improve-push.md
+++ b/CHANGELOG-improve-push.md
@@ -1,0 +1,1 @@
+- Improve push script specificity.


etc/build/push.sh
```diff
@@ -49,7 +49,7 @@ if ls CHANGELOG-*.md; then
   ) > CHANGELOG.md.new
   mv CHANGELOG.md.new context/app/markdown/CHANGELOG.md
   git rm CHANGELOG-*.md
-  git add .
+  git add context/app/markdown/CHANGELOG.md
```
john-conroy, line 52: We also have to commit the file deletions at the project root?

etc/build/push.sh
```diff
@@ -34,7 +34,7 @@ echo "Version: $VERSION"
 
 ./grab-dependencies.sh
 
-git add .
+git add context/package*.json
```
john-conroy, line 37: I don't suggest doing this here, but we could consider pulling some of the git version/git tag logic out of this script and into package.json? https://docs.npmjs.com/cli/v7/commands/npm-version details some of the options.

john-conroy (commented):
Thanks!

etc/build/push.sh
NickAkhmetov: `git rm` stages the removals; I just tested this locally to be sure:

```
(portal) nikolay@NikolayPCWin:~/projects/hubmap/portal-ui$ git rm example-app.conf 
rm 'example-app.conf'
(portal) nikolay@NikolayPCWin:~/projects/hubmap/portal-ui$ git status
On branch nickakhmetov/hmp-348/misc-dependency-updates
Your branch is up to date with 'origin/nickakhmetov/hmp-348/misc-dependency-updates'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        deleted:    example-app.conf
```

etc/build/push.sh
NickAkhmetov: I think that'd be a good idea! I made a ticket: https://hms-dbmi.atlassian.net/browse/HMP-410

john-conroy (approved):

PR #3264, opened 2023-09-14
Nickakhmetov/HMP-409 Remove `marked` in favor of `react-markdown`

### LINKED ISSUE #3067: Replace marked with react-markdown throughout

NickAkhmetov:
This PR removes all usages of `marked` in favor of using `react-markdown`, which does not require us to dangerously set inner HTML. I tested the places in the application where this was in use (organ pages' `reference analysis` pane, global alert, and markdown pages such as /changelog) and confirmed functionality is unaffected.

Fixes #3067

### DIFF
--- a/CHANGELOG-hmp-409-markdown.md
+++ b/CHANGELOG-hmp-409-markdown.md
@@ -1,0 +1,1 @@
+- Remove `marked` in favor of `react-markdown`.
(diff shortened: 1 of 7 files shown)


john-conroy (approved):
Also fix #3067 which we never moved to Jira.

PR #3266, opened 2023-09-18
Nickakhmetov/HMP-372 Remove sentry in favor of Faro

NickAkhmetov:
This PR
- Replaces `@sentry/react` with `@grafana/faro-react`
- Replaces the sentry error boundaries with Faro error boundaries
- Sets up the React Faro integration and removes code related to sentry integration
- Deduplicates code related to styling the Vitessce visualization container's other states

The `sentryEnv` environment variable is left untouched for the time being, but we can rename it if need be.

Local data is getting captured in Faro:

image


context/app/static/js/helpers/init-trackers.js
```diff
@@ -1,41 +1,37 @@
...
         capturePerformanceTimeline: true,
-        captureConsoleDisabledLevels: [LogLevel.INFO, LogLevel.DEBUG],
+        // Only capture errors and warnings in grafana
+        captureConsoleDisabledLevels: [LogLevel.INFO, LogLevel.DEBUG, LogLevel.LOG],
       }),
 
       // Initialization of the tracing package.
       // This packages is optional because it increases the bundle size noticeably. Only add it if you want tracing data.
       new TracingInstrumentation(),
+      new ReactIntegration(),
```
john-conroy, line 34: I believe you previously mentioned not using the `ReactIntegration` what motivated the change?

context/app/static/js/components/detailPage/visualization/VisualizationWrapper/VisualizationError.tsx
```diff
@@ -1,48 +1,39 @@
...
-      onError={(error) => {
-        faro.logError('VisualizationErrorBoundary', { error });
-      }}
-      fallback={VisualizationFallback({ ...rest })}
-    >
-      {children}
-    </SentryErrorBoundary>
-  );
+function VisualizationErrorBoundary({ children }: PropsWithChildren) {
+  return <FaroErrorBoundary fallback={VisualizationFallback}>{children}</FaroErrorBoundary>;
```
john-conroy, line 36: Does faro discern between the two error boundaries similar to sentry with its `location`?

context/app/static/js/helpers/init-trackers.js
NickAkhmetov: Now that we no longer have the `sentry` error boundary component, the `faro` error boundary component has become a relevant benefit to adding the react integration; previously, we did not immediately gain anything by adding this integration, as we were not immediately looking to start profiling performance (though we could add profiling now?), we do not use `react-router`, and already had an `ErrorBoundary` component that `faro`'s would have conflicted with.

context/app/static/js/components/detailPage/visualization/VisualizationWrapper/VisualizationError.tsx
NickAkhmetov: It looks like that's possible, albeit less straightforward than with Sentry. Annoyingly enough, the `pushError` fired by ErrorBoundary does not allow for any configuration for that call, which would allow me to specify the error boundary via error options.

context/app/static/js/helpers/init-trackers.js
```diff
@@ -1,41 +1,37 @@
...
-
-  sentryInit({
-    dsn: sentryDsn,
-    environment: sentryEnv,
-    enabled: enableTracking,
-    release,
-  });
+  const version = `portal-ui-react@${PACKAGE_VERSION}`;
+  const environment = sentryEnv;
+  const paused = !['prod', 'prod-stage'].includes(environment);
```
lchoy, line 12: Can we distinguish what's coming from prod vs stage? If not, perhaps we should only show prod so it's clear what's being monitored and reported on.

NickAkhmetov: There is an `app_environment` value reported when events are pushed in grafana:
image

It seems like it's not being correctly populated from our environment variable, however, since all my events are being reported as `local` while having the relevant environment variable set to `prod`, so I'll experiment with the configuration more.

context/app/static/js/components/detailPage/visualization/VisualizationWrapper/VisualizationError.tsx
NickAkhmetov: I opened a PR to add per-errorBoundary `pushErrorOptions` configurations, and we may be able to use my fork while we wait for the new release.

context/app/static/js/helpers/init-trackers.js
NickAkhmetov: I confirmed that the problem wasn't with the value being passed to `app_environment`, but with the `paused` prop not stopping reporting like I expected it to. I've revised the approach to suppressing events to use the `beforeSend` middleware instead, which checks if the current environment is prod or prod-stage and only reports to faro in those envs.

john-conroy (approved):
Thanks for the updates!

PR #3269, opened 2023-09-19
Nickakhmetov/HMP-406 Remove EUI-related dependencies

NickAkhmetov:
This PR reimplements the RangeSliderFacet to use a customized MUI slider instead of EUI. This allows us to remove `@elastic/eui`, `@elastic/datemath`, `@emotion/react`, and `@searchkit/elastic-ui`. I did my best to replicate the display/functionality of the `levels` display from the EUI component as well.

image


NickAkhmetov:
The lint issues that break CI for this PR are fixed in #3270, once it's merged into here that will go away

john-conroy (approved):
Looks good to me. Should we take a look at the ts failures picked up by `GitHub Actions /build`?

NickAkhmetov:
> Should we take a look at the TS failures

Those were resolved by the changes to the `tsconfig` which I made in #3270; merging that dismissed these errors.

The issue turned out to be that the tsconfig base path settings led to it using the emotion engine for types instead of StyledComponents; as a result, it couldn't figure out how to properly style MUI components (though the styling for non-MUI components like `'div'` worked fine with this setup, oddly enough).

PR #3277, opened 2023-09-27
NickAkhmetov/HMP-430 Fix Faro SWR logs

NickAkhmetov:
This PR fixes the calls to `faro.logError` and `faro.logWarning` - these functions don't exist and I didn't catch this in my initial implementation because `Providers` is a `.jsx` file (and thus ESLint does not throw errors when a non-existing property is referenced).

I've revised these calls to `faro.api.pushError` and `faro.api.pushEvent`, respectively.

### DIFF
--- a/CHANGELOG-HMP-430.md
+++ b/CHANGELOG-HMP-430.md
@@ -1,0 +1,1 @@
+- Fix Faro logging function calls for query errors and slow loading warnings.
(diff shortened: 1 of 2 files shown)


john-conroy (dismissed):
Thanks for the quick fix!

lchoy (approved):

PR #3278, opened 2023-09-28
Nickakhmetov/HMP-404 Upgrade Zustand/Immer, convert stores to TypeScript

NickAkhmetov:
We do this not because it is easy, but because we thought it would be easy.

I'm still holding out hope that I'll be able to reconcile at least some of the `.js` and `.ts` files that got unlinked from each other; I started out by doing renames and updates in separate commits, but Git still detected some of these actions as a delete and add for the PR as a whole due to the size of the changes overall. Whenever more than 50% of the file is changed in a single PR __AND__ the file is renamed, the change is seen as a remove/add instead of a rename.

This PR:
- Upgrades from Zustand v3 to Zustand v4 
  - Uses `useSyncExternalStore` for improved React 18 performance
  - Updated TS types
  - Improved `persist` middleware so we don't have to do as much localstorage management
    - May present more opportunities for store persistence?
  - Supports non-object states
  - Deprecates existing zustand context approach
- Upgrades from Immer v9 to Immer v10
  - dropped IE support for improved performance
- Adds helper functions for generating zustand stores with configured middleware
  - These abstract away some of the type complexity
- Adds helper functions for generating contexts, providers, and hooks for non-global Zustand stores
  - This more or less reintroduces the approach used for v3 context stores, which are necessary in order to instantiate stores with initial properties
  - There are packages which also recreate this functionality or do something similar (e.g. zustand-tools and zustand-di), but I'm concerned about stability/long term support given that these projects are not affiliated with Poimandres.
- Converts stores throughout the application to use TypeScript.
- Converts redux-style `reducer` stores to standard `zustand` stores to enhance readability
- Enables ESLint exception for `immer` stores' param reassignment to encourage more widespread use
  - Potential improvement - renaming the `state` variables to `draftState` and enabling it only for that word instead would likely be safer/more self explanatory.
- Enables use of TypeScript `jest` tests
  - Hooray, type safety!
  - I disabled the `no unsafe return` rule since it didn't like our `test-utils/functions` not having type annotations.
- Switches the Snackbars to a global store (`createStore` -> `create`)
  - No point in wrapping the entire application in a context layer to replicate default global state
- Makes minor revisions to the entity search's `SearchWrapper` to not hardcode the `Search` child
  - The tests for the `useGroupedFacets` hook were failing because the search wrapper did not accept children, so this change was made. The `Search` component was put in the children of the search wrapper in all appropriate places.
- Adds type definitions for the .yaml files so we can import them in .ts files


NickAkhmetov:
Docker is down, so Cypress tests aren't running successfully: https://downdetector.com/status/docker/
image

With that said, I think the functionality is finally all in place and implemented with my helpers, so I am opening this for review.

john-conroy (approved):
Thanks for getting through all of this! My understanding is that the vanilla `createStore` is only needed when used with `Context` or a `ref` and we should prefer `create` otherwise?

We should probably all commit to QA once this is deployed to the lower envs.

NickAkhmetov:
@john-conroy Yes, that's precisely it! States scoped to a specific context use `createStore`, while fully global states use `create`.

> We should probably all commit to QA once this is deployed to the lower envs.

Definitely!

john-conroy (approved):

PR #3279, opened 2023-09-29
NickAkhmetov/HMP-418 Migrate `@mui/styles` styling to `@mui/material/styles`

NickAkhmetov:
This PR removes the legacy `@mui/styles` engine (Prior to v5, Material UI used @mui/styles as a JSS wrapper) and replaces all of its styling with the styled util from `@mui/material/styles` and theme values, as appropriate.

---

Note - I found that instead of providing arg-less functions to `styled` when we don't need a theme value to define the styles, we can provide style objects directly; e.g., instead of 
```typescript
const ModalContentWrapper = styled('div')(() => ({
  position: 'absolute',
  top: '50%',
  left: '50%',
  transform: 'translate(-50%, -50%)',
}));
```

We should do
```typescript
const ModalContentWrapper = styled('div')({
  position: 'absolute',
  top: '50%',
  left: '50%',
  transform: 'translate(-50%, -50%)',
});
```

This matches the expected [type definitions in the MUI `styled` API](https://mui.com/system/styled/#api).


NickAkhmetov:
Note for myself upon review - need to fix appearance of accordions on publication pages

john-conroy (approved):
When should we prefer `styled` over `sx`?

NickAkhmetov:
> When should we prefer `styled` over `sx`?

What I've generally seen advised while reading up in the MUI docs throughout this process:

- If a styling change is meant to affect _every_ component of that type, modifying `theme` is helpful; entering the changes as `styleOverrides` applies it to every instance of a component, while extending `variants` allows us to opt in accordingly
  - e.g. I added the accordion styles to reduce the extra margins as `variants`; I originally included them as `styleOverrides`, but that broke the publication page vignettes' appearance as I noted above.
- If a component is meant to be reusable, `styled` works well since naming the component makes its semantic meaning/purpose clearer. 
  - e.g. If I have a special button style that's used in multiple places in a component, defining it as a styled component decreases the total amount of code relative to defining those styles in an `sx` in each of those instances
- If the style is not meant to be reused elsewhere / is being applied for utility only / has a lot of dependencies on local state/context, `sx` is more wieldy.
  - e.g. if we need a `flex` container, I prefer using `<Box sx={{display: 'flex', flexDirection: 'row'}}>` to adding another `const Flex = styled('div')({ display: 'flex', flexDirection: 'row'});`
- ⚠️ The only caveat is that we should count any component that appears multiple times on a page as reused, since `sx` has a slightly higher performance overhead than `styled`.

Some articles that go into more detail:
- https://mui.com/system/getting-started/usage/#when-to-use-mui-system
- https://smartdevpreneur.com/when-to-use-the-mui-sx-prop-styled-api-or-theme-override/

PR #3283, opened 2023-10-03
Nickakhmetov/HMP-428 Improve CCF UI integration in Data Portal

NickAkhmetov:
This PR updates our approach to mounting the CCF UI organ component to no longer require the use of an iframe for internal data portal usage. 

image

No `<iframe>`! 🎉

## Why not continue using `iframe` approach?
The current iframe approach is problematic for several reasons:
- A second instance of the React application gets mounted as well, which severely degrades performance due to high demand of the main browser thread by the page and the iframe simultaneously.
- Two-way data flow is impossible, as JS scope is constrained inside/outside the iframe (i.e. the iframe JS and the main page JS are sandboxed and unaware of each other)

Therefore, in order to improve the performance of organ pages and enable future development of features which require two-way data flow, we need to mount the CCF UI components without using an iframe, while keeping existing iframe routes functional for use outside of the portal UI.

## Changes

- I created hooks to dynamically add/remove the necessary scripts and stylesheets from the document head in order to import the components on demand
- I extended the global JSX interface to include various CCF web components.
- I updated the CCF UI component versions to `@3`, which uses the latest stable release.
- The `ccf-organ-info` component mounts and points at the correct organ.
- I used `reactify-wc` to wrap the web component for `ccf-organ-info` to allow us to intercept and reflect the selected sex/side. In the future, this will allow us to filter down samples etc. to reflect the currently selected side/sex.
  - The `nodeClicked` event does not appear to be correctly firing/being intercepted, but we will be able to use this in the future as well.
  - This pattern can be reused for the other CCF UI components and other web components.
- Updated Vitessce to v3.3.2 to include the updates to luma/deck.gl to match the CCF components.

<details><summary>Resolved blockers</summary>
<p>

- One of the main motivations of making this revision is to enable two-way data flow; however, [Angular private zones](https://angular.io/guide/zone) appear to be suppressing event emission outside of the Angular runtime, which prevents us from processing the corresponding inputs.
  - I attempted to work around the limitations React has when working with web components by manually adding event listeners to the CCF Organ and CCF Body components in a `useEffect`, but the events are not triggering console output.
  - **Potential solution:** I've reached out to Bruce with a potential solution from their side - [this article](https://stribny.name/blog/2018/01/how-to-communicate-with-angular-components-outside-angular/) suggests attaching a global pubsub to the browser `window` to enable interoperability between Angular components and other front-end JS, and am pending feedback on this suggestion.
- The `luma.gl` and `deck.gl` dependencies shared by Vitessce and the CCF UI are slightly different
  - CCF UI 3.6 required an older version of deck.gl than Vitessce, so I tried bumping to 3.7
  - CCF UI 3.7 required a slightly different minor version of luma.gl than Vitessce (8.5.12 vs 8.5.16)
  - I've contacted Bruce about this as well; they are bumping dependencies on their end right now and are setting luma.gl to `8.5.20`. 
  - I am opening a corresponding PR on the Vitessce side to pin the luma.gl version to `8.5.20` there as well.
  - There are a few long-term solutions to prevent this from recurring:
    - Publishing the web-component in the npm registry so that minor dependency mismatches on this scale are caught/resolved by the package manager is one potential approach which would also have the benefit of allowing react devs to more idiomatically use these components with wrappers like [`@microsoft/fast-react-wrapper`](https://www.npmjs.com/package/@microsoft/fast-react-wrapper).
    - Mark has suggested packaging a distribution of Vitessce without a baked-in `deck.gl`/`luma.gl`, allowing them to be provided as peer dependencies instead. This would allow us to locally pin those dependencies to match the CCF UI.

</p>
</details>


context/app/static/js/components/HRA/CCFOrganInfo.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+
+import reactifyWc from 'reactify-wc';
```
john-conroy, line 3: Are we concerned that the last publish was 3 years ago?

https://www.npmjs.com/package/reactify-wc

NickAkhmetov: I'm open to changing this down the line; the more recent wrappers I found wanted the component's JS Class as well, which would require @bherr2's team to publish the components on `npm`.

- https://www.npmjs.com/package/@microsoft/fast-react-wrapper
  - API constructs a wrapper using the current React runtime then wraps it around the component's Class
- https://www.npmjs.com/package/@lit/react
  - API includes `elementClass`
- https://www.npmjs.com/package/wc-react
  - Only other wrapper I found where the custom component's tag is wrapped instead of the class, but it's similarly unmaintained (one year vs. three)

Hopefully by the time that the age of this dependency begins affecting us, React will have mainline support for web-components, as the `experimental` version currently does and we will be able to remove it altogether: https://custom-elements-everywhere.com/libraries/react-experimental/results/results.html

PR #3290, opened 2023-10-16
John conroy/launch workspace template

john-conroy:
Towards [HMP-398](https://hms-dbmi.atlassian.net/browse/HMP-398). A few more items including overall polish need to be done.

- [ ] Add table to dialog launched on search pages. Add specific content, including link to search page, for landing and dataset pages.
- [ ] Replace existing dialog on workspace landing page and dataset pages.
- [ ] Refactor other multi select inputs to use shared component. (Cells gene/protein marker select).
- [ ] Update eslint-config to avoid in-line disabling of `@typescript-eslint/no-misused-promises` for the submit callback.
- [ ] Handle 200s from WS/User Templates APIs with `success:false`.
- [ ] Refactor workspace utils, hooks and types.

[HMP-398]: https://hms-dbmi.atlassian.net/browse/HMP-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


context/app/static/js/shared-styles/inputs/MultiAutocomplete/MultiAutocomplete.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useState } from 'react';
+
+import Autocomplete, { AutocompleteProps } from '@mui/material/Autocomplete';
+import TextField, { TextFieldProps } from '@mui/material/TextField';
+import { InputLabelProps as inputLabelProps } from '@mui/material/InputLabel';
+
+interface MultiAutocompleteTypes extends AutocompleteProps {
```
NickAkhmetov, line 7: The AutocompleteProps type is generic (its properties' types depend on the value type, whether it's a multiple autocomplete, whether it's clearable, whether it accepts freetext inputs, and optionally, the Chip component that it should use). I suggest modifying the interface:
```
interface MultiAutocompleteTypes<Value> extends AutocompleteProps<Value, true, false, false>
```
The component itself will also need a Value type parameter to support this:
```
function MultiAutocomplete<Value>({
```

context/app/static/js/shared-styles/inputs/MultiAutocomplete/MultiAutocomplete.tsx
```diff
@@ -0,0 +1,49 @@
...
+  tagComponent: React.ReactNode;
```
NickAkhmetov, line 8: `tagComponent` should be typed as `ComponentType`  instead of `ReactNode`, since `ComponentType` refers to the actual component name that would be passed around while `ReactNode` is the result of a component being rendered.

e.g.:
```
function Foo() {
    return <div>bar</div>
}
```
In this example, `Foo` is a `ComponentType` while `<div>bar</div>` is a `ReactNode`

context/app/static/js/shared-styles/inputs/MultiAutocomplete/MultiAutocomplete.tsx
```diff
@@ -0,0 +1,49 @@
...
+  renderInputProps: TextFieldProps;
+}
+
+interface RenderInputTypes {
+  InputLabelProps: inputLabelProps;
+  textFieldProps: TextFieldProps;
+}
+
+function MultiAutocomplete({ renderInputProps, tagComponent: TagComponent, ...rest }): MultiAutocompleteTypes {
```
NickAkhmetov, line 17: The `MultiAutocompleteTypes` should be inside the parentheses in order to type the props. If the type is written outside the parentheses, it marks the return type of the function.

```suggestion
function MultiAutocomplete<Value>({ renderInputProps, tagComponent: TagComponent, ...rest }: MultiAutocompleteTypes<Value>) {
```

context/app/static/js/shared-styles/inputs/MultiAutocomplete/MultiAutocomplete.tsx
```diff
@@ -0,0 +1,49 @@
...
+          name="substring"
+          variant="outlined"
+          onChange={handleChange}
+          onKeyDown={(e: React.KeyboardEvent<HTMLInputElement>) => {
+            e.stopPropagation();
+          }}
+          {...textFieldProps}
+        />
+      )}
+      {...rest}
```
NickAkhmetov, line 44: `rest` may contain `renderTags` and `renderInput`, since they are not extracted in the props above. As a result, the `renderTags` and `renderInput` values above might get overwritten. Moving the `...rest` spread above `renderTags` should resolve this concern

context/app/static/js/shared-styles/inputs/MultiAutocomplete/MultiAutocomplete.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useState } from 'react';
+
+import Autocomplete, { AutocompleteProps } from '@mui/material/Autocomplete';
+import TextField, { TextFieldProps } from '@mui/material/TextField';
+import { InputLabelProps as inputLabelProps } from '@mui/material/InputLabel';
+
+interface MultiAutocompleteTypes extends AutocompleteProps {
+  tagComponent: React.ReactNode;
+  renderInputProps: TextFieldProps;
```
NickAkhmetov, line 9: `TextFieldProps` is also generic based on which variant prop is used; since we're using `outlined`, we should be able to set this to `TextFieldProps<'outlined'>`.
```suggestion
  renderInputProps: TextFieldProps<'outlined'>;
```

context/app/static/js/shared-styles/inputs/MultiAutocomplete/MultiAutocomplete.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  return (
+    <Autocomplete
+      renderTags={(value, getTagProps) =>
+        value.map((option, index) => {
+          return <TagComponent option={option} {...getTagProps({ index })} key={option} />;
+        })
+      }
+      renderInput={({ InputLabelProps, ...textFieldProps }: RenderInputTypes) => (
```
NickAkhmetov, line 31: `renderInput` has an expected param type provided by MUI (`AutocompleteRenderInputParams`) so we shouldn't need to define these types. `ctrl + click`ing on the `renderInput` prop name opens the definition from the MUI types, then `ctrl + click`ing the params type in there shows what you have access to:

```
export interface AutocompleteRenderInputParams {
  id: string;
  disabled: boolean;
  fullWidth: boolean;
  size: 'small' | undefined;
  InputLabelProps: ReturnType<ReturnType<typeof useAutocomplete>['getInputLabelProps']>;
  InputProps: {
    ref: React.Ref<any>;
    className: string;
    startAdornment: React.ReactNode;
    endAdornment: React.ReactNode;
  };
  inputProps: ReturnType<ReturnType<typeof useAutocomplete>['getInputProps']>;
}
```

context/app/static/js/components/workspaces/NewWorkspaceDialog/NewWorkspaceDialog.tsx
```diff
@@ -0,0 +1,229 @@
...
+              component="form"
+              sx={{
+                display: 'grid',
+                gap: 2,
+                marginTop: 1,
+              }}
+              // eslint-disable-next-line @typescript-eslint/no-misused-promises
+              onSubmit={handleSubmit(submit)}
+            >
+              <WorkspaceField
```
NickAkhmetov, line 151: Since the `value` prop is destructured from the props and `WorkspaceField` doesn't have type annotations, TypeScript treats `value` as a required property. Adding a default value of `undefined` fixes this; I can't make the suggestion on the `WorkspaceField` component since this PR didn't touch it, but here's what it would look like:
```
function WorkspaceField({ control, name, label, errors, value = undefined, ...rest }) {
```

Converting WorkspaceField to `.tsx` is another option.

NickAkhmetov (commented):
I have some findings from troubleshooting the errors in `NewWorkspaceDialog`:

context/app/static/js/components/workspaces/NewWorkspaceDialog/NewWorkspaceDialog.tsx
```diff
@@ -0,0 +1,202 @@
...
+      aria-describedby="scroll-dialog-description"
+      maxWidth="lg"
+    >
+      <Box mb={2}>
+        <DialogTitle id="scroll-dialog-title" variant="h3">
+          Launch New Workspace
+        </DialogTitle>
+        <Box sx={{ px: 3 }}>
+          <Paper sx={{ p: 2 }}>
+            <Typography sx={{ mb: 2 }}>
```
NickAkhmetov, line 91: I think if we use `<Stack gap={2} sx={{ p: 2 }} component={Paper} direction="col">` we could cut out the inner `sx={{ mb: 2 }}` here; I think we could also add `mx: 3` and remove the outer wrapping Box on line 89 as well, without affecting styles.

context/app/static/js/components/workspaces/NewWorkspaceDialog/NewWorkspaceDialog.tsx
```diff
@@ -0,0 +1,202 @@
...
+              any of the datasets selected, which will be indicated by an error banner.
+            </Typography>
+          </Paper>
+        </Box>
+      </Box>
+      <DialogContent dividers>
+        <Step title="Edit Datasets Selection" index={0}>
+          {children}
+          <Paper sx={{ p: 2 }}>
+            <Typography sx={{ mb: 2 }}>
```
NickAkhmetov, line 106: Similar to lines 90/91, `Stack` might simplify spacing here as well

context/app/static/js/hooks/useSelectItems.ts
```diff
@@ -0,0 +1,58 @@
...
+    case 'removeItem':
+      state.delete(action.payload);
+      return new Set([...state]);
+    case 'setSelectedItems':
+      return new Set(action.payload);
+    default:
+      console.warn('Unrecognized action type', action);
+      return state;
+  }
+}
```
NickAkhmetov, line 33: Let's revisit this reducer - I think it's worth trying the `produce` approach with `toggleItem`, that should prevent the strict mode issue you experienced with the original implementation since it will keep the previous state Set untouched. This implementation masks the issues caused by using the `add` and `delete` functions on the previous state directly since calling them twice in a row since calling those operations with the same input on a Set twice doesn't have any additional effect, but it may still cause unexpected bugs later on since it mutates the previous state.

context/app/static/js/shared-styles/chips/SelectableChip/SelectableChip.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import Chip, { ChipProps } from '@mui/material/Chip';
+import CheckRoundedIcon from '@mui/icons-material/CheckRounded';
+
+interface SelectableChipProps extends ChipProps {
+  isSelected: boolean;
+}
+
+function SelectableChip({ isSelected, ...rest }: SelectableChipProps) {
```
NickAkhmetov, line 9: There is a component in `FileBrowser` called `ControlChip` which is very similar + contains a minor styling change to make the width of the chip consistent when selected and deselected to avoid layout shifts; can we merge that functionality into this? I should have split that out into a separate file when I did that conversion.

context/app/static/js/shared-styles/cards/SelectableCard/SelectableCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <Card
+      sx={[
+        {
+          minWidth: 275,
+          backgroundColor: `${colorVariant}.main`,
+        },
+        ...(Array.isArray(sx) ? sx : [sx]),
+      ]}
+    >
```
NickAkhmetov, line 48: Making this a styled component like the `SelectableCardText` would simplify the `sx`-related logic

context/app/static/js/shared-styles/cards/SelectableCard/SelectableCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <CardContent component={Stack} direction="column" sx={{ height: '100%' }}>
+        <Stack direction="row" sx={{ justifyContent: 'space-between', alignItems: 'center' }}>
```
NickAkhmetov, line 50: `Stack` has `justifyContent` and `alignItems` as native props: https://mui.com/material-ui/react-stack/#interactive-demo

context/app/static/js/components/workspaces/WorkspaceField/WorkspaceField.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { FieldValues, Path, PathValue, useController, UseControllerProps } from 'react-hook-form';
+import Box from '@mui/material/Box';
+import TextField, { TextFieldProps } from '@mui/material/TextField';
+
+interface WorkspaceFieldProps<FormType extends FieldValues>
```
NickAkhmetov, line 6: After seeing how you have to assert the form type if you don't pass the `value`, I'm not as much of a fan of this as I was initially; I can revisit this to see if we can make usage less of a pain.

I wonder if we could substitute passing some of these through props with https://react-hook-form.com/docs/useformcontext and https://react-hook-form.com/docs/formprovider?

context/app/static/js/hooks/useSelectItems.ts
NickAkhmetov: Let's revise this reducer before merging

NickAkhmetov (approved):
Great work on resolving those comments! My only remaining concern is the reducer since it mutates the previous state.

context/app/static/js/hooks/useSelectItems.ts
john-conroy: Using a zustand store with initial state would require using `createStore` which seemed to be a lot of boilerplate for its worth and `immerReducer` also required creating a temp copy of draft to have a consistent return which somewhat defeats the purpose of introducing it.

Opted for the simplest approach to just add a temp copy of state to the existing reducer.

NickAkhmetov (approved):
👍🏻 This approach fixes my main concern, i.e. mutating the previous state.

I can investigate restoring the `toggleItem` action while integrating these changes into my branch.

PR #3292, opened 2023-10-17
NickAkhmetov/HMP-442 Workspaces polish

NickAkhmetov:
This PR updates the workspaces list styles to match the updated designs in Figma.

image

- Delete button stays disabled until an item is disabled
- Both buttons have tooltips
- `Launch Workspace` button has same functionality as clicking on the workspace name, is visible if no jobs are currently running
- `Stop Jobs` button kept previous functionality, is visible if there are jobs currently running
- `Deleting` button takes priority over other buttons if the current workspace state is `deleting`. This only appears if a user deletes a workspace then quickly refreshes the page.

I added an `elevated` button variant to reflect the Material 3 buttons used in the designs. I considered setting up the `@mui/material-next` package so we could directly use the variant defined there, but decided to keep that outside of the scope of this PR.


NickAkhmetov:
> Verify workspace launch/stop jobs still work as expected

Clicking launch workspace updates job status: image

`Please Wait` page then hangs after being opened in new tab; will need to troubleshoot this.

NickAkhmetov:
^ above issues seem to be related to end of workday shutdowns of workspace API, everything works as expected today.

context/app/static/js/shared-styles/sections/Description/Description.jsx
```diff
@@ -7,7 +7,9 @@ function Description({ padding, children, ...props }) {
   return (
     <StyledPaper $padding={padding} {...props}>
       <StyledInfoIcon color="primary" />
-      <Typography variant="body1">{children}</Typography>
+      <Typography variant="body1" as="div">
```
john-conroy, line 10:
```suggestion
      <Typography variant="body1" component="div">
```
Is `as` a Typography prop or html attribute I'm not familiar with?

context/app/static/js/theme/theme.tsx
```diff
@@ -248,6 +252,19 @@ const theme = createTheme({
...
     },
+    MuiButton: {
+      variants: [
+        {
+          props: { variant: 'elevated' },
+          style: {
+            boxShadow:
+              '0px 1px 3px 0px rgba(0, 0, 0, 0.20), 0px 2px 2px 0px rgba(0, 0, 0, 0.12), 0px 0px 2px 0px rgba(0, 0, 0, 0.14)',
+            borderRadius: '8px',
+            fontWeight: 300,
```
john-conroy, line 263: Does it make sense to include the `borderRadius` and `fontWeight` in the new variant? These seem to be specific to each use?

context/app/static/js/components/workspaces/WorkspacesList.jsx
```diff
@@ -1,25 +1,39 @@
...
-  } = useWorkspacesList();
+  const { workspacesList, handleCreateWorkspace, handleDeleteWorkspace } = useWorkspacesList();
+
+  const { selectedItems, toggleItem } = useSelectItems();
+  const { toastError } = useSnackbarActions();
+
+  const isDeleting = useRef(false);
+
+  const handleDeleteSelected = async () => {
+    const workspaceIds = Array.from(selectedItems);
```
john-conroy, line 26: Benefit of `Array.from` versus spreading?

context/app/static/js/components/workspaces/WorkspacesList.jsx
```diff
@@ -1,25 +1,39 @@
...
+    isDeleting.current = true;
+    try {
+      await Promise.all(workspaceIds.map((workspaceId) => handleDeleteWorkspace(workspaceId)));
+    } catch (e) {
+      toastError('Error deleting workspaces');
```
john-conroy, line 31:
```suggestion
      toastError('Error deleting workspaces.');
```

context/app/static/js/components/workspaces/WorkspacesList.jsx
```diff
@@ -1,25 +1,39 @@
...
-    handleCreateWorkspace,
-    handleStopWorkspace: handleStop,
-    handleStartWorkspace,
-  } = useWorkspacesList();
+  const { workspacesList, handleCreateWorkspace, handleDeleteWorkspace } = useWorkspacesList();
+
+  const { selectedItems, toggleItem } = useSelectItems();
+  const { toastError } = useSnackbarActions();
+
+  const isDeleting = useRef(false);
```
john-conroy, line 23: Perhaps I'm mistaken, but updating the ref doesn't cause a re-render and the button wouldn't show to be disabled while `isDeleting` is true?

john-conroy (commented):
Great! I was getting 500s from the workspaces API which prevented me from inspecting some of the flow so I have a few questions.

context/app/static/js/components/workspaces/WorkspacesList.jsx
```diff
@@ -30,17 +44,18 @@ function WorkspacesList() {
...
+              disabled={selectedItems.size === 0 || isDeleting.current}
+              tooltip="Delete selected workspaces"
             >
-              <DeleteIcon color="primary" />
-            </StyledButton>
-            <CreateWorkspaceButton handleCreateWorkspace={handleCreateWorkspace} />
-          </>
+              <DeleteRounded />
+            </WorkspaceButton>
+            <WorkspaceButton onClick={handleCreateWorkspace} tooltip="Create workspace">
```
john-conroy, line 55: Is this just a placeholder until the new dialog is available? `handleCreate` workspace expects a `workspaceName` arg,

context/app/static/js/shared-styles/sections/Description/Description.jsx
NickAkhmetov: I think I mistakenly used `as` because I'm more used to that prop name than `component`. It doesn't appear to be documented anywhere, but it did work as expected (i.e. it turned the container for the description into a div, so we didn't have nested `p > p` or `p > div` errors)

I'll swap this over to `component` though, I have no idea how this was functional.

context/app/static/js/components/workspaces/WorkspacesList.jsx
NickAkhmetov: After looking into it, it looks like spreading is actually a bit more performant than Array.from on arrays of under 10 elements, and equal in performance above that, so I've swapped this over to a spread.

context/app/static/js/theme/theme.tsx
NickAkhmetov: I moved these out into a styled component 👍🏻

context/app/static/js/components/workspaces/WorkspacesList.jsx
NickAkhmetov: This should now be handled appropriately by updating all workspaces list actions to use `useSWRMutation`, which lets us use the loading state provided by the hook.

context/app/static/js/components/workspaces/WorkspacesList.jsx
NickAkhmetov: This is just a placeholder for now, yes; once your PR is in, I'll update this to launch the dialog instead.

context/app/static/js/components/workspaces/api.ts
```diff
@@ -0,0 +1,316 @@
...
+  const stopJobCallback = useCallback(
+    async (jobId: number) => {
+      await trigger({
+        jobId,
+        url: apiURLs.stopJob(jobId),
+        headers,
+      });
+    },
+    [apiURLs, headers, trigger],
+  );
```
john-conroy, line 95: Pull `apiURLs` and `headers` into the useCallback so they don't have to be dependencies?

context/app/static/js/components/workspaces/WorkspaceListItem.tsx
```diff
@@ -0,0 +1,102 @@
...
+      </StyledButton>
+    );
+  }
+  if (workspace.jobs.length > 0) {
+    return (
+      <StyledButton
+        type="button"
+        variant="elevated"
+        size="small"
+        disabled={workspace.jobs.length === 0}
```
john-conroy, line 53: Should we disable the button if there are no `running` jobs?

context/app/static/js/components/workspaces/WorkspaceListItem.tsx
```diff
@@ -0,0 +1,102 @@
...
+      </StyledButton>
+    );
+  }
+  return (
+    <StyledButton
+      type="button"
+      variant="elevated"
+      href={typedGetWorkspaceLink(workspace)}
+      size="small"
+      rel="noopener noreferrer"
```
john-conroy, line 71: We're linking to the loading page, is this necessary?

john-conroy (commented):
Thanks! Still working through it. The diff became out of date while reviewing so you must have pushed some changes. A few comments for now.

context/app/static/js/components/workspaces/hooks.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  async function handleStopWorkspace(workspaceId: number) {
+    await stopWorkspace(workspaceId);
+    await mutate();
+  }
+
+  async function handleCreateWorkspace({ workspaceName }: { workspaceName: string }) {
+    await createAndLaunchWorkspace({ path: 'blank.ipynb', body: { workspace_name: workspaceName } });
+    await mutate();
+  }
```
john-conroy, line 44: I don't believe we're still using this?

john-conroy (commented):
The text for each workspace is somewhat misaligned.

Screenshot 2023-10-25 at 4 31 33 PM

context/app/static/js/components/workspaces/api.ts
NickAkhmetov: Both `apiURLs` and `headers` are provided by hooks which take care of providing the appropriate `AppContext` values (i.e. the workspaces endpoint and token.) Pulling them into the `useCallback` would result in those AppContext values being dependencies instead.

context/app/static/js/components/workspaces/WorkspaceListItem.tsx
NickAkhmetov: This was present in the previous implementation since it used an Outbound Link; I don't see the necessity for it, but I carried it over to minimize unintentional changes in functionality.

john-conroy:
The UI is allowing me to attempt to delete a running workspace which results in an error.

context/app/static/js/components/workspaces/WorkspaceListItem.tsx
NickAkhmetov: Done in https://github.com/hubmapconsortium/portal-ui/pull/3292/commits/cfa368b1aaea397977edf1b3b5b4df9223fb516f

NickAkhmetov:
> The UI is allowing me to attempt to delete a running workspace which results in an error.

I've added logic to disable selection for workspaces with running jobs along with an explanatory tooltip:

image

image

It also toggles the tooltip off if the user starts the workspace up while it's selected.

context/app/static/js/components/workspaces/NewWorkspaceDialog/NewWorkspaceDialog.tsx
```diff
@@ -59,7 +59,7 @@ function Description({ blocks }: { blocks: (string | ReactElement)[] }) {
   return (
     <Stack gap={2} sx={{ p: 2 }} component={Paper} direction="column">
       {blocks.map((block) => (
-        <Typography>{block}</Typography>
+        <Typography key={String(block)}>{block}</Typography>
```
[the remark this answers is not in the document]
NickAkhmetov: Good call, I made a ticket to add this: https://hms-dbmi.atlassian.net/browse/HMP-462

john-conroy (dismissed):
Nothing blocking. We can tweak the deletion and selection UI/UX down the road if needed.

PR #3295, opened 2023-10-18
NickAkhmetov/HMP-440 Workspaces Tracking

NickAkhmetov:
This PR adds tracking for the following:

- [x] Duration of workspace startup
  - [x] Total Duration
  - [x] Job Pending Duration
  - [x] Jupyter Env Startup Duration
- [x] Number of datasets selected on launch.
- [x] Errors due to too many selected and how many were selected.
- [x] Attempts to launch with protected data

The following items depend on John's PR and are taken care of in #3297:
- Count of people who used a template
- Which templates users are selecting when launching workspaces

The following items are tracked on the back-end:
- Count of people who have launched a workspace
- Count of workspaces created

The following items require some clarification re: what they refer to:
- Session duration

Additionally,
- Resolved issue with 400 error being thrown for reported web vitals
  - all faro events must be strings per findings in https://github.com/grafana/faro-web-sdk/issues/269
- Added user tracking to `faro`
  - User's email, type, and groups now included in initialization


john-conroy:
I think we want to track three different measures for workspace startup.

1. The total duration of the startup which you've captured here.
2. The duration the workspace job has the `pending` status. This is the time the job is waiting for resources to be allocated.
3. The duration the workspace job status changes from `pending` to `running`, but before a url is returned to launch the workspace. This is the start up time for the Jupyter Lab env.

NickAkhmetov:
Logged attempt to create workspace with 20 datasets, 10 of which were protected:
image

Logged removal of protected datasets:
image

Logged successful creation of workspace from search page:
image

Logged measurement of opening an already-launched workspace:
image

Measurement of opening an un-launched workspace:
image

Note that the jupyter URL is being received simultaneously with the first `running` status - we may want to decrease the polling time from 5 seconds. Here's what the measurement results look like with the polling interval decreased to 1000ms instead of 5000ms:

image

For now, I'm pushing the polling interval change so that we can more accurately observe the loading process; I can revert this if desired.

NickAkhmetov:
Reverting this to draft pending revisions in #3292 ; since the changes in that branch will heavily affect this branch, I'll need to rebase onto it.

context/app/static/js/components/workspaces/api.ts
```diff
@@ -256,11 +272,38 @@ export function useStartWorkspace() {
...
 
 async function createWorkspaceFetcher(_key: string, { arg: { body, url, headers } }: { arg: CreateWorkspaceArgs }) {
+  trackEvent({
+    category: 'Workspace Creation',
+    action: 'Create Workspace',
+    value: {
+      name: body.name,
+      description: body.description,
+      globus_groups_token: body.workspace_details.globus_groups_token,
+      files: body.workspace_details.files.map((f) => f.name),
```
john-conroy, line 303: Should we track workspace template names and uuids instead of the files? The files can include the templates strings which could be quite large/uninformative.

john-conroy (commented):
Nice! One comment.

context/app/static/js/components/workspaces/api.ts
john-conroy: Realizing now that's already in place!

john-conroy (approved):

PR #3297, opened 2023-10-19
NickAkhmetov/Add tracking for workspace creation with templates

NickAkhmetov:
This PR adds tracking for workspace creation with templates. In order to avoid faro errors, I've stringified all the passed values; this extra stringification can be removed once we merge my other PR in, which adds automatic handling of stringification for Faro.


NickAkhmetov:
Closing in favor of #3295 given that this was never merged into the templates branch while it was active.

PR #3300, opened 2023-10-23
John conroy/fix nav click HMP-452

john-conroy:
The previous appearance seems to be maintained. I believe this also resolves HMP-451 which was misdiagnosed. 

Screenshot 2023-10-23 at 3 15 39 PM


NickAkhmetov (approved):
Nicely done!

PR #3302, opened 2023-10-24
John conroy/collections without doi HMP-437

john-conroy:
Collections are being expanded to allow consortium members to create 'custom' collections which will not have a DOI. These collections only exist as mock data on DEV/TEST. While we are working on a designs, we have have opted to not display these collections on the landing page.

Filed 460/461 design issues.

@tsliaw tagging you for awareness.


context/app/static/js/pages/Collections/types.ts
```diff
@@ -0,0 +1,17 @@
+interface Collection {
+  uuid: string;
+  title: string;
+  hubmap_id: string;
+  datasets: { hubmap_id: string }[];
+}
+
+interface CollectionHit {
+  _source: Collection;
+}
```
NickAkhmetov, line 10: Let's keep this for now, once we start moving away from searchkit in earnest I think this would be a prime case for a reusable generic type:

```
interface Hit<T> {
  _source: T
}
```

NickAkhmetov (approved):

context/app/static/js/pages/Collections/types.ts
john-conroy: Agreed, I didn't want to get too far into it here.

PR #3305, opened 2023-10-26
John conroy/fix dependencies in push

john-conroy:
Fix push script to once again stage changes to dependencies markdown. Remove un-merged changelog due to `.MD` extension.


NickAkhmetov (approved):
Ah okay, this makes sense now - I had found the `dependencies.md` file during the big sweep of updates, but since it was very outdated I removed it; it makes sense if it's supposed to be populated automatically on deploy.

PR #3308, opened 2023-11-01
HMP-475 Fix changelog version and exit build if Docker is not running

lchoy:
Adds a check at the start of the push.sh script to check if Docker is running. The push should exit early to avoid the changelog version getting out of sync with the docker tag.

Updates erroneous tag in last CHANGELOG entry.


NickAkhmetov (approved):
We should add a `version number skipped` entry for `v0.82.2`; otherwise, looks good to me! Great way to avoid this issue going forward.

NickAkhmetov (approved):
I added the extra entry myself, should be all set now!

PR #3313, opened 2023-11-03
John conroy/ws dialog tables

john-conroy:
Towards HMP-398. Introduces the datasets table to the dialog.

Screenshot 2023-11-08 at 10 06 54 AM

### DIFF
--- a/context/app/static/js/components/organ/Samples/Samples.tsx
+++ b/context/app/static/js/components/organ/Samples/Samples.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable no-underscore-dangle */
 import React, { useMemo } from 'react';
 import Button from '@mui/material/Button';
 
--- a/context/app/static/js/components/organ/Samples/Samples.tsx
+++ b/context/app/static/js/components/organ/Samples/Samples.tsx
@@ -1,64 +1,25 @@
 /* eslint-disable no-underscore-dangle */
 import React, { useMemo } from 'react';
-import format from 'date-fns/format';
 import Button from '@mui/material/Button';
 
-import { InternalLink } from 'js/shared-styles/Links';
 import SectionContainer from 'js/shared-styles/sections/SectionContainer';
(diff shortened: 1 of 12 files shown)


context/app/static/js/components/organ/Samples/Samples.tsx
```diff
@@ -1,64 +1,25 @@
 /* eslint-disable no-underscore-dangle */
```
NickAkhmetov, line 1: We should be able to remove this eslint-disable - `_id` and `_source` are global exceptions to this rule

context/app/static/js/components/workspaces/WorkspaceDatasetsTable/WorkspaceDatasetsTable.tsx
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-underscore-dangle */
```
NickAkhmetov, line 1: Ditto here

context/app/static/js/components/workspaces/WorkspaceDatasetsTable/WorkspaceDatasetsTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+              <Delete
+                onClick={() => removeDatasets([...selectedRows])}
+                tooltip="Remove selected datasets."
+                disabled={selectedRows.size === 0}
+              />
+            )}
+          </Stack>
+        }
+      />
+      <EntitiesTables<DatasetDocument> entities={[{ query, columns, entityType: 'Dataset' }]} />
```
NickAkhmetov, line 56: This is more of a note for myself, but I'm going to investigate whether we can avoid passing in types as part of the HTML element by deriving the type from the `entities` property.

NickAkhmetov (approved):
Looks great! I really like your strategy of handling promise rejection inside the synchronous `handleClick` callback - it looks much cleaner than including a `.catch` as part of the `onClick` prop itself, so I'll adopt this as well.

PR #3316, opened 2023-11-08
NickAkhmetov/HMP-492 Prevent unnecessary calls to workspaces API by users without permissions

NickAkhmetov:
This PR prevents noisy calls to fetch workspaces/jobs by users without access to the workspaces API.


john-conroy (approved):
Great, thanks!

PR #3317, opened 2023-11-08
NickAkhmetov/HMP-446 Add type-check steps to precommit and node CI

NickAkhmetov:
This PR adds type checking to the `node-test` CI workflow and to the `lint-staged` precommit script.

It also resolves all type errors reported when running the type check locally.

The `lint-staged` script uses `tsc-files` to lint only the staged files instead of rechecking unchanged files.


john-conroy (approved):
Thanks!

PR #3320, opened 2023-11-09
NickAkhmetov/HMP-478 Fix HTML rendering inside markdown

NickAkhmetov:
This PR adds the `rehype-raw` plugin to parse HTML inside markdown sections, creates a global markdown configuration component, and replaces all current calls to `ReactMarkdown` with this configured instance to ensure this parsing occurs in every relevant location.

Before:
image

After:
image


john-conroy (approved):
Nicely done!

PR #3321, opened 2023-11-09
NickAkhmetov/HMP-439 Convert homepage images to .webp

NickAkhmetov:
This PR

- Adjusts the homepage carousel to point to .webp sources and fall back to .png
- Adjusts the related util to generate srcset entries for different file types
- Adds a script to bulk convert a directory of images to .webp

Before changes:
image

After changes:
image

Overall size comparison:

WEBP: 1.26 MB total for all slides/size variations
image

PNG: 3.99 MB total for all slides/size variations
image

---

I additionally ran into some issues using the global `CDN_URL` variable Webpack provides, so I added that variable (along with all the values provided by Flask) to `global.d.ts`. I also moved `global.d.ts` to the `typings` folder.


john-conroy (approved):
Nice! Should we create a ticket to extend the script to create the resolutions we expect and optimize the images?

NickAkhmetov:
> Nice! Should we create a ticket to extend the script to create the resolutions we expect and optimize the images?

I've created [HMP-494](https://hms-dbmi.atlassian.net/browse/HMP-494) to track this, good call!

PR #3323, opened 2023-11-10
Updated the HRA API endpoint to point to the HuBMAP-only version

bherr2:
For both EUI and the organ-info web component, we are switching the API endpoint to a hubmap-only version of the HRA-API (formerly known as the CCF-API). It is exactly the same code as in production, but just limited to HuBMAP data.

This has been a request for a while and should provide good performance now. It can be merged at any time.


NickAkhmetov (approved):
Thanks, Bruce! I tested it out and it works perfectly. 👍🏻

PR #3328, opened 2023-11-14
NickAkhmetov/HMP-495 Gene Detail Page-Related Enhancements

NickAkhmetov:
This PR: (expand each detail for relevant screenshots)

<details><summary>Adds support for viewing datasets containing a cell type</summary>
<p>

image

</p>
</details> 
 

<details><summary>Adds a message to the search page to indicate when only a certain cell type's related datasets are being displayed</summary>
<p>

image
Note that only 35 datasets are available.

</p>
</details> 

<details><summary>Adds the ontology API status to the status page</summary>
<p>

image

</p>
</details> 

<details><summary>Causes a 404 error message to be displayed if an invalid gene name is used for detail lookups</summary>
<p>

image

</p>
</details> 

Miscellaneous changes made to support the above:
- The `LookupEntity` component has been removed; this was previously only used to provide `useEntityData` results to `SearchNote` by cloning its child `SearchNote` component and providing the entity as a prop. This has been phased out in favor of using the `useEntityData` hook in the message component itself, and was necessary in order to support displaying search page messages that do not look up an entity.
- The `fetcher` and `multiFetcher` utilities have been extended to allow providing custom error messages depending on the received status code. The `ErrorBoundary` logic has been extended to account for this as well.
- Additional util functions for compressing/decompressing a list of strings have been added to `js/helper/compress` in order to shorten the search params passed when viewing all datasets related to a cell type. The maximum recommended size for URLs is 2000 characters, beyond which point browser support varies. This should work as an interim solution to avoid excessively long URL's; however, we will need to revisit this since it's still prone to eventually exceeding the maximum URL length.
  - As a sample - the UUID's for the 35 datasets containing `endothelial cell of lymphatic vessel` are 1154 characters long without compression and are shrunk to 1028 characters via LZSString's `compressToBase64`. I tried using the window `atob` and `btoa` methods for this, but those inflated the size to 1500 characters instead.
  - I split the compression helpers out into their own file so that `LZString` isn't imported on every page that uses `js/helpers/functions`.
  - A potential future improvement would be to pass only the cell type and fetch the UUID's to use while filtering once on the search page. I wanted to avoid messing with searchkit more than necessary, so I took the approach of passing it in the URL params in the meantime.
- An additional `routes_cell_types` file is used for cell_type-related routes as the cell type functionality does not appear to be functional in prod yet; an explanatory todo is provided.

Other Notes:
- The 404 error causes an overlay to be displayed when thrown during development in addition to triggering the error boundary. From what I've found, this is an expected/standard behavior for dev mode.


yuanzhou:
FYI @NickAkhmetov, the ontology-api status is available in gateway's `/status.json` endpoint across the board.

NickAkhmetov:
@yuanzhou Thanks for the heads up! I've updated the screenshot accordingly.

john-conroy (approved):

PR #3333, opened 2023-11-16
John conroy/workspace detail HMP-465

john-conroy:
- Add workspace detail page.
- Increase gap between templates in template grid.
- Remove tsc-files from our pre-commit type checks as it was not using the project's tsconfig. 

Currently we're relying on `request_workspace_details` due to inconsistencies with `current_workspace_details`. Juan is working to add `dataset_uuid` to the symlinks in `current_workspace_details` and to ensure files are always objects with a `name` and `path`.

I'll open up a follow-up PR to show the datasets table with an alert message when no datasets are symlinked to the workspace.
 
Screenshot 2023-11-16 at 12 53 43 PM


context/app/static/js/components/workspaces/hooks.ts
```diff
@@ -146,4 +175,33 @@ export function useCreateAndLaunchWorkspace() {
...
+
+  if (!matchedWorkspace) {
+    return false;
+  }
+
+  const timeLeft = getWorkspaceTimeLeft(matchedWorkspace);
+
+  if (!timeLeft) {
+    return false;
+  }
```
NickAkhmetov, line 193: I think we may be able to avoid the extra call to `getWorkspaceTimeLeft` after a matched workspace is found; what do you think about this approach?

```suggestion
type FoundPair = [MergedWorkspace, number];

function useSessionWarning(workspaces: MergedWorkspace[]) {
  const [matchedWorkspace, timeLeft] = workspaces.reduce<FoundPair | undefined>((acc, ws) => {
  	if (acc) return acc; // Avoids extra calls to `getWorkspaceTimeLeft` after workspace has been found since we can't break a .reduce() 
    const time = getWorkspaceTimeLeft(ws);
    return time ? [ws, time]: acc; // If a time is found, return the workspace with time, otherwise continue iteration
  }, undefined);
  
  if (!matchedWorkspace || !timeLeft) {
  	return false;
  }
```

context/app/static/js/pages/Workspace/Workspace.tsx
```diff
@@ -0,0 +1,119 @@
...
+import TemplateGrid from 'js/components/workspaces/TemplateGrid';
+import SummaryData from 'js/components/detailPage/summary/SummaryData';
+import SectionPaper from 'js/shared-styles/sections/SectionPaper';
+import LabelledSectionText from 'js/shared-styles/sections/LabelledSectionText';
+import { condenseJobs } from 'js/components/workspaces/utils';
+import JobStatus from 'js/components/workspaces/JobStatus';
+import { useWorkspaceDetail, useSessionWarning } from 'js/components/workspaces/hooks';
+import { Alert } from 'js/shared-styles/alerts';
+import WorkspaceLaunchStopButton from 'js/components/workspaces/WorkspaceLaunchStopButton';
+import { Typography } from '@mui/material';
```
NickAkhmetov, line 19:
```suggestion
import Typography from '@mui/material/Typography';
```

context/app/routes_workspaces.py
```diff
@@ -16,11 +16,21 @@ def index():
...
     )
+
+
+@blueprint.route('/workspaces/<workspace_id>')
+def workspace_detail(workspace_id):
+    flask_data = {**get_default_flask_data()}
+    return render_template(
+        'base-pages/react-content.html',
+        flask_data=flask_data,
+        title='Workspace'
```
NickAkhmetov, line 35: We should consider setting the current workspace name as the page title once it's loaded on the front-end - this can be an additional improvement 👍🏻

context/app/static/js/shared-styles/cards/SelectableCard/SelectableCard.tsx
```diff
@@ -32,7 +32,7 @@ function SelectableCard({
           <SelectableCardText variant="subtitle1" $colorVariant={colorVariant}>
             {title}
           </SelectableCardText>
-          <Checkbox checked={isSelected} onChange={(e) => selectItem(e)} value={cardKey} />
+          {selectItem && <Checkbox checked={isSelected} onChange={(e) => selectItem(e)} value={cardKey} />}
```
NickAkhmetov, line 35: This approach should work too:

```suggestion
          {selectItem && <Checkbox checked={isSelected} onChange={selectItem} value={cardKey} />}
```

NickAkhmetov (approved):
Looks great, nice work! I tested things out locally and they seemed to look as expected. There is additional room for enhancements as previously discussed (e.g. adding loading skeletons, refactoring some of the shared components that have odd props) but the core functionality looks great and the finer details are exactly what iterative development is for.

I did have a few tiny nitpicks on the smaller stuff, but nothing substantial.

context/app/static/js/components/workspaces/hooks.ts
john-conroy: I think this solution is also running into similar issues.

NickAkhmetov (approved):

PR #3335, opened 2023-11-20
NickAkhmetov/HMP-457 Add workspace session renewal

NickAkhmetov:
This PR adds an alert to the top of the workspaces list which allows for the renewal of active workspace sessions.

https://github.com/hubmapconsortium/portal-ui/assets/19957804/9f2aa718-3693-4dae-be85-13c96e833998

Ideas for additional improvements:
- Adding a threshold time at which to start displaying that alert; right now, it displays this alert whenever any workspace is active
- Linking to the running workspace's detail page as part of the alert

### DIFF
--- a/CHANGELOG-hmp-457.md
+++ b/CHANGELOG-hmp-457.md
@@ -1,0 +1,1 @@
+- Add a prompt to renew workspace sessions to the top of the workspaces screen.
--- a/context/app/static/js/components/workspaces/WorkspaceSessionWarning.tsx
+++ b/context/app/static/js/components/workspaces/WorkspaceSessionWarning.tsx
@@ -1,0 +1,45 @@
+import React from 'react';
+import { Alert } from 'js/shared-styles/alerts';
+import Button from '@mui/material/Button';
+import { useSnackbarActions } from 'js/shared-styles/snackbars';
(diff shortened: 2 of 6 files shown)


john-conroy (commented):
The alignment of the text and button looks slightly off?

NickAkhmetov:
> The alignment of the text and button

This was due to the alert applying some additional padding around the action by default; it should be resolved now.

image
image

john-conroy (approved):
> Adding a threshold time at which to start displaying that alert; right now, it displays this alert whenever any workspace is active

I think this is okay for now. As users will be generally unfamiliar with the workspaces feature and its limitations we should make this information immediately available. Especially since opening a workspace takes you out of the portal.

> Linking to the running workspace's detail page as part of the alert

All for it!

We might also want to think about adding a confirm dialog of some sort.

NickAkhmetov:
> Linking to the running workspace's detail page as part of the alert

Created HMP-514 to track this 👍🏻 

> We might also want to think about adding a confirm dialog of some sort.

We have HMP-502 reported for confirming deletion, so I'm a bit more hesitant with this since renewing isn't as destructive an action; introducing another confirmation for renewing the session might cause confirmation fatigue for users, which can cause accidental deletions due to user error/autopilot.

PR #3337, opened 2023-11-20
John conroy/hide templates HMP-507

john-conroy:
Do not display workspace templates with designated `is_hidden` bool.


NickAkhmetov (approved):
LGTM 👍🏻

PR #3338, opened 2023-11-20
NickAkhmetov/HMP-501 Link to organ pages from gene detail page

NickAkhmetov:
This PR adds links from the gene detail page's `cell types` table to the corresponding organ.

image

During development, I discovered we don't currently handle the edge case where we don't have any available cell types/organs to display and created https://hms-dbmi.atlassian.net/browse/HMP-512 to track this.

### DIFF
--- a/context/app/static/js/components/genes/CellTypes/CellTypes.tsx
+++ b/context/app/static/js/components/genes/CellTypes/CellTypes.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 
 import Table from '@mui/material/Table';
 import TableHead from '@mui/material/TableHead';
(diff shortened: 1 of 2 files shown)


context/app/static/js/components/genes/CellTypes/CellTypes.tsx
```diff
@@ -33,6 +34,21 @@ export function TableSkeleton({ numberOfCols = columnLabels.length }: { numberOf
   );
 }
 
+function OrgansCell({ organs }: { organs: { name: string }[] }) {
+  const contents = !organs ? (
+    <>&mdash;</>
+  ) : (
+    organs.map((o, idx) => (
```
john-conroy, line 41: Slight preference for `i` over `idx` for consistency with the repo.

john-conroy: If you're only using `organ.name` destructure it in the callback args?

john-conroy (dismissed):
Small comments, but definitely not blocking. Thanks!

john-conroy (approved):

PR #3340, opened 2023-11-20
John conroy/inconsistent workspace files

john-conroy:
Temp fix until we narrow down file structure with Juan.


NickAkhmetov (approved):
👍🏻 Solid fix!

Should we include the reasoning as a comment on the type as well? Just in case we end up not coming back to this for a while/for the benefit of new team members.

john-conroy:
Hoping to have clarity tomorrow, but I'll add a comment!

NickAkhmetov (approved):

PR #3341, opened 2023-11-21
NickAkhmetov/HMP-503-HOTFIX Fix workspace launch form lock

NickAkhmetov:
My previous PR (#3336) introduced a regression where submitting the workspace launch dialog and having a validation failure occur locked the user out of continuing the workspace creation process until they reopened the dialog to reset form state.

This PR fixes this regression by using `isSubmitSuccessful` instead of `isSubmitted` to prevent queueing multiple workspace launches. The former is only toggled to true after validation succeeds, while the latter is set to true as soon as the user submits - even if the submit fails runtime validation.

The original issue of duplicate workspace creation requests being created is still handled.

This PR also fixes an error finding the appropriate file path to provide when launching a workspace.


john-conroy (commented):
I'm getting `/workspaces/start/439?notebook_path=` when checking this out locally.

john-conroy:
The workspace passed to `startAndOpenWorkspace` only has files in `request_workspace_details` not `current_workspace_details`.

NickAkhmetov:
I did notice the leading slash is missing, but it didn't seem to affect anything functionality-wise; for the sake of consistency, I will add logic to prepend a `/` if one is not already present.

john-conroy (approved):
It works! Thanks!

PR #3351, opened 2023-11-30
Update Node version in Github Action workflow scripts

lchoy:
Modifies the version string in .nvmrc to remove the 'v' prefix to enable compatibility with GitHub Action workflow scripts. Update action workflow scripts to use the node-version-file.


NickAkhmetov:
The actions should already be using the node version specified in the .nvmrc file; was there an issue of some sort?

lchoy:
The version in the .nvmrc is not supported. See https://github.com/actions/setup-node#supported-version-syntax. As a result, the workflow produces this warning about overriding the default to 16: 
portal-ui-workflow

I opened the PR and set to draft to force the workflow to run. Goal is to use a consistent version of the action and node-version. Perhaps `node-version: latest` would be the desired option?

NickAkhmetov:
> The version in the .nvmrc is not supported.

The version in the .nvmrc is `v20.7.0`; maybe the `v` prefix is causing some issues? Could we try removing the `v` and seeing if GitHub actions properly detect it? It seems like that would [still match the .nvmrc spec](https://github.com/nvm-sh/nvm#nvmrc) while matching the expected format indicated in the `setup-node` readme.

Regardless of whether the action uses the .nvmrc file or not (I still strongly prefer using the `node-version-file` approach to avoid having multiple sources of truth), can we avoid regressing to Node 18?

PR #3354, opened 2023-12-04
John conroy/workspace tutorial HMP-425

john-conroy:
We'll need to revisit the styling when there are more tutorials, but this works for now. We should also refactor the tutorial prompt and convert it to ts at a later date.

I imagine we'll want to add the tutorials link to navigation once workspaces is live?

Screenshot 2023-12-04 at 10 44 19 AM

Screenshot 2023-12-04 at 10 44 12 AM

Screenshot 2023-12-04 at 10 44 01 AM

### DIFF
--- a/CHANGELOG-workspace-tutorial.md
+++ b/CHANGELOG-workspace-tutorial.md
@@ -1,0 +1,2 @@
+- Add tutorials landing page.
+- Add workspaces tutorial.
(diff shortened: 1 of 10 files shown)


NickAkhmetov (approved):
Nicely done, thanks for getting this out so quickly!

As an eventual enhancement, it could be nice to eventually externalize the tutorials list so we can add more tutorials without a deploy, but this isn't relevant to our immediate concerns 👍🏻

PR #3361, opened 2023-12-12
NickAkhmetov/HMP-549 Revise to use assaytype endpoint

NickAkhmetov:
This PR adjusts the generation of vitessce visualizations to be done using the `assaytype` endpoint instead of the deprecated TypeClient.

The `SOFT_ASSAY_ENDPOINT` environment variable has been added to configs by Zhou. Since it is a part of the ingest-api, I've adjusted the service status table to display the ingest API URL in use by trimming the `/assaytype` suffix from the soft assay endpoint.

TODO:
- ~~Remove type service env variables and add assaytype endpoint env variable~~
- ~~Troubleshoot seqfish datasets only displaying descendant pyramids~~
- ~~Update portal-visualization version~~


example-app.conf
NickAkhmetov, line 1: I adjusted the spacing of the commented out endpoints so they're all aligned with the `USER_TEMPLATES_ENDPOINT` - this was purely an aesthetic change. The meaningful additions are the different `SOFT_ASSAY_ENDPOINT` values for the `TEST`, `DEV` and `PROD` environments on lines 38, 49, and 60.

lchoy (approved):

PR #3363, opened 2023-12-13
Dataset Type Hierarchical Facet Example

john-conroy:
Example for Tiffany/Samson to inspect.


NickAkhmetov:
We could probably close this in favor of the merged-in implementation?

PR #3365, opened 2023-12-21
Update portal images cloud formation

john-conroy:
Largely adapted from the 'CORS-With-Preflight' policy. Unfortunately you can't combine an existing `CloudFrontResponseHeadersPolicy` with a managed policy of the same type.

### DIFF
--- a/cloudformation-templates/portal-images.yml
+++ b/cloudformation-templates/portal-images.yml
@@ -30,9 +30,11 @@ Resources:
           AllowedMethods:
             - "HEAD"
             - "GET"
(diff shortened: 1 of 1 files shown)


NickAkhmetov (approved):
Ah, okay, this makes sense:

- `OPTIONS` wasn't a previously allowed method, so the preflight calls that were being sent when attempting to fetch the markdown were being rejected outright
- CORS headers were also not enabled since images and plaintext requests are handled differently by the browser

lchoy (approved):

PR #3367, opened 2024-01-05
Add explicit logging for assaytype lookup failures + Fix endpoint

NickAkhmetov:
This PR adds error logging to display the parameters of requests to the assaytype service.


context/app/api/client.py
```diff
@@ -211,7 +211,17 @@ def get_vitessce_conf_cells_and_lifted_uuid(self, entity, marker=None, wrap_erro
                 def get_assaytype(entity):
                     uuid = entity.get('uuid')
                     url = f"{current_app.config['SOFT_ASSAY_ENDPOINT']}/{uuid}"
-                    return requests.get(url, headers=self._get_headers()).json()
+                    headers = self._get_headers()
+                    try:
+                        request = requests.get(url, headers=headers)
```
lchoy, line 219: This is a nitpick, but technically, requests.get() returns a requests.Response object.
```suggestion 
                        response = requests.get(url, headers=headers)
```

context/app/api/client.py
```diff
@@ -211,7 +211,17 @@ def get_vitessce_conf_cells_and_lifted_uuid(self, entity, marker=None, wrap_erro
...
+                        return request.json()
+                    except Exception as e:
```
lchoy, line 221: The exception is raised when the response doesn't contain json. Should there be a check for the response status?

lchoy (changes requested):

context/app/api/client.py
NickAkhmetov: Done - correct/logical naming is important for clarity 😄

context/app/api/client.py
```diff
@@ -211,7 +211,17 @@ def get_vitessce_conf_cells_and_lifted_uuid(self, entity, marker=None, wrap_erro
...
+                        current_app.logger.error(
+                            f'Failed to fetch from {url} with headers {headers}: {e}'
```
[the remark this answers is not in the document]
NickAkhmetov: Done; I've added logic to update `Authorization`'s value to `REDACTED` if it's present in the headers.

context/app/api/client.py
NickAkhmetov: I figured we could implicitly handle any errors thrown by `requests.get` by including it in the try/except and catching all `Exception`s, not just the `JSONDecodeError` we were encountering before: https://docs.python-requests.org/en/latest/_modules/requests/exceptions

I have added `status` to the logged error, and also updated the error to be logged in the form of an object to make it easier to identify fields.

lchoy (approved):

PR #3369, opened 2024-01-05
NickAkhmetov/CAT-212 Implement gene detail page feedback

NickAkhmetov:
- Updated section order and titles
- Added entity header
- Added gene icon where appropriate

TODOs from Figma designs:
- [x] Gracefully handle missing cell types/organs information
- ~~Add cell types section description~~
- [x] Update organ section description
- [x] Redo organ section as accordions
- [x] Rename "Nuclei in reference" in azimuth component to `Cells in Reference`
- [x] Add `Azimuth Overview` tooltip 
- [ ] Add context textbox to search results

Other general TODOs:
- [x] Organize Cells API and UBKG API calls/overall logic


context/app/static/assets/svg/workspaces.svg
NickAkhmetov, line 1: This got swept up in linting while I was confirming that its structure matched the new gene icon's structure

NickAkhmetov:
Current state:
image

Remaining TODOs:

- I still want to parallelize the cells API queries via asyncio (looking at [flask docs](https://flask.palletsprojects.com/en/2.3.x/async-await/) it seems like we will need to install some additional dependencies to support this so this may be a later improvement.) The main benefits here would be visible for iterative queries like e.g. getting the cell type counts for each organ on the cell types page.
- Nils asked me to try preselecting genes in the azimuth visualization when we presented the first iteration of the gene detail page - this may be a stretch goal, but is something I can work on if the above is not possible.
- We still need to restore the cell types table, which is currently blocked by the cells API lookups for cells by gene expression levels not working properly on dev/test.

NickAkhmetov:
Latest updates:
- Cell types section will not be implemented yet, but I am leaving my current related code intact
- I parallelized the requests to the cells API
- I reorganized the related code to make it a bit easier to follow

TODO:
- Waiting for updated organ section description from CMU
- Should take another look at preselecting marker genes
- Need to fix node test failure related to my fix for CAT-527

NickAkhmetov:
I've opened this for review and closed the cell detail page PR; since that page stopped working yesterday as a result of the cells API no longer accepting CL ID's as inputs, I'll need to revisit that, but this will at least allow us to get the gene pages out of their current broken state.

This branch contains a lot of minor improvements I've made along the way while working in this branch/its previous parent over the past month or two, including some pixel changes we made during a mockup review session with Tiffany, a fix for CAT-527 (no organs displaying in molecular data query results), and the conversion of the stacked bar chart to TypeScript.

The current state of the gene detail page is visible below:

image

Remaining TODOs:
- Update organ section description - waiting for CMU to provide this
- Marker gene preselection for azimuth visualizations - working on this, may become separate PR
- Search page context box - since the related functionality (cell types table) is not enabled, I did not get around this; I think tackling it as a separate ticket/PR would be more logical than trying to tack it on to this.
- Any requested revisions from reviews

context/app/static/js/shared-styles/charts/hooks.ts
```diff
@@ -0,0 +1,138 @@
+import { MouseEvent, useMemo } from 'react';
```
john-conroy, line 1: Are the changes to the charts components/hooks just typescript conversions or also a mix of meaningful changes?

john-conroy (commented):
Looks good! A few questions, but nothing blocking.

context/app/routes_cell_types.py
```diff
@@ -1,30 +1,359 @@
...
+
+# Converts a resultset to a list
+def _unwrap_result_set(result_set):
+    length = len(result_set)
+    result_list = result_set.get_list()._get(length, 0)
+    return result_list
+
+
+# Combines results from multiple modalities
+def _combine_results(modality1, modality2):
```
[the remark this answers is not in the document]
NickAkhmetov: I'll add handling for this case as well 👍🏻

context/app/routes_cell_types.py
```diff
@@ -1,30 +1,359 @@
...
+        fields = ['uuid', 'hubmap_id', 'origin_samples_unique_mapped_organs',
+                  'sample_category', 'last_modified_timestamp']
+        samples = client.get_entities(plural_lc_entity_type='samples',
+                                      query_override=query_override, non_metadata_fields=fields)
+        samples = [{'uuid': sample['uuid'],
+                    'hubmap_id': sample['hubmap_id'],
+                    'organ': sample['origin_samples_unique_mapped_organs'],
+                    'sample_category': sample['sample_category'],
+                    'last_modified_timestamp': sample['last_modified_timestamp']
+                    } for sample in samples]
```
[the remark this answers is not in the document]
NickAkhmetov: I think the main goals were to rename the mapped organs key and remove any unused fields, but I'll double check. In the long run we should be able to move this into a front-end ES lookup with the datasets as a constraint, this was an artifact from trying to prototype the cell types detail page when we initially believed we were ready to implement it (i.e. when we were getting the cell type/biomarker mapping info from the manually-populated UBKG index).

context/app/static/js/shared-styles/charts/hooks.ts
NickAkhmetov: Full summary of revisions to related components:

# VerticalStackedBarChart.jsx -> VerticalStackedBarChart.tsx

- Allow setting custom X/Y scales via `getXScaleRange`/`getYScaleRange` to allow changing scales
- Styling updates for labels to make them match the latest designs
- Allow formatting Y axis tick values with the `getTickValues` prop
- Allow using log scales in general

# hooks.js -> hooks.ts

- Added hooks to create linear, log, band, and ordinal d3 scales

# types.ts
- Added util function to distinguish between tooltips with and without bar data (i.e. tooltips on axis labels vs. tooltips on data)

# utils.js -> utils.ts

- Improved type safety of `trimStringWithMiddleEllipsis` so it doesn't crash when passed a null or undefined value
- Added `defaultXScaleRange` / `defaultYScaleRange`

context/app/static/js/hooks/useCrossModalityApi.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Samples corresponding to datasets that contain the requested feature.
+ */
+export interface FeatureSample {
+  hubmap_id: string;
+  last_modified_timestamp: number;
+  organ: string[];
+  sample_category: string;
+  uuid: string;
```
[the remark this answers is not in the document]
NickAkhmetov: Created a JIRA ticket: https://hms-dbmi.atlassian.net/browse/CAT-572

context/app/routes_cell_types.py
```diff
@@ -1,30 +1,359 @@
...
+    rna_datasets, atac_datasets = await gather(
+        to_thread(get_datasets_for_modality, 'rna'),
+        to_thread(get_datasets_for_modality, 'atac'))
+    datasets = _combine_results(rna_datasets, atac_datasets)
+    datasets = _unwrap_result_set(datasets)
+    datasets = list(map(lambda x: x['uuid'], datasets))
+    return datasets
+
+
+async def _get_organs_for_cell_type(client, cell_type):
```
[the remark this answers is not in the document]
NickAkhmetov: Good call, this has been done 👍🏻

context/app/static/js/components/cell-types/CellTypesBiomarkersTable.tsx
```diff
@@ -1,25 +1,43 @@
...
 }
 
-function BiomarkerTableRow({ biomarker, type }: { biomarker: CellTypeBiomarkerInfo; type: 'genes' | 'proteins' }) {
+function otherTableKey(tableKey: TableKey): TableKey {
+  const otherKey = tableKeys.find((key) => key !== tableKey);
+  if (!otherKey) {
+    throw new Error(`Could not find other table key for ${tableKey}. This should never happen.`);
+  }
+  return otherKey;
+}
```
john-conroy, line 38: This will get picked up by the route level error boundary? Should we handle it more gracefully in case it ever does happen?

john-conroy (approved):
A comment, but not blocking. Thanks!

context/app/static/js/components/cell-types/CellTypesBiomarkersTable.tsx
NickAkhmetov: This is a type guard function used to safely access `biomarkers.genes` or `biomarkers.proteins` below, it should never reach the error case since those keys should always be present. This was added in an effort to deduplicate the two tables' code per your previous suggestion.

These tables are also not currently functional due to the cells API not having this functionality hooked up and the UBKG API no longer caching this data for us.

PR #3375, opened 2024-01-23
NickAkhmetov/[CAT-533] Disable searchkit timeout, update faro version

NickAkhmetov:
This PR disables the default 5000ms searchkit timeout and updates the faro web sdk version.

Tested locally with "regular 2g" speed setting in network tools:

image
image


john-conroy (approved):
Did you discover `0` disabled the timeout through experimentation or was it somewhere in those legacy docs?

NickAkhmetov:
I couldn't find relevant documentation about this prop's configuration, so I tried out 0 since I've seen that used as a "disable timeout" option in many other contexts; I tested it out on all three search pages to confirm it worked as expected.

PR #3378, opened 2024-01-25
John conroy/fix matomo events

john-conroy:
- Fix matomo event tracking to provide name field where applicable.
- Track entity page visits by HuBMAP ID not UUID.

I also added tracking for publication pages and cleaned up an old changelog file that wasn't compiled as it was missing a `.md` extension.

### DIFF
--- a/context/app/static/js/helpers/trackers.js
+++ b/context/app/static/js/helpers/trackers.js
@@ -85,7 +85,8 @@ const makeFaroSafe = (obj) =>
   Object.fromEntries(Object.entries(obj).map(([key, value]) => [key, `${JSON.stringify(value)}`]));
 
 function trackEvent(event) {
-  tracker.trackEvent(event);
+  const name = event?.label ?? {};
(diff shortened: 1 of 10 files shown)


lchoy (approved):

NickAkhmetov (dismissed):
One question, looks good otherwise

context/app/static/js/helpers/trackers.js
```diff
@@ -85,7 +85,8 @@ const makeFaroSafe = (obj) =>
   Object.fromEntries(Object.entries(obj).map(([key, value]) => [key, `${JSON.stringify(value)}`]));
 
 function trackEvent(event) {
-  tracker.trackEvent(event);
+  const name = event?.label ?? {};
```
NickAkhmetov, line 88: Should the fallback here be `''` instead of `{}`? Otherwise I think it would show up in matomo as `[Object Object]`

NickAkhmetov (approved):

PR #3379, opened 2024-02-01
John conroy/cleanup tracking

john-conroy:
- Extends `trackEvent` to prepend an ID to the event label when provided. 
- Fixes existing events to send event values as event labels when the value is not a number. 
  - Since Faro handles whatever keys/values you send there should be little impact there.
  - I was wrong about there being character limits for event properties in Matomo. There are only limitations on custom dimensions. We can send stringified objects as event labels.
- Removes web vitals tracking from Matomo to stop polluting events. These metrics will continue to be tracked with Faro albeit with only a two week lookback.


context/app/static/js/components/detailPage/visualization/VisualizationTracker/utils.ts
```diff
@@ -109,7 +109,7 @@ export function fireVitessceEvent(ref: MutableRefObject<VitessceInteraction>) {
     trackEvent({
       category: 'Visualization',
       action: 'Interaction',
-      value: stringifyVitessceInteraction(ref.current),
+      label: stringifyVitessceInteraction(ref.current),
```
NickAkhmetov, line 112: I'm not sure this will work well for these events since they spit out long digests of activity (i.e. this wouldn't be categorizable by label)

john-conroy: Do you have a suggestion for how we can track this? If it's not a number `value` won't work either.

john-conroy: Should we file a ticket to revisit this?

lchoy: Filed https://hms-dbmi.atlassian.net/browse/CAT-546

context/app/static/js/components/workspaces/api.ts
```diff
@@ -319,10 +325,8 @@ async function createWorkspaceFetcher(_key: string, { arg: { body, url, headers
   trackEvent({
     category: 'Workspace Creation',
```
lchoy, line 326: Should this be 'Workspace Action'?

context/app/static/js/components/workspaces/api.ts
```diff
@@ -379,20 +383,25 @@ export interface UpdateWorkspaceBody {
...
-      globus_groups_token: body?.workspace_details?.globus_groups_token,
-      files: body?.workspace_details?.files?.map((f) => f.name),
-      symlinks: body?.workspace_details?.symlinks?.map((s) => s.name),
+async function updateWorkspaceFetcher(
+  _key: string,
+  { arg: { body, url, headers, workspaceId } }: { arg: UpdateWorkspaceArgs },
+) {
+  trackEvent(
+    {
+      category: 'Workspace Update',
```
lchoy, line 395: Should this be 'Workspace Action' ?

context/app/static/js/components/workspaces/api.ts
john-conroy: I didn't really take a look at the strings for the existing events. If anything this should be 'Workspaces' and let the `action` differentiate the events.

lchoy: From a grouping perspective, I'm advocating for a consistent category value.  'Workspaces" sounds good to me. I'm seeing three categories currently, 'Workspace Action', 'Workspace Creation', 'Workspace Update'. It may be valuable to have all the Workspace events categorized together, still with `action` differentiating the events.

image

john-conroy: Sorry, edited my comment likely while you were responding. I'm in full agreement and can pull those changes into this PR.

lchoy (approved):

PR #3381, opened 2024-02-05
John conroy/fix launch ws dataset page CAT-469

john-conroy:
Fixes workspace creation from dataset detail pages.


NickAkhmetov (approved):
This should work! 

I was contemplating the `NotebookButton` for a little while to see if there are any cleaner options/ways to avoid having the `tooltip` be dependent on the disabled status, but settled on concurring with the current implementation since it's a local component as opposed to something sitewide.

PR #3382, opened 2024-02-06
John conroy/fix entity page tracker CAT-553

john-conroy:
Fixes `useTrackEntityPageEvent` to handle undefined `entity_type` and `hubmap_id` inside the flask data context. We could consider defining a new field in the flask data for detail pages which would be the `entity_type` for search-api dependent detail pages and manually set for other pages like workspaces.

I also went on a little bit of a tangent to pull the tracker out of `SummaryData`.


NickAkhmetov (approved):
👍🏻 Looks good!

PR #3387, opened 2024-02-20
[CAT-159] NickAkhmetov / VisLifting revision, Visium support

NickAkhmetov:
This PR:

- Revises vis lifting approach to only request the actual parent assay type instead of iterating through all ancestors
- Added support for visium dataset visualizations
- Bumped portal-visualization to 0.2.0
- Bumped vitessce to 3.3.6


context/app/api/client.py
```diff
@@ -268,7 +272,7 @@ def get_descendant_to_lift(self, uuid):
                     "must": [
                         {
                             "term": {
-                                "vitessce-hints": "is_support"
+                                "vitessce-hints": "is_image"
```
john-conroy, line 275: Sorry I know this is still a draft. This will fix the broken dataset visualizations, but will break the publication vignettes. I imagine you'll add the hint as a param or similar?

NickAkhmetov: Yes, good callout - as discussed on Slack, this was the approach I took. A new boolean arg has been added and is now used to distinguish calls for publications and image pyramids.

context/app/api/client.py
```diff
@@ -312,7 +320,9 @@ def get_publication_ancillary_json(self, entity):
         '''
         publication_json = {}
         publication_ancillary_uuid = None
-        publication_ancillary_descendant = self.get_descendant_to_lift(entity["uuid"])
+        publication_ancillary_descendant = self.get_descendant_to_lift(
+            entity["uuid"], is_publication=True)
+        print(publication_ancillary_descendant)
```
lchoy, line 325: Is the print() on line 325 meant for debugging?

lchoy (changes requested):

john-conroy (dismissed):
Great, thanks!

john-conroy (approved):

PR #3389, opened 2024-02-26
John conroy/globus all statuses CAT-579

john-conroy:
Show Globus link for datasets regardless of their status.

@tsliaw Until now all consortium access datasets showed the now removed 'These data are still being prepared, processed, or curated...' message since they all have a non-published status. This PR makes it so the public message shows for those datasets. Is that ok or do you want a different message?

### DIFF
--- a/context/app/static/js/components/detailPage/BulkDataTransfer/usePanelSet.jsx
+++ b/context/app/static/js/components/detailPage/BulkDataTransfer/usePanelSet.jsx
@@ -204,7 +204,7 @@ export const usePanelSet = () => {
   const hasNoAccess = globusURLStatus === 403;
   const isNonConsortium = !isHubmapUser;
 
-  if (accessType === 'Public') {
+  if (accessType !== 'Protected') {
     return getGlobusPanel({ status: globusURLStatus, panel: PUBLIC_DATA, isLoading: globusURLIsLoading });
   }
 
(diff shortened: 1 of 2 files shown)


NickAkhmetov (approved):
👍🏻 LGTM!

context/app/static/js/components/detailPage/BulkDataTransfer/usePanelSet.jsx
```diff
@@ -208,17 +196,15 @@ export const usePanelSet = () => {
...
 
   const hasDbGaPStudyURL = Boolean(dbgap_study_url);
   const { status: globusURLStatus, isLoading: globusURLIsLoading } = useFetchProtectedFile(uuid);
   const hasNoAccess = globusURLStatus === 403;
   const isNonConsortium = !isHubmapUser;
-  const unfinalizedStatuses = ['New', 'Error', 'QA', 'Processing', 'Invalid'];
-  const isNotFinalized = unfinalizedStatuses.includes(status);
 
-  if (accessType === 'Public') {
+  if (accessType !== 'Protected') {
```
lchoy, line 207: The access types are [protected, consortium, or public]. Has the user already been authenticated at this point? Is it ok to show this if the accessType is consortium and the user is not authenticated?

john-conroy: The user has to have been authenticated to reach this point. Unlike protected datasets, which you can visit but not access files, I believe you can only visit consortium datasets if you have access. To be safe we could rely on the status of the Globus call as we do for protected access and just adapt the message.

NickAkhmetov: Correct me if I'm wrong - my understanding is that if a user is viewing a consortium dataset, they are already logged in

lchoy (approved):

tsliaw:
@john-conroy Let's keep the public Globus message for these datasets.

PR #3390, opened 2024-03-04
NickAkhmetov/CAT-546 Revise Vitessce tracking approach

NickAkhmetov:
This PR:
- Indicates where a user interacted with Vitessce (dataset vs. publication vs. preview) as part of the event category 
  - `Visualization - Dataset`, `Visualization - Publication`, `Visualization - Preview`, `Visualization - Genes`, etc. `Visualization - Unknown` is used for any cases not matched by other cases in `useVitessceEventMetadata`.
- Indicates the entity ID or preview name as the `name` of the event
- Indicates event actions in `Action Component` format, e.g. `Left Click Download obsLocations`
  - A separator or wrapper could be included here if desired @tsliaw 
- Breaks up the event digest format to fire individual events for each interaction
  - This increases the amount of data in total but should be much easier to work with during analysis
- Indicates search filter terms by firing a typing event 5 seconds afterwards


context/app/static/js/components/detailPage/visualization/VisualizationTracker/hooks.ts
```diff
@@ -1,90 +1,102 @@
...
+  const flaskData = useFlaskDataContext();
 
-export const useVisualizationTracker = () => {
-  // Track when the visualization is first mounted
-  useTrackMount('Visualization', 'Vitessce Mounted');
-  // Create an object to track Vitessce interactions
-  // This list is populated with a summary of every interaction a user has with the visualization
-  const vitessceInteraction = useRef<VitessceInteraction>([]);
+  let category = 'Unknown';
+  let name = location; // Fall back to the full URL if we can't find a better name
```
john-conroy, line 13: Can we use case/switch or similar to avoid the `let`?

context/app/static/js/components/detailPage/visualization/VisualizationTracker/hooks.ts
```diff
@@ -1,90 +1,102 @@
...
+
+  // Handle preview pages
+  if (location.includes('preview')) {
+    const { title } = flaskData;
+    category = 'Preview';
+    name = title;
```
john-conroy, line 19: I believe we should pass this as `label`. `name` gets set to the value of `label` so it works with both GA/Matomo.

context/app/static/js/components/detailPage/visualization/VisualizationTracker/hooks.ts
NickAkhmetov: I ran into this eslint rule with the switch approach: https://eslint.org/docs/latest/rules/no-case-declarations

If I remember correctly, we'd need to return the category/name object separately in each switch statement without separately instantiating the variables with `const`.

context/app/static/js/components/detailPage/visualization/VisualizationTracker/hooks.ts
NickAkhmetov: Will do!

context/app/static/js/components/detailPage/visualization/VisualizationTracker/hooks.ts
john-conroy: Ah gotcha, this is actually ringing a bell and I think I've dealt with it before. 

Would it make sense to move to a function which returns an object with category and name to be destructured? Sorry if this is just tedious, feel free to ignore!

context/app/static/js/components/detailPage/visualization/VisualizationTracker/hooks.ts
john-conroy: It happens [here](https://github.com/hubmapconsortium/portal-ui/blob/292509d5af96772cef7016e94c5c58d1bc5a060a/context/app/static/js/helpers/trackers.js#L105-L123).

context/app/static/js/components/detailPage/visualization/VisualizationTracker/hooks.ts
NickAkhmetov: Good call on the helper function, that was a very good way to clean up the logic a bit (and make that individual piece testable).

context/app/static/js/components/detailPage/visualization/VisualizationTracker/hooks.ts
NickAkhmetov: Thanks for pointing this out! I switched `name` to `label` where appropriate.

PR #3393, opened 2024-03-06
John conroy/fix dev search hierarchical

john-conroy:
Fix hierarchical facets on dev search.


NickAkhmetov (approved):
Makes sense to me 👍🏻

PR #3394, opened 2024-03-07
NickAkhmetov/CAT-591 Fix tile search view

NickAkhmetov:
This PR adds the missing field that was causing the tile view of search pages to crash if viewing any query but the default.

image


john-conroy (commented):
We should add it for the sample tile config too.

lchoy (approved):

PR #3391, opened 2024-03-06
NickAkhmetov/CAT-595 Contributors v2 Support

NickAkhmetov:
This PR:
- Adds a function to normalize legacy and CEDAR schema contributors and tests for said function
- Lays a foundation to add support for future schema revisions
- Converts some components to TypeScript

Visium dataset contributors (a6e6ba302926dff92c599199ea1a7d6c on `dev`):
image

PAS dataset contributors (c9d21ebf6e8a3b9f6e40fe45d026b304 on `dev`):
image

PR #3395, opened 2024-03-07
NickAkhmetov/CAT-593 Support display of legacy contact information without `version` field

NickAkhmetov:
The `contacts` in data collections do not have a `version` field, while `contributors` do. Since the adapter I had built out for this as part of #3391 assumed that all legacy contacts have a `version` field while all CEDAR contacts have the `metadata_schema_id` field, I had used the presence of the `version` field as the discriminator to determine which adapter function to use. As this assumption has been proven incorrect, I have reversed this logic - if the metadata schema ID field is present, we know it's a CEDAR schema contributor, while if the metadata schema ID is missing, we know it's legacy data.

image

### DIFF
--- a/CHANGELOG-cat-593.md
+++ b/CHANGELOG-cat-593.md
@@ -1,0 +1,1 @@
+- Support display of legacy contact information which does not have `version` field available.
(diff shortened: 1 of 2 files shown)


john-conroy (approved):
Thanks!

PR #3401, opened 2024-03-13
John conroy/fix publication test

john-conroy:
The test broke as the data underlying the publication changed. This should be less fragile. We should look into why the data for the publication changed.


NickAkhmetov (approved):

NickAkhmetov:
For context, I remember having to switch these values a few weeks ago when the tests started failing during another reindex, so I think this could be related to the entities that failed to index altogether. Realistically, these more permissive tests should be less flaky.

PR #3404, opened 2024-03-22
Explicitly set container image architecture.

lchoy:
Explicitly sets the container image architecture to amd64.


NickAkhmetov (approved):
LGTM!

PR #3405, opened 2024-03-22
Set platform environment variable.

lchoy:
Docker will automatically select an image variant which matches your OS and architecture. To instruct docker to pass the --platform flag to build our Docker image for an x86/amd64 platform, we set the environment variable `DOCKER_DEFAULT_PLATFORM` in the build script.


NickAkhmetov (approved):
This definitely seems more idiomatic 👍

PR #3412, opened 2024-04-09
NickAkhmetov/convert `useSearchData` to TypeScript

NickAkhmetov:
I tried to keep the git history from `useSearchData.js` but it still saw the changes as two separate files.

Remaining todos

- [x] Pull `query` and `queries` out of the config arg so that they can be provided separately and the interface is more wieldy
- [x] Update hook usage throughout the application to match


context/app/static/js/hooks/useSearchData.ts
```diff
@@ -0,0 +1,339 @@
...
+  query: SearchRequest;
+  useDefaultQuery: boolean;
+}
+function useRequestInit({ query, useDefaultQuery }: RequestInitArgs) {
+  const authHeader = useAuthHeader();
+  return useMemo(() => {
+    const body = createSearchRequestBody({ query, useDefaultQuery });
+    return buildSearchRequestInit({ body, authHeader });
+  }, [query, useDefaultQuery, authHeader]);
+}
```
john-conroy, line 76: One of the situations where we might not need to memoize since swr performs deep comparisons?

context/app/static/js/hooks/useSortState.ts
```diff
@@ -77,6 +77,7 @@ export const useSortState = (
       : sortState.columnId;
 
   // The doc _id is used as a tiebreaker which is necessary for search_after queries.
+  // TODO: Update types to improve s
```
john-conroy, line 80: Seems to be incomplete?

context/app/static/js/hooks/useSearchData.ts
```diff
@@ -0,0 +1,339 @@
...
+  isLoading: boolean;
+}
+
+interface UseHitsData<Documents> {
+  searchHits: NonNullable<Required<SearchHit<Documents>>[]>;
+  isLoading: boolean;
+}
+
+const defaultConfig = {
+  useDefaultQuery: false,
```
john-conroy, line 94: Looking through the git history, I'm not sure why `useDefaultQuery` was set to false. Perhaps a misnomer, but `useDefaultQuery` extends the es query passed as an arg to avoid docs with `next_revision_uuid` (previous dataset versions) and `sub_status`. We should have this set to true by default. 

Do you remember why you changed it to false here?

context/app/static/js/hooks/useSearchData.ts
```diff
@@ -0,0 +1,339 @@
...
+        const body = createSearchRequestBody({ query, useDefaultQuery });
+        return buildSearchRequestInit({ body, authHeader });
+      }),
+    [queries, useDefaultQuery, authHeader],
+  );
+  return inits;
+}
+
+const defaultMultiFetchConfig = {
+  useDefaultQuery: false,
```
john-conroy, line 151: Should be `true` as above.

context/app/static/js/hooks/useSearchData.ts
```diff
@@ -0,0 +1,339 @@
...
+  return { searchData, isLoading };
+}
+
+// Infinite search
+
+function getTotalHitsCount(results?: SearchResponseBody<unknown, unknown>) {
+  const total = results?.hits?.total;
+  if (typeof total === 'number') {
+    return total;
+  }
```
john-conroy, line 182: I imagine this is based on es types where total is a number or some interface?

context/app/static/js/hooks/useSearchData.ts
```diff
@@ -0,0 +1,339 @@
...
+    fetcher,
+    swrConfig,
+  );
+
+  return { searchData, isLoading };
+}
+
+// Infinite search
+
+function getTotalHitsCount(results?: SearchResponseBody<unknown, unknown>) {
```
john-conroy, line 178: Are we using <unknown, unknown> here and below because the generic doesn't affect the function's return type?

john-conroy (changes requested):
The primary concern is `useDefaultQuery` set to `false`, but that comes from a much earlier PR. 

We also discussed switching to destructured args for the hooks, but understand that will require updates to all of their instances. Should that be done here or later on?

context/app/static/js/hooks/useSearchData.ts
NickAkhmetov: Good call! I will remove the extra useMemo here.

context/app/static/js/hooks/useSortState.ts
NickAkhmetov: Yup, I think I got distracted by the portal-containers work halfway through writing this comment 😅

context/app/static/js/hooks/useSearchData.ts
NickAkhmetov: I'm not sure why it was set to `false`, I'll update this accordingly 👍🏻

context/app/static/js/hooks/useSearchData.ts
NickAkhmetov: Yes - in the ES types, the `total` can be either a raw number or an object with a `.value` key that holds the number.

NickAkhmetov:
> We also discussed switching to destructured args for the hooks, but understand that will require updates to all of their instances. Should that be done here or later on?

I went through the repository and checked all the references to `useSearchData` and `useSearchHits` - I believe this piece of it should be all set.

john-conroy (approved):
Thanks for the changes!

PR #3416, opened 2024-04-18
John conroy/migrate husky

john-conroy:
Migrates husky from v4 -> v9 according to https://typicode.github.io/husky/migrate-from-v4.html. Confirmed to still pick up linting and type errors in the pre-commit hooks.


NickAkhmetov (approved):
🙌 thanks for the fix!

PR #3417, opened 2024-04-22
Update CCFOrganInfo.tsx to improve performance

bherr2:
Hi Nick,

I updated CCFOrganInfo to use the parameters provided previously for the IFrame version here: https://github.com/hubmapconsortium/portal-ui/blob/main/context/app/static/js/pages/Iframe/Iframe.jsx#L20-L24 . This is a much faster implementation. Please test locally before merging though as I cannot test the code directly.

Thanks!


NickAkhmetov (approved):
Confirmed it works for me locally - thanks for the fix, @bherr2!

PR #3420, opened 2024-05-06
John conroy/ws r

john-conroy:


NickAkhmetov (approved):
Looks good to me!

john-conroy:
@NickAkhmetov I added a small change to update the workspace `job_type` after starting per Juan's suggestion. For now the dialog remains open while updating after the workspace is launched in a new tab. I'll revisit that behavior in a later PR after discussing with UI/UX team.

PR #3424, opened 2024-05-09
John conroy/fix ws create errors

john-conroy:
Fix unhandled errors if any of the parallel requests to the templates API fail. Filed a ticket to make toasts a stack.


NickAkhmetov (approved):
🙏

PR #3426, opened 2024-05-09
John conroy/update ws loading message CAT-225

john-conroy:


context/app/static/js/components/Routes/Routes.jsx
```diff
@@ -262,7 +262,7 @@ function Routes({ flaskData }) {
     const workspaceId = urlPath.split('/').pop();
     return (
       <Route>
-        <WorkspacePleaseWait workspaceId={workspaceId} />
+        <WorkspacePleaseWait workspaceId={parseInt(workspaceId, 10)} />
```
NickAkhmetov, line 265: Should we do the `parseInt` call on line 262 instead? I don't think it particularly matters but it would keep this a bit cleaner.

NickAkhmetov (approved):

NickAkhmetov (approved):
🚀

PR #3427, opened 2024-05-09
John conroy/ws access CAT-257

john-conroy:


context/app/static/js/components/Header/UserLinks/UserLinks.jsx
```diff
@@ -10,13 +11,14 @@ import { TruncatedSpan, WarningDropdownLink } from './style';
...
 
+  const userType = getUserType();
   return (
     <Dropdown
       title={<TruncatedSpan>{isAuthenticated ? userEmail || 'User' : 'User Profile'}</TruncatedSpan>}
       data-testid="user-profile-dropdown"
     >
       <DropdownLink href="/my-lists">My Lists</DropdownLink>
-      {isWorkspacesUser && <DropdownLink href="/workspaces">My Workspaces</DropdownLink>}
+      {(isWorkspacesUser || userType === 'internal') && <DropdownLink href="/workspaces">My Workspaces</DropdownLink>}
```
NickAkhmetov, line 21: Perhaps adjusting the "isWorkspacesUser" logic in the app context instead could help avoid unexpected regressions in other places?

john-conroy: We are only changing the behavior for the dropdown to be available to workspace users or users who have authenticated in the past (also covers currently authenticated, non-workspace users). The auth guard will behave the same as it has. Maybe I'm misinterpreting your comment.

john-conroy: Added a var to make this more clear per your suggestion.

NickAkhmetov (dismissed):
Thanks for clearing up the confusion 🙏

NickAkhmetov (approved):

PR #3436, opened 2024-05-15
NickAkhmetov/CAT-298 Implement profile page

NickAkhmetov:
This PR implements the user profile page.

<details><summary>Logged in HuBMAP user with access to workspaces</summary>
<p>

image

</p>
</details> 

<details><summary>Logged in HuBMAP user with no access to workspaces</summary>
<p>

image

</p>
</details> 

<details><summary>Logged in external user</summary>
<p>

image

</p>
</details>

<details><summary>Unauthenticated user</summary>
<p>

image

</p>
</details> 

<details><summary>Minor Details</summary>
<p>

I made the buttons more "call to action"-y if there are no existing lists or workspaces. 

If no lists are available:
image

If no workspaces have been created but user has access to workspaces:
image

</p>
</details>

### DIFF
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -181,3 +181,15 @@ def tutorial_detail(tutorial_name):
         title=tutorial_name,
         flask_data=flask_data
     )
+
+
+@blueprint.route('/profile')
+def profile():
+    # We could fetch the globus groups info here, but
+    # the auth helper doesn't currently provide group descriptions
+    flask_data = {**get_default_flask_data()}
+    return render_template(
+        'base-pages/react-content.html',
+        title='Profile',
+        flask_data=flask_data
+    )
--- a/context/app/routes_main.py
+++ b/context/app/routes_main.py
@@ -185,8 +185,6 @@ def tutorial_detail(tutorial_name):
 
 @blueprint.route('/profile')
 def profile():
-    # We could fetch the globus groups info here, but
-    # the auth helper doesn't currently provide group descriptions
     flask_data = {**get_default_flask_data()}
     return render_template(
         'base-pages/react-content.html',
--- a/context/app/static/js/components/profile/Summary/Description.tsx
+++ b/context/app/static/js/components/profile/Summary/Description.tsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import { Alert } from 'js/shared-styles/alerts';
+import ContactUsLink from 'js/shared-styles/Links/ContactUsLink';
+import { useAppContext } from 'js/components/Contexts';
+
+export function ProfileDescription() {
+  const { userEmail } = useAppContext();
+  // ensure that the user only sees this message once per account per device
+  const localStorageKey = `${userEmail}-has-seen-profile-description`;
+
+  const hasSeenProfileDescriptionCookie = Boolean(localStorage.getItem(localStorageKey));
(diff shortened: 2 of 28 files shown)


context/app/routes_main.py
```diff
@@ -181,3 +181,15 @@ def tutorial_detail(tutorial_name):
         title=tutorial_name,
         flask_data=flask_data
     )
+
+
+@blueprint.route('/profile')
+def profile():
+    # We could fetch the globus groups info here, but
+    # the auth helper doesn't currently provide group descriptions
```
NickAkhmetov, line 189: On further thought, maybe requesting this data once and keeping it in a `@cache` could work?

context/app/static/js/shared-styles/alerts/LoginAlert.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import Button from '@mui/material/Button';
+import { Alert } from './Alert';
+
+interface LoginAlertProps {
+  featureName: string;
+}
+export default function LoginAlert({ featureName }: LoginAlertProps) {
```
NickAkhmetov, line 9: Tiffany indicated that I should reuse the workspaces login alert - I figured this would be an easy way to make it reusable in other contexts.

context/app/routes_main.py
NickAkhmetov: I added an API route to fetch the globus groups and cache them once; the front-end code still uses swr to fetch this data in case we ever end up needing it anywhere else.

context/app/static/js/components/profile/Summary/Description.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  const hasSeenProfileDescriptionCookie = Boolean(localStorage.getItem(localStorageKey));
+  const [hasSeenProfileDescription, setHasSeenProfileDescription] = React.useState(hasSeenProfileDescriptionCookie);
+  if (hasSeenProfileDescription) {
+    return null;
+  }
+  const onClose = () => {
+    localStorage.setItem(localStorageKey, 'true');
+    setHasSeenProfileDescription(true);
+  };
```
john-conroy, line 19: We could pull this into a reusable hook as it seems to be popping up throughout the designs.

context/app/static/js/components/profile/Summary/hooks.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  return {
+    data,
+    error,
+    isLoading: !data && !error,
+  };
+}
+
+export function useCurrentUserGlobusGroups() {
+  const groupsCookie = readCookie('user_groups');
```
john-conroy, line 37: We could consider adding `user_groups` passed from flask to the context.

john-conroy (approved):
Looks good to me, thanks!

context/app/static/js/components/profile/Summary/Description.tsx
NickAkhmetov: Good idea - I've created https://hms-dbmi.atlassian.net/browse/CAT-663 to track this.

context/app/static/js/components/profile/Summary/hooks.ts
NickAkhmetov: We could; I figured we should have this be consistent with the existing approach for tracking where we use the cookie. Are there any drawbacks you can think of for this approach?

PR #3439, opened 2024-05-20
NickAkhmetov/CAT-641 Fix unexpected visualizations on unprocessed datasets

NickAkhmetov:
This PR fixes visualizations appearing on primary datasets with image descendants. Instead of only checking for `is_support` or `is_image`, image pyramid cases need to check for the presence of both hints (which is also why we can't use a `terms`	condition here, since that returns if any of the matched terms are met).

AF Dataset:
image

CODEX (non-processed) dataset:
image

CODEX (processed) dataset:
image

### DIFF
--- a/CHANGELOG-cat-641.md
+++ b/CHANGELOG-cat-641.md
@@ -1,0 +1,1 @@
+- Fix unexpected visualizations on unprocessed datasets.
--- a/context/app/api/client.py
+++ b/context/app/api/client.py
@@ -269,16 +269,24 @@ class ApiClient():
         returns the doc of the most recent descendant
         that is in QA or Published status.
         '''
-        hints = "is_support" if is_publication else "is_image"
+
+        hints = [{
+            "term": {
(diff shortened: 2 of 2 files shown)


john-conroy (approved):
Alternatively you could use a `terms` query, but it shouldn't block.

NickAkhmetov:
> Alternatively you could use a terms query, but it shouldn't block.

@john-conroy I tried that approach at first; the derived CODEX dataset was still getting picked up since `terms` queries work with `or` logic, i.e. if any of the terms in the hints array matched, the descendant was getting picked up.

Per the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-query.html):
> Returns documents that contain one or more exact terms in a provided field.

There is a [`terms_set` query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-set-query.html), but that would need to be pulled out of the `bool.must` nesting.

PR #3441, opened 2024-05-26
John conroy/replace searchkit

john-conroy:
Feature branch for the effort to replace searchkit.


NickAkhmetov (commented):
Great work getting through this gargantuan effort 🙌🏻

I went through the search pages on local. My observations are noted below; I've also gone through the code and made suggestions where things stood out to me. I anticipate making a few more passes through the code as we iterate.

# Datasets

- The whole table header cell shows a `pointer` cursor, but only clicking the arrow applies the sort. The unselected arrow is also very low contrast relative to the background: https://webaim.org/resources/contrastchecker/?fcolor=7979791f&bcolor=FFFFFF

https://github.com/user-attachments/assets/ce7bf8b3-4787-44bf-980c-106325b161eb

- It might be helpful to show a loading indicator in the `show more` button after it's clicked in case the user has a slow connection.
- The page seems to hang for a second or two when filters are changed:

https://github.com/user-attachments/assets/cdd07472-7c2e-4442-8778-5935a10c240e

- Expanding/collapsing a search facet without changing any filters appears to change the displayed search results - this may be related to the issue above.

https://github.com/user-attachments/assets/4c7b477b-b000-4dc7-b6c8-e4eee5ab92ac

  - Somewhat funny example of the same issue, the order of two results changes when I expand/collapse the `group`: 

https://github.com/user-attachments/assets/2b95dcec-bfa6-4bd5-9b68-9950546a610d

# Donors



- Donor page results' column sizing is a bit off - the same amount of space is allotted to each column, despite `Group` being a much more text-heavy field than `Donor BMI` or `Donor Age`. The donor age tooltips from #3548 are also missing.
image

- Related to the above, Donor Age slider should be capped at 90; the visualization also appears somewhat misaligned, since the last bin in the histogram lines up with `80`. 
image
- Also, should we consider naming the columns "Age" and "BMI" to be a bit less redundant?


# Publications

The chronological sort order applied in #3565 is no longer present/the publications list is no longer sorted:
image

# Organ pages
- Search page links from assay graphs need to be updated to use the new `buildSearchLink` util:
image

# Homepage
Same as the organ page, the graph's links need to be updated to use the new util.
image

The collections count fix from #3564 should be restored:
image

context/package.json
NickAkhmetov, line 1: Should we also remove `searchkit`? (line 59)

john-conroy: I haven't removed `searchkit` or cleaned up the old components yet.

john-conroy:
The publications/collections issues are just because I'm behind main here. Should be fixed with a merge.

context/app/static/js/components/search/es-mapping.json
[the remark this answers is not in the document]
john-conroy: I'll move this into a hook to request once.

NickAkhmetov:
Looking good! The only thing that's missing from prod is this little loading indicator in the searchbar: 

https://github.com/user-attachments/assets/ff62add1-ba3f-40d2-ad15-f72c254265df

Another potential small improvement - maybe we can set the starting BMI at 10 instead of 0? The lowest-BMI result in prod appears to have a BMI of 15.
image

NickAkhmetov:
Some minor issues I noticed which we can file for later resolution:

## Dismissing the "Ancestor ID" filter chip leaves a Skeleton behind

* Go to a dataset page
* Click "view derived datasets"
* Clear the filter by removing the "ancestor ID" chip

After the above steps, the skeleton in that section persists:

https://github.com/user-attachments/assets/87b6abd1-1080-4271-a1e3-81725dcce27b

The issue does not occur if using the `clear filters` button.

## Date Picker colors for selectable/unselectable dates are difficult to discern

I found it a bit difficult to tell at a glance which values in the date picker were and weren't disabled:

https://github.com/user-attachments/assets/769fd7bf-6586-4da0-8080-d775a791541f

The selectable dates are #444a67, while the disabled dates render as #666666 on a white background (i.e. black with 0.6 opacity). 
image

Could we try adjusting the opacity down to `0.38` (which I believe is the `disabled` text color in the theme) would help? The current style is the `secondary` text color. CC @tsliaw


## Date Picker "chin"

Is it possible for us to hide this empty whitespace under the months? Since there's a known max size of 12 months and it shouldn't expand any further, it might look a bit neater without it.

image

## Date picker info/error messages pop up after other filters are applied

I noticed this error text appeared after I first set some start/end dates then adjusted BMI/age ranges:

image

This seems to pop up if the max value is after the last dataset that otherwise matches the search query, so while this doesn't hide any data it could potentially be a bit confusing. Maybe we don't need the extra validation on the end range?

PR #3442, opened 2024-05-29
NickAkhmetov/Homepage Graphs

NickAkhmetov:
image

This PR:
- Improves the HorizontalStackedBarChart component
- Implements the updated homepage stacked bar chart component
- Touches up the organ page since it used the same base component and the revisions I made to said base component propagated elsewhere
- Resolved issues from prior development on homepage updates which I had deferred or missed:
  - Fixed broken infinite scrolling
  - Fixed publication panel overflow issues after introduction of icons
  - Fixed test failures due to revisions to organ endpoint and panel props helper

Remaining TODOs:
- [x] Make sure storybook is OK
- [x] Double check against Figma

Future improvements:
- Unify horizontal/vertical stacked bar charts
- Deduplicate ordinal legend component and related logic for when to show it


context/app/static/js/components/home/ExploreTools/ExploreTools.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { Typography } from '@mui/material';
+
```
john-conroy, line 4: Why did you choose to include this in the current PR?

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.tsx
```diff
@@ -0,0 +1,202 @@
...
+import {
+  AggregatedDatum,
+  getKeysFromAggregatedData,
+  useAggregatedChartData,
+  useDatasetTypeMap,
+  useOrganOrder,
+  useSearchDataRange,
+} from './hooks';
+
+const margin = { top: 40, right: 50, bottom: 64, left: 100 };
```
john-conroy, line 37: Perhaps comment why these values were chosen?

context/app/static/js/components/home/HuBMAPDatasetsChart/hooks.ts
```diff
@@ -0,0 +1,96 @@
...
+  type QueryAggs,
+} from './queries';
+
+export type AggregatedDatum = {
+  organ: string | number; // Actually just a string, but the union with Record below is necessary for TS to infer the type correctly
+} & Record<string, number>;
+
+export type AggregatedData = AggregatedDatum[];
+
+// Categorizes the data by organ, showing a count for each paired key
```
john-conroy, line 17: We could benefit from some tests for these utils?

context/app/static/js/pages/Organ/Organ.tsx
```diff
@@ -12,15 +12,20 @@ import Samples from 'js/components/organ/Samples';
...
-  const samplesId = 'Samples';
+interface OrganProps {
+  organ: OrganFile;
+}
+
+const summaryId = 'Summary';
+const hraId = 'Human Reference Atlas';
+const referenceId = 'Reference-Based Analysis';
+const assaysId = 'Assays';
+const samplesId = 'Samples';
```
john-conroy, line 26: Would this make more sense as an obj?

context/app/static/js/components/home/HuBMAPDatasetsChart/queries.ts
```diff
@@ -0,0 +1,168 @@
...
+      doc_count: number;
+      key: string;
+    }[];
+  };
+}
+
+// Query for looking up the mapping of dataset types to assay display names
+const datasetTypeMapQuery: SearchRequest = {
+  size: 0,
+  query: includeOnlyDatasetsClause,
```
john-conroy, line 27:
```
  size: 0,
  query: includeOnlyDatasetsClause,
```
seems to be used in most of these search request queries. Could it be pulled out?

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.tsx
```diff
@@ -0,0 +1,202 @@
...
+  const colors = useChartPalette();
+  const [selectedColorDataIndex, setSelectedColorDataIndex] = useSelectedDropdownIndex(0);
+
+  const { organOrder } = useOrganOrder();
+
+  const assayBuckets = useAggregatedChartData<AssaysQueryKey>(assayTypeQuery);
+  const donorSexBuckets = useAggregatedChartData<DonorSexQueryKey>(donorSexQuery);
+  const donorRaceBuckets = useAggregatedChartData<DonorRaceQueryKey>(donorRaceQuery);
+  const analyteClassBuckets = useAggregatedChartData<AnalyteClassQueryKey>(analyteClassQuery);
+  const processingStatusBuckets = useAggregatedChartData<ProcessingStatusQueryKey>(processingStatusQuery);
```
john-conroy, line 74: Are we fetching the buckets for each view upfront?

john-conroy (commented):
Thanks! A few questions.

context/app/static/js/components/home/ExploreTools/ExploreTools.tsx
NickAkhmetov: I stubbed these in to get the CSS for the homepage grid into place and make sure that the chart's positioning was all set; since there's a feature branch in between, I figured it'd be okay to keep these placeholders so there's a solid starting point when we start working on the tools section.

context/app/static/js/components/home/HuBMAPDatasetsChart/hooks.ts
NickAkhmetov: I'll add some 👍🏻

context/app/static/js/pages/Organ/Organ.tsx
NickAkhmetov: These were defined as instance variables inside the component previously, I pulled them out in the same form they were already in. They are referenced individually in the method body when providing the ID to each section.

context/app/static/js/components/home/HuBMAPDatasetsChart/queries.ts
NickAkhmetov: I can make that pair of attributes an object and spread it where needed, similar to the `overflowCss` in panel styles 👍🏻

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.tsx
NickAkhmetov: I settled on this approach so that all the different label keys can be provided to the chart wrapper. This allows us to prevent the legend from resizing when changing between bucket sets.

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.tsx
NickAkhmetov: Good call; most of these initial values were inherited from the original implementation. I've adjusted the values for all directions except the bottom to match Figma (32px) and added an explanation for why the bottom margin is greater than the others.

image

john-conroy (approved):

PR #3443, opened 2024-05-31
NickAkhmetov/Post-YAML cleanup

NickAkhmetov:
This PR:
- Removes `js-yaml` and `yaml-loader` dependencies which were only used for the IVT yaml files
- Removes remaining references to the IVT git submodule in the VS Code tasks config and dev-start script
- Reorganizes the dependencies page so that the empty list of git submodules is at the bottom of the page and includes explanatory text instead of being an empty heading.


context/app/markdown/dependencies.md
```diff
@@ -4,15 +4,10 @@ The portal has many dependencies:
 Git submodules and Python and NodeJS packages are summarized here;
```
john-conroy, line 4:
```suggestion
Python and NodeJS packages are summarized here;
```

context/app/markdown/dependencies.md
```diff
@@ -109,3 +104,7 @@ lxml>=4.9.1
 "zustand": "^4.5.2"
 }
 ```
+
+## Git submodules
+
+The portal UI does not currently depend on any git submodules.
```
john-conroy, line 110:
```suggestion
```

john-conroy (commented):
I don't think there's a need to reference submodules.

NickAkhmetov:
@john-conroy Makes sense - I was wavering re: whether to keep the section or not and kept it in just in case we had any git submodules in the future. 👍🏻

PR #3446, opened 2024-06-03
NickAkhmetov/CAT-698 Remove "other groups" from profile pages

NickAkhmetov:
This PR removes the `other groups` section of the profile page:

image


john-conroy (approved):
Nice, thanks!

PR #3454, opened 2024-06-21
NickAkhmetov/CAT-716 Display dbGaP link to logged in users if one is available and user does not have access to Globus for protected data

NickAkhmetov:
Updated image: 

image

<details><summary>Old image</summary>
<p>

image

</p>
</details>


john-conroy (commented):
Last I remember we were not supposed to show the dbGaP link for authenticated, consortium members without access and instead instruct them to request access through consortium means.

NickAkhmetov:
@john-conroy We discussed this during a call on Friday since we received a bug report about this behavior; we should discuss further before merging in to make sure we're on the same page.

context/app/static/js/shared-styles/icons/TooltipIcon.tsx
```diff
@@ -3,22 +3,28 @@ import { styled } from '@mui/material/styles';
 import { InfoIcon } from 'js/shared-styles/icons';
 import { SecondaryBackgroundTooltip } from '../tooltips';
 
-const StyledInfoIcon = styled(InfoIcon)(({ theme }) => ({
-  marginLeft: theme.spacing(0.5),
+interface StyledInfoIconProps {
```
john-conroy, line 6: Do we want to make this prop apply to `TooltipIcon` at large?

context/app/static/js/components/detailPage/BulkDataTransfer/usePanelSet.jsx
```diff
@@ -14,6 +15,8 @@ const dbGaPText = {
     'The database of Genotypes and Phenotypes archives and distributes data and results from studies that have investigated the interaction of genotype and phenotype in humans.',
 };
 
+const dbGaPTooltip = <InfoTooltipIcon iconTooltipText="Database of Genotypes and Phenotypes" noMargin />;
```
john-conroy, line 18: Should we just add the `dbGaP` text here to avoid repeating it below?

john-conroy (approved):

context/app/static/js/shared-styles/icons/TooltipIcon.tsx
NickAkhmetov: I'm not sure it's necessary in all cases; I'd rather have it be enabled by default as it was before and allow us to opt out in the cases where the margin is unnecessary.

context/app/static/js/components/detailPage/BulkDataTransfer/usePanelSet.jsx
NickAkhmetov: I'm not sure where the repetition you're referring to is occurring, to be honest - I believe you may be referring to the slightly different cases for `dbGapPanel` and `noGlobusAccessWhileLoggedInPanel`?

john-conroy: The `dbGaPTooltip` is generally preceded by the word `dbGaP`. Don't worry about it!

john-conroy (approved):

PR #3456, opened 2024-06-25
NickAkhmetov/CAT-616 Implement navigation updates

NickAkhmetov:
This PR:
- Implements new navigation
- Updates icons to use `createSvgIcon` MUI util for consistency in styling/props


context/app/static/js/components/Header/staticLinks/staticLinks.tsx
NickAkhmetov, line 1: Only remaining TODO is to verify that all of these links point to where they should go.

context/app/static/js/components/Header/HeaderContent/style.ts
```diff
@@ -0,0 +1,27 @@
+import { styled } from '@mui/material/styles';
+import Button from '@mui/material/Button';
+import Divider from '@mui/material/Divider';
+import SvgIcon from '@mui/material/SvgIcon';
+
+const Spacer = styled('div')({
```
john-conroy, line 6: I imagine these are in-place conversions, but we could replace the divs here with `Box`/`Stack`.

context/app/static/js/components/Header/staticLinks/staticLinks.tsx
```diff
@@ -1,116 +1,277 @@
...
-    },
-  ],
-};
+export const resourceLinks: DrawerSection[] = [
+  {
+    title: 'User Guides',
+    items: [
+      {
+        label: 'Tutorials',
+        description: 'Learn more about how to explore the data portal',
```
john-conroy, line 29: Do we want punctuation on these descriptions or is that handled elsewhere?

context/app/static/js/components/home/ExploreTools/ExploreTools.tsx
```diff
@@ -23,7 +23,7 @@ const workspaceCTAGuest = {
 const cards = [
   {
     title: 'Analyze data in Workspaces',
-    icon: <entityIconMap.Workspace color={theme.palette.info.main} fontSize="1.5rem" />,
```
john-conroy, line 26: The icon changes allowed for the earlier fix to be reverted?

context/app/static/js/components/home/Hero/const.tsx
```diff
@@ -11,7 +11,7 @@ export const HOME_TIMELINE_ITEMS: TimelineData[] = [
     titleHref: '/cells',
     description: 'Search datasets by cell type name or Cell Ontology ID.',
     date: 'May 2024',
-    img: <entityIconMap.CellType />,
+    img: <entityIconMap.CellType fontSize="1.5rem" />,
```
john-conroy, line 14: Pull out the fontSize for reuse?

context/app/static/js/shared-styles/Drawer/hooks.ts
```diff
@@ -0,0 +1,17 @@
+import { useMemo, useState } from 'react';
+
+export function useDrawerState() {
+  const [open, setOpen] = useState(false);
+
+  const toggle = useMemo(() => () => setOpen((prev) => !prev), []);
+  const onOpen = useMemo(() => () => setOpen(true), []);
+  const onClose = useMemo(() => () => setOpen(false), []);
```
john-conroy, line 8: Memo rather than callback?

john-conroy (dismissed):
Great! A few small comments, but nothing blocking.

context/app/static/js/components/Header/staticLinks/staticLinks.tsx
NickAkhmetov: The original mockups don't seem to have descriptions: 
image

context/app/static/js/components/home/ExploreTools/ExploreTools.tsx
NickAkhmetov: Yes - with the `createSvgIcon` util, theme values like `info` are properly recognized, so the actual value doesn't need to be passed in anymore. This should help avoid annoying related regressions in the future.

context/app/static/js/components/Header/HeaderContent/style.ts
NickAkhmetov: I removed some of the other unused styles as well 👍🏻

context/app/static/js/components/home/Hero/const.tsx
NickAkhmetov: I created a `timelineIconProps` object which we can pass any additional props into in the future 👍🏻

john-conroy (approved):
Looks good to me. I assume cypress is being flakey. We should make sure to re-run it and it passes before making a release.

lchoy:
> Looks good to me. I assume cypress is being flakey. We should make sure to re-run it and it passes before making a release.

The publication document did not successfully index onto Test environment. I reindexed and kicked off a rerun of the failed tests.

PR #3457, opened 2024-06-27
John conroy/fix entity tiles

john-conroy:
Updated the type for the entity passed to the variants of `checkEntityType` so this should be caught in the future.


NickAkhmetov (approved):
Good catch!

PR #3469, opened 2024-07-16
Fixed samples table header count

tkakar:


CHANGELOG-fix-support-headers.md
NickAkhmetov, line 1: These changelogs appear to be present in `main`/haven't been included in a release yet, so we should keep them around. In the previous PR, since the feature branch was started before a release, the changelog entries for that release had been reintroduced.

context/app/static/js/hooks/useSearchData.ts
```diff
@@ -191,9 +191,15 @@ export function useMultiSearchData<Documents, Aggs>(
   const requestInits = useRequestInits({ queries, useDefaultQuery });
   const { elasticsearchEndpoint } = useAppContext();
 
+  const inlineFetcher = () =>
+    fetcher({
+      urls: [elasticsearchEndpoint],
+      requestInits,
+    }) as Promise<SearchResponseBody<Documents, Aggs>[]>;
+
```
NickAkhmetov, line 199: I think if we provide the function directly in the `useSWR` hook call we can avoid needing to cast its return type, but I could be mistaken

NickAkhmetov (commented):
Great job with the fix! :raised_hands: If we restore the removed changelogs it should be good to go - my other comment is non-blocking.

NickAkhmetov (approved):

john-conroy:
@tkakar are you ready to merge this PR?

PR #3470, opened 2024-07-18
austenem/CAT-734 update workspace landing page

austenem:
This PR updates the workspaces landing page for users who are not logged in with updated information and a list of templates. 

TODO: Set a max height for the template grid to match intended designs and add href to “log in” text.

### DIFF
--- a/CHANGELOG-workspace-landing-page.md
+++ b/CHANGELOG-workspace-landing-page.md
@@ -1,0 +1,1 @@
+Update messaging on My Workspace landing page for users who are not logged in.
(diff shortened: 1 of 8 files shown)


john-conroy:
WorkspacesLandingPreview

We should try to move to https://mui.com/material-ui/react-grid2/ to avoid the alignment issues with the Grid's spacing.

NickAkhmetov (approved):
Looks good to me!

NickAkhmetov:
> We should try to move to [mui.com/material-ui/react-grid2](https://mui.com/material-ui/react-grid2/) to avoid the alignment issues with the Grid's spacing.

Agreed, we could also do this styling with normal CSS grid

PR #3473, opened 2024-07-18
John conroy/update table of contents CAT-755

john-conroy:
TableOfContents

TableOfContentsResponsive

Updates the table of contents to reflect updated design. The links for the processed data sections are just stubs until those sections can be implemented.


john-conroy:
@tsliaw There have been no changes in ui/ux since you last took a look.

tsliaw:
Couple notes

- It doesn't look like organ page's table of contents was updated? (only noticed since summary section doesn't have an icon, which might be the only relevant section with an icon)
Screenshot 2024-07-22 at 10 42 21 AM
-Not related to this PR maybe since this is on prod, but the "Derived" section is the only section not spelled out on sample pages. 
Screenshot 2024-07-22 at 10 41 11 AM

john-conroy:
Yeah, I noticed that. The organ page uses the `TableOfContents` component directly instead of the `DetailLayout`. I can make that change.

john-conroy:
@tsliaw organ page should be updated now if you want to check again.

john-conroy:
@tsliaw sample/donor derived data sections are now all labeled as 'Derived Data'.

context/app/static/js/components/Routes/Route/Route.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const rightRouteBoundaryID = 'right-route-boundary';
+
+function RouteBoundary({
+  id,
+  showBoundary,
+}: {
+  id: typeof leftRouteBoundaryID | typeof rightRouteBoundaryID;
+  showBoundary: boolean;
+}) {
+  return <Box id={id} flex="1 0" sx={{ padding: 2, ...(!showBoundary && { display: 'none' }) }} />;
```
NickAkhmetov, line 20: I think we can provide the padding/display properties directly here since they're system props

NickAkhmetov (commented):
Some minor organizational items, but the overall work is great! Can't think of any major issues.

NickAkhmetov (approved):
🚀

PR #3475, opened 2024-07-22
Austenem/cat 774 fix workspaces text

austenem:
This PR updates the login alert text on the workspaces landing page in order to match intended designs.


NickAkhmetov (approved):

john-conroy (changes requested):
Changelog entries need to begin with a `-`.

john-conroy (approved):
Thanks!

PR #3485, opened 2024-07-26
Austenem/CAT-711 add dataset limit warnings

austenem:
## Summary

This update adds alerts, errors, tooltips, and disabled buttons as needed to alert users whenever they exceed a dataset limit (alerts when the # of datasets > 25, and errors when the # of datasets > 150). This is the result of deciding that after 25 datasets are added to a workspace, workspace creation time is extended to the point where additional user communication is necessary.

This impacts all of the pages associated with creating or editing a workspace.

[CAT-711 JIRA ticket (dev)](https://hms-dbmi.atlassian.net/issues/?jql=project+%3D+%22CAT%22+AND+assignee+%3D+currentUser%28%29+ORDER+BY+created+DESC&atlOrigin=eyJpIjoiNTNmOGRhMGU5NDM0NGVmYjlhNzY0NjE4NGVkZjRhM2MiLCJwIjoiaiJ9)

[CAT-783 JIRA ticket (design)](https://hms-dbmi.atlassian.net/browse/CAT-783?atlOrigin=eyJpIjoiMGI1ZGY1Yzc4ZWEyNDNiMWFiNjdlZDY3NGE1M2ZmN2IiLCJwIjoiaiJ9)

[Figma mockups](https://www.figma.com/design/NgBjvs0jRbDHhiyn5nicDm/Workspace?node-id=2727-17336&m=dev)

## Testing

This update was manually tested by checking that the 25+ alert banner and the 150+ error banner appears on all pages associated with adding datasets to a workspace (see screenshots). The 150+ datasets error banner should also disable all buttons that would allow a workspace to be created/added to. 

The absence of these banners/disabled buttons was also checked in cases where a workspace had < 25 attached datasets.

## Screenshots/Video

_(note: for these documentation screenshots, the dataset max was set at 30, but is now set at 150)._

"Launch New Workspace" page:

Screenshot 2024-07-26 at 11 58 13 AM
Screenshot 2024-07-26 at 12 08 35 PM

"Add Datasets to Existing Workspace" page:

Screenshot 2024-07-26 at 12 14 34 PM
Screenshot 2024-07-26 at 12 15 55 PM
Screenshot 2024-07-26 at 12 16 58 PM

Workspace detail pages:

Screenshot 2024-07-26 at 12 25 20 PM
Screenshot 2024-07-26 at 12 27 39 PM
Screenshot 2024-07-26 at 12 28 39 PM

## Checklist

- [X] All existing tests pass

## Additional Notes

Event tracking for occurrences of selecting 25+ datasets has not been added.


context/app/static/js/components/workspaces/WorkspaceLaunchStopButtons/WorkspaceLaunchStopButtons.tsx
```diff
@@ -24,12 +26,17 @@ function StopWorkspaceButton({
...
   isStoppingWorkspace,
+  checkMaxDatasets = false,
 }: Omit<WorkspaceButtonProps, 'disableLaunch' | 'disableStop'>) {
   const { toastError } = useSnackbarActions();
+
   const currentWorkspaceIsRunning = isRunningWorkspace(workspace);
-  if (!currentWorkspaceIsRunning) {
+  const currentWorkspaceHasMaxDatasets = checkMaxDatasets && isWorkspaceAtDatasetLimit(workspace);
+
+  if (!currentWorkspaceIsRunning || currentWorkspaceHasMaxDatasets) {
```
john-conroy, line 36: A little confused here. We don't want to allow workspaces to stop the workspace if it has max datasets?

context/app/static/js/components/workspaces/WorkspaceLaunchStopButtons/WorkspaceLaunchStopButtons.tsx
```diff
@@ -52,9 +59,11 @@ function StopWorkspaceAlertButton(props: ButtonProps) {
 
 function StopWorkspaceAlert() {
   const { handleStopWorkspace, isStoppingWorkspace, workspacesList } = useWorkspacesList();
+
   const runningWorkspace = findRunningWorkspace(workspacesList);
+  const runningWorkspaceHasMaxDatasets = runningWorkspace && isWorkspaceAtDatasetLimit(runningWorkspace);
 
-  if (!runningWorkspace) {
+  if (!runningWorkspace || runningWorkspaceHasMaxDatasets) {
```
john-conroy, line 66: Same comment as above.

context/app/static/js/components/workspaces/WorkspaceListItem.tsx
```diff
@@ -38,11 +41,18 @@ function WorkspaceListItem({
...
   const isRunning = workspace.jobs.some((j) => !jobStatuses[j.status].isDone);
+  const hasMaxDatasets = checkMaxDatasets && isWorkspaceAtDatasetLimit(workspace);
 
-  const tooltip = isRunning ? 'Stop all jobs before selecting.' : undefined;
+  let tooltip;
+  if (hasMaxDatasets) {
+    tooltip = 'This workspace has reached the maximum number of datasets allowed. You cannot add any more datasets.';
+  } else if (isRunning) {
+    tooltip = 'Stop all jobs before selecting.';
+  }
```
john-conroy, line 55: Would prefer to avoid the `let`. Perhaps a `switch` statement?

context/app/static/js/shared-styles/alerts/AlertMessages/AlertMessages.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
```
john-conroy, line 1: Is this pattern of showing either warning or error messages something we want generalized in our shared components or is it specific to this case? If so could we use a more descriptive component name which captures that these are warning and error messages? Alerts can be used for purposes other than warnings and errors.

context/app/static/js/components/workspaces/WorkspaceLaunchStopButtons/WorkspaceLaunchStopButtons.tsx
```diff
@@ -24,12 +26,17 @@ function StopWorkspaceButton({
   handleStopWorkspace,
   button: ButtonComponent,
   isStoppingWorkspace,
+  checkMaxDatasets = false,
```
john-conroy, line 29: If this is necessary can we just pass `hasMaxDatasets` to avoid calling `isWorkspaceAtDatasetLimit` here and in its parent?

john-conroy (changes requested):
Thanks! A few comments.

context/app/static/js/shared-styles/alerts/AlertMessages/AlertMessages.tsx
```diff
@@ -0,0 +1,42 @@
...
+function AlertMessages({ errorMessages = [], warningMessages = [] }: AlertMessagesProps) {
+  if (errorMessages.length > 0 || warningMessages.length > 0) {
+    return (
+      <Box sx={{ display: 'grid', gap: 1, marginBottom: 3 }}>
+        {/* If there are error messages, do not display warning messages */}
+        {errorMessages.length > 0
+          ? errorMessages.map((message) => {
+              return (
+                <div key={message}>
+                  <Alert key={message} severity="error">
```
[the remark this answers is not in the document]
austenem: They aren't here, good catch!

context/app/static/js/shared-styles/alerts/AlertMessages/AlertMessages.tsx
austenem: That's a good point - I think a more specific name would be fitting (maybe`ErrorOrWarningMessages`).

context/app/static/js/components/workspaces/WorkspaceLaunchStopButtons/WorkspaceLaunchStopButtons.tsx
austenem: This is some logic required for separating the use cases of `WorkspaceListItem`. This component is used both in the 'Workspaces' page, for selecting workspaces for deletion, and in the 'Add Datasets to Workspace' dialog, for selecting workspaces to add datasets to.

In the former case, a workspace should be selectable for deletion regardless of the # of attached datasets. In the latter, if a workspace is at the max # of datasets, it needs a new tooltip and the 'stop jobs' button should be removed (as it should not be selectable regardless of whether jobs are running).

Maybe this prop can be renamed to describe these use cases, instead of what's actually being checked, to make this clearer? (for example, something like `isInAddDatasetsDialog`?)

john-conroy: Ah, I get it now. Can we just use the existing `disableStop` prop in `WorkspaceLaunchStopButtons`? Albeit `disableStop` is a misnomer as it actually prevents the button from being shown.

john-conroy: If the true intended behavior is for the button to be visible, but disabled, I would rename the existing `disable*` props to `show*` and add a new disabled prop for general use.

austenem: For this case, I'm hesitant to use a `disabled` prop because the `hasMaxDatasets` condition that requires disabling the button should also generate a specific tooltip. Using the `disabled` prop would mean checking for `hasMaxDatasets` in both components, unless the tooltip is also passed as a prop - is this preferable? Or is the `disabled` prop intended for future cases?

context/app/static/js/components/workspaces/WorkspaceLaunchStopButtons/WorkspaceLaunchStopButtons.tsx
austenem: No longer an issue with `showStop` prop 🎉

context/app/static/js/components/workspaces/WorkspaceLaunchStopButtons/WorkspaceLaunchStopButtons.tsx
john-conroy: I might try the following steps:

1. Add a `disabled` prop to `WorkspaceListItem`.
2. Add a `tooltip` prop to `WorkspaceListItem`. Do something like `const tooltipToDisplay = tooltip === undefined && isRunning ? 'Stop all jobs before selecting.' : tooltip` to prioritize tooltips.
3. Create a wrapper component for `WorkspaceListItem` in `AddDatasetsFromSearchDialog` to pass in the tooltip and disabled props.

austenem: Nice!

context/app/static/js/components/workspaces/WorkspaceListItem.tsx
```diff
@@ -39,20 +39,15 @@ function WorkspaceListItem({
   toggleItem,
   selected,
   ToggleComponent,
-  disableLaunch = false,
-  disableStop = false,
-  checkMaxDatasets = false,
+  showLaunch = false,
+  showStop = false,
+  disabled = false,
```
john-conroy, line 44: These default props are now inverted. I imagine you updated all of the existing cases of `WorkspaceListItem`? I see the change for `WorkspacesList`.

austenem: Yes, that's correct - the other use of that component is in `AddDatasetsFromSearchDialog`, which was updated to reflect the new props.

PR #3487, opened 2024-07-30
Fixed assays and samples sections for organs

tkakar:
## Summary

Fixes the assays and samples sections on the detail page for organs. Some organs (such as brain and uterus) had `search:[]` in the yaml files, so the search query was not executed.

Jira ticket: [CAT-770](https://hms-dbmi.atlassian.net/browse/CAT-770)

## Testing
In the Dev env, for organs with datasets, now the samples and assay sections are visible and vice-versa.

## Screenshots/Video
Screenshot 2024-07-30 at 10 24 40 AM

## Checklist

- [x] All existing tests pass
- [ ] Any relevant documentation in JIRA/Confluence has been updated to reflect the new feature

## Additional Notes
To avoid changes in the samples component that underneaths uses a shared component, I added the re-render logic in the Assays component. The idea is that search will be executed for all organs, and the page will be re-rendered  only for the organs that returns 0 searches. The table-of-content also updates based on the presence of those sections.

A side note: I have noticed the assay table and the barchart look a little off. Maybe the vertical grid lines need to be removed from the barchart, or we need a border at the bottom? 
For the table, the bottom looks off, could be inherent from the underlying react component, but when scrolling is needed, the last row is cut off (attached image below).

Screenshot 2024-07-30 at 10 28 26 AM

[CAT-770]: https://hms-dbmi.atlassian.net/browse/CAT-770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/context/app/static/js/pages/Organ/Organ.tsx
+++ b/context/app/static/js/pages/Organ/Organ.tsx
@@ -1,8 +1,9 @@
 import React, { useMemo } from 'react';
 import Typography from '@mui/material/Typography';
+
 import TableOfContents from 'js/shared-styles/sections/TableOfContents';
 import { getSections } from 'js/shared-styles/sections/TableOfContents/utils';
-import useSearchData from 'js/hooks/useSearchData';
+import useSearchData, { useSearchHits } from 'js/hooks/useSearchData';
 import Azimuth from 'js/components/organ/Azimuth';
 import Assays from 'js/components/organ/Assays';
 import Description from 'js/components/organ/Description';
@@ -10,8 +11,8 @@ import HumanReferenceAtlas from 'js/components/organ/HumanReferenceAtlas';
 import Samples from 'js/components/organ/Samples';
 import DatasetsBarChart from 'js/components/organ/OrganDatasetsChart';
 import Section from 'js/shared-styles/sections/Section';
-
 import { OrganFile } from 'js/components/organ/types';
+import { mustHaveOrganClause } from './queries';
 import { FlexRow, Content } from './style';
 
 interface OrganProps {
@@ -39,14 +40,11 @@ const samplesId = 'Samples';
 
 function Organ({ organ }: OrganProps) {
   const searchItems = useMemo(
-    // to avoid returning all datasets with organ.search, [organ.name] is added
     () => (organ.search.length > 0 ? organ.search : [organ.name]),
     [organ.search, organ.name],
   );
 
-  let shouldDisplaySearch = organ.search.length >= 0;
-
-  const query = useMemo(
+  const assaysQuery = useMemo(
     () => ({
       size: 0,
       aggs: {
@@ -59,19 +57,12 @@ function Organ({ organ }: OrganProps) {
                     'entity_type.keyword': 'Dataset',
                   },
                 },
-                {
-                  bool: {
-                    should: searchItems.map((searchTerm) => ({
-                      term: { 'origin_samples.mapped_organ.keyword': searchTerm },
-                    })),
-                  },
-                },
+                mustHaveOrganClause(searchItems),
               ],
             },
           },
           aggs: {
             'assay_display_name.keyword': { terms: { field: 'assay_display_name.keyword', size: 100 } },
-            'assay_display_name.keyword_count': { cardinality: { field: 'assay_display_name.keyword' } },
           },
         },
       },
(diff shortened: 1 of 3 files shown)


context/app/static/js/pages/Organ/Organ.tsx
```diff
@@ -26,7 +26,20 @@ const assaysId = 'Assays';
...
+
+  const getShouldDisplaySearch = () => {
+    if (searchHasData) return organ.search.length > 0 || organ.name;
+    return organ.search.length > 0;
+  };
+
+  const getSearchItems = () => {
+    if (searchHasData) return organ.search.length > 0 ? organ.search : [organ.name];
+    return organ.search;
+  };
```
john-conroy, line 39: Can we just do `const searchTerm = organ?.search ?? organ.name;`

context/app/static/js/components/organ/Assays/Assays.jsx
```diff
@@ -17,9 +17,11 @@ import { useDatasetTypeMap } from 'js/components/home/HuBMAPDatasetsChart/hooks'
 import { Flex, StyledInfoIcon, StyledDatasetIcon } from '../style';
 import { getSearchURL } from '../utils';
 
-function Assays({ organTerms }) {
+function Assays({ organTerms, onHandleSearchStatus }) {
   const assayTypeMap = useDatasetTypeMap();
 
+  const [bucketData, setBucketData] = useState([]);
```
john-conroy, line 23: Not sure why we would need to add this to state? We can continue to access the buckets as we have.

context/app/static/js/components/organ/Assays/Assays.jsx
```diff
@@ -54,9 +56,14 @@ function Assays({ organTerms }) {
...
-    ? searchData.aggregations.mapped_data_types['assay_display_name.keyword'].buckets
-    : [];
+
+  useEffect(() => {
+    if (searchData && searchData.aggregations) {
+      const buckets = searchData.aggregations.mapped_data_types['assay_display_name.keyword'].buckets || [];
+      setBucketData(buckets);
+      if (buckets.length === 0) onHandleSearchStatus(false);
+    }
+  }, [searchData, onHandleSearchStatus]);
```
john-conroy, line 66: Objects, like `searchData`, and arrays provided in the dependencies are compared by reference and will continuously trigger if not memoized, handled with a custom compare fn, or handled using a custom hook like `useDeepCompareEffect` which performs a deep comparison. In this case, if we wanted to track whether the `searchData` has buckets in state we could define `const hasBuckets = buckets.length` which could then be used in the `useEffect` without issue.

context/app/static/js/pages/Organ/Organ.tsx
```diff
@@ -68,15 +85,15 @@ function Organ({ organ }: OrganProps) {
             <Azimuth config={organ.azimuth} />
           </Section>
         )}
-        {shouldDisplaySection[assaysId] && (
+        {shouldDisplaySection[assaysId] && searchHasData && (
           <Section id={assaysId}>
-            <Assays organTerms={organ.search} />
-            <DatasetsBarChart search={organ.search} />
+            <Assays organTerms={searchItems} onHandleSearchStatus={handleDataStatus} />
```
john-conroy, line 90: I think we can guarantee that if there are datasets, there are samples, but possibly not the other way around.

context/app/static/js/components/organ/Assays/Assays.jsx
```diff
@@ -54,9 +56,14 @@ function Assays({ organTerms }) {
   );
 
   const { searchData } = useSearchData(query);
```
john-conroy, line 58: I think the simplest approach would just be to lift this to the parent for both `Assays` and `Samples` and avoid using the `useEffect` altogether?

john-conroy (changes requested):
Thanks! A few comments in hopes of simplifying this a bit. I'll ask @NickAkhmetov to chime in and maybe he has an approach to make this simpler than what I've suggested.

context/app/static/js/pages/Organ/Organ.tsx
```diff
@@ -68,15 +85,15 @@ function Organ({ organ }: OrganProps) {
             <Azimuth config={organ.azimuth} />
           </Section>
         )}
-        {shouldDisplaySection[assaysId] && (
+        {shouldDisplaySection[assaysId] && searchHasData && (
```
NickAkhmetov, line 88: Could `searchHasData` be included as part of the `shouldDisplaySection` logic above?

john-conroy (approved):
Great, thanks!

PR #3489, opened 2024-07-30
Austenem/cat 765 select default template

austenem:
## Summary

This update selects the blank template by default in the "Create New Workspace" dialog.

[CAT-765 JIRA ticket](https://hms-dbmi.atlassian.net/browse/CAT-765?atlOrigin=eyJpIjoiNGQ0MmMyMmFjNjNjNGNkZTllMzA2YTZlYTFjMDI3ZWIiLCJwIjoiaiJ9)

## Testing

The altered component appears in the "Create new workspace" dialog and the "Add templates" dialog (see documentation below).

Both dialogs were manually tested by ensuring that the blank template is only selected in the "Create new workspace" dialog and that the template is able to be deselected.

## Screenshots/Video
Screenshot 2024-07-30 at 5 10 44 PM
Screenshot 2024-07-30 at 5 11 22 PM

## Checklist

- [X] All existing tests pass

## Additional Notes

There is an ongoing issue with a `useEffect` dependency array in the `SelectableTemplateGrid` component - I'll drop a comment about this below, but would appreciate some feedback about the best way to handle this.

### DIFF
--- a/context/app/static/js/components/workspaces/SelectableTemplateGrid/SelectableTemplateGrid.tsx
+++ b/context/app/static/js/components/workspaces/SelectableTemplateGrid/SelectableTemplateGrid.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, ChangeEvent, useEffect } from 'react';
+import React, { useCallback, ChangeEvent } from 'react';
 import Typography from '@mui/material/Typography';
 import Button from '@mui/material/Button';
 import Box from '@mui/material/Box';
@@ -20,7 +20,6 @@ const inputName = 'templates';
 
 interface ControllerProps<FormType extends FormWithTemplates> {
   control: Control<FormType>;
-  defaultSelectedTemplateKey?: string;
 }
 
 function getActiveTemplates({ templates, disabledTemplates = {} }: TemplateGridProps) {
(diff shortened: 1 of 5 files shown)


context/app/static/js/components/workspaces/SelectableTemplateGrid/SelectableTemplateGrid.tsx
```diff
@@ -51,19 +54,37 @@ function SelectableTemplateGrid<FormType extends FormWithTemplates>({
...
-    [updateTemplates, selectedTemplates],
+    [selectOrUnselectTemplate],
   );
 
+  useEffect(() => {
+    if (defaultSelectedTemplateKey) {
+      selectOrUnselectTemplate(defaultSelectedTemplateKey, 'select');
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
austenem, line 86: This is the `useEffect` issue mentioned above. I want this hook to be triggered on component mount only, which is why the empty dependency array is necessary - however, as this violates eslint rules, I'm assuming there is likely a better way to accomplish this. If the included dependencies are added to the array, the component will re-render infinitely.

john-conroy (commented):
Can we add the default template key to `defaultValues.templates` in the workspace form hook?

john-conroy:
If the above comment works, I don't think we would need to make any other change.

NickAkhmetov:
I believe it should be possible to avoid using `useEffect`. Considering that the main goal is to set an initial selection, is there an approach that would allow us to define the initial state of the form to have this selected?

john-conroy (approved):
Nice, glad it worked!

austenem:
Yes to both comments - that ended up being a much simpler solution!

PR #3509, opened 2024-08-16
Austenem/CAT-840 fix protected dataset ids

austenem:
## Summary

Fix bug that prevented protected dataset ids from being displayed in "Launch New Workspace" dialog.

[CAT-840 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-840?atlOrigin=eyJpIjoiMDI3Mzg2MWMxNTBjNGRiZmI5MTkxNjZiMTRhYzVhYjciLCJwIjoiaiJ9)

## Testing

Tested by checking that the IDs only appear when one or more protected datasets are selected, that the datasets can be deleted via the provided method, and that the other Workspace field components work as expected (namely the "Workspace Name" field).

## Screenshots/Video

Update:
Screenshot 2024-08-16 at 3 11 06 PM

Previous bug:
Screenshot 2024-08-16 at 3 24 18 PM

## Checklist

- [X] All existing tests pass

## Additional Notes

This may not be the most elegant solution to this issue of updating the IDs after the initial render - if there's another recommended method, I'd be open to it.

### DIFF
--- a/context/app/static/js/components/workspaces/NewWorkspaceDialog/useCreateWorkspaceForm.ts
+++ b/context/app/static/js/components/workspaces/NewWorkspaceDialog/useCreateWorkspaceForm.ts
@@ -26,7 +26,7 @@ interface CreateWorkspaceFormTypes extends FormWithTemplates {
 interface UseCreateWorkspaceTypes {
   defaultName?: string;
   defaultTemplate?: string;
-  defaultProtectedDatasets?: string;
+  initialProtectedDatasets?: string;
 }
 
 const schema = z
@@ -34,9 +34,13 @@ const schema = z
   .partial()
   .required({ 'workspace-name': true, templates: true });
 
-function useCreateWorkspaceForm({ defaultName, defaultTemplate, defaultProtectedDatasets }: UseCreateWorkspaceTypes) {
+function useCreateWorkspaceForm({ defaultName, defaultTemplate, initialProtectedDatasets }: UseCreateWorkspaceTypes) {
   const [dialogIsOpen, setDialogIsOpen] = useState(false);
   const createTemplateNotebooks = useTemplateNotebooks();
+
+  const checkedWorkspaceName = defaultName ?? '';
+  const checkedProtectedDatasets = initialProtectedDatasets ?? '';
+  const checkedTemplates = [defaultTemplate ?? DEFAULT_TEMPLATE_KEY];
 
   const {
     handleSubmit,
@@ -46,9 +50,9 @@ function useCreateWorkspaceForm({ defaultName, defaultTemplate, defaultProtected
     trigger,
   } = useForm({
     defaultValues: {
-      'workspace-name': defaultName ?? '',
-      'protected-datasets': defaultProtectedDatasets ?? '',
-      templates: [defaultTemplate ?? DEFAULT_TEMPLATE_KEY],
+      'workspace-name': checkedWorkspaceName,
+      'protected-datasets': checkedProtectedDatasets,
+      templates: checkedTemplates,
       workspaceJobTypeId: DEFAULT_JOB_TYPE,
     },
     mode: 'onChange',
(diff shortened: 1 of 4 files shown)


context/app/static/js/components/workspaces/WorkspaceField/WorkspaceField.tsx
```diff
@@ -24,6 +24,12 @@ function WorkspaceField<FormType extends FieldValues>({
     defaultValue: value,
   });
 
+  useEffect(() => {
+    if (value) {
+      field.onChange(value);
+    }
+  }, [value, field]);
+
```
john-conroy, line 32: I'm a bit curious what introduced this bug. We were only experiencing it with the protected datasets field, but not the other instances of `WorkspaceField`.

austenem: From what I could tell, the other instances of WorkspaceField are only related to the workspace name field, which defaults to an empty string - this one defaults to the protected dataset ids, which are initially blank. This potentially changes later on depending on which datasets are selected, but the form wasn't updating the default field value with the new IDs after the initial render.

john-conroy: Well it definitely looks like I introduced this bug (sorry)! Can you try providing an optional initial value for `protected-datasets` to `useCreateWorkspaceForm` like we do in `/workspaces/AddDatasetsFromSearchDialog/hooks.ts`?

austenem: Yes, that works - however, I'm still seeing a need for a `useEffect` function in `hooks.ts` to monitor when the initial value changes from an empty string to a list of IDs. I'll push my changes in one moment.

context/app/static/js/components/workspaces/NewWorkspaceDialog/useCreateWorkspaceForm.ts
```diff
@@ -66,6 +67,17 @@ function useCreateWorkspaceForm({ defaultName, defaultTemplate }: UseCreateWorks
...
+  useEffect(() => {
+    if (defaultProtectedDatasets && defaultProtectedDatasets !== '') {
+      reset({
+        'workspace-name': defaultName ?? '',
+        'protected-datasets': defaultProtectedDatasets ?? '',
+        templates: [defaultTemplate ?? DEFAULT_TEMPLATE_KEY],
+        workspaceJobTypeId: DEFAULT_JOB_TYPE,
+      });
+    }
+  }, [defaultProtectedDatasets, reset, defaultName, defaultTemplate]);
```
john-conroy, line 79: After looking more into this, you were right this is necessary because the form is created before any selections have been made.

Let's keep this logic, but let's pull the default values e.g. ` defaultName ?? ''` into variables for reuse since they're used both in `defaultValues` and the `useEffect`. Also maybe  'initialProtectedDatasets' instead of 'defaultProtectedDatasets' since they depend on user selections?

context/app/static/js/components/workspaces/NewWorkspaceDialog/useCreateWorkspaceForm.ts
```diff
@@ -68,15 +72,16 @@ function useCreateWorkspaceForm({ defaultName, defaultTemplate, defaultProtected
...
-        'protected-datasets': defaultProtectedDatasets ?? '',
-        templates: [defaultTemplate ?? DEFAULT_TEMPLATE_KEY],
+        'workspace-name': checkedWorkspaceName,
+        'protected-datasets': checkedProtectedDatasets,
+        templates: checkedTemplates,
         workspaceJobTypeId: DEFAULT_JOB_TYPE,
       });
     }
-  }, [defaultProtectedDatasets, reset, defaultName, defaultTemplate]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
john-conroy, line 83: Why the eslint-disable? I think you need to put the other `checked*` variables into the dependency array.

john-conroy (approved):

PR #3514, opened 2024-08-21
Austenem/CAT-767 update workspace toast

austenem:
## Summary

Update language and add an action button to toasts for successful workspace creation. This also takes care of notifying users that workspaces are launched in a new tab and to watch out for their pop-up settings.

[CAT-767 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-767?atlOrigin=eyJpIjoiYmNmODRiYWEzOTFlNDk0YmIxNjEwMTZmYjQyNWQwMjUiLCJwIjoiaiJ9)

[Figma mockup](https://www.figma.com/design/NgBjvs0jRbDHhiyn5nicDm/Workspace?node-id=2944-3124&m=dev)

## Testing

Visual comparison to mockups, ensured that button links to the correct workspace detail page. 

## Screenshots/Video

Screenshot 2024-08-21 at 3 07 13 PM

## Checklist

- [X] All existing tests pass

### DIFF
--- a/context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
+++ b/context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
@@ -5,7 +5,7 @@ import Button from '@mui/material/Button';
 
 function WorkspaceLaunchSuccessToast(id: number) {
   return (
-    <Stack spacing={1} sx={{ maxWidth: '22rem' }}>
+    <Stack spacing={1} maxWidth="22rem">
       <Typography>
         Workspace successfully launched in a new tab. If the tab didn&apos;t open, please check your pop-up blocker
         settings and relaunch your workspace.
--- a/context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
+++ b/context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
@@ -5,16 +5,14 @@ import Button from '@mui/material/Button';
 
 function WorkspaceLaunchSuccessToast(id: number) {
   return (
-    <Stack sx={{ maxWidth: '22rem' }}>
+    <Stack spacing={1} sx={{ maxWidth: '22rem' }}>
       <Typography>
         Workspace successfully launched in a new tab. If the tab didn&apos;t open, please check your pop-up blocker
         settings and relaunch your workspace.
       </Typography>
-      <Stack direction="row" justifyContent="flex-end" marginTop={1}>
-        <Button href={`/workspaces/${id}`} variant="text" color="inherit">
-          View Workspace Detail Page
-        </Button>
-      </Stack>
+      <Button href={`/workspaces/${id}`} variant="text" color="inherit" sx={{ alignSelf: 'flex-end' }}>
+        View Workspace Detail Page
+      </Button>
     </Stack>
   );
 }
--- a/context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
+++ b/context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
@@ -3,7 +3,7 @@ import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
 import Button from '@mui/material/Button';
 
-function WorkspaceToastSuccessMessage(id: number) {
+function WorkspaceLaunchSuccessToast(id: number) {
   return (
     <Stack sx={{ maxWidth: '22rem' }}>
       <Typography>
@@ -11,20 +11,12 @@ function WorkspaceToastSuccessMessage(id: number) {
         settings and relaunch your workspace.
       </Typography>
       <Stack direction="row" justifyContent="flex-end" marginTop={1}>
-        <Button
-          href={`/workspaces/${id}`}
-          variant="contained"
-          sx={{
-            backgroundColor: 'transparent',
-            boxShadow: 'none',
-            ':hover': { backgroundColor: 'transparent', boxShadow: 'none' },
-          }}
-        >
-          <Typography variant="button">View Workspace Detail Page</Typography>
+        <Button href={`/workspaces/${id}`} variant="text" color="inherit">
+          View Workspace Detail Page
         </Button>
       </Stack>
     </Stack>
   );
 }
 
-export default WorkspaceToastSuccessMessage;
+export default WorkspaceLaunchSuccessToast;
--- a/context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
+++ b/context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import Stack from '@mui/material/Stack';
+import Typography from '@mui/material/Typography';
+import Button from '@mui/material/Button';
+
+function WorkspaceToastSuccessMessage(id: number) {
+  return (
+    <Stack sx={{ maxWidth: '22rem' }}>
+      <Typography>
+        Workspace successfully launched in a new tab. If the tab didn&apos;t open, please check your pop-up blocker
+        settings and relaunch your workspace.
+      </Typography>
+      <Stack direction="row" justifyContent="flex-end" marginTop={1}>
+        <Button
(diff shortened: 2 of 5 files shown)


context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+          href={`/workspaces/${id}`}
+          variant="contained"
+          sx={{
+            backgroundColor: 'transparent',
+            boxShadow: 'none',
+            ':hover': { backgroundColor: 'transparent', boxShadow: 'none' },
+          }}
+        >
+          <Typography variant="button">View Workspace Detail Page</Typography>
+        </Button>
```
NickAkhmetov, line 24: Would it be possible to extend the `message` object that's sent to the snackbar by the various `toast` functions to accept an `action` and pass that in to the `Alert` in `StyledSnackbar`? This would allow us to use the API provided by MUI/avoid needing to add these extra styles: https://mui.com/material-ui/react-alert/#actions

austenem: It looks like it is possible, but some restyling of the main `toast` component would be needed in order to match the designs. Right now, the `action` slot is being used for a close button, and any additional `action`s passed in will override that, so there would need to be logic to handle both cases. Additionally, the extra styling for this button would still need to be there to match designs.

If this seems like an option that we could reuse frequently, I think it's doable, but otherwise I'd stay with the current React Node `message`.

Screenshot 2024-08-21 at 3 56 30 PM

NickAkhmetov (approved):

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+        <Button
+          href={`/workspaces/${id}`}
+          variant="contained"
+          sx={{
+            backgroundColor: 'transparent',
+            boxShadow: 'none',
+            ':hover': { backgroundColor: 'transparent', boxShadow: 'none' },
+          }}
+        >
+          <Typography variant="button">View Workspace Detail Page</Typography>
```
john-conroy, line 23: I don't think we need another nested button, especially without an `onClick`, or the typography component?

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Stack sx={{ maxWidth: '22rem' }}>
+      <Typography>
+        Workspace successfully launched in a new tab. If the tab didn&apos;t open, please check your pop-up blocker
+        settings and relaunch your workspace.
+      </Typography>
+      <Stack direction="row" justifyContent="flex-end" marginTop={1}>
+        <Button
+          href={`/workspaces/${id}`}
+          variant="contained"
```
john-conroy, line 16: Would `variant="text"` be more appropriate? We likely wouldn't need many of the below styles.

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Stack from '@mui/material/Stack';
+import Typography from '@mui/material/Typography';
+import Button from '@mui/material/Button';
+
+function WorkspaceToastSuccessMessage(id: number) {
+  return (
+    <Stack sx={{ maxWidth: '22rem' }}>
```
john-conroy, line 8: Why this `maxWidth` in particular?

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <Typography>
+        Workspace successfully launched in a new tab. If the tab didn&apos;t open, please check your pop-up blocker
+        settings and relaunch your workspace.
+      </Typography>
+      <Stack direction="row" justifyContent="flex-end" marginTop={1}>
```
john-conroy, line 13: Do we need a second stack? Could we just apply these props to the outer stack and remove this? Perhaps add `alignItems="center"` for alignment?

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Stack from '@mui/material/Stack';
+import Typography from '@mui/material/Typography';
+import Button from '@mui/material/Button';
+
+function WorkspaceToastSuccessMessage(id: number) {
```
john-conroy, line 6: `WorkspaceLaunchSuccessToast` maybe?

john-conroy (changes requested):

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
austenem: Yes, that works - the weird styling was the result of attempting to change the button text color, which it seems can only be changed to a small number of themed options. I switched to using variant="text" and color="inherit", which worked.

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
austenem: This was from the designs - without a maxWidth, the toast stretches across the bottom of the screen and is hard to read. Is there a better method for setting width options like these?

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
austenem: I believe the second `Stack` is necessary as the text and button are aligned respectively to the left and right sides of the `Toast`. There may be a different approach to this, though

john-conroy: Try `justifyContent="space-between"`

john-conroy: As a prop on the outer stack.

john-conroy: You can use `alignSelf="flex-end"` to align the button on the right side. You might need to wrap the button in a `Box` but try it without first.

context/app/static/js/components/workspaces/WorkspaceToastSuccessMessage.tsx
NickAkhmetov: Would `fit-content` work? https://developer.mozilla.org/en-US/docs/Web/CSS/max-width#fit-content

austenem: That's a bit better than the 100% width - it still stretches out further than the designs, because I think the max-content is higher than 20rem. The `<length-percentage> `parameter doesn't seem to change anything, which is a bit odd:

Screenshot 2024-08-23 at 2 05 08 PM

context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Stack from '@mui/material/Stack';
+import Typography from '@mui/material/Typography';
+import Button from '@mui/material/Button';
+
+function WorkspaceLaunchSuccessToast(id: number) {
+  return (
+    <Stack spacing={1} sx={{ maxWidth: '22rem' }}>
```
NickAkhmetov, line 8: `maxWidth` can be provided as a `system` prop: 
```suggestion
    <Stack spacing={1} maxWidth="22rem">
```

- https://mui.com/material-ui/react-stack/#system-props
- https://mui.com/system/properties/

NickAkhmetov (commented):
Looks good to me! Thanks for experimenting with the other max width options. We can keep the current approach (with one minor adjustment).

NickAkhmetov (approved):

john-conroy (approved):
Thanks for the fixes!

PR #3515, opened 2024-08-21
NickAkhmetov/CAT-844 Migrate away from using `descendants` and `ancestors` in favor of `id` equivalents

NickAkhmetov:
## Summary

This PR removes the portal UI's dependencies on the direct ancestor/descendant information available in the entity API in favor of using the corresponding `ids`.

### Progress
- `ancestors`
	- [x] Dev search used `ancestors` to determine raw/processed status -> revised to use `processing` field
	- [x] Prov table used `ancestors` to construct the table -> revised to use `ancestor_ids`, corresponding revisions made to error display for donors
	- [x] Ancestors' `rui_location` field is used for `details_rui_json` on back-end and to determine whether a dataset is spatial in the dev search
	- [x] Ancestors not being datasets is currently a condition for the unified prov view lookup when looking up the top level dataset -> dropped the condition in favor of using `processing: raw`, there should be 0..1 raw datasets in any given dataset's ancestors anyway.
- `descendants`
  - Not in use.
- `immediate_ancestors`
  - Mentioned in an artillery config, but otherwise not present.
- `immediate_descendants`
  - ~~`useImmediateDescendantProv` is used for the `view derived entities` button~~ This actually uses the entity API endpoint; this can be improved to use the search API equivalent once the upstream change is made to the search API, but is not currently a blocker.

https://hms-dbmi.atlassian.net/browse/CAT-844

## Testing

Manually verified, confirmed functionality is analogous to current production behaviors.

## Screenshots/Video

Not very informative since this matches existing functionality, but for verification:

image

<details><summary>Dev search filters</summary>
<p>

image

image

image

</p>
</details> 

## Checklist

- [x] All existing tests pass

### DIFF
--- a/context/app/routes_browse.py
+++ b/context/app/routes_browse.py
@@ -139,18 +139,12 @@ def details_rui_json(type, uuid):
         abort(404)
     client = get_client()
     entity = client.get_entity(uuid)
-    # For Samples...
+    # For samples and datasets, the nearest RUI location is indexed with the entity itself.
+    # https://github.com/hubmapconsortium/search-api/pull/860
     if 'rui_location' in entity:
         return json.loads(entity['rui_location'])
-    # For Datasets...
-    if 'ancestors' not in entity:
-        abort(404)
-    located_ancestors = [a for a in entity['ancestors'] if 'rui_location' in a]
-    if not located_ancestors:
-        abort(404)
-    # There may be multiple: The last should be the closest...
-    # but this should be confirmed, when there are examples.
-    return json.loads(located_ancestors[-1]['rui_location'])
+    # Otherwise throw 404
+    abort(404)
 
 
 @blueprint.route('/sitemap.txt')
--- a/context/app/static/js/components/entity-tile/EntityTile/EntityTile.tsx
+++ b/context/app/static/js/components/entity-tile/EntityTile/EntityTile.tsx
@@ -1,6 +1,5 @@
 import React, { ComponentProps, ComponentType } from 'react';
 import Typography from '@mui/material/Typography';
-import IconButton from '@mui/material/IconButton';
 import ContentCopyIcon from '@mui/icons-material/ContentCopyRounded';
 
 import Tile from 'js/shared-styles/tiles/Tile/';
@@ -12,6 +11,7 @@ import ContactUsLink from 'js/shared-styles/Links/ContactUsLink';
 import { useHandleCopyClick } from 'js/hooks/useCopyText';
 import Paper from '@mui/material/Paper';
 import Stack from '@mui/material/Stack';
+import IconLink from 'js/shared-styles/Links/iconLinks/IconLink';
 import EntityTileFooter from '../EntityTileFooter/index';
 import EntityTileBody from '../EntityTileBody/index';
 import { StyledIcon } from './style';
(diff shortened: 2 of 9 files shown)


context/app/routes_browse.py
```diff
@@ -139,18 +139,12 @@ def details_rui_json(type, uuid):
         abort(404)
     client = get_client()
     entity = client.get_entity(uuid)
-    # For Samples...
+    # For samples and datasets, the nearest RUI location is indexed with the entity itself.
+    # https://github.com/hubmapconsortium/search-api/pull/860
     if 'rui_location' in entity:
         return json.loads(entity['rui_location'])
```
john-conroy, line 145: Datasets don't include an `rui_location` unless you've made some corresponding change in the `search-api`.

context/app/static/js/components/entity-tile/EntityTile/EntityTile.tsx
```diff
@@ -73,56 +73,25 @@ function ErrorTile({ entity_type, id }: Pick<EntityTileProps, 'id' | 'entity_typ
...
       variant="outlined"
     >
-      <Stack direction="row" gap={1} color="error.main" alignItems="center">
-        <StatusIcon status="error" sx={{ fontSize: '1.5rem' }} />
-        <Typography variant="subtitle1">{id}</Typography>
-        <IconButton onClick={() => copy(id)}>
-          <ContentCopyIcon color="info" />
-        </IconButton>
+      <Stack direction="row" gap={1}>
+        <StatusIcon status="error" sx={{ alignSelf: 'start', fontSize: '1.3rem' }} />
```
john-conroy, line 77: Maybe `1.25rem`, I believe we use that more commonly throughout. I can't recall using `1.3rem` before, but maybe you have?

context/app/static/js/components/types.ts
```diff
@@ -38,7 +38,9 @@ export interface Entity {
   contacts: ContactAPIResponse[];
   contributors: ContributorAPIResponse[];
   created_timestamp: number;
+  /** @deprecated Use `ancestor_ids` and `useEntitiesData` instead */
   ancestors: Entity[];
+  ancestor_ids: string[];
   // eslint-disable-next-line no-use-before-define -- Donor is defined later in the file and extends Entity
   donor: Donor;
   descendant_counts: { entity_type: Record<string, number> };
```
john-conroy, line 46: Should we add like comments for `descendants` and `descendant_counts` or do we plan to keep those for now? I don't think we have the immediate variants in the types.

context/app/static/js/hooks/useEntityData.ts
```diff
@@ -2,12 +2,30 @@ import { useMemo } from 'react';
...
 
   return [searchHits[0]?._source, isLoading];
 }
 
+export function useEntitiesData(uuids: string[], source?: string[]): [Entity[], boolean] {
+  const query = useEntityQuery(uuids, source);
+
+  const { searchHits, isLoading } = useSearchHits<Entity>(query);
+
+  return [searchHits.map((hit) => hit._source), isLoading];
```
john-conroy, line 28: Slightly prefer an obj so we can also include the other returned values from `useSearchHits`.

context/app/static/js/pages/search/DevSearch.jsx
```diff
@@ -88,10 +88,8 @@ function DevSearch() {
         checkboxFilter('no_metadata', 'No metadata?', BoolMustNot(ExistsQuery('metadata.metadata'))),
         checkboxFilter('has_files', 'Has files?', ExistsQuery('files')),
         checkboxFilter('no_files', 'No files?', BoolMustNot(ExistsQuery('files'))),
-        checkboxFilter('has_rui_sample', 'Spatial Sample?', ExistsQuery('rui_location')),
-        checkboxFilter('no_rui_sample', 'Not Spatial Sample?', BoolMustNot(ExistsQuery('rui_location'))),
-        checkboxFilter('has_rui_dataset', 'Spatial Dataset?', ExistsQuery('ancestors.rui_location')),
-        checkboxFilter('no_rui_dataset', 'Not Spatial Dataset?', BoolMustNot(ExistsQuery('ancestors.rui_location'))),
+        checkboxFilter('is_spatial', 'Spatial?', ExistsQuery('rui_location')),
+        checkboxFilter('not_spatial', 'Not Spatial?', BoolMustNot(ExistsQuery('rui_location'))),
```
john-conroy, line 92: Like above, I think this won't work for datasets unless there was some change I'm not aware of?

john-conroy (commented):
Do we plan to keep `entity_type` in the descendants for `descendants_counts`?

context/app/routes_browse.py
NickAkhmetov: I did make this change: https://github.com/hubmapconsortium/search-api/blob/dev-integrate/src/hubmap_translation/addl_index_transformations/portal/translate.py#L307

context/app/static/js/components/types.ts
NickAkhmetov: I think we do plan to keep `descendant_counts`,for the time being, but good call on adding this comment for `descendants`!

context/app/static/js/hooks/useEntityData.ts
NickAkhmetov: >other returned values

From what I can tell, `useSearchHits` only returns the `searchHits` and `isLoading` values, so I stuck with the same format that `useEntityData` uses above. Are you interested in us keeping the top-level `hit` data?

context/app/static/js/pages/search/DevSearch.jsx
NickAkhmetov: This is now supported with the change I linked earlier.

john-conroy (approved):

PR #3526, opened 2024-08-29
Austenem/CAT-867 update workspace toasts

austenem:
## Summary

Updates workspace success toasts to indicate whether the workspace has launched in a new tab. Additionally, updates logic to show this toast not just when a workspace is created, but any time a workspace is launched.

[CAT-867 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-867?atlOrigin=eyJpIjoiMGFlMzJhYjE4NDdmNGZhZGEzZDRlNGVjYTEwY2RkYTMiLCJwIjoiaiJ9)

## Testing

Tested various entry points for the toast:
* Creating a new workspace
  * When another workspace is running: `"Workspace successfully created."` + launch dialog, then if launched `"Workspace successfully launched..."`
  * When no workspaces are running: `"Workspace successfully launched..."`
* Launching an existing workspace
  * When another workspace is running: launch dialog, then if launched `"Workspace successfully launched..."`
  * When no workspaces are running: `"Workspace successfully launched..."`

## Screenshots/Video

Screenshot 2024-08-29 at 3 24 19 PM

Screenshot 2024-08-29 at 3 24 08 PM

## Checklist

- [X] All existing tests pass

### DIFF
--- a/context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
+++ b/context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
@@ -3,16 +3,27 @@ import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
 import Button from '@mui/material/Button';
 
-function WorkspaceLaunchSuccessToast(id: number) {
+interface WorkspaceLaunchSuccessToastProps {
+  id: number;
+  workspaceLaunched: boolean;
+}
+
+function WorkspaceLaunchSuccessToast({ id, workspaceLaunched }: WorkspaceLaunchSuccessToastProps) {
   return (
     <Stack spacing={1} maxWidth="22rem">
-      <Typography>
-        Workspace successfully launched in a new tab. If the tab didn&apos;t open, please check your pop-up blocker
-        settings and relaunch your workspace.
-      </Typography>
-      <Button href={`/workspaces/${id}`} variant="text" color="inherit" sx={{ alignSelf: 'flex-end' }}>
-        View Workspace Detail Page
-      </Button>
+      {workspaceLaunched ? (
+        <>
+          <Typography>
+            Workspace successfully launched in a new tab. If the tab didn&apos;t open, please check your pop-up blocker
+            settings and relaunch your workspace.
+          </Typography>
+          <Button href={`/workspaces/${id}`} variant="text" color="inherit" sx={{ alignSelf: 'flex-end' }}>
+            View Workspace Detail Page
+          </Button>
+        </>
+      ) : (
+        <Typography>Workspace successfully created.</Typography>
+      )}
     </Stack>
   );
 }
--- a/CHANGELOG-update-workspace-toasts.md
+++ b/CHANGELOG-update-workspace-toasts.md
@@ -1,0 +1,1 @@
+- Update workspace success toast to indicate whether the workspace has launched in a new tab or just been created.
--- a/context/app/static/js/components/workspaces/hooks.ts
+++ b/context/app/static/js/components/workspaces/hooks.ts
@@ -205,6 +205,7 @@ function useLaunchWorkspace() {
   const { open, setWorkspace } = useLaunchWorkspaceStore();
 
   const { handleUpdateWorkspace } = useHandleUpdateWorkspace();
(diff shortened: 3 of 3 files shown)


context/app/static/js/components/workspaces/WorkspaceLaunchSuccessToast.tsx
```diff
@@ -3,16 +3,27 @@ import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
 import Button from '@mui/material/Button';
 
-function WorkspaceLaunchSuccessToast(id: number) {
+interface WorkspaceLaunchSuccessToastProps {
+  id: number;
+  workspaceLaunched: boolean;
+}
+
+function WorkspaceLaunchSuccessToast({ id, workspaceLaunched }: WorkspaceLaunchSuccessToastProps) {
```
john-conroy, line 11: I think I'd just prefer two functions `WorkspaceLaunchSuccessToast` and `WorkspaceCreateSuccessToast` in a central `WorkspaceToasts` file rather than depending on the `workspaceLaunchedProp`.

@NickAkhmetov what do you think?

NickAkhmetov: I agree, this would also tie in with https://hms-dbmi.atlassian.net/browse/CAT-49

austenem: Agreed! Pushed an update with this + the remaining workspace toasts in a central file.

john-conroy:
Does it make sense to make these all into hooks and move the calls to `useSnackbarActions`?

austenem:
> Does it make sense to make these all into hooks and move the calls to `useSnackbarActions`?

So something like:

```
const useWorkspaceUpdateSuccessToast = () => {
    return () => toastSuccess('Workspace successfully updated.');
  };
```

for each of these in `useSnackbarActions()`?

john-conroy:
Yeah, but likely using `useCallback` with `useSnackbarActions` in its dependency array. What do you think? It would push all of the logic into the one file, but we don't necessarily have to address it in this PR.

austenem:
Would having the toast functions be in the dependency array be cleaner? For example, having the following in `useSnackbarActions()`:

```
  const useWorkspacesDeleteErrorToast =  useCallback(() => {
    return (names: string) => toastError(`Error deleting workspaces: ${names}`);
  }, [toastError]);

```
So that they can be all be pulled out of `useSnackbarActions()` instead of imported individually?

john-conroy:
Yep! That's what I meant, but struggled to communicate. Thanks!

john-conroy:
Something like: 

```
  const useWorkspacesDeleteErrorToast =  return useCallback( (names: string)  => {
    toastError(`Error deleting workspaces: ${names}`);
  }, [toastError]);
```

austenem:
Ah, ok - so just to clarify, in the example you gave it looks like these hooks would be defined outside of `useSnackbarActions()` and imported individually?

john-conroy:
I think I prefer your suggestion. A central hook which returns the various functions. Thanks! Sorry for the confusion.

context/app/static/js/shared-styles/snackbars/store.tsx
```diff
@@ -0,0 +1,163 @@
...
+    toastError: store.toastError,
+    toastInfo: store.toastInfo,
+    toastWarning: store.toastWarning,
+    toastSuccess: store.toastSuccess,
+    closeSnackbar: store.closeSnackbar,
+  }));
+
+  /** *********************************
+   *           Error Toasts           *
+   ********************************** */
```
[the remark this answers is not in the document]
austenem: That makes sense to me! I'd lean towards this hook living in `/workspaces`, unless there's an argument for keeping snackbar objects in a single place?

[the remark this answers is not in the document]
NickAkhmetov: I'm fine with that approach as well, we definitely don't have to have it all in one file, I just figured that might be less annoying to refactor for the time being 👍🏻

NickAkhmetov (approved):
🚀

austenem:
Thanks for reviewing!

PR #3528, opened 2024-09-03
NickAkhmetov/Hotfix descendants lookup query

NickAkhmetov:
## Summary

This PR fixes the query used to look up the processed descendants of a primary dataset so that it returns either Dataset descendants or supporting image pyramid descendants. This fixes the regression caused by a related change in #3519.

N/A

## Testing

Tested with the pages that the issue was observed on staging.

## Screenshots/Video

<details><summary>Details</summary>
<p>

image

</p>
</details> 

## Checklist

- [x] All existing tests pass

### DIFF
--- a/context/app/static/js/helpers/queries.ts
+++ b/context/app/static/js/helpers/queries.ts
@@ -35,9 +35,16 @@ export const includeDatasetsAndImageSupports: QueryDslQueryContainer = {
   bool: {
     should: [
       {
-        terms: { entity_type: ['Dataset', 'Support'] },
+        term: { 'entity_type.keyword': 'Dataset' },
       },
-      { terms: { 'vitessce-hints': ['pyramid', 'is_image'] } },
+      {
+        bool: {
+          must: [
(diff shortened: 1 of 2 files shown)


context/app/static/js/helpers/queries.ts
```diff
@@ -35,9 +35,16 @@ export const includeDatasetsAndImageSupports: QueryDslQueryContainer = {
   bool: {
     should: [
       {
-        terms: { entity_type: ['Dataset', 'Support'] },
```
john-conroy, line 38: I think just appending the `.keyword` to the field would work here as well?

john-conroy (approved):

context/app/static/js/helpers/queries.ts
NickAkhmetov: That approach leads to non-dataset support descendants being pulled in, e.g. publication ancillary data, since it matches against the `Support` entity type clause:

image

With the separate `entity_type.keyword` clauses, we ensure that the returned entities are either:
- a dataset
- an image pyramid support

image

PR #3535, opened 2024-09-09
NickAkhmetov/CAT-888 Improve combined provenance logic to safeguard against duplicate nodes/edges

NickAkhmetov:
## Summary

This PR reinforces the logic used to combine provenance graphs to guard against duplicate edges more effectively. Since edge IDs are not globally unique, collisions in edges don't necessarily mean they're referring to the same object - however, this led to cases where two edges with unique ID's led to duplicate connections between nodes, which caused the `cwlInput` conversion to throw errors.

With these changes, duplicate edges should no longer propagate to the graph from the combining logic. I've also improved the error thrown by the `makeCwlInput` function so that it is easier to track down exactly what's going wrong.

https://hms-dbmi.atlassian.net/browse/CAT-888

## Testing

Manually tested with the dataset that this issue was occuring on:
- Locally, opening the `graph` tab while logged in no longer causes crashes: http://localhost:5001/browse/dataset/438f7ba3b54f4020e8cbd852594aaca3?redirected=True#provenance
- On prod, the crashes are present while logged in: https://portal.hubmapconsortium.org/browse/dataset/438f7ba3b54f4020e8cbd852594aaca3?redirected=True#provenance

## Screenshots/Video

image

## Checklist

- [x] All existing tests pass

### DIFF
--- a/context/app/static/js/components/detailPage/provenance/utils.ts
+++ b/context/app/static/js/components/detailPage/provenance/utils.ts
@@ -9,12 +9,12 @@ export function createProvDataURL(uuid: string, entityEndpoint: string) {
 }
 
 function findDifference<T extends Record<string, unknown>, U extends Record<string, unknown>>(obj1: T, obj2: U) {
-  const diffKeys = Object.keys({ ...obj1, ...obj2 }).reduce((acc, key) => {
+  const diffKeys = Object.keys({ ...obj1, ...obj2 }).reduce<string[]>((acc, key) => {
     if (key in obj1 && key in obj2 && obj1[key] === obj2[key]) {
       return acc;
     }
     return [...acc, key];
-  }, [] as string[]);
+  }, []);
   return diffKeys;
 }
 


context/app/static/js/components/detailPage/provenance/utils.ts
```diff
@@ -21,11 +42,35 @@ export function nonDestructiveMerge<T extends Record<string, unknown>, U extends
...
     if (merged[key]) {
-      // Confirm that properties are actually different
+      // Heuristic to check if the objects are the same
       if (JSON.stringify(merged[key]) === JSON.stringify(value)) {
         return;
       }
+
+      if (typeof merged[key] === 'object' && typeof value === 'object') {
+        // Perform a more in-depth check to see if objects are the same when excluding certain fields
+        const difference = findDifference(merged[key] as object, value!).filter(
```
john-conroy, line 65: Forgive my ignorance, but what does the righthand `!` operator do?

NickAkhmetov: The `!` operator asserts that the value is non-null; despite having the `typeof value === 'object'` type guard above, this was necessary to avoid it being treated as null in the moment.

I've adjusted the guard statement to be more strict instead.

context/app/static/js/components/detailPage/provenance/utils.ts
```diff
@@ -8,6 +8,27 @@ export function createProvDataURL(uuid: string, entityEndpoint: string) {
...
+  Object.keys(obj2).forEach((key) => {
+    // @ts-expect-error -- We're specifically checking for the key in obj1
+    if (!(key in obj1) || obj1[key] !== obj2[key]) {
+      if (!diffKeys.includes(key)) {
+        diffKeys.push(key);
+      }
+    }
+  });
+  return diffKeys;
+}
```
john-conroy, line 28: Just brainstorming a potentially simpler method. Feel free to take or leave.

```
const diffKeys = Object.keys({...obj1, ...obj2}).reduce((acc, k) => {
 const v1 = obj1?.[key];
 const v2 = obj2?.[key];

if(v1 === v2){
   return acc;
}

return [...acc, key];
}, [])
```

context/app/static/js/components/detailPage/provenance/utils.ts
```diff
@@ -8,6 +8,27 @@ export function createProvDataURL(uuid: string, entityEndpoint: string) {
   return `${entityEndpoint}/entities/${uuid}/provenance`;
 }
 
+function findDifference(obj1: object, obj2: object) {
```
john-conroy, line 11: Why `object` as the types here rather than something like `Record<string, unknown>`? Is it because the later used `merged[key]` is `unknown` and we can only use a type guard for `object`?

context/app/static/js/components/detailPage/provenance/utils.ts
```diff
@@ -21,11 +42,35 @@ export function nonDestructiveMerge<T extends Record<string, unknown>, U extends
 ): T {
   const merged = { ...a } as Record<string, unknown>;
```
john-conroy, line 43: Can we provide stricter types for the provenance data?

john-conroy (approved):
I don't want to block the fix, but I'm curious why this case necessitates type assertion?

context/app/static/js/components/detailPage/provenance/utils.ts
NickAkhmetov: The types are currently provided via the generics, i.e. this function gets called with `ProvData` objects, then gets cast back to the ProvData type at the end with the `return merged as T`. Prior to adding this most recent prov-only logic, this was a more generic method we could reuse elsewhere; if we end up needing a similar functionality in the future, we could generify this approach by extracting the prov-specific logic into a function param.

context/app/static/js/components/detailPage/provenance/utils.ts
NickAkhmetov: Thanks, good call. I was definitely overthinking this.

context/app/static/js/components/detailPage/provenance/utils.ts
NickAkhmetov: I've generified this to match `nonDestructiveMerge`, thanks for pointing out the improvement!

context/app/static/js/components/detailPage/provenance/utils.ts
```diff
@@ -8,6 +8,18 @@ export function createProvDataURL(uuid: string, entityEndpoint: string) {
   return `${entityEndpoint}/entities/${uuid}/provenance`;
 }
 
+function findDifference<T extends Record<string, unknown>, U extends Record<string, unknown>>(obj1: T, obj2: U) {
+  const diffKeys = Object.keys({ ...obj1, ...obj2 }).reduce((acc, key) => {
```
john-conroy, line 12:
```suggestion
  const diffKeys = Object.keys({ ...obj1, ...obj2 }).reduce<string[]>((acc, key) => {
```
Would this avoid the assertion below?

john-conroy (approved):
Looks good to me, thanks!

PR #3542, opened 2024-09-18
Deployment updates for STAGE delete

yuanzhou:
In our current setup:
- `portal-prod.stage` goes against ALL PROD APIs and Neo4j data as well as PROD indices
- `portal.stage` goes against a local entity-api and search-api pointing a separate STAGE indices, but the backend data is from PROD

In the new setup:
- `portal-prod.test` will go against ALL PROD APIs and Neo4j data as well as PROD indices
- the existing `portal.test` stays the same

So we'll have DEV/TEST/PROD eventually. The only difference is we'll no longer use STAGE environment and the separate STAGE indices.

This PR will need to go hand in hand with https://github.com/hubmapconsortium/gateway/pull/318


lchoy (approved):
Thanks @yuanzhou !

The remaining references to the stage environment are in legacy artillery tests. Since these are no longer in use, it is not blocking.

`end-to-end/artillery/scenarios/assets.yml`
`end-to-end/artillery/skip-scenarios/entity-globus.yml`
`end-to-end/artillery/skip-scenarios/entity-provenance.yml`
`end-to-end/artillery/skip-scenarios/portal-details.yml`
`end-to-end/artillery/skip-scenarios/portal-docs.yml`
`end-to-end/artillery/skip-scenarios/search-ccf.yml`
`end-to-end/artillery/skip-scenarios/search.yml`

yuanzhou:
@lchoy @john-conroy I couldn't test the `redeploy.sh` myself, but once you merge this PR you should be able to deploy on DEV and TEST with that script. No STAGE. Then I'll enable the routing piece to the new `portal-prod.test` instance.

PR #3548, opened 2024-09-25
Austenem/CAT-873 add donor age messaging

austenem:
## Summary

Add messaging where donors with ages > 89 are displayed on search and detail pages, and cap age filter on search page at 90 years.

[CAT-873 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-873?atlOrigin=eyJpIjoiYzMyOGEzNzcyMjk5NDg3ZDhlZDRlOWYwZmFiOGQ3MDkiLCJwIjoiaiJ9)

## Testing

Tested for both 90+ aged donors.

## Screenshots/Video

Donor detail page:
Screenshot 2024-09-25 at 12 26 08 PM
Dataset detail page:
Screenshot 2024-09-25 at 12 26 58 PM
Tooltips:
Screenshot 2024-09-25 at 12 25 58 PM
Screenshot 2024-09-25 at 12 27 16 PM
Screenshot 2024-09-25 at 12 27 39 PM
Screenshot 2024-09-25 at 12 30 27 PM

## Checklist

- [X] All existing tests pass

### DIFF
--- a/context/app/static/js/components/searchPage/ResultsTable/utils.jsx
+++ b/context/app/static/js/components/searchPage/ResultsTable/utils.jsx
@@ -1,4 +1,7 @@
+import React from 'react';
+import Stack from '@mui/material/Stack';
 import { get } from 'js/helpers/nodash';
+import DonorAgeTooltip from 'js/shared-styles/tooltips/DonorAgeTooltip';
 
 const donorMetadataPath = 'mapped_metadata';
 const sampleMetdataPath = 'metadata';
@@ -48,6 +51,14 @@ function getByPath(hitSource, field) {
   }
 
   if (Array.isArray(fieldValue)) {
+    if (field?.id === 'mapped_metadata.age_value') {
+      return (
+        <Stack direction="row">
+          {fieldValue.join(' / ')}
+          <DonorAgeTooltip donorAge={fieldValue.join(' / ')} />
+        </Stack>
+      );
+    }
     return fieldValue.join(' / ');
   }
 
--- a/CHANGELOG-add-donor-age-messaging.md
+++ b/CHANGELOG-add-donor-age-messaging.md
@@ -1,0 +1,1 @@
+- Add messaging where donors with ages > 89 are displayed on search and detail pages.
--- a/context/app/static/js/components/detailPage/BulkDataTransfer/const.ts
+++ b/context/app/static/js/components/detailPage/BulkDataTransfer/const.ts
@@ -20,5 +20,3 @@ export const SRA_EXPERIMENT_TEXT = {
 
 export const BULK_DATA_DESCRIPTION_TEXT =
   'This section explains how to download data in bulk from raw and processed datasets. Processed datasets have separate download directories in Globus or dbGaP, distinct from the raw dataset.';
-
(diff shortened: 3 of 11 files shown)


NickAkhmetov (commented):
Just a few minor tweaks, great work otherwise! Also, would it be possible to add this logic to the `MetadataTable` component as well to make sure we cover it wherever necessary?

john-conroy (commented):
I should have included this in my PR, but we need to make `mapped_metadata` potentially undefined in the `Donor` type.

austenem:
> Just a few minor tweaks, great work otherwise! Also, would it be possible to add this logic to the `MetadataTable` component as well to make sure we cover it wherever necessary?

I can definitely do that, but the metadata table already has this language in the top info banner - I'd be in favor of replacing that with an inline tooltip like the rest of the messaging, or including both. @tsliaw?

tsliaw:
@austenem I would be fine with a tooltip for the metadata table to keep it consistent. Let's remove the messaging in the info textbox. I think the messaging appears in enough places to be clear about what's going on in this scenario.

context/app/static/js/components/detailPage/MetadataTable/MetadataTable.tsx
```diff
@@ -32,7 +34,13 @@ function MetadataTable({ tableRows = [] as MetadataTableRow[], columns = default
             {tableRows.map((row) => (
               <TableRow key={row.key}>
                 <IconTooltipCell tooltipTitle={row?.description}>{row.key}</IconTooltipCell>
-                <TableCell>{row.value}</TableCell>
+                <TableCell>
+                  <Stack direction="row">
+                    {row.value}
+                    {row.key === 'age_value' ||
+                      (row.key === 'donor.age_value' && <DonorAgeTooltip donorAge={row.value} />)}
```
austenem, line 41: These were the two matching values for this row I was able to find - not sure if we have all of the expected keys for the metadata table somewhere?

john-conroy: It seems like this logic is nested in one of the utils and doesn't reflect the es doc. You could use `endsWith`?

austenem: Good idea!

context/app/static/js/components/searchPage/ResultsTable/utils.jsx
```diff
@@ -48,6 +51,14 @@ function getByPath(hitSource, field) {
   }
 
   if (Array.isArray(fieldValue)) {
+    if (field?.id === 'mapped_metadata.age_value') {
+      return (
+        <Stack direction="row">
+          {fieldValue.join(' / ')}
+          <DonorAgeTooltip donorAge={fieldValue.join(' / ')} />
```
john-conroy, line 58: Will this be able to be converted with `Number`?

austenem: Yes, because the `fieldValue` array only contains one element - but it might be a bit safer to take only the first element of `fieldValue`.

john-conroy (approved):
Thanks!

PR #3558, opened 2024-10-07
NickAkhmetov/CAT-881 deduplicate errored processed datasets' urls

NickAkhmetov:
## Summary

This PR adds logic to prevent duplicate URL anchors on unified dataset pages that have multiple errored runs, such as https://portal.hubmapconsortium.org/browse/dataset/9e9085a732ae35d4a46b68d86167f475?redirected=True#summary-salmon%2barchr%2bmuon-error

Error datasets are the only duplicate state we've observed so far in the system, but I have implemented this such that we can extend this logic at will in the future.

https://hms-dbmi.atlassian.net/browse/CAT-881

## Testing

Unit tests have been added for the related utility function, including a case specifically covering duplicate dataset statuses.

## Screenshots/Video

Sample URL after changes:
image

## Checklist

- [x] All existing tests pass

### DIFF
--- a/context/app/static/js/pages/Dataset/utils.ts
+++ b/context/app/static/js/pages/Dataset/utils.ts
@@ -1,5 +1,7 @@
 import { ProcessedDatasetInfo } from './hooks';
 
+const potentialDuplicateStates = ['error'];
+
 export function datasetSectionId(
   dataset: Pick<ProcessedDatasetInfo, 'hubmap_id' | 'status'> & { pipeline?: string },
   prefix = '',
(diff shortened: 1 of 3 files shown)


context/app/static/js/pages/Dataset/utils.ts
```diff
@@ -1,11 +1,16 @@
 import { ProcessedDatasetInfo } from './hooks';
 
+const potentialDuplicateStates = ['error'];
+
 export function datasetSectionId(
-  dataset: Pick<ProcessedDatasetInfo, 'pipeline' | 'hubmap_id' | 'status'>,
+  dataset: Pick<ProcessedDatasetInfo, 'hubmap_id' | 'status'> & { pipeline?: string },
```
austenem, line 6: Is there a new potential use case where a `pipeline` is not given, or is this for testing purposes only?

austenem (approved):
Looks good to me, just one question. Tests look thorough as well!

context/app/static/js/pages/Dataset/utils.ts
NickAkhmetov: I came across some processed dataset descendants in new/error states that did not have pipeline info on dev during the initial implementation of unified views, so I figured that it might be good to document this possibility in the type signature of this utility to make sure that isn't lost. We already had the `??` logic in place to fall back from the pipeline to the hubmap id if the pipeline info was missing.

john-conroy (approved):

context/app/static/js/pages/Dataset/utils.ts
austenem: Got it!

PR #3570, opened 2024-10-11
NickAkhmetov/CAT-925 ensure visualizations for support image pyramid datasets use parent dataset UUID for notebooks

NickAkhmetov:
## Summary

This PR adjusts the logic for the "download jupyter notebook" button so that it uses the parent UUID instead of the support dataset's UUID.

https://hms-dbmi.atlassian.net/browse/CAT-925

## Testing

Manually tested with `/browse/dataset/8690897fced9931da34d66d669c1d698` - was able to successfully download notebook with parent HBM ID as its name.

Regression tested with `/browse/dataset/3d14dcc3d7c3e0cd339c9366e34b37c7` to confirm it was still working as expected with non-Support datasets.

## Screenshots/Video

Downloads kicked off from the pages listed above:

image

## Checklist

- [x] All existing tests pass

## Additional Notes
This endpoint is probably in need of some revisions. It appears to be left over from a previous implementation of workspaces - i.e. there's some workspaces-related language included which might not be relevant anymore, since I believe those notebooks are now fetched from the templates API. We will also likely need to revisit this for EPICs, once we have visualizations ready for them.

I also noticed that the version of vitessce in the created notebook was `3.2.8`, which matches the  - we may want to bump the version that `portal-vis` depends on sometime soon to make sure we don't fall behind the functionalities provided upstream.


context/app/static/js/components/detailPage/visualization/VisualizationWrapper/VisualizationWrapper.tsx
```diff
@@ -38,6 +38,19 @@ function VisualizationWrapper({
...
+    if (Array.isArray(vitData)) {
+      const vitDataArray = vitData as object[];
+      const found = vitDataArray.find((data) => 'parentUuid' in data) as { parentUuid: string } | undefined;
+      return found?.parentUuid;
+    }
+    if ('parentUuid' in vitData) {
+      return (vitData as { parentUuid: string }).parentUuid;
+    }
+    return undefined;
+  }, [vitData]);
```
austenem, line 52: Since `vitData` is passed to `Visualization`, could this be calculated there instead of being passed as an extra prop?

austenem (commented):
Looks great, just one comment!

context/app/static/js/components/detailPage/visualization/VisualizationWrapper/VisualizationWrapper.tsx
NickAkhmetov: Good call! No need to have the extra layer of prop drilling.

austenem (approved):
Looks great!

PR #3578, opened 2024-10-22
Austenem/CAT-957 Fix shaking viewport

austenem:
## Summary

Fixes bug where scrolling to the top of a processed dataset section led to a shaking viewport.

**Explanation:** The processed dataset helper panel appears when any part of the processed dataset accordion is visible. In narrower browser windows, opening the helper panel means there is less width for the accordion, so it is moved further down the page. The panel then closes, which causes the accordion to be visible, which opens it again and triggers an infinite loop.

This was resolved by only showing the helper panel when 10% or more of the accordion is visible, and only hiding the panel when the accordion is completely off-screen.

[CAT-957 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-954?atlOrigin=eyJpIjoiYWQ2NTkzNGI2MzVkNGJjYWE4MGM0MDBkZWE4MTg1ZWQiLCJwIjoiaiJ9)

## Testing

Attempted to reproduce bug in Chrome/Safari on browsers of varying widths on pages with one or several processed datasets.

## Screenshots/Video

<details>
<summary>Original bug on prod:</summary>

https://github.com/user-attachments/assets/fc621103-d8a6-4d48-85f6-200e537774f1

</details>

<details>
<summary>Local:</summary>

https://github.com/user-attachments/assets/11f0bda5-c44d-418b-a869-e38235467638

</details>

## Checklist

- [X] All existing tests pass

### DIFF
--- a/CHANGELOG-fix-shaking-viewport.md
+++ b/CHANGELOG-fix-shaking-viewport.md
@@ -1,0 +1,1 @@
+- Fix bug where expanding a processed dataset section leads to a shaking viewport.
--- a/context/app/static/js/components/detailPage/ProcessedData/ProcessedDataset/ProcessedDatasetAccordion.tsx
+++ b/context/app/static/js/components/detailPage/ProcessedData/ProcessedDataset/ProcessedDatasetAccordion.tsx
@@ -18,6 +18,7 @@ import useProcessedDataStore from '../store';
 import { DetailPageSection } from '../../DetailPageSection';
 
 const iconPlaceholder = <Skeleton variant="circular" width={24} height={24} animation="pulse" />;
+const inViewThreshold = 0.1;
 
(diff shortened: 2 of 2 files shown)


NickAkhmetov (approved):
Clever fix, nicely caught!

I am curious if it's possible to animate the changing dimensions for the page content so the panel doesn't pop in as abruptly, but it's definitely on the "nice to have" side of things rather than a blocker.

austenem:
> Clever fix, nicely caught!
> 
> I am curious if it's possible to animate the changing dimensions for the page content so the panel doesn't pop in as abruptly, but it's definitely on the "nice to have" side of things rather than a blocker.

That's a good idea! I'll look into that as a future improvement.

PR #3580, opened 2024-10-23
Austenem/CAT-959 Update epic labels

austenem:
## Summary

Use `assay_display_name` instead of `pipeline` when labelling EPIC dataset sections, which do not have `pipeline` values.

[CAT-959 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-959?atlOrigin=eyJpIjoiN2RlZjUzNWU0Y2Q5NDIxNjg1ZDYzZTllOTc5ZDBkZTQiLCJwIjoiaiJ9)

## Testing

Tested with the Segmentation Mask EPIC currently on dev: http://localhost:5001/browse/dataset/22901da5f080b219a514e38381acbb0e?redirected=True&redirectedFromId=HBM474.PFWF.998#section-hbm474.pfwf.998-qa

## Screenshots/Video

Test instance:
Screenshot 2024-10-23 at 12 45 47 PM

Local:
Screenshot 2024-10-23 at 12 09 57 PM

## Checklist

- [X] All existing tests pass


context/app/static/js/components/detailPage/ProcessedData/HelperPanel/HelperPanel.tsx
```diff
@@ -80,7 +80,9 @@ function HelperPanelBody() {
           {currentDataset.description}
         </HelperPanelBodyItem>
       )}
-      <HelperPanelBodyItem label="Pipeline">{currentDataset.pipeline}</HelperPanelBodyItem>
+      <HelperPanelBodyItem label="Pipeline">
```
john-conroy, line 83: I think you need to update the label as well?

john-conroy (changes requested):

context/app/static/js/components/detailPage/ProcessedData/HelperPanel/HelperPanel.tsx
austenem: Are we using different terminology for the EPIC processing types?

john-conroy: @tsliaw what should the label be for EPIC type?

tsliaw: I'm not sure, pipeline doesn't seem completely accurate. Maybe `Processing Type` `Externally Processing Type` something like that, but I'm really not sure. @lchoy Do you have any insights for what this label should be?

lchoy: My suggestion is "Analysis" or "Analysis Type" so as to work for all cases - those datasets that underwent pipeline processing and epics that are "transformed".

I had added a comment in the Jira: https://hms-dbmi.atlassian.net/browse/CAT-959?focusedCommentId=60307, before I came across this thread here.

lchoy: The Processed Data Summary section should be updated too, as it lists "Pipelines" and is showing "unknown" for epics.

austenem: Updated language:
Screenshot 2024-10-24 at 10 46 37 AM
Screenshot 2024-10-24 at 10 47 06 AM

context/app/static/js/components/detailPage/ProcessedData/hooks.spec.ts
```diff
@@ -1,13 +1,13 @@
...
-import { useSortedSearchHits, createdByCentralProcess, datasetIsPublished, usePipelineCountsInfo } from './hooks';
+import { useSortedSearchHits, createdByCentralProcess, datasetIsPublished, useAnalysesCountInfo } from './hooks';
 
 const testDatasets: {
   _id: string;
   _index: string;
   _score: number;
   _source: {
-    assay_display_name: string[];
+    assay_display_name: string;
```
NickAkhmetov, line 10: I think we should leave these types as is - in HelperPanel we're accessing `assay_display_name[0]`, so having these be consistent is preferable.

austenem: All of these functions get their prop types from the `ProcessedDatasetInfo` interface, which defines the `assay_display_name` as a string - I think it makes sense to update these tests accordingly, but I also see your point about accessing indices in `HelperPanel`. I updated that function so that it treats `assay_display_name` as a string.

austenem: Looking back on this because it is continuing to cause issues with testing elsewhere on the detail pages - it might make more sense to instead update the `ProcessedDatasetInfo` interface to have a string array for `assay_display_name`, so that it is more accurate to the query results. I implemented this in #3582, along with making the `pipeline` field optional (as it is absent from the new EPICs).

context/app/static/js/components/detailPage/ProcessedData/hooks.ts
```diff
@@ -40,26 +40,26 @@ export function useSortedSearchHits(datasets: ReturnType<typeof useProcessedData
...
-      acc[pipeline] = (acc[pipeline] || 0) + 1;
+export function useAnalysesCountInfo(datasets: Pick<ProcessedDatasetInfo, 'pipeline' | 'assay_display_name'>[]) {
+  const analyses = datasets.map((dataset) => dataset.pipeline ?? dataset.assay_display_name);
+  const analysesCount = analyses.reduce(
+    (acc, analysis) => {
+      acc[analysis] = (acc[analysis] || 0) + 1;
       return acc;
     },
     {} as Record<string, number>,
   );
```
john-conroy, line 55: You could bring the logic from the `map` into the `reduce` to avoid iterating twice. Not worth addressing by itself.

context/app/static/js/pages/Dataset/hooks.ts
```diff
@@ -146,10 +147,11 @@ function getProcessedDatasetSection({
   };
 
   const sectionsToDisplay = Object.entries(shouldDisplaySection).filter(([_k, v]) => v === true);
+  const sectionTitle = pipeline ?? assay_display_name[0] ?? hubmap_id;
```
john-conroy, line 150: In which cases will a dataset not have an `assay_display_name` is that for EPICs?

john-conroy (commented):
Just a quick question, thanks!

context/app/static/js/pages/Dataset/hooks.ts
austenem: I believe the `hubmap_id` was a safeguard in case of datasets without a `pipeline` value - I kept it because I wasn't sure whether all datasets were guaranteed to have an `assay_display_name`. If they are, then this can be removed.

john-conroy: Ok great, it's fine to leave here I was more so curious if you ever intended for it to return the `hubmap_id`.

john-conroy (approved):

PR #3593, opened 2024-11-01
NickAkhmetov/CAT-969 Increase homepage counts query specificity

NickAkhmetov:
## Summary

This PR:
- Excludes datasets in a non-QA/Published state from being counted on the homepage.
- Excludes samples/donors without a descendant dataset from being counted on the homepage.

## Design Documentation/Original Tickets

https://hms-dbmi.atlassian.net/browse/CAT-969

## Testing

I tested by comparing homepage counts while logged in on local and production environments, and confirming that the dataset total count matches a manually calculated count from the search page.

## Screenshots/Video

Localhost counts:
image

Prod counts:
image

## Checklist

- [x] Code follows the project's coding standards
  - [x] Lint checks pass locally
  - [x] New `CHANGELOG-your-feature-name-here.md` is present in the root directory, describing the change(s) in full sentences.
- [x] Unit tests covering the new feature have been added
- [x] All existing tests pass
- [x] Any relevant documentation in JIRA/Confluence has been updated to reflect the new feature
- [x] Any new functionalities have appropriate analytics functionalities added

## Additional Notes

There is a related ticket to exclude these samples/donors from the search page results (https://hms-dbmi.atlassian.net/browse/CAT-968); I did not include that change in this PR because of the pending searchkit changes, but these changes should probably be deployed in tandem to minimize confusion from the donor/sample count mismatches.

PR #3595, opened 2024-11-04
NickAkhmetov/CAT-634 Fix selection of samples on organ page when more than 500 are present

NickAkhmetov:
## Summary

The Kidney page on prod currently fails to select all results when using the "select all" checkbox of the samples table. This issue was caused by the `useAllSearchIds` hook being limited to only fetching the first 500 results. Lifting this limit/removing the "number of pages to fetch" from the ID lookup resolved this issue by enabling lookup of all 634 results in the table.

https://hms-dbmi.atlassian.net/browse/CAT-634

## Testing

Tested on the kidney page as well as other organ pages.

## Screenshots/Video

https://github.com/user-attachments/assets/345dc850-6b0c-404e-8df7-a60bb7c0b031

## Checklist

- [x] All existing tests pass

## Additional Notes

I'm not sure if this PR overlaps with the changes in the searchkit PR - it can definitely wait if so.

### DIFF
--- a/context/app/static/js/hooks/useSearchData.ts
+++ b/context/app/static/js/hooks/useSearchData.ts
@@ -234,7 +234,7 @@ export function useAllSearchIDs(
   const { elasticsearchEndpoint, groupsToken } = useAppContext();
 
   const { searchData } = useSearchData(
-    { ...query, track_total_hits: true, size: 0, ...sharedIDsQueryClauses },
+    { ...query, track_total_hits: true, size: 10000, ...sharedIDsQueryClauses },
     {
       useDefaultQuery,
       fetcher,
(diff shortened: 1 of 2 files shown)


context/app/static/js/hooks/useSearchData.ts
```diff
@@ -243,21 +243,14 @@ export function useAllSearchIDs(
   );
 
   const totalHitsCount = getTotalHitsCount(searchData);
-  const numberOfPagesToRequest = totalHitsCount ? Math.ceil(10000 / totalHitsCount) : undefined;
```
john-conroy, line 246: Does this continue to handle > 10k results?

NickAkhmetov: That is definitely worth testing to be safe in the long run - I can test it out with a query for all samples without any filters and confirm whether it works as expected.

The 10k results limitation is definitely worth keeping in mind in general, but I am curious - per #3593, it seems like we aren't as close to the 10k result threshold as we thought once we filter samples without descendant datasets, so maybe this could be a premature optimization?

john-conroy: We have ~7k datasets with `HuBMAP Read`.

NickAkhmetov:
Following John's feedback, I've made some improvements to the `fetchAllIds` functionality, using an async generator to sequentially request all the IDs for a given search request's results.

On login with HuBMAP Read, there are >20,000 kidney samples, so I was able to confirm the fix:

https://github.com/user-attachments/assets/3be82ec0-05f5-41b3-96fe-abc6c7234fc0

It is worth noting that adding the selected to a list caused the page to freeze. This might become a bigger issue if we start persisting lists in a key-value store.

john-conroy (approved):
Looks good to me! We could likely avoid the additional `forEach` call, but not a big deal. I wasn't aware of the `_` convention in numbers for readability so thanks for teaching me that!

PR #3600, opened 2024-11-07
NickAkhmetov/CAT-976 Visualization link in table of contents

NickAkhmetov:
## Summary

This PR adjusts the logic for whether or not to display the `visualization` link in the table of contents.

The current logic just checks whether there is a loaded conf in the SWR cache; this approach works well for fast-running configs (such as self-contained processed datasets) but runs into issues with configs that are slower to build (such as image pyramids with multiple slides or other multi-conf cases like 10X).

The updated approach adds a heuristic condition to check whether the dataset is an image pyramid; since image pyramids always have visualizations, this is a safe assumption to make.

https://hms-dbmi.atlassian.net/browse/CAT-976

## Testing

Tested with image pyramid datasets and 10X datasets to confirm the visualization link shows up when expected. Tested with non-image pyramid processed datasets to confirm the visualization link does not show up when one is not present.

## Screenshots/Video

image

## Checklist

- [x] All existing tests pass

## Additional Notes

An alternative way to approach this would be to mark the image pyramids themselves as having a visualization, which would make the boolean check in line 162 sufficient and allow us to remove the dependency on the SWR cache; this solution would require revising search indexing to mark image pyramids as having a visualization.

### DIFF
--- a/CHANGELOG-cat-976.md
+++ b/CHANGELOG-cat-976.md
@@ -1,0 +1,1 @@
+- Use `Image Pyramid` pipeline name as a heuristic for whether a visualization link should be present in unified dataset table of contents.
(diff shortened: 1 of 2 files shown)


austenem (approved):
Looks good to me!

john-conroy (approved):

PR #3604, opened 2024-11-13
Austenem/CAT-983 MVP Bulk File Download

austenem:
## Summary

First iteration of the Bulk Download dialog, which adds a dialog to the Search, Publication, and Collection pages that generates a download manifest for files from selected datasets. An option to download the corresponding dataset metadata is also included.

In this iteration, all files from selected datasets must be downloaded; in an upcoming iteration, the dialog will include an "Advanced Selection" accordion that will allow users to select specific file types from centrally processed datasets.

[CAT-983 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-983?atlOrigin=eyJpIjoiM2VjZTFkNjI3YWI2NDQwODk5MGI3MjEyMzlmYmZhNjAiLCJwIjoiaiJ9)

[Figma mockups](https://www.figma.com/design/bHP9yKluRqoINaBVc6tTQ9/Bulk-File-Download?node-id=0-1&t=lzxJlo9OAtWhnfpd-1)

## Testing

Tested the following cases:

- General
    - Dialog is available for authenticated and non-authenticated users
    - Download manifest is formatted appropriately for CLT tool and contains relevant datasets
    - Metadata download option works and resulting file contains relevant datasets
    - One or more download options must be selected before download is permitted

- Search page
    - **Select one or more dataset from each of the download options** (centrally processed, externally processed, raw) - should show one download option, and no "select all" option
    - **Select datasets from all three download options** - should show all download options and a "select all" option
    - **Select protected datasets** - should show warning banner and not allow a download until the datasets are removed. If only protected datasets were selected and subsequently removed, a "no files to download" warning banner should appear.
    - **Select no datasets** - should include all datasets based on the current filters

- Collection pages
    - Open dialog and download datasets from table

- Publication pages
    - Open dialog and download datasets from table

## Screenshots/Video

<details>
<summary>Search</summary>

Screenshot 2024-11-13 at 3 18 32 PM

https://github.com/user-attachments/assets/b0956a80-040e-495d-ac08-fdbf32f40642

</details>

<details>
<summary>Collections</summary>

Screenshot 2024-11-13 at 3 16 46 PM

</details>

<details>
<summary>Publications</summary>

Screenshot 2024-11-13 at 3 14 18 PM

</details>

<details>
<summary>Error cases</summary>

No available files:

Screenshot 2024-11-13 at 3 35 10 PM

Protected datasets selected:

Screenshot 2024-11-13 at 3 48 59 PM

Download failure:

Screenshot 2024-11-13 at 3 39 16 PM

</details>

## Checklist

- [X] All existing tests pass


context/app/static/js/components/bulkDownload/bulkDownloadFormFields.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from 'zod';
+import { withCustomMessage } from 'js/helpers/zod/withCustomMessage';
+
+const bulkDownloadOptionsField = {
+  bulkDownloadOptions: z
+    .array(z.string(), {
+      errorMap: withCustomMessage('My special error message'),
```
NickAkhmetov, line 7: This seems like it might be a placeholder?

context/app/static/js/components/workspaces/formHooks.ts
```diff
@@ -37,37 +16,32 @@ const errorHelper = {
...
-  }
-  const protectedHubmapIds = protectedRows?.map((row) => row?._source?.hubmap_id).join(', ');
+    return errorHelper.protectedDatasets(protectedRows);
+  }, []);
 
-  const removeProtectedDatasets = useCallback(() => {
-    deselectRows(protectedRows.map((r) => r._id));
-    toastSuccessRemoveProtectedDatasets();
-  }, [deselectRows, protectedRows, toastSuccessRemoveProtectedDatasets]);
+  const { ...rest } = useProtectedDatasetsForm({
```
NickAkhmetov, line 37: We could probably return the result of the hook instead:

```suggestion
  return useProtectedDatasetsForm({
```

context/app/static/js/components/workspaces/formHooks.ts
```diff
@@ -37,37 +16,32 @@ const errorHelper = {
...
+    selectedRows,
+    deselectRows,
+    protectedDatasetsErrorMessage,
+    trackEventHelper,
+  });
 
-  return { errorMessages, protectedHubmapIds, removeProtectedDatasets, protectedRows, selectedRows };
+  return { ...rest };
```
NickAkhmetov, line 44: In conjunction with the previous suggestion:
```suggestion
```

context/app/static/js/components/bulkDownload/hooks.ts
```diff
@@ -0,0 +1,199 @@
...
+    errors,
+    reset,
+    trigger,
+  };
+}
+
+interface BulkDownloadDialogProps {
+  deselectRows?: (uuids: string[]) => void;
+}
+function useBulkDownloadDialog({ deselectRows }: BulkDownloadDialogProps) {
```
NickAkhmetov, line 60: Perhaps this could work to avoid the empty object prop required to use this without providing a `deselectRows` callback?

```suggestion
function useBulkDownloadDialog(deselectRows?: (uuids: string[]) => void;) {
```

Alternatively, if you foresee potentially adding other props to this hook, the empty object could be a default value instead:

```suggestion
interface BulkDownloadDialogProps {
  deselectRows?: (uuids: string[]) => void;
}
function useBulkDownloadDialog(props: BulkDownloadDialogProps = {}) {
```

context/app/static/js/components/bulkDownload/hooks.ts
```diff
@@ -0,0 +1,199 @@
...
+  const { isOpen, uuids, open, close, setUuids } = useBulkDownloadStore();
+  const { control, handleSubmit, errors, reset, trigger } = useBulkDownloadForm();
+  const { toastErrorDownloadFile, toastSuccessDownloadFile } = useBulkDownloadToasts();
+
+  // Fetch datasets for the selected uuids
+  const datasetQuery = {
+    query: getIDsQuery([...uuids]),
+    _source: ['hubmap_id', 'processing', 'uuid', 'files', 'processing_type'],
+    size: 1000,
+  };
```
NickAkhmetov, line 70: It seems potentially worthwhile to put this into a `useMemo` to avoid recalculating it unless the uuids change.

This raised some other questions for me: 

- Is there an upper bound to how many datasets can be selected for bulk download at once? 
	- If so, are we communicating this to users? 
- If the limit is higher than 1,000, could we increase the size to `10_000` to handle selections of >1,000 datasets? We could also use `size: uuids.size()`.

context/app/static/js/components/bulkDownload/hooks.ts
```diff
@@ -0,0 +1,199 @@
...
+    (uuidsToRemove: string[]) => {
+      if (deselectRows) {
+        deselectRows([...uuidsToRemove]);
+      }
+      setUuids(new Set([...uuids].filter((uuid) => !uuidsToRemove.includes(uuid))));
+    },
+    [deselectRows, setUuids, uuids],
+  );
+
+  const { ...protectedDatasetsFields } = useProtectedDatasetsForm({
```
NickAkhmetov, line 91: The extra destructuring here doesn't seem necessary, we can reintroduce it in the future if we need to extract a specific field and keep it separate from the `protectedDatasetsFields` object. 

```suggestion
  const protectedDatasetsFields = useProtectedDatasetsForm({
```

context/app/static/js/components/bulkDownload/hooks.ts
```diff
@@ -0,0 +1,199 @@
...
+  const downloadOptions = ALL_BULK_DOWNLOAD_OPTIONS.map((option) => ({
+    ...option,
+    count: datasets.filter((dataset) => option.isIncluded(dataset)).length,
+  })).filter((option) => option.count > 0);
+
+  // Remove selected uuids from the list and deselect them in the table if needed
+  const removeUuidsOrRows = useCallback(
+    (uuidsToRemove: string[]) => {
+      if (deselectRows) {
+        deselectRows([...uuidsToRemove]);
```
NickAkhmetov, line 84: I think we don't need to destructure in this case since `uuidsToRemove` is already an array.

```suggestion
        deselectRows(uuidsToRemove);
```

NickAkhmetov (commented):
Great work! Some minor notes to begin; I will pull the branch down to test more thoroughly tomorrow.

context/app/static/js/components/bulkDownload/bulkDownloadFormFields.ts
austenem: Sure is, good catch! :)

context/app/static/js/components/bulkDownload/hooks.ts
austenem: Good idea!

context/app/static/js/components/bulkDownload/hooks.ts
austenem: This is a good point and I like the suggestion of using `size: uuids.size()` - in regard to an upper bound, we haven't discussed this as far as I know. The load time for the dialog is definitely noticeable after a few hundred datasets, so it seems like some additional messaging should be provided to users who have selected a large number of datasets (maybe an alert similar to the one used in the workspaces dialog) letting them know about the wait time. @tsliaw what do you think?

NickAkhmetov:
One minor issue I noticed while testing: the "download" button as soon as the page loads shows a "files are not available" view:

https://github.com/user-attachments/assets/70c28e59-eba6-4b48-a968-55d07578a57c

Perhaps we could disable the "download" button and show a loading message on hover if
* All UUIDS have not yet loaded, and
* There is no selection

austenem:
> One minor issue I noticed while testing: the "download" button as soon as the page loads shows a "files are not available" view:
> 
>  Screen.Recording.2024-11-14.134655.mp4 
> 
> Perhaps we could disable the "download" button and show a loading message on hover if
> 
> * All UUIDS have not yet loaded, and
> * There is no selection

Good idea! It's now disabled during that initial loading period.

context/app/static/js/components/bulkDownload/BulkDownloadDialog/BulkDownloadDialog.tsx
```diff
@@ -0,0 +1,224 @@
...
+import RelevantPagesSection from 'js/shared-styles/sections/RelevantPagesSection';
+import LabelledSectionText from 'js/shared-styles/sections/LabelledSectionText';
+import { Alert } from 'js/shared-styles/alerts';
+import ErrorOrWarningMessages from 'js/shared-styles/alerts/ErrorOrWarningMessages';
+
+function DownloadDescription() {
+  return (
+    <SectionDescription>
+      <Stack spacing={2}>
+        <Box>
```
john-conroy, line 29: Should these be `Typography` instead of `Box`?

context/app/static/js/components/bulkDownload/BulkDownloadDialog/BulkDownloadDialog.tsx
```diff
@@ -0,0 +1,224 @@
...
+      />
+    </Stack>
+  );
+}
+
+function DownloadOptionsDescription() {
+  return (
+    <SectionDescription>
+      <Stack spacing={2}>
+        <Box>Select raw and/or processed files to download.</Box>
```
john-conroy, line 85: Typography here as well?

context/app/static/js/components/bulkDownload/BulkDownloadDialog/BulkDownloadDialog.tsx
```diff
@@ -0,0 +1,224 @@
...
+    );
+  }
+
+  if (downloadOptions.length === 0) {
+    return (
+      <Box paddingTop={1}>
+        <Alert severity="warning">
+          <Typography>Files are not available for any of the selected datasets.</Typography>
+        </Alert>
+      </Box>
```
john-conroy, line 138: Do we need the `Box` is the parent a flex container?

context/app/static/js/components/bulkDownload/BulkDownloadDialog/BulkDownloadDialog.tsx
```diff
@@ -0,0 +1,224 @@
...
+    );
+  }
+
+  return (
+    <Box>
```
john-conroy, line 143: Same question as above regarding the `Box`.

context/app/static/js/components/bulkDownload/BulkDownloadOptionsField/BulkDownloadOptionsField.tsx
```diff
@@ -0,0 +1,81 @@
...
+        {/* All download options */}
+        <Stack paddingLeft={multipleOptionsAvailable ? 4 : 0}>
+          {downloadOptions.map(({ key, label, count }) => (
+            <FormControlLabel
+              key={key}
+              control={
+                <Checkbox
+                  size="small"
+                  checked={(field.value as string[]).includes(key)}
+                  onChange={(e) => {
```
john-conroy, line 64: We could pull the `onChange` functions out into callbacks with `useCallback`.

context/app/static/js/components/bulkDownload/hooks.ts
```diff
@@ -0,0 +1,196 @@
...
+    _source: ['hubmap_id', 'processing', 'uuid', 'files', 'processing_type'],
+    size: 1000,
+  };
+  const { searchHits, isLoading } = useSearchHits<BulkDownloadDataset>(datasetQuery);
+  const datasets = searchHits.map(({ _source }) => _source);
+
+  // Which options to show in the dialog
+  const downloadOptions = ALL_BULK_DOWNLOAD_OPTIONS.map((option) => ({
+    ...option,
+    count: datasets.filter((dataset) => option.isIncluded(dataset)).length,
```
john-conroy, line 74: We could potentially push this logic into the search-api call by using a composite aggregations query with `processing` and `processing_type`. Or we could make separate requests to get the datasets for each option.

context/app/static/js/components/bulkDownload/hooks.ts
```diff
@@ -0,0 +1,196 @@
...
+    close();
+  }, [close, reset]);
+
+  const onSubmit = useCallback(
+    ({ bulkDownloadOptions, bulkDownloadMetadata }: BulkDownloadFormTypes) => {
+      const datasetsToDownload = datasets.filter((dataset) =>
+        bulkDownloadOptions.some((option) =>
+          ALL_BULK_DOWNLOAD_OPTIONS.find(({ key }) => key === option)?.isIncluded(dataset),
+        ),
+      );
```
john-conroy, line 150: This could also be avoided by separately requesting each option? Or maybe at the least we can iterate over the datasets once to separate them to avoid doing it here and counting them above?

john-conroy (commented):
Looks great! A couple of thoughts as to how we could avoid iterating over the datasets as many times which is likely contributing to the slowness.

context/app/static/js/components/bulkDownload/hooks.ts
john-conroy: @NickAkhmetov thoughts?

context/app/static/js/components/bulkDownload/hooks.ts
austenem: @john-conroy It's server side - once the search hits are returned the dialog loads almost immediately.

austenem: @tsliaw @john-conroy I tried timing the wait time and it varies pretty substantially depending on what filters are applied - I'm guessing this is because of how the datasets are indexed?

For filters that include ~1000 datasets, the wait time is < 2 seconds for filters on dataset type, status, and analyte class, ~30 seconds for filters on sample category, and just over two minutes for filters on organs.

context/app/static/js/components/bulkDownload/BulkDownloadDialog/BulkDownloadDialog.tsx
austenem: The `Box` is needed to prevent the `spacing` style from the parent `Stack` from applying to the children of the `Step` component.

context/app/static/js/components/bulkDownload/BulkDownloadDialog/BulkDownloadDialog.tsx
austenem: The `Box` here is needed to prevent the `spacing` style from the parent `Stack` from applying to the children of the `Step` component.

context/app/static/js/components/bulkDownload/BulkDownloadDialog/BulkDownloadDialog.tsx
austenem: The `Box` was to apply the `paddingTop` styling somewhere since `sx` doesn't work on `Alerts` - however it looks like the padding is actually unnecessary here, so I'll remove it.

context/app/static/js/components/bulkDownload/hooks.ts
austenem: Good idea!

john-conroy (approved):
Nice catch removing 'files' from the source!

PR #3620, opened 2024-11-25
Linting and type fixing

tkakar:
## Summary
This PR addresses the accessibility issue of `Ensuring links having discernible text`  by adding aria-labels to any links used in the code base, particularly, the tiles and stacked barchart components.

[CAT-1029](https://hms-dbmi.atlassian.net/browse/CAT-1029)

## Testing
Manually tested with Axe DevTools

## Screenshots/Video

Screenshot 2024-11-25 at 10 53 00 AM

Include screenshots or video demonstrating any significant visual or behavioral changes.

## Checklist

- [ ] Code follows the project's coding standards
  - [ ] Lint checks pass locally
  - [ ] New `CHANGELOG-your-feature-name-here.md` is present in the root directory, describing the change(s) in full sentences.
- [ ] All existing tests pass
- [ ] Any relevant documentation in JIRA/Confluence has been updated to reflect the new feature

## Additional Notes

Please specify any additional information or context relevant to this PR.

### DIFF
--- a/context/app/static/js/shared-styles/charts/StackedBar/StackedBar.tsx
+++ b/context/app/static/js/shared-styles/charts/StackedBar/StackedBar.tsx
@@ -1,5 +1,7 @@
 import React, { SVGProps } from 'react';
 
+import LZString from 'lz-string';
+import { FiltersType } from 'js/components/search/store';
 import { StyledRect } from './style';
 
 type Direction = 'vertical' | 'horizontal';
@@ -55,8 +57,31 @@ function StackedBar({ direction = 'vertical', bar, hoverProps, href }: StackedBa
     <StyledRect fill={bar.color} {...mappedProps} $showHover={Boolean(hoverProps) || Boolean(href)} {...hoverProps} />
   );
   if (href) {
+    const page = href.split('?')[0].replace('/search/', '');
+    const encodedURI = href.split('?')[1];
+    interface searchObj {
+      filters: FiltersType;
+    }
+
+    const decodedHref = JSON.parse(LZString.decompressFromEncodedURIComponent(encodedURI)) as searchObj;
+
+    // Extracting the filters from the searchURI
+    const extractedValues = (() => {
+      const results: string[] = [];
+      Object.values(decodedHref.filters).forEach(({ values }) => {
+        if (Array.isArray(values)) {
+          results.push(...values);
+        } else if (typeof values === 'object' && values !== null) {
+          results.push(...Object.keys(values));
+        }
+      });
+      return results;
+    })();
+
+    const labelString = `${page} page for the selected bar representing ${extractedValues.join(', ')}`;
+
     return (
-      <a href={href} target="_parent">
+      <a href={href} target="_parent" aria-label={labelString}>
         {rect}
       </a>
     );
--- a/context/app/static/js/shared-styles/tiles/Tile/Tile.tsx
+++ b/context/app/static/js/shared-styles/tiles/Tile/Tile.tsx
@@ -33,7 +33,11 @@ function Tile({
     </StyledPaper>
   );
   if (href) {
-    return <a href={href}>{tile}</a>;
+    return (
+      <a href={href} aria-label={href}>
+        {tile}
+      </a>
+    );
   }
   return tile;
 }
--- a/CHANGELOG-links-with-discernible-text.md
+++ b/CHANGELOG-links-with-discernible-text.md
@@ -1,0 +1,1 @@
+- Added discernible text to the links on stacked barcharts.


context/app/static/js/shared-styles/tiles/Tile/Tile.tsx
```diff
@@ -33,7 +33,11 @@ function Tile({
     </StyledPaper>
   );
   if (href) {
-    return <a href={href}>{tile}</a>;
+    return (
+      <a href={href} aria-label={href}>
+        {tile}
```
tkakar, line 38: Not sure if we want to parse the hrefs in this case by extracting keywords from them. But it would need some kind of mapping, easy for organs, difficult for datasets page with long uuids that are not meaningful

context/app/static/js/shared-styles/charts/StackedBar/StackedBar.tsx
```diff
@@ -55,8 +57,31 @@ function StackedBar({ direction = 'vertical', bar, hoverProps, href }: StackedBa
...
   );
   if (href) {
+    const page = href.split('?')[0].replace('/search/', '');
+    const encodedURI = href.split('?')[1];
+    interface searchObj {
+      filters: FiltersType;
+    }
+
+    const decodedHref = JSON.parse(LZString.decompressFromEncodedURIComponent(encodedURI)) as searchObj;
+
```
tkakar, line 67: I took a first pass on this, this extracts the filter (assaytype, etc) and organ, we can also include the count (shown on hover) but not sure if that's important as hover does provide that info?
Also, would be happy to move it to some utils file, or adding more meaningful content to the aria-label

NickAkhmetov: Perhaps we could pass the aria-label text generation logic to the stacked bar chart, then directly pass the label as a prop here? That seems simpler than reverse engineering the aria label from the href, and will work for non-search page links as well.

context/app/static/js/shared-styles/tiles/Tile/Tile.tsx
NickAkhmetov: Ditto here - I think constructing a descriptive text would be more straightforward if we were to define that logic from the parent element that uses these tiles.

context/app/static/js/shared-styles/charts/hooks.ts
```diff
@@ -154,6 +155,21 @@ function useOrdinalScale(domain: string[], config: Omit<OrdinalScaleConfig<strin
...
+    }
+    const extractedValues = Object.entries(tooltipData.bar.data)
+      .filter(([, value]) => typeof value === 'string')
+      .map(([key, value]) => [key, value as string]);
+    const extractedValuesString = extractedValues.join('  ');
+    const count = tooltipData.bar.data[tooltipData.key] ?? '';
+
+    return `Bar represents ${extractedValuesString} with count ${String(count)}`;
+  }, [tooltipData]);
+};
```
NickAkhmetov, line 171: This is a good iteration over the first approach, but there's still a few things that stand out to me:

1. Since this is based on the currently hovered tooltip data:
  a. If there is no hovered bar, every bar shows up with the `Bar represents` text and nothing else. This is also encountered when hovering tick labels.
image
  b. When a bar is hovered, every bar gets the same aria label: 
image

2. The fallthrough to an incomplete label (`"Bar represents"`) is problematic since it makes it sound like the screenreader stopped reading halfway through; this may frustrate users even more than a blank label.

3. The approach of iterating over all bar data may lead to irrelevant or unformatted fields being pulled in to the label if they are not removed from the input data. Additionally, even fields which are relevant might still be read wrong - e.g. `Bar represents datasetType,Slide-seq with count 46` might not be read correctly due to the `datasetType` word.

---

I would suggest trying an alternative approach: each stacked bar chart could be provided with a function prop which takes in the same generic `Datum` being used as the type constraint for the chart overall and returns a string. This would allow us to define well-formatted labels which will be guaranteed to be relevant to the situation at hand, e.g. adding something like this to the end of `HorizontalStackedBarChartProps` and `VerticalStackedBarChartProps`, similar to how we provide `getBarHref`:

```
  getAriaLabel?: (d: TooltipData<Datum>) => string;
``` 

As an example of how we'd be able to apply this - for the homepage bar chart, we would extend the `ColorOption` array to have unique `getAriaLabel`s for each selected aggregation, e.g.

```
    {
      key: 'assay_type',
      data: assayBuckets,
      dropdownLabel: 'Assay Type',
      keys: getKeysFromAggregatedData(assayBuckets),
      getBarHref: (d) => {
        const organTerms = [String(d.bar.data.organ)];
        return getSearchURL({
          entityType: 'Dataset',
          organTerms,
          assay: d.key,
          assayTypeMap: datasetTypeMap,
        });
      },
      getAriaLabel: (d) => {
        const organ = String(d.bar.data.organ);
        const assay = d.key;
        const count = d.bar.data[d.key];
        return `${count} ${organ} datasets with ${assay} dataset type.`
      }
    },
```

This approach ensures that the same datum is used to generate both the href for the specific bar and the aria label.

context/app/static/js/components/search/Results/ResultsTiles.tsx
```diff
@@ -97,6 +97,7 @@ function Tile({ hit }: { hit: SearchHit<Partial<Entity>> }) {
       uuid={hit?._source?.uuid}
       id={hit?._source?.hubmap_id}
       entityData={hit?._source}
+      ariaLabelText={`Tile representing ${hit?._source?.entity_type} ${hit?._source?.uuid}`}
```
NickAkhmetov, line 100: I think the HBM ID would be preferable to the UUID here since the screenreader would have to read off all 32 characters of the UUID.

We may also want to be more descriptive depending on the result type - e.g. right now this will say `Tile representing Dataset blahblahblah` or `Tile representing Sample blahblahblah`, but in the future we could provide e.g. the assay type and source organ for datasets, sample category and source organ for samples, and descriptive info for donors.

context/app/static/js/components/detailPage/provenance/ProvTable/ProvTable.tsx
```diff
@@ -92,6 +92,7 @@ function ProvEntityColumnContent({
                 uuid={item.uuid}
                 id={item.hubmap_id}
                 entity_type={item.entity_type}
+                ariaLabelText={`Tile representing ${item.entity_type} ${item.uuid}`}
```
NickAkhmetov, line 95: I wonder if a `useEntityTileAriaLabel` hook could help? That would allow us to implement more complex logic depending on the content type (dataset/sample/donor).

Similar concern as below with using UUID vs. HBM ID in aria labels.

NickAkhmetov (changes requested):
Thanks for pulling in the `lang="en"` fixes!

These are some great iterations over the initial version, but I think there are still more improvements which could be made here - let me know if I can clarify any of my feedback! 😄

context/app/static/js/shared-styles/charts/hooks.ts
tkakar: Thanks for reviewing. These are some great suggestions, will incorporate them.
I was also not comfortable with the fallthrough text, but thought that the the screenReader would only read the label when the link is hovered/clicked.

context/app/static/js/components/detailPage/provenance/ProvTable/ProvTable.tsx
```diff
@@ -92,6 +102,7 @@ function ProvEntityColumnContent({
                 uuid={item.uuid}
                 id={item.hubmap_id}
                 entity_type={item.entity_type}
+                ariaLabelText={ariaLabelText[j]}
```
tkakar, line 105: This one is sorted, so fixing it now.

context/app/static/js/components/home/HuBMAPDatasetsChart/HuBMAPDatasetsChart.tsx
```diff
@@ -93,6 +94,12 @@ function HuBMAPDatasetsChart() {
           assayTypeMap: datasetTypeMap,
         });
       },
+      getAriaLabel: (d) => {
+        const organ = String(d?.bar?.data?.organ);
+        const assay = d.key;
+        const count = d.key ? d?.bar?.data[d.key] : null;
+        return `${count} ${organ} datasets with assay type ${assay}.`;
```
NickAkhmetov, line 101: We should account for `null`'s and `undefined`'s string representations in template strings being `"null"/"undefined"` e.g.:
```js
const test = `${null} ${"Stomach"} datasets with assay type ${"scRNA-seq"}` 
console.log(test)
> "null Stomach datasets with assay type scRNA-seq"
const test2 = `${undefined} ${"Stomach"} datasets with assay type ${"scRNA-seq"}` 
console.log(test2)
> "undefined Stomach datasets with assay type scRNA-seq"
```

Maybe this could work?

```js
// We can use optional chaining with indexed objects like this:
const count = d?.bar?.data?.[d.key];
if (count) {
  return `${count} ${organ} datasets with assay type ${assay}.`;
}
return `${organ} datasets with assay type ${assay}.`
```

NickAkhmetov (commented):
Thanks, this looks great! Just a few minor finishing touches:

NickAkhmetov (approved):
👍🏻 Nicely done!

PR #3671, opened 2025-01-22
Austenem/CAT-754 Refactor "My Lists"

austenem:
## Summary

Broad update to the "My Lists" feature across the portal that leverages the new UKV API.

Major updates include:

- Lists are now tied to a user's profile.
  - Lists will now persist across devices (previously, these were stored locally).
  - As a result, lists will be restricted to authenticated users going forward. 
- "Save to list" buttons are now present on all search pages, dataset tables, and processed dataset sections.
- Language on all "My Lists" pages and alerts have been updated.
- All associated files have been converted to TypeScript.

[CAT-754 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-754?atlOrigin=eyJpIjoiMDg2YTBiYjk3ZGY3NDllN2IzZjUxMDc3MDI1MzE4YjMiLCJwIjoiaiJ9)

## Testing

Tested using the following checklist:

- [x]  When logging in for the first time on a device with saved entities and lists, these are copied over.
- [x]  Create new list with title and/or description
- [x]  Save entities
    - [x]  From search pages
    - [x]  From entity header of detail pages
    - [x] From processed dataset summary section of dataset detail pages
    - [x]  From derived entities tables
    - [x]  From Collections dataset tables
- [x]  Delete entities
- [x]  Add entities to list
    - [x]  From lists page
    - [x]  From entity header of detail pages
    - [x] From processed dataset summary section of dataset detail pages
- [x]  Remove entities from list
    - [x]  From list detail page
    - [x]  From entity header of detail pages
    - [x] From processed dataset summary section of dataset detail pages
- [x]  View list detail page without entities
- [x]  View list detail page with entities
- [x]  Edit list title/description
- [x]  Delete list

## Screenshots/Video

Updated Search page
Screenshot 2025-01-21 at 6 11 22 PM

Updated Collections dataset table
Screenshot 2025-01-22 at 12 41 41 PM

Updated "My Lists" page for logged-out users
Screenshot 2025-01-28 at 5 23 52 PM

Example of banner shown after transfer of local lists to a user's profile
Screenshot 2025-01-28 at 12 40 19 PM

## Checklist

- [X] All existing tests pass


context/app/static/js/components/savedLists/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const SAVED_ENTITIES_DEFAULT = {
+  title: 'My Saved Items',
+  description: '',
+  dateSaved: Date.now(),
+  dateLastModified: Date.now(),
```
john-conroy, line 5: Perhaps rename these to `saved_timestamp` and `last_modified_timestamp`. I think of a date object or string when thinking of a date.

context/app/static/js/components/savedLists/api.ts
```diff
@@ -0,0 +1,329 @@
...
+      Authorization: `Bearer ${token}`,
+      'Content-Type': 'application/json',
+    },
+    body,
+  });
+
+  if (!response.ok) {
+    console.error('UKV API failed with error: ', await response.text());
+  }
+}
```
john-conroy, line 56: This is pattern used elsewhere so no fault in introducing it, but could we use the fetcher in the swr helpers?

context/app/static/js/components/savedLists/hooks.ts
```diff
@@ -0,0 +1,212 @@
...
+  deleteQueuedLists: state.deleteQueuedLists,
+  deleteList: state.deleteList,
+  editList: state.editList,
+});
+
+/* Hook that wraps the savedEntities store and provides functions for saving, deleting, and managing lists. */
+function useSavedLists() {
+  const urls = useUkvApiURLs();
+  const { groupsToken, isAuthenticated } = useAppContext();
+  const { setTransferredToProfile } = useSavedListsAlertsStore();
```
john-conroy, line 50:
```suggestion
  const setTransferredToProfile = useSavedListsAlertsStore((state) => setTransferredToProfile);
```
We should slice the store when accessing specific fields. This avoids updates when state outside of the field you're picking changes. 

https://github.com/pmndrs/zustand?tab=readme-ov-file#selecting-multiple-state-slices

context/app/static/js/components/savedLists/hooks.ts
```diff
@@ -0,0 +1,212 @@
...
+  createListRemote,
+  addEntityToListRemote,
+  addEntitiesToListRemote,
+  removeEntityFromListRemote,
+  removeEntitiesFromListRemote,
+  deleteListRemote,
+  editListRemote,
+} from 'js/components/savedLists/api';
+
+const savedEntitiesSelector = (state: SavedEntitiesStore) => ({
```
john-conroy, line 25: Is this a subset of the store or the entire thing? If it's the entire store we don't need the selector.

context/app/static/js/components/savedLists/hooks.ts
```diff
@@ -0,0 +1,212 @@
...
+      copySavedItemsToRemoteStore({ savedEntitiesLocal, savedListsLocal, urls, groupsToken });
+      store.setEntities(savedEntitiesLocal);
+      store.setLists(savedListsLocal);
+    } else {
+      // If the user already has saved entities and lists in their remote store, we need to copy those over to the local store.
+      store.setEntities(savedEntities);
+      store.setLists(savedLists);
+    }
+    storeHasBeenSetRef.current = true;
+    wasLoadingRef.current = isLoading;
```
john-conroy, line 85: Do we need both `storeHasBeenSetRef` and `wasLoadingRef`, I believe they will always have the same value?

context/app/static/js/components/savedLists/hooks.ts
```diff
@@ -0,0 +1,212 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isAuthenticated, isLoading]);
```
john-conroy, line 87: Is this to avoid putting the store into the dependencies?

context/app/static/js/components/savedLists/hooks.ts
```diff
@@ -0,0 +1,212 @@
...
+      setTransferredToProfile(true);
+      const { savedEntities: savedEntitiesLocal, savedLists: savedListsLocal } = useLocalSavedEntitiesStore.getState();
+
+      copySavedItemsToRemoteStore({ savedEntitiesLocal, savedListsLocal, urls, groupsToken });
+      store.setEntities(savedEntitiesLocal);
+      store.setLists(savedListsLocal);
+    } else {
+      // If the user already has saved entities and lists in their remote store, we need to copy those over to the local store.
+      store.setEntities(savedEntities);
+      store.setLists(savedLists);
```
john-conroy, line 82: Why are we still using a store for the saved lists/entities. Wouldn't we be fetching this data from the API using `swr1?

john-conroy (commented):
I'm a little confused why we are continuing to use a store for the saved lists/entities. Maybe it would be helpful to have a chat between the three of us?

context/app/static/js/components/savedLists/constants.ts
austenem: That's a good point - however I'm a bit hesitant to change these, as the locally stored entities we are copying over to the remote store use them. Changing them would require supporting both versions of this type in order to convert the locally stored entities to the updated type, and this seems like it could become messy quickly. If the naming is significantly confusing, though, I'm open to the swap.

context/app/static/js/components/savedLists/EditSavedStatusDialog/EditSavedStatusDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+        console.error(error);
+      });
+    });
+
+    const unselectedLists = Object.keys(savedListsAndEntities).filter((listUUID) => !selectedLists.has(listUUID));
+    unselectedLists.forEach((listUUID) => {
+      handleRemoveEntitiesFromList({ listUUID, entityUUIDs: new Set([uuid]) }).catch((error) => {
+        console.error(error);
+      });
+    });
```
john-conroy, line 39: Could you explain whats going on here?

context/app/static/js/components/savedLists/SaveEntitiesButton/SaveEntitiesButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const entityTypes = `${entity_type.toLowerCase()}s`;
+  const entityLabel = useSelectableTableTooltips ? `selected ${entityTypes}` : entityTypes;
+
+  let tooltip = '';
+
+  if (disabled) {
+    tooltip = allInSavedEntities ? `All ${entityLabel} are already saved.` : `Select ${entityTypes} to save.`;
+  } else {
+    tooltip = `Save ${entityLabel}.`;
+  }
```
john-conroy, line 42: Could we use a ternary and define the tooltip with `const`?

context/app/static/js/components/savedLists/SaveEntitiesButton/SaveEntitiesButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  return (
+    <WhiteBackgroundIconTooltipButton
+      onClick={() => {
+        trackSave({ action: 'Save To List', label: generateCommaList(Array.from(uuids)) });
+        handleSaveEntities({ entityUUIDs: uuids }).catch((error) => {
+          console.error(error);
+        });
+      }}
```
john-conroy, line 51: Can Nick, you and I agree on a pattern for `useCallback`?

context/app/static/js/hooks/useSavedEntityData.ts
```diff
@@ -1,21 +1,41 @@
...
+    return dateB - dateA;
+  });
+}
+
 function useSavedEntityData(savedEntities: SavedEntities, source: string[]) {
   const query = useMemo(
     () => ({
       query: getIDsQuery(Object.keys(savedEntities)),
       _source: source,
+      size: Object.keys(savedEntities).length,
```
john-conroy, line 30: I would either bump this to 10000 when we are returning a source or set it to 0 for purely aggregations queries.

context/app/static/js/components/savedLists/SavedListPanel/hooks.ts
```diff
@@ -0,0 +1,18 @@
...
+  const { searchHits } = useSavedEntityData(listSavedEntities, source);
+  const counts = getSearchHitsEntityCounts(searchHits);
+
+  return {
+    donors: counts.Donor || 0,
+    samples: counts.Sample || 0,
+    datasets: counts.Dataset || 0,
+  };
+}
+
```
john-conroy, line 17: Could we just use an aggregations query if the only purpose of this request is to get the counts by `entity_type`?

context/app/static/js/components/savedLists/SavedListsContent/style.ts
```diff
@@ -0,0 +1,22 @@
...
+  marginBottom: theme.spacing(3),
+}));
+
+const StyledHeader = styled(Typography)(({ theme }) => ({
+  marginBottom: theme.spacing(1),
+}));
+
+const PageSpacing = styled('div')(({ theme }) => ({
+  marginBottom: theme.spacing(5),
+}));
```
john-conroy, line 20: If these are the only styles we could use the margin props or `sx`. I defer to @NickAkhmetov here.

context/app/static/js/components/savedLists/api.ts
```diff
@@ -0,0 +1,167 @@
...
+    buildKey({ url: listsURL }),
+    ([url, head]) =>
+      fetcher<{ key: string; value: SavedEntitiesList }[]>({
+        url,
+        requestInit: { headers: head },
+        errorMessages: {
+          404: listUUID ? `No list with UUID ${listUUID} found.` : 'No lists for the current user were found.',
+        },
+      }),
+    { revalidateOnFocus: hasAccess, refreshInterval: 1000 * 60 },
```
john-conroy, line 85: Do we need to refresh this on an interval or only once the user has performed some action which might affect what's returned?

context/app/static/js/hooks/useSavedEntityData.ts
```diff
@@ -1,21 +1,41 @@
...
+    const entityA: SavedEntity = savedEntities[a._id] ?? { dateSaved: 0 };
+    const entityB: SavedEntity = savedEntities[b._id] ?? { dateSaved: 0 };
+
+    const dateA = entityA.dateSaved ?? 0;
+    const dateB = entityB.dateSaved ?? 0;
+
+    return dateB - dateA;
+  });
+}
+
```
john-conroy, line 24: Can we apply sort to the ES query?

context/app/static/js/components/savedLists/hooks.ts
```diff
@@ -0,0 +1,336 @@
...
+  let savedEntities: SavedEntitiesList = { ...SAVED_ENTITIES_DEFAULT };
+  let savedLists: Record<string, SavedEntitiesList> = {};
+
+  if (!isLoading) {
+    savedEntities = savedListsAndEntitiesRecord.savedEntities || { ...SAVED_ENTITIES_DEFAULT };
+
+    savedLists = Object.fromEntries(
+      Object.entries(savedListsAndEntitiesRecord).filter(([key]) => key !== 'savedEntities'),
+    );
+  }
```
john-conroy, line 146: Can we avoid the `let`s here?

context/app/static/js/components/savedLists/SavedListsContent/style.ts
NickAkhmetov: `sx` would be preferable for these based off of the MUI 6 docs.

context/app/static/js/components/savedLists/SaveEntitiesButton/SaveEntitiesButton.tsx
NickAkhmetov: Maybe we can pull this out into a helper function instead to keep things more readable? e.g.

```
// Outside of the component
function createTooltip(entityLabel: string, allInSavedEntities: boolean, disabled: boolean) {
  if (disabled) {
		return `Save ${entityLabel}.`
	}
	if (allInSavedEntities) {
		return `All ${entityLabel} are already saved.`
	}
	return `Select ${entityLabel}s to save.`
}

...

// Back inside the component
const tooltip = createTooltip(entityLabel, allInSavedEntities, disabled)
```

That would also let us add some unit tests if need be.

context/app/static/js/components/savedLists/SaveEntitiesButton/SaveEntitiesButton.tsx
NickAkhmetov: In my experience, `useCallback` is necessary whenever passing event handler functions down, since it breaks memoization otherwise.

We could also use `useEventCallback` from MUI (https://usehooks-ts.com/react-hook/use-event-callback - same hook, different library) which was designed to be easier to use than `useCallback` and address issues like https://github.com/facebook/react/issues/14099.

NickAkhmetov (commented):
Great improvements! I will also need to test this locally 😄

context/app/static/js/components/detailPage/summary/SummarySaveEntityButton/SummarySaveEntityButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { handleSaveEntities } = useSavedLists();
+  const trackSave = useTrackEntityPageEvent();
+
+  return (
+    <WhiteRectangularTooltipIconButton
+      onClick={() => {
+        handleSaveEntities({ entityUUIDs: new Set([uuid]) }).catch((error) => {
+          console.error(error);
+        });
+        trackSave({ action: 'Save To List', label: uuid });
```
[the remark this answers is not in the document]
austenem: Good idea!

context/app/static/js/components/savedLists/EditSavedStatusDialog/EditSavedStatusDialog.tsx
austenem: This is the logic for removing the selected entities from the lists that are un-checked, in the case that someone is using the dialog to remove the selected entities from a list. The naming of this function is a bit confusing - maybe something like `updateLists` would clarify.

john-conroy: I'm a little confused by the user flow and the fact that we are handling both cases in a single function..Maybe I'll have a better idea once I take a look locally.

context/app/static/js/components/savedLists/SaveEntitiesButton/SaveEntitiesButton.tsx
austenem: I like the idea of the helper function!

context/app/static/js/components/savedLists/SaveEntitiesButton/SaveEntitiesButton.tsx
austenem: I wasn't familiar with `useEventCallback`, but big fan of no dependency arrays. I'd be in favor of using that to wrap these in a `handleClick` function instead of defining them inline.

context/app/static/js/hooks/useSavedEntityData.ts
austenem: Got it - is this just a performance concern?

context/app/static/js/components/savedLists/api.ts
austenem: The refresh interval isn't needed - just removed. This was based on the workspaces fetcher function.

context/app/static/js/hooks/useSavedEntityData.ts
austenem: I don't know of a good way to do this in the query (apart from maybe passing in a custom comparator function for the sort) - the dateSaved property is only accessible by looking up the UUID in the savedEntities object, and not from any info returned by ES. What would you recommend?

context/app/static/js/components/detailPage/summary/SummarySaveEntityButton/index.ts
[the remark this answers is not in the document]
austenem: Sure was, good catch!

context/app/static/js/components/savedLists/SaveEntitiesButton/SaveEntitiesButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+    tooltip = `Save ${entityLabel}.`;
+  }
+
+  return (
+    <WhiteBackgroundIconTooltipButton
+      onClick={() => {
+        trackSave({ action: 'Save To List', label: generateCommaList(Array.from(uuids)) });
+        handleSaveEntities({ entityUUIDs: uuids }).catch((error) => {
+          console.error(error);
+        });
```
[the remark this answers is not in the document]
austenem: I've put together a list of lists-related events to track that Tiffany and I will go over together - that will be a separate PR ([CAT-1113](https://hms-dbmi.atlassian.net/browse/CAT-1113?atlOrigin=eyJpIjoiODhkMDJkZDMyNmNhNDBlZmIwYTZmYTZiZGE3NTg1MTAiLCJwIjoiaiJ9)). I don't believe based on the event tracking doc that we are currently tracking errors for most events, but that's definitely an option to consider.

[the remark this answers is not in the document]
tsliaw: We don't typically track error states, so I don't see the need to track this one unless we expect this error to occur often.

context/app/static/js/components/savedLists/SavedListPanel/hooks.ts
```diff
@@ -0,0 +1,40 @@
...
+  };
+}
+function useSavedEntityCountsData(savedEntities: Record<string, SavedEntity>) {
+  const query = useMemo(
+    () => ({
+      size: 0,
+      query: getIDsQuery(Object.keys(savedEntities)),
+      aggs: {
+        entity_counts: {
+          terms: { field: 'entity_type.keyword', size: 10 },
```
john-conroy, line 18:
```suggestion
          terms: { field: 'entity_type.keyword', size: 10000 },
```
I would just use the max value to be safe. If for whatever reason the number of `entity_type` variations were to exceed 10 the aggregations might not include donor, sample or dataset.

context/app/static/js/components/savedLists/SavedListPanel/hooks.ts
```diff
@@ -0,0 +1,40 @@
...
+
+function useSavedEntityTypeCounts(listSavedEntities: Record<string, SavedEntity>) {
+  const { searchData } = useSavedEntityCountsData(listSavedEntities);
+
+  const buckets = searchData?.aggregations?.entity_counts?.buckets ?? [];
+  return {
+    donors: buckets.find((bucket) => bucket.key === 'Donor')?.doc_count ?? 0,
+    samples: buckets.find((bucket) => bucket.key === 'Sample')?.doc_count ?? 0,
+    datasets: buckets.find((bucket) => bucket.key === 'Dataset')?.doc_count ?? 0,
+  };
```
john-conroy, line 37: Can we just iterate once? Perhaps `reduce`?

context/app/static/js/components/savedLists/SavedListPanel/hooks.ts
```diff
@@ -1,18 +1,43 @@
...
-    samples: counts.Sample || 0,
-    datasets: counts.Dataset || 0,
-  };
+  const counts = { donors: 0, samples: 0, datasets: 0 };
+
+  searchData?.aggregations?.entity_counts?.buckets?.forEach((bucket) => {
+    if (bucket.key === 'Donor') counts.donors = bucket.doc_count;
+    else if (bucket.key === 'Sample') counts.samples = bucket.doc_count;
+    else if (bucket.key === 'Dataset') counts.datasets = bucket.doc_count;
+  });
```
NickAkhmetov, line 38: This should work! I'm a bit of a stickler for minimizing mutations outside of hooks, so I took a stab at wrapping this in a useMemo (might need some linting/formatting fixes):

```suggestion
  const counts = useMemo(() => {
  	return searchData?.aggregations?.entity_counts?.buckets?.reduce((acc, bucket) => { 
	  	const { key, doc_count } = bucket;
	  	switch(key) {
	  		case "Donor": 
	  			acc.donors = doc_count;
	  			break;
				case "Sample":
					acc.samples = doc_count;
					break;
				case "Dataset":
					acc.datasets = doc_count;
					break;
	  	}
	    return acc;
	  }, 
	  { donors: 0, samples: 0, datasets: 0 }, [searchData])
```

NickAkhmetov (commented):
Still doing local tests, but spotted a potential minor improvement

NickAkhmetov (commented):
I encountered a bug with transferring saved items - is it reproduceable on your end?
* Switch to `main`, save an item to local storage, confirm the list is there
* Switch back to this branch, go to `my-lists` page
* The "your lists have been transferred" alert displays, but saved items disappear

I tried record a reproduction of the issue - the item is initially displayed in the recording, then is cleared from the page after a refresh. When I initially encountered the issue, the item was completely absent.

https://github.com/user-attachments/assets/fcec2990-18a0-4135-b122-2b0062943864

 

Overall, I think this will be a solid foundation for us to continue to build on! There are some minor improvements we could probably make in terms of UX, but these can be follow-up tickets:
* It may be useful to show optimistic UI while waiting for actions to be confirmed by the server, e.g. 
	* turning list items into skeletons while they're being deleted or being worked on, showing a skeleton on the "my lists" page while it's being initially populated (to avoid flashing "0 lists and potentially confusing users", especially on slow networks)
* It may be worthwhile for us to eventually add the ability to add directly to a list from the search table - the "save entity -> go to my lists -> transfer entities from saved entities to list -> clear them from saved entities" workflow might be annoying.
* Should we consider displaying the character limit for descriptions?

austenem:
> I encountered a bug with transferring saved items - is it reproduceable on your end?

@NickAkhmetov Interesting - I'm having trouble recreating this on my end, but that's worrisome. The items disappearing after you leave the page is definitely strange.

One reason some of that might be happening is if you logged into the portal on this branch _before_ switching back to main and trying this out. I designed the lists transfer logic under the assumption that the user would not already have saved lists/entities in the UKV store - because the transfer is supposed to happen just once. If that were not the case, and the transfer happened multiple times, there could be some weird behavior as the existing entities would be overwritten.

During my testing of the transfer process, I've been manually deleting any UKV keys for my account beforehand. If it's helpful, I can share the UKV Postman collection I've been using.

I agree with all the UI ideas you've listed - I think those are great follow-up tickets. You actually _can_ add entities directly to a list from the search pages right now, but it's a bit non-intuitive (you save the entity, then click the save button again to "edit" the saved entity, which allows you to save to a list.)

john-conroy (approved):
Good to merge!

PR #3688, opened 2025-02-11
NickAkhmetov/CAT-1078 Fix unselectable collections tabs

NickAkhmetov:
## Summary

This PR makes unselectable collections tabs clickable, so that the content of their tab panel changes accordingly.

https://hms-dbmi.atlassian.net/browse/CAT-1078

## Testing

Manual testing

## Screenshots/Video

https://github.com/user-attachments/assets/80501d5b-9ba5-4aa6-b18c-7790ec95c3e7

## Checklist

- [x] All existing tests pass


NickAkhmetov:
Maybe revisiting the fetching strategy for collections data is necessary - i.e., hoisting the fetch of all collection-related data to the container to pre-filter which tabs should and shouldn't be displayed before the first render of the Tabs component altogether

context/app/static/js/hooks/useDatasetsCollections.ts
```diff
@@ -11,6 +14,7 @@ function buildCollectionsWithDatasetQuery(datasetUUIDs: string[]): SearchRequest
         'datasets.uuid': datasetUUIDs,
       },
     },
+    _source: ['uuid', 'title', 'hubmap_id', 'datasets'],
```
john-conroy, line 17: Do we need everything in `datasets`?

context/app/static/js/components/detailPage/CollectionsSection/CollectionsSection.tsx
```diff
@@ -1,76 +1,35 @@
...
-          <OutlinedAlert severity="info">The raw dataset is not referenced in any existing collections.</OutlinedAlert>
-        </TabPanel>
-      );
-    }
-    return null;
+    return (
+      <TabPanel value={value} index={index}>
+        <OutlinedAlert severity="info">The raw dataset is not referenced in any existing collections.</OutlinedAlert>
+      </TabPanel>
+    );
```
austenem, line 32: This is pre-existing, but what is the case for this panel being returned? It's my understanding that if the dataset is not referenced by a collection, there shouldn't be any tab present for that dataset.

austenem (approved):
LGTM, just one question.

context/app/static/js/hooks/useDatasetsCollections.ts
NickAkhmetov: I'll revert this, this was a leftover

context/app/static/js/components/detailPage/CollectionsSection/CollectionsSection.tsx
NickAkhmetov: I've rearranged the logic so that the collections data is fetched once for all datasets on the page, then sorted into buckets by dataset on the client side. The resulting map starts with an empty entry for the raw dataset; all other datasets with no matching collections are not included whatsoever. With this approach, it's known that the only reason why there would be a panel displayed with no collections data to show for it is because it is the primary dataset.

john-conroy (approved):

PR #3700, opened 2025-02-24
Austenem/CAT-1150 Workspaces landing page

austenem:
## Summary

Updates workspaces landing page with tables for workspace invitations and owned workspaces. First PR for in-progress workspace sharing implementation.

[CAT-1150 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-1150?atlOrigin=eyJpIjoiN2Q0NmI0MzMwYzM5NGM5MGIyZTg3ZDM2ODk2M2RlM2MiLCJwIjoiaiJ9)

## Testing

Tested via interacting with endpoints directly to share workspaces back and forth, as sharing dialogs are not yet implemented. 

## Screenshots/Video

Screenshot 2025-02-24 at 4 07 10 PM

## Checklist

- [X] All existing tests pass


context/app/static/js/components/workspaces/utils.ts
```diff
@@ -370,6 +374,48 @@ function unconvert(value: number, conversionFactor: number) {
...
+ * Get the value of a field from a workspace or workspace invitation.
+ * @param item The workspace or workspace invitation to get the field value from.
+ * @param field The field to get the value of.
+ * @param prefix The prefix to use if the field is a workspace invitation.
+ * @returns The value of the field.
+ */
+function getFieldValue({ item, field, prefix }: { item: WorkspaceItem; field: string; prefix?: string }) {
+  // datetime_last_job_launch starts as null, so we fall back to datetime_created if the workspace has never been launched
+  if (field === 'datetime_last_job_launch') {
+    const launchDate = get(item, field, '');
```
john-conroy, line 401: Do you prefer the `get` syntax rather than `item?.[field] ?? ''` ?

context/app/static/js/components/workspaces/hooks.ts
```diff
@@ -134,6 +136,36 @@ function useWorkspacesList() {
...
+
+  const workspacesWithInvitationInfo = workspacesList.map((workspace) => {
+    const user_id = receivedInvitations.find((invitation) => invitation.original_workspace_id.id === workspace.id)
+      ?.shared_workspace_id.user_id;
+
+    return {
+      ...workspace,
+      user_id,
+    };
+  });
```
john-conroy, line 160: Could it make sense to first make a map of the invitations by ID to avoid the nested looping? Is there a situation in which you have multiple invitations with the same `original_workspace_id`?

context/app/static/js/components/workspaces/WorkspacesList.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const [inputValue, setInputValue] = useState('');
+
+  // Filter workspaces in list based on search input
+  const filteredWorkspaces = useMemo(() => {
+    if (!inputValue.trim()) return workspacesList;
+    return workspacesList.filter(
+      (ws) =>
+        ws.name.toLowerCase().includes(inputValue.toLowerCase()) || ws.id.toString().includes(inputValue.toLowerCase()),
+    );
+  }, [inputValue, workspacesList]);
```
john-conroy, line 60: Should we introduce some `debounce` to avoid filtering the workspaces on every change to the input?

context/app/static/js/components/workspaces/Tables/WorkspacesTable/hooks.ts
```diff
@@ -0,0 +1,50 @@
...
+    const own = workspacesList.filter((workspace) => !workspace.user_id).length;
+    return { ownCount: own, sharedCount: workspacesList.length - own };
+  }, [workspacesList]);
+
+  const [showOwn, setShowOwn] = useState(true);
+  const [showShared, setShowShared] = useState(true);
+
+  useEffect(() => {
+    setShowShared(sharedCount > 0);
+  }, [sharedCount]);
```
john-conroy, line 16: Can we just pass the `sharedCount > 0` as the default value to the `useState` for `showShared`?

context/app/static/js/components/workspaces/Tables/WorkspacesTable/hooks.ts
```diff
@@ -0,0 +1,50 @@
+import { useEffect, useMemo, useState } from 'react';
+import { WorkspaceWithUserId } from 'js/components/workspaces/types';
+import { TableFilter } from 'js/components/workspaces/Tables/WorkspaceItemsTable/types';
+
+function useWorkspacesTable(workspacesList: WorkspaceWithUserId[]) {
+  const { ownCount, sharedCount } = useMemo(() => {
+    const own = workspacesList.filter((workspace) => !workspace.user_id).length;
+    return { ownCount: own, sharedCount: workspacesList.length - own };
+  }, [workspacesList]);
```
john-conroy, line 9: Instead of returning the lengths here could we filter/split the workspaceList by own and shared here? Then we wouldn't have to filter each time the showOwn and showShared state change.

context/app/static/js/components/workspaces/Tables/WorkspaceItemsTable/hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+      return;
+    }
+
+    items.forEach((item) => {
+      const itemId = 'id' in item ? item.id.toString() : item.original_workspace_id.id.toString();
+      if (selectedItemIds.size === items.length) {
+        toggleItem(itemId);
+      } else if (!selectedItemIds.has(itemId)) {
+        toggleItem(itemId);
+      }
```
john-conroy, line 47: Can we add a `toggleAllItems` or similar and avoid toggling each item individually?

context/app/static/js/components/workspaces/Tables/InvitationsTable/hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const filteredInvitations = useMemo(
+    () =>
+      [...invitations].filter((invitation) => {
+        if (showPending && !invitation.is_accepted) return true;
+        if (showAccepted && invitation.is_accepted) return true;
+        return false;
+      }),
+    [invitations, showPending, showAccepted],
+  );
```
john-conroy, line 26: Same comment as the one in `workspaces/Tables/WorkspacesTable/hooks.ts`.

john-conroy (commented):
Nice progress! I understand everything is in flux so no need to address these immediately.

context/app/static/js/components/workspaces/WorkspaceButton/WorkspaceButton.tsx
```diff
@@ -4,12 +4,12 @@ import { TooltipIconButton } from 'js/shared-styles/buttons/TooltipButton';
...
   color: theme.palette.primary.main,
-  borderRadius: theme.spacing(1),
+  borderRadius: theme.spacing(0.5),
   border: `1px solid ${theme.palette.divider}`,
 }));
 
 WorkspaceButton.defaultProps = {
-  size: 'large',
+  size: 'small',
 };
```
NickAkhmetov, line 13: `defaultProps` is deprecated ahead of React 19 - could we provide this as part of the styled component instead? e.g. something like:

```
const WorkspaceButton = styled((props: React.ComponentProps<TooltipIconButton> => <TooltipIconButton size="large" {...props} />)(({ theme }) => ({
  backgroundColor: theme.palette.white.main, 
  color: theme.palette.primary.main,  
  borderRadius: theme.spacing(0.5),
  border: `1px solid ${theme.palette.divider}`, 
}));
```

NickAkhmetov (commented):
Looks good! Agreed with John's comments + a minor request on my part.

context/app/static/js/components/workspaces/utils.ts
```diff
@@ -370,6 +374,48 @@ function unconvert(value: number, conversionFactor: number) {
   return value * conversionFactor;
 }
 
+/**
+ * Extract the prefix from a field if it is a workspace invitation.
+ * @param field The field to extract the prefix from.
+ * @returns 'original_workspace_id.' or 'shared_workspace_id.' if the field is a workspace invitation, else an empty string.
+ */
+function getFieldPrefix(field: string) {
```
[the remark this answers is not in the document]
austenem: It's used in the `CellContent` component.

context/app/static/js/components/workspaces/utils.ts
austenem: In this case it was helpful - the `'datetime_last_job_launch'` field doesn't exist on `WorkspaceInvitations`, so trying to access it here causes a TS error that I'd need to otherwise include a type guard for. Additionally, this syntax is already in use within this function to access nested fields.

context/app/static/js/components/workspaces/hooks.ts
austenem: That's a good call - that situation is possible (if someone shared the same workspace multiple times), but in that case the user info would always be the same.

context/app/static/js/components/workspaces/Tables/WorkspacesTable/hooks.ts
austenem: Good idea!

context/app/static/js/components/workspaces/Tables/WorkspacesTable/hooks.ts
austenem: Unfortunately this doesn't work as the `workspacesList` isn't loaded initially, so `showShared` will always be set to false. And because the filter should be disabled when there are no shared workspaces, `showShared` shouldn't be set to true initially either without some update after the `workspacesList` has loaded.

context/app/static/js/components/workspaces/WorkspaceButton/WorkspaceButton.tsx
austenem: Nice! Good to know.

PR #3703, opened 2025-03-04
Austenem/CAT-1176 Add data products

austenem:
## Summary

Adds a data products section to Organ pages, incorporating the [CMU data products API](https://data-products.cmu.hubmapconsortium.org/data_products/) into the portal ui.

[CAT-1176 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-1176?atlOrigin=eyJpIjoiYmY2MTE5MjhjOWM2NGZkN2I0ZTE3MmQ3ZDEyMTI0ZTYiLCJwIjoiaiJ9)

## Screenshots/Video

Screenshot 2025-03-04 at 11 37 09 AM

## Checklist

- [X] All existing tests pass


context/app/static/js/components/organ/DataProducts/DataProducts.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <Paper>
+          <EntitiesTable
+            headerCells={[
+              { headerId: 'organ', label: 'Organ' },
+              { headerId: 'assay-type', label: 'Assay Type' },
+              { headerId: 'raw-download', label: 'Raw Download' },
+              { headerId: 'processed-download', label: 'Processed Download' },
+              { headerId: 'shiny-app', label: 'Shiny App' },
+              { headerId: 'creation-date', label: 'Creation Date' },
+              { headerId: '', label: '' },
```
john-conroy, line 51: You could pull the header cells definition outside of the component since it doesn't depend on other state/prop variables.

john-conroy: Hmm, do we need a something to identify the last cell for accessibility?

context/app/static/js/helpers/functions.ts
```diff
@@ -304,3 +304,19 @@ export function getEntityCreationInfo({
...
+ * ex: 'https://example.com/file.txt' => 'file.txt'
+ * ex: 'None' => ''
+ * @author Austen Money
+ */
+export function getFileName(url: string) {
+  const name = url.split('/').pop() ?? '';
+
+  if (name.toLowerCase() === 'none') {
+    return '';
+  }
```
john-conroy, line 319: Is this logic specific to the data products or do we expect "None" elsewhere?

context/app/static/js/pages/Organ/hooks.ts
```diff
@@ -153,3 +158,26 @@ export function useLabelledDatasetsQuery(searchItems: string[]) {
...
+  const dataProductsUrl = `${dataProductsEndpoint}/api/data_products/${organName}`;
+
+  const { data } = useSWR<OrganDataProducts[]>(
+    dataProductsUrl,
+    (url: string) =>
+      fetcher({
+        url,
+      }),
+    { fallbackData: [] },
+  );
```
john-conroy, line 173: Do we need to return `isLoading` to handle loading state and avoid layout shift?

john-conroy (commented):
Nice, a couple of comments, but nothing major. If you agree they are worth addressing, we could file issues to tackle them later.

context/app/static/js/components/organ/DataProducts/DataProducts.tsx
austenem: That's a good point - I think adding a distinct ID and an `aria-label` would help contextualize the column.

context/app/static/js/helpers/functions.ts
austenem: This is specific to the data products as that's the placeholder they use. This could potentially be expanded into an array of placeholders to check against (like `['none', 'na', ...]`), if this helper is used in other places that expect something different.

context/app/static/js/pages/Organ/hooks.ts
austenem: Good idea!

context/app/static/js/helpers/functions.ts
john-conroy: Because this is in the top level utils, you would expect it to be generalized for use throughout the application. You could either handle cases via additional params or just add a function located closer to the data products which calls this function then handles the case.

john-conroy (approved):

PR #3708, opened 2025-03-11
Austenem/CAT-1157 Update workspace dialogs

austenem:
## Summary

Adds dialogs for new workspace sharing actions and updates existing workspace dialogs.

[CAT-1157 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-1157?atlOrigin=eyJpIjoiODQyOWU3MjNlNWExNDBiYTgwYjFlM2VjZDQ2Y2Y3MjIiLCJwIjoiaiJ9)

[CAT-1156 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-1156?atlOrigin=eyJpIjoiNTAxNGQxZDcwYTEwNDhhYmIxMzNhNWQxMjYyODhjYzAiLCJwIjoiaiJ9)

[Figma designs](https://www.figma.com/design/NgBjvs0jRbDHhiyn5nicDm/Workspace?node-id=3392-1923&t=poVarH2IwMFj8GZE-1)

## Screenshots/Video

New dialogs:

Screenshot 2025-03-11 at 2 49 48 PM
Screenshot 2025-03-11 at 2 49 38 PM
Screenshot 2025-03-11 at 2 52 37 PM
Screenshot 2025-03-11 at 2 52 48 PM

Updated dialogs:

Screenshot 2025-03-11 at 2 50 25 PM
Screenshot 2025-03-11 at 2 52 20 PM

## Checklist

- [X] All existing tests pass


context/app/static/js/components/workspaces/ConfirmDeleteWorkspacesDialog/ConfirmDeleteWorkspacesDialog.tsx
```diff
@@ -1,96 +1,49 @@
...
 
   return (
-    <Dialog
-      open={dialogIsOpen}
-      onClose={handleClose}
-      scroll="paper"
-      aria-labelledby="delete-workspace-dialog"
-      maxWidth="lg"
+    <ConfirmationDialog
+      title={`Delete Workspace ${selectedWorkspaceIds.size > 1 ? 's' : ''}`}
```
NickAkhmetov, line 41: Tiny fix - the plural case would be displayed as `Delete Workspace s` due to the extra space.
```suggestion
      title={`Delete Workspace${selectedWorkspaceIds.size > 1 ? 's' : ''}`}
```

context/app/static/js/components/workspaces/Tables/WorkspaceItemsTable/hooks.ts
```diff
@@ -15,7 +15,11 @@ function useWorkspaceItemsTable<T extends WorkspaceItem>({
   toggleAllItems,
 }: WorkspaceItemsTableProps<T>) {
   const [sortField, setSortField] = useState<SortField>(initialSortField);
-  const [numVisibleItems, setNumVisibleItems] = useState(showSeeMoreOption ? 3 : items.length);
+  const [numVisibleItems, setNumVisibleItems] = useState(3);
+
+  useEffect(() => {
+    setNumVisibleItems(showSeeMoreOption ? 3 : items.length);
+  }, [items, showSeeMoreOption]);
```
NickAkhmetov, line 22: If this is the only place where we use the `setNumVisibleItems` setter and it is fully derived from other variables, we can avoid making this a stateful variable altogether. A quick `ctrl + F` didn't find any usage of this function outside of this hook, so I think this should work and avoid the extra rerender.

```suggestion
  const numVisibleItems = showSeeMoreOption ? 3 : items.length;
```

context/app/static/js/components/workspaces/WorkspaceField/WorkspaceField.tsx
```diff
@@ -24,20 +27,31 @@ function WorkspaceField<FormType extends FieldValues>({
     defaultValue: value,
   });
 
+  const [charCount, setCharCount] = useState((field?.value as string)?.length ?? 0);
```
NickAkhmetov, line 30: I believe we can derive the character count from the field state to avoid separately tracking it, e.g.

```jsx
const charCount = (watch(name) ?? '').length
```

Where `watch` is a callback provided by react-hook form: https://www.react-hook-form.com/api/usewatch/

We might also be able to set the max length validation using react-hook-form: https://www.react-hook-form.com/get-started/#Applyvalidation

NickAkhmetov (commented):
Nicely done, clean code! Some minor improvements came to mind while reading through but nothing huge.

context/app/static/js/components/workspaces/Tables/WorkspaceItemsTable/hooks.ts
austenem: Unfortunately this setter is in use by the "See more" button (in order to increase the number of visible rows by three at a time).

context/app/static/js/components/workspaces/WorkspaceField/WorkspaceField.tsx
austenem: This is good to know, and this is possible - but from what I can tell implementing this requires adding some messy typing & prop drilling (as the `watch` function needs to be passed down from the initial `useForm` call). Unless there are some larger benefits (or a simpler way to incorporate this function) I'd prefer to keep the component state as-is for readability purposes.

NickAkhmetov: > as the `watch` function needs to be passed down from the initial `useForm` call

Using a `FormProvider` would help alleviate the prop drilling concerns: https://react-hook-form.com/docs/useformcontext

I believe we can also provide the top-level form type to `useWatch` with e.g. `useWatch<FormType>({control, ...})`, but I could be mistaken.

Either way, for the sake of simplicity, we can keep this as-is.

NickAkhmetov (approved):

context/app/static/js/components/workspaces/WorkspaceField/WorkspaceField.tsx
austenem: Good point! A `FormProvider` could be useful for some of the other workspace form props as well.

PR #3728, opened 2025-04-25
Austenem/CAT-1226 Add homepage tracking

austenem:
## Summary

Adds tracking to the homepage.

[CAT-1226 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-1226?atlOrigin=eyJpIjoiZjkzOGQ3MzM4NWM5NGU2ZmJhYTYyY2QwMGZiZmVkNTAiLCJwIjoiaiJ9)

[Homepage trackable events doc](https://docs.google.com/spreadsheets/d/18hF0VDrepN58GiDX22_jJ6vsnfGwKE_GiL8hdD_Art4/edit?pli=1&gid=0#gid=0)

## Testing

Tested by logging all events to the console, then verifying against trackable events doc.

## Checklist

- [X] All existing tests pass


context/app/static/js/components/home/Hero/HeroTab.tsx
```diff
@@ -1,23 +1,39 @@
...
+  const handleClick = useEventCallback(() => {
+    trackEvent({
+      category: 'Homepage',
+      action: `Hero / ${tabTitle}`,
+      label: `${title} Button`,
+    });
+
+    if (onClick) {
+      onClick();
+    }
```
NickAkhmetov, line 30: Alternatively:
```suggestion
    onClick?.();
```
But this is optional.

In other circumstances, I would also consider adjusting `handleClick` and the type definition for `onClick` so that the underlying event still gets forwarded; since `onClick` was being passed directly to the chip before, there is an implicit click event arg that still gets passed. In some circumstances the parent component may want to use this event directly. Since the onClick was already defined as an argless void in this case it's probably fine, though.

context/app/static/js/components/home/EntityCounts/EntityCounts.tsx
```diff
@@ -30,6 +32,14 @@ interface EntityCountsProps {
 
 function EntityCounts({ organsCount }: EntityCountsProps) {
   const entityCounts = useEntityCounts();
+  const handleTrack = useEventCallback((type: string) => {
```
NickAkhmetov, line 35: This likely shouldn't cause any issues, but we should keep in mind that these entity counts are embedded in an iframe on the homepage of the consortium - https://hubmapconsortium.org/. I think that matomo should be able to properly distinguish any events that are fired from there instead of on one of our main sites, but it may be something worth tracking separately in case it doesn't.

Could we try adding some logic to distinguish whether the event is being fired from an iframe? e.g. `const inIframe = window.self !== window.top` [should work](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_script_api_access) to determine if the current JS frame is remotely embedded, and we could modify the `action` or `label` to reflect that it's in an iframe.

austenem: Ah good to know!! Added that logic in to conditionally change the category.

PR #3749, opened 2025-06-03
Austenem/CAT-1264 OpenKeyNav integration

austenem:
## Summary

Includes new profile preferences section with a boolean on/off switch, which is connected to a new value in the UKV store for user preferences. 

Next step is to only initialize the open key nav instance when that boolean is true & vice versa.

### DIFF
--- a/context/app/static/js/components/App.jsx
+++ b/context/app/static/js/components/App.jsx
@@ -1,6 +1,7 @@
 import React, { StrictMode } from 'react';
 import { pdfjs } from 'react-pdf';
 import { enableMapSet } from 'immer';
+import OpenKeyNav from 'openkeynav';
 
 import StyledSnackbar from 'js/shared-styles/snackbars';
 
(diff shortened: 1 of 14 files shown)


john-conroy:
Screenshot 2025-06-05 at 3 07 38 PM
Screenshot 2025-06-05 at 3 07 32 PM

context/app/static/js/components/OpenKeyNavInitializer.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+function useInitializeOpenKeyNav() {
+  const { savedPreferences } = useSavedPreferences();
+
+  if (savedPreferences.enableOpenKeyNav) {
+    const openKeyNav = new OpenKeyNav();
+    openKeyNav.init({
+      keys: {
+        menu: '/',
+        modifierKey: 'metaKey',
```
NickAkhmetov, line 14: I still think using the meta key instead of control might be uncomfortable for windows users who are not accustomed to shortcuts using the windows key. Could we consider using control based off of the user agent OS, similarly to how we display a specific warning for Vitessce performance on Firefox?

john-conroy: Would you be open to getting without this suggested change and letting Larry and other folks test it out?

PR #3768, opened 2025-07-03
Austenem/CAT-1269 Allow access to protected datasets

austenem:
## Summary

Updates workspace and bulk data transfer dialogs to allow access to protected datasets. 

Note: A major part of this update is a shift towards differentiating between "Protected" and "Restricted" datasets. Protected datasets are datasets marked a data access level of "Protected", meaning a given user might not have access to them depending on their group permissions. Restricted datasets are datasets that the current user does not have access to based on their group permissions.

We are differentiating between these two types because at present, there is no way to tell whether a user has permission to access a given dataset - however, there is an endpoint that will eventually become available to retrieve these details.

The current behavior is to provide a warning when any protected dataset is selected, but there is logic in place to provide an error and an option to remove any restricted datasets once the endpoint is available.

[CAT-1269 Jira ticket](https://hms-dbmi.atlassian.net/browse/CAT-1269?atlOrigin=eyJpIjoiZmFkMzZmOTIxYzMyNGFkYmJkNzZiNTU1ZGQ1NmUxYjIiLCJwIjoiaiJ9)

## Testing

Tested both warning and error logic by temporarily changing the conditions of `checkDatasetAccess()`.

## Screenshots/Video
Screenshot 2025-07-02 at 3 46 24 PM
Screenshot 2025-07-03 at 3 08 14 PM
Screenshot 2025-07-03 at 3 09 26 PM
Screenshot 2025-07-03 at 3 45 58 PM

## Checklist

- [X] All existing tests pass


NickAkhmetov (commented):
Looks good/makes sense to me.

Can we include also update the logic and messaging for the `isNotPublic` case here? https://github.com/hubmapconsortium/portal-ui/blob/main/context/app/static/js/shared-styles/tables/columns.tsx#L17-L64

NickAkhmetov (approved):

PR #3999, opened 2026-05-19
NickAkhmetov/CAT-1546 Add support datasets to ancestor ID-based queries in Search

NickAkhmetov:
## Summary

This PR adds support datasets to the search page when users navigate to the search page via the "view derived datasets" button in the provenance table. This avoids displaying misleading counts in the provenance table.

I originally planned to exclude supports from the counts in the prov table, but this proved impossible as the entity API reports image pyramids as datasets - accurately making those counts in the search index stage would require either changing the reported entity type in the entities API, or making an excessive amount of extra API calls during indexing to accurately determine whether each entity's dataset descendants are datasets or supports. Fixing this on the front-end would add extra search API calls to every single detail page.

Support entities are otherwise never visible - they are only ever displayed/included if an Ancestor ID query clause is included, which only occurs from the provenance view.

https://hms-dbmi.atlassian.net/browse/CAT-1546

## Testing

Added unit tests.

## Screenshots/Video

image

## Checklist

- [x] All existing tests pass

### DIFF
--- a/context/app/static/js/components/search/Results/CellContent.tsx
+++ b/context/app/static/js/components/search/Results/CellContent.tsx
@@ -10,11 +10,13 @@ import { VitessceIcon } from 'js/shared-styles/icons';
 import { SecondaryBackgroundTooltip } from 'js/shared-styles/tooltips';
 import DonorAgeTooltip from 'js/shared-styles/tooltips/DonorAgeTooltip';
 import { Entity } from 'js/components/types';
+import Typography from '@mui/material/Typography';
 
 export interface HuBMAPIdDisplayInfo {
   hasVisualization?: boolean;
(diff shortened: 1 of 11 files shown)


context/app/static/js/components/search/Results/CellContent.tsx
```diff
@@ -91,6 +94,21 @@ export function RetractedChip() {
   );
 }
 
+export function SupportChip() {
+  return (
+    <SecondaryBackgroundTooltip title="Support entities have only been processed to aid visualizations on the web and do not contain additional analysis.">
```
tsliaw: Update text to "Support datasets are created to enable visualizations and do not contain independent analysis results."

john-conroy (approved):

PR #4024, opened 2026-06-15
NickAkhmetov/CAT-1589 Update/regenerate organ yamls, update organs pages to match other bio entity functionalities

NickAkhmetov:
## Summary

The `make_organ_dir.py` generator (which produces the per-organ YAML files rendered on organ pages) had gone stale — every external source except Elasticsearch had drifted, and the script could no longer run. This PR repairs it, switches organ/laterality data to the authoritative HuBMAP ontology endpoint, drops the dead Azimuth integration, and regenerates the organ data (adding four new organ pages).

## Why

- **Azimuth** reference data was broken and its UI was already hardcoded off — dead weight in
  every organ file.
- The `search` terms (organ names used to query datasets) were grouped by brittle fuzzy name
  matching that missed real cases (e.g. dataset organ `"Bladder"` never matched the page name
  `"Urinary Bladder"`).
- The **CCF** source now returns SPARQL `results.bindings`, which the old parsing crashed on;
  it has also grown well beyond the portal's organ set.
- The **OLS3** description API stopped returning definitions (returned `null`), crashing the run.

### Generator (`etc/dev/organ-utils/`)
- **Drop Azimuth**: removed the fetch/processing code, the `--azimuth_url` arg, and the
  `azimuth` block from `organ-schema.yaml`.
- **Ontology endpoint** (`/organs?application_context=hubmap`): use its `category` field to
  group lateralized organ names (e.g. `Kidney (Left)`/`Kidney (Right)` → Kidney) and to resolve
  CCF's paired organs (kidney/eye/ureter/fallopian via UBERON, knee/mammary/tonsil via FMA) to
  their parent — replacing the fuzzy matching and the hand-maintained `expected_unmatched`
  allowlist. Elasticsearch still supplies the literal organ names.
- **CCF parsing**: read the current SPARQL `results.bindings` shape.
- **OLS3 → OLS4**: migrate the description source; pull the formal definition from
  `obo_definition_citation` rather than `description[0]`, which mixed in editorial comments and
  taxon notes (this had surfaced odd text for Heart, Lung, Placenta, and Intervertebral Disc).
- **Unmatched guardrail**: CCF tracks ~22 anatomical structures the portal doesn't surface
  (whole systems, etc.); these now emit a warning instead of raising, so the script stays
  runnable as CCF grows.

### New organ pages (`descriptions.yaml`)
Added **Adipose Tissue, Intervertebral Disc, Larynx, and Tonsil** — organs the ontology
endpoint lists/implies that the portal lacked.

### Frontend cleanup
Removed the now-unused Azimuth organ component (`components/organ/Azimuth/`), the `AzimuthConfig`
type and `azimuth` field on `OrganFile`, and the `reference-based-analysis` section wiring in
`Organ.tsx`. Unrelated external "Azimuth" links (header, home page) are untouched.

### Regenerated data
All 31 existing organ YAMLs updated (azimuth removed, corrected OLS4 definitions, lateralities
regrouped) plus 4 new files.

## Reviewer notes
- **Adipose Tissue** and **Intervertebral Disc** currently have no datasets (`search: []`) and
  no CCF component, so those pages will be sparse; Larynx and Tonsil are more complete.
- Adipose Tissue has no icon/ASCTB version yet.

## Testing
- `ruff check` + `ruff format` clean; module doctests pass.
- Generator runs end-to-end; jsonschema validation passes for all 35 organs.
- Frontend: `pnpm tsc` and `pnpm exec eslint` clean.
- `pytest` organ route tests pass (56).


NickAkhmetov:
Additional fixes that I realized are possible:
* There is a more up to date icon github repo to reference
* There are updates to asctb tables

john-conroy (approved):

PR #4062, opened 2026-08-06
NickAkhmetov/CAT-1655 SEO fixes

NickAkhmetov:
## Summary

Improves how dataset pages are surfaced in Google Dataset Search. An external investigation found the portal has low coverage there, with many dataset pages excluded as duplicates, and recommended canonical tags, richer structured data, and a `citation` field. This PR does those three things, and addresses two additional causes found while tracing the code.

**The schema.org markup is now server-rendered.** It was previously injected client-side by `schema.org.ts` from `index.tsx`, after the deferred module bundle executed — so discovery depended entirely on Googlebot's render pass. Generation moves to `_get_dataset_ld` in `routes_browse.py` and is emitted in `base.html`, which also lets it reuse the `_format_dataset_title` / `_format_donor_title` helpers that already existed in Python. `schema.org.ts` and its spec are deleted (200 lines), removing a second, diverging copy of the title logic that carried three live bugs.

**Processed datasets now 301 rather than 302 to their raw ancestor.** A 302 tells Google the *source* URL remains canonical, so each processed-dataset URL — with `?redirected=True&redirectedFromId=…` appended — was eligible for indexing as its own near-duplicate of the primary page. Same change for HuBMAP ID URLs and entity-type normalization. `latest_redirect` stays 302, since its target changes as new revisions are registered.

**Canonical tags** are added in `base.html` using `request.base_url`, which is the request URL without its query string — precisely what strips the redirect's tracking params. It covers every page extending the base template, not just entity pages.

**The structured data** now carries `identifier` (HuBMAP ID + DOI), `sameAs`, `datePublished`/`dateModified`, `keywords`, `measurementTechnique`, `includedInDataCatalog`, `funder`, `isAccessibleForFree`, `citation` (NLM format), and `creator` built from `contributors[]` with ORCID `sameAs` and affiliations — replacing `created_by_user_displayname`, which is the submitter's login display name. `identifier` matters most: Google clusters dataset replicas by persistent ID.

**Titles and descriptions now include the HuBMAP ID** for datasets, samples, and donors. Two entities sharing a donor, organ, and assay previously produced byte-identical titles and descriptions.

**The sitemap** becomes an XML index (`/sitemap.xml`) with per-type children carrying `<lastmod>`. This fixes two bugs: `get_uuids` raised past 10k hits inside a `try` whose `finally: return uuids` swallowed the exception, silently truncating results; and every lookup was `@cache`d for process lifetime, so entities published since the last deploy never appeared at all.

### Implications

- **`robots.txt` now advertises `/sitemap.xml`.** `/sitemap.txt` 301s to it, since Search Console and external referrers hold the old URL. The new sitemap needs resubmitting in Search Console after deploy.
- **Nothing user-visible changes.** All of this lives in `<head>`, in HTTP status codes, or in `/sitemap*`. Browser tab titles gain a ` (HBM…)` suffix.
- **301s are cached indefinitely by browsers.** See Additional Notes.

https://hms-dbmi.atlassian.net/browse/CAT-1655
https://docs.google.com/document/d/1-yL4a4KQld6fR1m91sldr2loL-nCGr3HbJgGwlcmP9M/edit?tab=t.0#heading=h.q6elnzoeehaf

### Automated

- **`context/app/test_routes_browse.py` is new** — 32 tests, where there was previously no test file for this route module at all. Coverage:
  - LD field-by-field: `identifier`, `sameAs`, dates, `keywords`, `measurementTechnique`, `citation`, `creator` with ORCID, `funder`.
  - `isAccessibleForFree` is `False` for `Protected`, `Consortium`, and a missing access level.
  - `_get_dataset_ld` returns `None` for donor/sample/collection/publication, so the `<script>` is omitted server-side.
  - An empty `orcid_id` must not produce a bare `https://orcid.org/` link.
  - **End to end:** `test_dataset_ld_is_in_the_served_html` parses the `ld+json` out of the rendered HTML and asserts the canonical link and `<title>` agree with it. This is the assertion that would have caught the original client-side-only problem.
  - Canonical URL has no query string when the page is fetched with `?redirected=True&redirectedFromId=…`.
  - No meta title or description contains a run of 2+ whitespace characters, across six entity shapes including ones with no metadata at all.
  - A donor's ID does not leak into dataset or sample titles (see Additional Notes).
  - Sitemap index, per-type children, `<lastmod>` value, `/sitemap.txt` → 301, unknown type → 404.
- **15 doctests** in `routes_browse.py`, run under `--doctest-modules`. Existing ones updated for the ID suffix; new ones on `_with_hubmap_id`, `_format_donor_page_title`, `_format_donor_details`, `_build_nlm_citation`, `_contributor_ld`, `_timestamp_to_iso`, `_sitemap_query`.
- **`test_routes_main.py`**: `test_302_redirect` split, with the entity-type normalization cases moved to a new `test_301_entity_type_normalization` that also asserts the redirect target.
- `uv run pytest context/app` — **357 passed** (was 321 on `main`).
- `pnpm test` — 910 passed, 7 skipped. Confirms nothing else imported `schema.org`.
- `pnpm tsc`, `pnpm lint`, `uv run ruff check context`, `uv run ruff format` — all clean.

### Manual

- Rendered the dataset page through the Flask test client with the bundle absent, and read the raw `<head>` to confirm the canonical link, `<title>`, meta description, and JSON-LD are all in the served HTML with no JS involved.
- Dumped the generated JSON-LD and the sitemap index and reviewed both by hand (output below).
- Verified both candidate funder ROR IDs against the ROR v2 API rather than trusting recall: `01cwqze88` is `ror_display: "National Institutes of Health"`; `001d55x84` is `ror_display: "NIH Common Fund"`, active, FundRef `100015326`. Used the latter, since HuBMAP is specifically a Common Fund program and it matches the `name` field.

### Not yet done

- **Google's Rich Results Test / Schema Markup Validator has not been run** — it needs a publicly reachable URL, so this has to happen against a deployed environment. The markup is well-formed JSON and structurally correct, but worth a pass on dev before merge.
- **Search Console verification** is post-deploy: URL Inspection on a few dataset URLs to confirm "Duplicate, Google chose different canonical" clears, and resubmit `/sitemap.xml`.

## Screenshots/Video

No visual changes — everything is in `<head>`, HTTP status codes, or `/sitemap*`. The served `<head>` for a dataset page:

```html
<title>Histology of Kidney (HBM123.ABCD.456) | Dataset | HuBMAP</title>
<meta name="description" content="Explore the Histology of Kidney (HBM123.ABCD.456) dataset from HuBMAP. Access and download metadata, visualizations, and analysis tools for research.">
<link rel="canonical" href="http://localhost/browse/dataset/fake-uuid">
<script type="application/ld+json">{ … }</script>
```

The JSON-LD, from the test fixture:

```json
{
  "@context": "https://schema.org/",
  "@type": "Dataset",
  "name": "Histology of Kidney (HBM123.ABCD.456)",
  "description": "Histology of Kidney from White Male, 70 years old, with a medical history of Diabetes. Too short.",
  "url": "http://localhost/browse/dataset/fake-uuid",
  "identifier": ["HBM123.ABCD.456", "https://doi.org/10.35079/HBM123.ABCD.456"],
  "sameAs": "https://doi.org/10.35079/HBM123.ABCD.456",
  "citation": "Lovelace A, Hopper G. Histology of Kidney (HBM123.ABCD.456) [Internet]. HuBMAP Consortium; 2019.",
  "creator": [
    {
      "@type": "Person",
      "name": "Ada Lovelace",
      "sameAs": "https://orcid.org/0000-0002-1825-0097",
      "affiliation": { "@type": "Organization", "name": "Analytical Engine Institute" }
    },
    { "@type": "Organization", "name": "Mock TMC" }
  ],
  "datePublished": "2019-10-31",
  "dateModified": "2024-07-19",
  "keywords": ["Kidney", "Histology [Image Pyramid]", "Histology", "HuBMAP"],
  "measurementTechnique": ["Histology [Image Pyramid]"],
  "includedInDataCatalog": {
    "@type": "DataCatalog",
    "name": "HuBMAP Data Portal",
    "url": "https://portal.hubmapconsortium.org"
  },
  "funder": { "@type": "Organization", "name": "NIH Common Fund", "sameAs": "https://ror.org/001d55x84" },
  "license": "https://creativecommons.org/licenses/by/4.0/",
  "isAccessibleForFree": true
}
```

Before and after, for the meta text:

| | Before | After |
| --- | --- | --- |
| Dataset title | `Histology of Kidney \| Dataset` | `Histology of Kidney (HBM123.ABCD.456) \| Dataset` |
| Sample title | `Section from Kidney of  \| Sample` | `Section from Kidney (HBM333.SAMP.444) \| Sample` |
| Donor title | ` \| Donor` (no demographics) | `HBM9 \| Donor` |
| Meta description | `…from HuBMAP.                 Access and…` | `…from HuBMAP. Access and…` |

## Checklist

  - [x] New `CHANGELOG-dataset-search-seo.md` is present in the root directory, describing the change(s) in full sentences.
- [x] All existing tests pass
- [x] Any new functionalities have appropriate analytics functionalities added — N/A, no new user-facing functionality; all changes are in `<head>`, HTTP status codes, and `/sitemap*`

### Deliberate omissions

- **No `distribution`.** Bulk download goes through a per-user Globus URL fetched at runtime. The only static alternatives were one `DataDownload` per file (hundreds of entries) or a `contentUrl` pointing back at this page, which is not a data location. `isAccessibleForFree` carries the access signal instead, still gated on `mapped_data_access_level == 'Public'` so protected and consortium datasets never advertise availability. Recorded in a `ponytail:` comment on `_get_dataset_ld`.
- **Google composes its own "how to cite" block** from `creator`, `name`, `datePublished`, and `identifier` — it does not render an arbitrary `citation` string verbatim. `citation` is added as the report asked, but the visible citation guidance comes mostly from those four fields, which this PR also fills in.
- **No structured data for donors, samples, or collections.** Google Dataset Search only consumes `Dataset`/`DataCatalog`/`DataDownload`, and marking a donor as a `Dataset` would be wrong.
- **Publication and collection titles are untouched** — they're curated and already distinct, so they don't need the ID suffix.

### Why the ID suffix is applied at the callers

`_format_donor_title` is shared three ways: it renders the donor's own page title *and* the donor clause inside dataset and sample titles. Appending the ID inside it would produce `Histology of Kidney of 70 years old Male (HBM111.DONOR.222) (HBM555.DS.666)` — two identifiers in one title, the donor's being the misleading one. So the suffix goes on at the caller via a shared `_with_hubmap_id` helper, plus `_format_donor_page_title` for the donor-page-only case. `test_donor_id_does_not_leak_into_dataset_and_sample_titles` pins this.

### Pre-existing bugs fixed along the way

1. **`_get_all_template_keys` had `finally: return response_json['data'].keys()`** — a request failure was logged and then immediately raised `KeyError`, **500-ing the entire sitemap**. It was never caught because the old `sitemap.txt` route had no test. If user-templates-api had gone down, the sitemap would have gone down with it. Now returns `[]`.
2. **`medicalHistory` vs `medical_history`** — the client-side LD read a camelCase key Elasticsearch never populates, so every dataset description claimed the donor had no medical history. Real history now appears, which incidentally differentiates descriptions further.
3. **`_format_sample_title` emitted `"Section from Kidney of "`** — a dangling preposition and double space whenever donor metadata was absent.
4. **Meta descriptions contained baked-in source indentation.** `_get_entity_description` used backslash line-continuation inside f-strings, so the value carried 16+ literal spaces. Same defect fixed in the `routes_main.py` and `routes_file_based.py` descriptions.
5. **`DonorMappedMetadata` typed `mapped_metadata` values as scalars** when ES returns lists; the `as unknown as DatasetForLD` cast at the old `index.tsx:39` call site hid it. Moot now that generation is in Python, which already indexes `[0]`.

### Implementation notes

- `_paginate_search_after` is imported from `portal_visualization.client` rather than reimplemented. It's private, but so is the `_request` this module already calls, and both come from the pinned dependency. Sorted on `uuid.keyword`, which is unique and already the tiebreaker for search-result sorting in `js/components/search/utils.ts`.
- The sitemap `@cache` TTL is an hour-bucket cache key (`int(time.time() // 3600)`), which avoids adding a caching dependency for this.
- `_build_nlm_citation` is a deliberate six-line twin of `buildNLMCitation` in `Citation.tsx`, marked with a `ponytail:` comment. Sharing one implementation across the Python/TS boundary costs more than keeping them in sync by hand.
- `tojson` is HTML-safe (`htmlsafe_dumps`), which is what makes it safe to inline the LD inside `<script>` — the same escaping `react-content.html` already relies on for `flaskData`.

### Follow-ups, deliberately not in this PR

- **`DatasetCitation` in `SummaryBody.tsx` renders only a bare DOI link**, while `CollectionCitation` uses the full `Citation` component with an NLM string. Datasets already have `contributors`, `title`, and `published_timestamp`, and `buildNLMCitation` already exists — so wiring it up would give on-page citation text matching the new LD `citation` field. Since this is a UI change, I did not include this.
- If canonical + LD doesn't fix coverage issues, lifting more to the server-side will be necessary.


lchoy:
Do you have a source for this?

> Google composes its own "how to cite" block from creator, name, datePublished, and identifier — it does not render an arbitrary citation string verbatim. citation is added as the report asked, but the visible citation guidance comes mostly from those four fields, which this PR also fills in.

Based on Google Developer documentation, the `citation` field 

> Identifies academic articles that are recommended by the data provider be cited in addition to the dataset itself. Provide the citation for the dataset itself with other properties, such as name, identifier, creator, and publisher properties. For example, this property can uniquely identify a related academic publication such as a data descriptor, data paper, or an article for which this dataset is supplementary material for. 

and 
> Additional guidelines
> * Don't use this property to provide citation information for the dataset itself. It is intended to identify related academic articles, not the dataset itself. 

Reference: https://developers.google.com/search/docs/appearance/structured-data/dataset#structured-data-type-definitions

NickAkhmetov:
@lchoy Thanks for the catch! That stemmed from an AI hallucination I failed to fact-check.

Per my understanding, the cited documentation indicates that the `citation` field should list additional articles to cite _in addition_ to the current dataset; which is occasionally present for datasets that are present in publications. I'm editing it to use that functionality.

I've had it fix and apply the following changes:
* `publisher` (HuBMAP Consortium) is added — the one property from Google's list for the dataset's own citation (name, identifier, creator, publisher) that we weren't populating.
* `_build_nlm_citation` is deleted from Python; it was only feeding the misused field. buildNLMCitation stays in Citation.tsx, which is UI and unaffected.
* `citation` now contains references to related publications.

context/app/routes_browse.py
```diff
@@ -426,114 +543,471 @@ def _get_publication_data_types_and_organs(uuid: str):
...
+        identifier = f'https://doi.org/{doi}'
+    elif uuid:
+        identifier = f'{HUBMAP_PORTAL_URL}/browse/publication/{uuid}'
+    else:
+        return None
+
+    names = [
+        name for name in map(_contributor_name, publication.get('contributors') or []) if name
+    ]
+    author = f'{names[0]}, et al' if len(names) > 1 else (names[0] if names else '')
```
lchoy, line 899: Should author be derived from `contacts`, not `contributors`? Is the author guaranteed to always be listed first in `contributors`?

context/app/routes_browse.py
```diff
@@ -426,114 +543,471 @@ def _get_publication_data_types_and_organs(uuid: str):
...
+    ]
+
+    ld = {
+        '@context': 'https://schema.org/',
+        '@type': 'Dataset',
+        'name': title,
+        'description': _get_dataset_ld_description(entity),
+        'url': request.base_url,
+        'license': CC_BY_4_0,
+        'isAccessibleForFree': entity.get('mapped_data_access_level') == 'Public',
```
lchoy: Would this result in `false` if the access level is not `Public`? As `isAccessibleForFree` is an optional field, this can be omitted to avoid confusion.

context/app/routes_browse.py
```diff
@@ -426,114 +543,471 @@ def _get_publication_data_types_and_organs(uuid: str):
...
+            'name': 'NIH Common Fund',
+            'sameAs': 'https://ror.org/001d55x84',
+        },
+    }
+
+    optional = {
+        # Google clusters dataset replicas by identifier, so the DOI is the single most
+        # valuable field for having this page recognized as a distinct dataset.
+        'identifier': [value for value in [hubmap_id, doi_url] if value],
+        'sameAs': doi_url,
```
lchoy: I'm unclear as to whether this is the correct usage of this field. 

> Use the [sameAs](https://schema.org/sameAs) property to link to the canonical page if you add structured data to multiple copies of the dataset, such as listings in search results pages. 

> Use the [sameAs](https://schema.org/sameAs) property to indicate the most canonical URLs for the original in cases when the dataset or description is a simple republication of materials published elsewhere.

context/app/routes_browse.py
```diff
@@ -180,39 +197,92 @@ def details_rui_json(type, uuid):
...
+    """
+    Renders a <urlset> sitemap from an iterable of (loc, lastmod) pairs, where lastmod is
+    an epoch-millisecond timestamp or None.
+    """
+    urls = list(urls)
+    if len(urls) > SITEMAP_MAX_URLS:
+        current_app.logger.warning(
+            f'Sitemap has {len(urls)} URLs, over the {SITEMAP_MAX_URLS} limit: '
+            'the file will be rejected and needs splitting.'
+        )
```
john-conroy, line 276: Are we just warning if we exceed the 50k limit? Should we handle the case to generate multiple sitemaps? Could be easy to miss log warnings.

context/app/routes_browse.py
NickAkhmetov: 1. `contacts` is present in the current dataset schema, but may be unavailable for legacy datasets where contacts are indicated via a boolean field on the objects in the contributors list.
2. `contributors[0]` is used for the citation component elsewhere, e.g. for publications, so this inherits the assumption.

context/app/routes_browse.py
NickAkhmetov: Omitted, thanks for the catch!

context/app/routes_browse.py
NickAkhmetov: Omitted, thanks for the catch!

context/app/routes_browse.py
NickAkhmetov: Added pagination to handle this.

context/app/routes_browse.py
```diff
@@ -543,18 +606,46 @@ def _get_publication_data_types_and_organs(uuid: str):
...
+            {
+                'bool': {
+                    'filter': [{'term': {'processing.keyword': 'raw'}}],
+                    # Components are Multi-Assay Split datasets, which search-api leaves as
+                    # `processing: raw`, so the raw filter alone lets them through. must_not
+                    # rather than `term: false`: `is_component` is only set on multi-assay
+                    # datasets, and a false-match would drop every single-assay dataset.
+                    'must_not': [{'term': {'is_component': True}}],
+                }
+            },
```
john-conroy, line 629: The intent here is that the matching dataset is either integrated or both raw and not a component?

package.json
```diff
@@ -20,8 +20,11 @@
   },
   "pnpm": {
     "onlyBuiltDependencies": [
+      "@danmarshall/deckgl-typings",
```
john-conroy, line 23: Based on the github repo for this package, it was last updated 6 years ago? Can we use the typings directly exported from deckgl https://deck.gl/docs/get-started/using-with-typescript?

package.json
```diff
@@ -20,8 +20,11 @@
...
       "@swc/core",
       "canvas",
+      "core-js",
+      "core-js-pure",
```
john-conroy, line 27: Are these imported anywhere?

context/app/routes_browse.py
NickAkhmetov: Yes, since component datasets and processed non-integrated datasets dont' have their own page

package.json
NickAkhmetov: They're subdependencies of other dependencies, there was a message from pnpm saying that we hadn't enabled build for them.

package.json
NickAkhmetov: We aren't directly using these, they're transient from Vitessce, which still uses the old typings since they haven't updated to using the internal deckgl types.

package.json
john-conroy: Is it worth documenting this somewhere? Could be hard to trace back why these were introduced without any imports.

john-conroy (approved):

PR #4078, opened 2026-09-15
NickAkhmetov/CAT-1663 Add 8px gap between the header and page content

NickAkhmetov:
## Summary

Page content currently starts flush against the bottom of the header. This adds an 8px gap sitewide.

Every page routes through the single `Route` wrapper, so the margin goes on its `GridWrapper` once rather than being repeated per page. On entity pages the entity header is rendered inside `Header`, above `Routes`, so the gap lands below whichever header is currently on screen — the app bar normally, the entity header once it appears. The gap applies to the left and right route boundaries too, so the table of contents and the helper panel stay aligned with the content.

The homepage opts out through a new `disableTopMargin` prop, since its hero is meant to sit flush against the header. When a sitewide alert is present, the alert is already sitting between the header and the hero, so the homepage takes the margin like every other page.

One page is deliberately left alone: the error page is not wrapped in `Route`. It renders a full-bleed colored `Background` that fills the remaining viewport height, and adding a gap above it would show a strip of page background over the red, which reads as a rendering bug rather than as spacing. Happy to include it if you would rather it be consistent.

https://hms-dbmi.atlassian.net/browse/CAT-1663

## Testing

- New `Route.spec.tsx` asserts the computed `margin-top` on the wrapper: 8px by default, 0 with `disableTopMargin`.
- `pnpm tsc`, `pnpm lint` — clean.
- `pnpm test` — 949 passed, 7 skipped (135 files).

Not verified in a running dev server, so the homepage-with-alert case has only been checked through the unit test and by reading the layout — worth a quick look when reviewing.

## Screenshots/Video

None — see the note above about not having run this against a dev server.

## Checklist

  - [x] New `CHANGELOG-content-top-margin.md` is present in the root directory, describing the change(s) in full sentences.
- [x] All existing tests pass
- [x] Any relevant documentation in JIRA/Confluence has been updated to reflect the new feature — N/A
- [x] Any new functionalities have appropriate analytics functionalities added — N/A

## Additional Notes

The two calls worth confirming are the error page exclusion and the homepage rule keyed on `globalAlertMd`.


john-conroy (approved):
Any particular reason to add margin top to the route rather than margin bottom to the header?

NickAkhmetov:
The reasoning for adding margin top to the route instead of adding margin bottom to the header is to allow for exceptions to the margin rule per route - i.e. the homepage only gets the extra margin if there's a global alert of some sort (since the hero should be flush with the header otherwise).

PR #3521, opened 2024-08-27
NickAkhmetov/CAT-859 Remove raw usage of `styled-components`

NickAkhmetov:
## Summary

This PR removes the remaining references to the raw `styled-components` package. All styling is now handled with `@mui/material/styles`, and all `style.js` files have been converted to TypeScript.

## Design Documentation/Original Tickets

[CAT-859](https://hms-dbmi.atlassian.net/browse/CAT-859)

## Testing

Manually confirmed each feature matched as I worked.

## Screenshots/Video

N/A - no behavioral changes should occur.

## Checklist

- [x] Code follows the project's coding standards
  - [x] Lint checks pass locally
  - [x] New `CHANGELOG-your-feature-name-here.md` is present in the root directory, describing the change(s) in full sentences.
- [x] Unit tests covering the new feature have been added
- [x] All existing tests pass
- [x] Any relevant documentation in JIRA/Confluence has been updated to reflect the new feature
- [x] Any new functionalities have appropriate analytics functionalities added

## Additional Notes

Please specify any additional information or context relevant to this PR.

[CAT-859]: https://hms-dbmi.atlassian.net/browse/CAT-859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4081, opened 2026-09-16
Swap MUI's styled engine from styled-components to Emotion

NickAkhmetov:
## Summary

MUI has used **styled-components** as its CSS-in-JS engine in this app since the MUI 4 → 5 migration, wired up through the `@mui/styled-engine` → `@mui/styled-engine-sc` alias. Since #3521 removed the last direct `styled-components` imports, that alias bought us nothing except a second CSS-in-JS runtime to keep working. This PR moves to **Emotion**, MUI's default engine: it is the configuration upstream actually tests, and it is already installed as a peer of `@mui/material`, so the swap removes a dependency instead of adding one.

It also fixes a pin that had quietly drifted. The alias resolved `@mui/styled-engine` to `@mui/styled-engine-sc@latest`, so the app was running a **9.x** styled engine against MUI **7.3** — `pnpm ls @mui/styled-engine` on `main` shows `npm:@mui/styled-engine-sc@9.4.0`. Removing the alias re-couples the engine to `@mui/system`.

This is the **first of two stacked PRs**. Keeping the engine swap separate from the version bump means any visual regression can be bisected to one change or the other. The second PR upgrades MUI 7 → 9 (v8 was skipped upstream) and is based on this branch.

### What changed

**Dependencies and build config**

- Removed `styled-components`, `@mui/styled-engine-sc`, the `@mui/styled-engine` npm alias, and the `@swc/plugin-styled-components` SWC plugin.
- Added `@emotion/react` and `@emotion/styled` as direct dependencies. They were already in the lockfile as MUI peers, so the install is a net removal of four packages.
- Dropped the engine alias from `vite.config.mts`, `vite.config.maintenance.mts`, `.storybook/main.ts` and `tsconfig.json`; `dedupe` and `optimizeDeps.include` now name the Emotion packages.

**Transient props**

The one difference that touches app code. styled-components strips `$`-prefixed ("transient") props before forwarding them; Emotion does not, so they reach the wrapped component and end up on the DOM as React unknown-prop warnings. Rather than rename ~56 props across 36 files, a shared helper keeps the convention:

```ts
// js/helpers/styled.ts
export function shouldForwardProp(prop: PropertyKey) {
  return typeof prop !== 'string' || (!MUI_INTERNAL_PROPS.has(prop) && !prop.startsWith('$'));
}
```

The 32 `styled(Component)` call sites that take `$` props now pass it. `styled('div')`-style string-tag sites need nothing — Emotion already filters those through `@emotion/is-prop-valid`. Two sites that hand-rolled a single-prop `shouldForwardProp` now use the helper, and `shared-styles/alerts/Alert.tsx` had `shouldForwardProp: () => true`, which would have forwarded every `$` prop straight to the DOM.

The `MuiChip` theme variant keyed on `$borderRadius: 'halfRound'` has no such escape hatch, since the prop has to reach the variant resolver on the way through. It is replaced by `sx={{ borderRadius: '4px' }}` at its three call sites, and the `ChipOwnProps` module augmentation is gone.

**Smaller Emotion differences**

- **No component selectors.** Emotion does not support `` [`& ${Component}`] `` without its Babel plugin. The homepage pill bar was the only place using them; those four nested selectors are now child-element selectors (`'& > nav'`, `'& > button'`).
- **`OverridableComponent` and the `component` prop.** `styled()` over an overridable MUI component drops the `component` prop from the resulting type, so `HeaderIcon` and `StyledHeaderIcon` are cast back to `OverridableComponent<SvgIconTypeMap>`.
- **Stricter prop types.** Emotion's `StyledComponent` types surfaced three pre-existing problems rather than introducing new ones: `SelectableCard`'s `ReactNode` title collided with the DOM `title` attribute (now `Omit`ted), and `TutorialList` and `TemplateGrid` were passing `route` and `jobTypes` props that no component consumed and that landed on the card's root element.
- `SectionButtonRow`'s `leftText` is now optional, which is what its `= <div />` default already implied.
- `usePublications`'s tab handler takes a `SyntheticEvent` rather than a DOM `Event`, matching MUI's `Tabs` `onChange` signature.

### Implications

Emotion injects its `<style>` tags at runtime exactly as styled-components did, so nothing changes for Flask, the CSP (there is none), or `vendors.css`. `styled-components` still appears in the vendors bundle, but transitively, via `vitessce → @react-three/xr → @iwer/devui`; it is no longer our dependency, and no MUI code path reaches it.

Dev-mode class names lose the `displayName` labels the SWC plugin generated. MUI's own `Mui*` classes are unaffected, and no test or Cypress spec selects on generated class names.

No JIRA ticket — groundwork for the MUI 9 upgrade in the stacked follow-up PR.

## Testing

Automated, all run locally on this branch:

| Check | Result |
| --- | --- |
| `pnpm tsc` | clean |
| `pnpm lint` | clean |
| `pnpm test` | 984 passed, 7 skipped, 137 files |
| `pnpm build` | succeeds; Emotion present in `vendors.js`, no `@mui/styled-engine-sc` |
| `pnpm storybook build` | succeeds |

The test run is the useful signal for the transient-prop work: the suite renders most of these components, and `grep` over its output finds **zero** `React does not recognize the ... prop` warnings, which is what a missed `shouldForwardProp` would produce.

Manual testing still to do, and the reason this is separate from the version bump: a pass over the homepage (hero pill bar, parallax and carousel slides), search (facets, result tiles and table), a dataset detail page (accordions, the Visualization section's expand and fullscreen states, bulk data transfer), a publication page (timeline, collections), the tutorials page (chips) and the workspaces template grid, watching for both visual drift and console warnings.

## Screenshots/Video

No intended visual change. Screenshots to follow with the manual pass if anything moves.

## Checklist

  - [x] New `CHANGELOG-emotion-styled-engine.md` is present in the root directory, describing the change(s) in full sentences.
- [ ] Unit tests covering the new feature have been added — no new behavior to cover; the existing 984 tests are the regression net, and the absence of unknown-prop warnings in their output is the specific check for this change.
- [x] All existing tests pass
- [ ] Any relevant documentation in JIRA/Confluence has been updated to reflect the new feature — `CLAUDE.md`'s Styling and Vite Path Aliases sections are updated; it also still said "MUI v6", which is now corrected to v7.
- [ ] Any new functionalities have appropriate analytics functionalities added — not applicable.

## Additional Notes

Two things worth a reviewer's attention:

1. **The `$` convention is now ours, not the library's.** With styled-components the `$` prefix was meaningful to the engine. Under Emotion it is meaningful only because `js/helpers/styled` says so, which means a new `styled(Component)<{ $foo }>` that forgets to pass `shouldForwardProp` will leak `$foo` to the DOM. Worth an ESLint rule if it bites; I did not add one speculatively.
2. **Removing `route` and `jobTypes`** from the two `SelectableCard` callers is a small behavioral cleanup rather than a pure refactor. Neither prop was read by `SelectableCard`, so both were being spread onto the card's root `div`. Removing them is the fix, but flagging it in case either was intended to do something.


NickAkhmetov:
## QA round: five fixes pushed

Clicking through the app on this branch turned up five things. All are in `f8671a02`.

**The Data and Resources navigation drawer icons had collapsed to nothing.** This is the interesting one. `DrawerListItemIcon` sizes its child with

```css
& > * { padding: 8px; width: 100%; height: 100% }
```

which has exactly the same specificity as MUI's own `.MuiSvgIcon-root { width: 1em; height: 1em }`. A specificity tie is broken by stylesheet order, and Emotion inserts these in the opposite order from styled-components, so MUI's `1em` won the width while the slot's `8px` padding still applied. With `box-sizing: border-box` that is a 16px box with 16px of padding, i.e. a zero-size icon. Confirmed by dumping the generated rules out of `document.styleSheets` in a scratch render, before and after.

The rule now uses `&&` to win on specificity instead of on order. The two other containers with the same shape (`StyledIconContainer`, `StyledExternalImageIconContainer`) only ever wrap `ExternalImageIcon`, which renders a plain `` with no competing MUI class, so they are left alone.

Worth flagging for review generally: **any `& > *` rule that sets a property MUI also sets on the child's own class is now order-dependent.** That is a new hazard under Emotion. There is one more instance in `Facets/style.ts` that only contests `padding`/`margin`/`color`; nothing looked wrong there, so I have not touched it.

**`LabelledSectionText` leaked a transient prop.** Its `TextContainer` filtered `bottomSpacing`, but the transient prop is `$bottomSpacing` — the unprefixed name is destructured by the component and never reaches the styled call. So `$bottomSpacing` was forwarded and React rejected it as an invalid attribute name. It now uses the shared `shouldForwardProp` helper. This was my miss: I explicitly left this file alone in the first pass, having read the filter as already correct.

**`globalStyles` used kebab-case CSS property names.** styled-components accepted those; Emotion warns and names the camelCase spelling it expects.

**Two rules used `:first-child`,** which Emotion flags as unsafe for server-side rendering. The file browser's chip rule becomes `:first-of-type`, which is equivalent there because its only siblings are chips. The search bar's is **not** equivalent and I did not change it that way: `& > :first-child` resolves to the label whenever one is supplied, and `:first-of-type` would have matched both the label and the input wrapper. It targets `.MuiInputBase-root` by class instead.

**The build was recommending `@vitejs/plugin-react` on every run,** because removing `@swc/plugin-styled-components` left no SWC plugins at all. Swapped, which also retires the `@swc/core` and `@swc/helpers` pins that existed only to match the WASM plugin's ABI.

### One thing I could not fix

Our own `styled()` components still report as `Styled(div)` in React component stacks, with the frame pointing into `@emotion/react` rather than our source. The SWC plugin's `displayName` option used to name them. Reproduced with a scratch error boundary:

```
at div (<anonymous>)
at Styled(Component) (.../@emotion/react/dist/emotion-element-...development.cjs.js:60:23)
at Styled(div) (.../@emotion/react/dist/emotion-element-...development.cjs.js:60:23)
```

The fix is Emotion's Babel plugin, which sets a `label` that Emotion then uses as the `displayName`. Two blockers: `@vitejs/plugin-react` v6 dropped its `babel` option under Vite 8 (it uses oxc now), and routing it through `@rolldown/plugin-babel` resolves Babel 8 while `@emotion/babel-plugin` is a Babel 7 plugin — I tried it end to end against the dev server and it silently transformed nothing. I backed that out rather than leave four dependencies that don't do anything.

If readable component stacks matter more than the plugin swap, the alternative is to stay on `@vitejs/plugin-react-swc` and add `@swc/plugin-emotion`, which is the direct analogue of the plugin we removed. That reintroduces the exact-version `@swc/core` pinning this repo was already carrying, which is why I did not pick it.

Re-verified after the fixes: `pnpm tsc`, `pnpm lint`, `pnpm test` (984 passing), `pnpm build` and `pnpm storybook build` all clean, and the build no longer prints the plugin recommendation.

NickAkhmetov:
One more note on the drawer icon sizing, since it is a real visual delta rather than a pure restore.

The icon slot sets `width: theme.spacing(6)` (48px). Before the upgrade MUI 7's `ListItemIcon` had `min-width: 56px`, which beat that explicit width, so the slot was actually 56px and the icon rendered at 40px after its 8px padding. v9 lowers that default to 36px (`theme.spacing(4.5)`), so the explicit 48px now governs and the icon comes out at **32px**.

I left it at 48px because that is what the code asks for — the old 56px was MUI's default winning by accident. But it does mean these icons are ~20% smaller than on `main`, so it is worth eyeballing against production. If they should match the old size, it is one line:

```diff
-  width: theme.spacing(6),
+  width: theme.spacing(7),
```

PR #4082, opened 2026-09-16
Upgrade Material UI from v7 to v9

NickAkhmetov:
## Summary

Upgrades Material UI from **v7 to v9**. MUI skipped v8, so v9 is the next major. This also moves `@mui/icons-material` and `@mui/system` to 9, `@mui/lab` to `9.0.0-beta.9` (the only v9-compatible lab release — lab has no stable v9), and `@mui/x-date-pickers` from 8 to 9.

**This PR is stacked on #4081** and should be reviewed and merged after it. That PR swaps MUI's styled engine from styled-components to Emotion; splitting the two means a visual regression can be attributed to the engine change or the version bump, not both at once.

Most of the diff is mechanical and came from MUI's own codemods:

```bash
npx @mui/codemod@latest v9.0.0/system-props app/static/js
npx @mui/codemod@latest deprecations/all app/static/js
npx @mui/x-codemod@latest v9.0.0/pickers/preset-safe app/static/js
```

### System props → `sx` (the bulk of the diff)

v9 stops accepting shorthand style props (`mt`, `flexGrow`, `alignItems`, `bgcolor`, `display`, …) directly on `Box`, `Stack`, `Typography`, `Link`, `Grid` and `DialogContentText`. They move into each component's `sx`. Genuine component props are untouched — Stack still takes `direction`, `spacing`, `useFlexGap` and `divider`, and those counts are unchanged from `main`.

The codemod only matches the MUI component names it knows about, so it was run twice more with `--jsx=` naming this repo's wrappers around those primitives, whose call sites relied on the same shorthands: `BodyCell`, `HeaderCell`, `StackTemplate`, `LabelledSectionText`, `StyledItemLink`, `AnimatedStack`, `PageTitle`, `URLSvgIcon`, `InternalLink`, `Directory`, `ImageWrapper`, `ExternalLinkContainer`, `StyledStack`, `SectionItem`, `IconPageTitle`, `TileGrid`, `PanelBox` and `SectionHeader`.

`OrganIcon` was deliberately **excluded** from that list. Its `fontSize` is its own prop with a default that also sizes the loading `Skeleton`, so pushing it into `sx` would have broken the skeleton. It is declared on the component instead.

### Deprecated props → `slots` / `slotProps`

The `deprecations/all` codemod handled `renderTags` → `renderValue`, `ListboxComponent` → `slotProps.listbox.component`, `Typography`'s `paragraph`, and most `inputProps`. The rest were hand-written:

- **`AutocompleteRenderInputParams` no longer carries `InputProps` / `InputLabelProps`.** Those values now live under a `slotProps` object on the params. The five `renderInput` implementations (`AutocompleteEntity`, `GenePathwaysAutocomplete`, `MultiAutocomplete`, `WorkspacesAutocomplete`, `AddDatasetsTable`) destructure `slotProps` off the params and merge it into the `TextField`'s own `slotProps`, rather than letting the explicit prop clobber the spread one.
- **`inputProps`** on `Switch`, `Checkbox` and `TextField` becomes `slotProps.input` (the input component) or `slotProps.htmlInput` (the actual `<input>`, which is where `maxLength` belongs).

### Other hand-written changes

- **Grid `direction="column"` is removed in v9**, so the LineUp page's transfer list uses a `Stack` for its button column. This is the one intentional DOM change in the PR.
- **`mergeSx` helper** in `js/helpers/styled`. The codemod inlines `...(Array.isArray(sx) ? sx : [sx])` at every wrapper that merges a caller's `sx` into its own. That spread trips `@typescript-eslint/no-unsafe-assignment`, and one typed helper reads better than eight `eslint-disable` comments. Eight sites use it.
- **`SectionItem`** now extends `BoxProps` so it can forward `sx`, and its `flexBasis` prop is deleted — the codemod moved its only two callers to `sx`, leaving the prop dead.
- **`<Box component="dl" my={0}>`** in the data products list, which the codemod missed (the `component` prop makes Box polymorphic, so TypeScript accepted `my` as a DOM attribute). Left as-is it would have forwarded `my` to the DOM and silently dropped the margin reset.
- `usePublications`' tab handler and a few `sx` array entries needed `false` instead of `undefined`, which v9's `SxProps` types no longer accept.

The pickers codemod found **nothing** to change: the date-range facet already used `slotProps`, and none of the removed picker props (`enableAccessibleFieldDOMStructure`, `unstableFieldRef`, `PickersDay`) are used here. Its only output was redundant parentheses around unrelated `as` casts, which were reverted.

No JIRA ticket. Migration guides used:

- https://mui.com/material-ui/migration/upgrade-to-v9/
- https://mui.com/x/migration/migration-pickers-v8/

## Testing

Automated, all run locally on this branch:

| Check | Result |
| --- | --- |
| `pnpm tsc` | clean (was 45 errors immediately after the codemods) |
| `pnpm lint` | clean |
| `pnpm test` | 984 passed, 7 skipped, 137 files — same as `main` |
| `pnpm build` | succeeds |
| `pnpm storybook build` | succeeds |

`pnpm tsc` is the load-bearing check here: v9 deletes the deprecated props from its type definitions, so every missed call site is a compile error rather than a silent runtime change. The test suite renders most of these components and reports zero React unknown-prop warnings.

Two greps were used to confirm the codemods were complete rather than merely successful: no shorthand style props remain on any bare `Box` / `Stack` / `Typography` / `Link` / `Grid` (that grep is how the `component="dl"` case was found), and none of `inputProps`, `InputProps`, `InputLabelProps`, `renderTags`, `ListboxComponent`, `componentsProps`, `TransitionComponent` or `PaperProps` survive anywhere in the tree.

### Manual testing still needed

**Vitessce is the one I would not merge without.** See Additional Notes — it now renders against a MUI major it does not declare support for, and nothing in CI covers it. Worth loading a dataset detail page with a visualization and exercising the Vitessce chrome (layout controls, the data grid in the biomarker/feature-list panels, fullscreen, the light/dark theme switch).

Beyond that, the ~200-file `sx` conversion needs eyes on the pages it touches, because a mis-scoped style is invisible to the type checker:

- Homepage: hero pill bar, the parallax and carousel slides, related links.
- Search: facets, result tiles, the results table, the facet combobox's virtualised listbox, and the date-range facet (pickers v9).
- A dataset detail page: summary, accordions, the file browser and data products list, bulk data transfer, processed-data sections.
- Panel lists: organs, collections, publications, cell types — these all go through `StackTemplate` / `BodyCell` / `HeaderCell`, which the second codemod pass rewrote.
- Workspaces: the template grid, the workspace list checkboxes, the "add datasets" autocomplete.
- Saved lists: the title and description fields, whose `maxLength` moved to `slotProps.htmlInput`.
- Tutorials page, LineUp transfer list (the `Grid` → `Stack` swap).

Also worth a specific look: **`ListItemIcon`'s default minimum width drops from 56px to 36px in v9**, which tightens every icon list site-wide. That is an upstream default change, not something in this diff, so it will not show up in review.

## Screenshots/Video

To follow with the manual pass. No intended visual change except the `ListItemIcon` default above and the LineUp `Grid` → `Stack` swap.

## Checklist

  - [x] New `CHANGELOG-mui-9.md` is present in the root directory, describing the change(s) in full sentences.
- [ ] Unit tests covering the new feature have been added — no new behavior to cover; the existing 984 tests plus `pnpm tsc` are the regression net.
- [x] All existing tests pass
- [ ] Any relevant documentation in JIRA/Confluence has been updated to reflect the new feature — `CLAUDE.md`'s Styling section now says MUI v9.
- [ ] Any new functionalities have appropriate analytics functionalities added — not applicable.

## Additional Notes

**Vitessce and `scellop` now render against MUI 9 without declaring support for it.** `@vitessce/styles` depends on `@mui/material@^7`, `@mui/lab@7.0.0-beta.17` and `@mui/x-data-grid@8.0.0-beta.3`; `scellop` brings `@mui/material@6`. `pnpm` installs those copies, but Vite's `resolve.dedupe` collapses `@mui/material` to a single instance, and the built bundle confirms it is ours: v9-only markers are present (`nativeButton`, `MuiPickerDay`) and v7-only ones are absent (`MuiGridLegacy`, `MuiPickersDay`).

Before this PR that dedupe was benign, because our version and Vitessce's were the same 7.3.11. Now it forces a two-major jump on a third-party library, and v9 removed a lot of the deprecated props Vitessce's vendored MUI 7 code may still pass. `pnpm install` reports the unmet peer, and the build and tests pass, but no test renders Vitessce's own UI — so this is unverified rather than known-good.

If it does break, the fix is to drop `@mui/material` from `dedupe` in `vite.config.mts` and let Vitessce keep its own copy. Vitessce ships its own `ThemeProvider` and `StyledEngineProvider`, so it does not need ours. That costs bundle size, which is why I have not done it pre-emptively — it is a trade worth making deliberately, and only if the browser check shows it is needed.

Two smaller notes:

1. **`@mui/lab` is a beta.** `9.0.0-beta.9` is the only lab release compatible with material v9; there is no stable v9 lab. We use it for `LoadingButton` (6 call sites) and the `Timeline` family. Both are still exported.
2. **The browser baseline rises** to Chrome 117, Firefox 121, Safari 17 and Edge 121, up from Chrome 109 / Firefox 115 / Safari 15.4. There is no `browserslist` config in this repo to update, so this is informational.


NickAkhmetov:
## QA round: two fixes pushed, and this branch was rebased

Two more v9 consequences turned up while clicking through the app. Both are in `e84075c3`.

**Column widths were still being forwarded to the DOM.** The organ, collection, publication, biomarker and cell type panel lists keep their widths in a module-level object and spread it onto each cell:

```tsx
const desktopConfig = { name: { flexBasis: '75%', flexGrow: 1 }, /* ... */ };
<BodyCell {...desktopConfig.name} aria-label="Name">
```

The system-props codemod rewrites style props it can see in JSX, and these are hidden inside an object literal, so `flexBasis`, `flexGrow` and `flexShrink` survived the upgrade and reached the DOM, where React reported them as unknown attributes. The configs are now plain style objects passed as `sx`. Three cells combined a config with their own `sx`; those spread the two together rather than letting one silently win.

This is the general gap in the codemod worth knowing about: **it cannot see style props that reach a component through a spread.** The greps in the PR description missed these for the same reason, since they match on JSX attributes.

**Collapsible detail page sections nested a button inside a button.** `CollapsibleDetailPageSection` renders its `buttons` prop inside the accordion summary, and AccordionSummary is a ButtonBase that defaults to a native `<button>`. This affects every detail page that passes `buttons`, not only Collections where it was spotted.

Setting `component="div"` fixes it. Note that `nativeButton={false}` alone does **not**, which is the non-obvious part: reading ButtonBase, the rendered element comes from `component`, and `nativeButton` only declares whether that element is natively a button. With a div, ButtonBase infers non-native and supplies `role="button"`, `tabIndex` and its own Enter/Space handling. Verified in a scratch render that the summary is a `div` with `role="button"`, `tabindex="0"` and `aria-expanded`, and that no native button is nested inside another.

### This branch was rebased

Most of the QA findings were Emotion-engine bugs rather than v9 bugs, so they went onto #4081 where they belong — otherwise #4081 would be standalone-broken if it is ever merged and deployed to TEST on its own. This branch was rebased onto the updated #4081 and force-pushed (with `--force-with-lease`). Your `always skip scan for kill project task` commit is preserved in the new history. You will need to reset your local copy of this branch.

The lockfile conflicted during the rebase and was regenerated from the merged `package.json` rather than resolved by hand.

Re-verified after the rebase and fixes: `pnpm tsc`, `pnpm lint`, `pnpm test` (984 passing, zero unknown-prop warnings), `pnpm build` and `pnpm storybook build` all clean. The greps are clean too: no shorthand style props remain on any bare `Box` / `Stack` / `Typography` / `Link` / `Grid`, and no style props remain inside the spread config objects.

**Still unverified, and still the thing I would not merge without: Vitessce.** See the Additional Notes in the description. Nothing in this QA round exercised it.

